PR #8, opened 2019-10-10
Check for exact forms of OptionTable parse errors

sjavora:


harlanhaskins:
@swift-ci please test

harlanhaskins:
Thanks @sjavora!

Sources/SwiftDriver/Options/OptionParsing.swift
```diff
@@ -9,7 +9,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
-public enum OptionParseError : Error {
+public enum OptionParseError : Error, Equatable {
```
brentdax: You don't need it for what you're doing here, but why not `Hashable`?

sjavora: That's precisely why. 🙂  Also, synthesizing `Hashable` would increase the code size by a tiny amount, wouldn't it?

brentdax: It would, but it's usually worth doing anyway because it allows `Dictionary` and `Set` membership. In any case, we can always add it later when someone needs it.

sjavora: Yes, I understand that. But to be honest, I wasn't even sure about adding `Equatable` because it would only be used in tests (for now, anyway).

harlanhaskins: I don't think there's too much concern here about the code size implications of `OptionParseError`. I don't necessarily think we need to add Hashable right now, but we shouldn't feel bad adding conformances to these internal types in support of tests.

PR #9, opened 2019-10-11
Cleanup dead branches in prefix trie on nil insert

tmspzz:
This PR addresses this [FIXME](https://github.com/apple/swift-driver/blob/6411e9656b7c730839d6fd2e9c2105869c999aff/Sources/SwiftDriver/Utilities/PrefixTrie.swift#L72) in PrefixTrie.swift


Sources/SwiftDriver/Utilities/PrefixTrie.swift
```diff
@@ -87,4 +105,17 @@ public struct PrefixTrie<Key: Collection, Value> where Key.Element: Hashable {
       }
     }
   }
+
+  public var nodeCount: Int {
```
harlanhaskins: Nitpick: Can you add a doc comment?

Sources/SwiftDriver/Utilities/PrefixTrie.swift
```diff
@@ -57,21 +57,39 @@ public struct PrefixTrie<Key: Collection, Value> where Key.Element: Hashable {
     set {
       var index = key.startIndex
       var current = root
+      var traversed: [(node: Node, step: Key.Element?)] = [(root, nil)]
```
harlanhaskins:
```suggestion
      // Keep track of which nodes we've visited along the way,
      // so we can walk back up this if we need to prune dead branches.
      // Note: This is only used (and is only appended to) if the new
      //       value being stored is `nil`.
      var traversed: [(node: Node, step: Key.Element?)] = [(root, nil)]
```

Sources/SwiftDriver/Utilities/PrefixTrie.swift
```diff
@@ -57,21 +57,39 @@ public struct PrefixTrie<Key: Collection, Value> where Key.Element: Hashable {
...
         current.value = newValue
+        // backtrack and remove dead branches
+        if newValue == nil {
+            var current = traversed.popLast()
+            while let parent = traversed.popLast(), let currentStep = current?.step {
+                if parent.node.next[currentStep]?.value == nil
+                  && parent.node.next[currentStep]?.next.keys.count == 0 {
+                    parent.node.next[currentStep] = nil
+                    current = parent
+                }
```
harlanhaskins: I think you can simplify this. You can just find the last parent in the list of traversed nodes that has a value or multiple steps, and just cut off its child for the provided step. The rest of the nodes will get deallocated.

```swift
for (parent, step) in traversed.reversed() {
  if parent.value != nil || !parent.next.keys.isEmpty {
    parent.next[step] = nil
    break
  }
}
```

Sources/SwiftDriver/Utilities/PrefixTrie.swift
tmspzz: sure

Sources/SwiftDriver/Utilities/PrefixTrie.swift
tmspzz: I don't believe this will work for the case of going all the way back to the root node. The root node is always contains no value.

harlanhaskins: `traversed.dropFirst().reversed()`? Or just don’t add the root at the beginning

tmspzz: Ok. I tried and imho this is a bit more convoluted. What do you think?

```diff
diff --git a/Sources/SwiftDriver/Utilities/PrefixTrie.swift b/Sources/SwiftDriver/Utilities/PrefixTrie.swift
index 302bb9f..b09e05f 100644
--- a/Sources/SwiftDriver/Utilities/PrefixTrie.swift
+++ b/Sources/SwiftDriver/Utilities/PrefixTrie.swift
@@ -61,13 +61,13 @@ public struct PrefixTrie<Key: Collection, Value> where Key.Element: Hashable {
       // so we can walk back up this if we need to prune dead branches.
       // Note: This is only used (and is only appended to) if the new
       //       value being stored is `nil`.
-      var traversed: [(node: Node, step: Key.Element?)] = [(root, nil)]
+      var traversed: [(parent: Node, step: Key.Element)] = []
 
       // Traverse as much of the prefix as we can, keeping track of the index
       // we ended on
       while index < key.endIndex, let next = current.next[key[index]] {
         if newValue == nil {
-            traversed.append((next, key[index]))
+            traversed.append((current, key[index]))
         }
 
         key.formIndex(after: &index)
@@ -79,14 +79,15 @@ public struct PrefixTrie<Key: Collection, Value> where Key.Element: Hashable {
         // Update the value in the trie with the new value
         current.value = newValue
         // backtrack and remove dead branches
-        if newValue == nil {
-            var current = traversed.popLast()
-            while let parent = traversed.popLast(), let currentStep = current?.step {
-                if parent.node.next[currentStep]?.value == nil
-                  && parent.node.next[currentStep]?.next.keys.count == 0 {
-                    parent.node.next[currentStep] = nil
-                    current = parent
-                }
+        // if the current child is a leaf
+        if newValue == nil && current.next.keys.count == 0 {
+            for (parent, step) in traversed.reversed() {
+              // find the first parent with a value or more than one child
+              // if we traversed up to the root then chop everything regarless
+              if parent.value != nil || parent.next.keys.count > 1 || parent === root {
+                parent.next[step] = nil
+                break
+              }
             }
         }
         return
```

harlanhaskins: Hm, it's a little more convoluted than my example, but it seems easier to grasp than the original example. I think if it works, this implementation would be good. Maybe it would be a good idea to throw this branch into a `pruneEmptyBranchesIfNeeded(_ traversed: [(Node, Key.Element)])` function, so we can isolate it

harlanhaskins:
@swift-ci please test

PR #10, opened 2019-10-11
A proper implementation of String.isSwiftIdentifier

broadwaylamb:
Stolen from [the Swift lexer](https://github.com/apple/swift/blob/1846a5957ee971979d86bf843a0e4885da2b5a46/lib/Parse/Lexer.cpp).

### DIFF
--- a/Tests/SwiftDriverTests/StringAdditionsTests.swift
+++ b/Tests/SwiftDriverTests/StringAdditionsTests.swift
@@ -1,0 +1,79 @@
+//===--------- StringAdditionsTests.swift - String Additions Tests --------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import SwiftDriver
+
+final class StringAdditionsTests: XCTestCase {
+
+    func testBasicIdentifiers() {
+        XCTAssertTrue("contains_Underscore".sd_isSwiftIdentifier)
+        XCTAssertTrue("_startsWithUnderscore".sd_isSwiftIdentifier)
+        XCTAssertTrue("contains_Number5".sd_isSwiftIdentifier)
+        XCTAssertTrue("_1".sd_isSwiftIdentifier)
+        XCTAssertFalse("5startsWithNumber".sd_isSwiftIdentifier)
+        XCTAssertFalse("contains space".sd_isSwiftIdentifier)
+        XCTAssertFalse("contains\nnewline".sd_isSwiftIdentifier)
+        XCTAssertFalse("contains\ttab".sd_isSwiftIdentifier)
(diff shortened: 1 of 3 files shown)


brentdax:
Thank you for working on this!

Before we merge this change, you will need to add tests of its functionality. What I would suggest you do is look at each identifier used in [Swift's test/Parse/identifiers.swift](https://github.com/apple/swift/blob/master/test/Parse/identifiers.swift) and determine if the test is checking that it is valid or invalid, then create a Tests/SwiftDriverTests/StringAdditionsTests.swift file with an XCTestCase that makes sure `isSwiftIdentifier` returns the appropriate result for each of them.

Once you've ported these test cases, think about if there are any others you should add. test/Parse/identifiers.swift actually doesn't seem to include some very basic tests, like identifiers containing numbers. For bonus points, you could backport these additional test cases to test/Parse/identifiers.swift, too!

To allow this test to be written, you may need to make `isSwiftIdentifier` public. Please don't use `@testable import`; we want to be able to run tests on release builds, and release builds can't enable testable import.

Thanks!

broadwaylamb:
>To allow this test to be written, you may need to make isSwiftIdentifier public.

Do I need to add a prefix to the property, since I'm extending a type I don't own with a declaration that doesn't involve types I own?

brentdax:
I don't think we have an official policy on prefixing extension methods, but I asked around and @aciidb0mb3r at least thinks it's a good idea. Please add an `sd_` prefix.

(Thank you for thinking of this—I sure didn't.)

broadwaylamb:
@brentdax I've added some tests. However, I'm not sure whether we want to consider Swift keywords as identifiers here. Right now `import`, `class` etc. are "identifiers".

brentdax:
@broadwaylamb You should include tests proving that `sd_isSwiftIdentifier` *accepts* keywords. The C++ driver uses `Lexer::isIdentifier()`, which accepts keywords, so that's the behavior we want in swift-driver too. What it's trying to do is make sure that the module name would be valid in an `import` statement, and keywords would be valid in an `import` as long as the user surrounded them with backticks, so accepting them is the right behavior here.

Tests/SwiftDriverTests/StringAdditionsTests.swift
```diff
@@ -0,0 +1,79 @@
...
+        XCTAssertTrue("var".sd_isSwiftIdentifier)
+        XCTAssertTrue("typealias".sd_isSwiftIdentifier)
+        XCTAssertTrue("class".sd_isSwiftIdentifier)
+        XCTAssertTrue("struct".sd_isSwiftIdentifier)
+        XCTAssertTrue("enum".sd_isSwiftIdentifier)
+        XCTAssertTrue("associatedtype".sd_isSwiftIdentifier)
+        XCTAssertTrue("prefix".sd_isSwiftIdentifier)
+        XCTAssertTrue("infix".sd_isSwiftIdentifier)
+        XCTAssertTrue("postfix".sd_isSwiftIdentifier)
+        XCTAssertTrue("_".sd_isSwiftIdentifier)
```
brentdax: Good, you've already included these tests.

brentdax:
@swift-ci please smoke test

brentdax:
@swift-ci please smoke test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #12, opened 2019-10-11
Keep xcrunFind on DarwinToolchain only

cristeahub:
Hi, this is my first pull request to this repo, so I'm not 100 % sure what direction you guys want everything to go in. Let me know of any changes you want.

This change removes xcrunFind from Toolchain, and only keeps it on
DarwinToolchain. There is a special case on GenericUnixToolchain where
we check if we are on macOS, and if so we use xcrunFind. This change
solves that by initializing a new DarwinToolchain to return the correct
path.

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -15,10 +15,12 @@ fileprivate func envVarName(forExecutable toolName: String) -> String {
   return "SWIFT_DRIVER_\(toolName.uppercased())_EXEC"
 }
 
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

DougGregor:
FYI, it's failing on Linux:

cristeahub:
I added back the linux hack. I ran this on my linux docker now and it passes. I'll look into removing the hack, but it might take a bit longer, so if this pull request gets merged I'll just create a new one.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #13, opened 2019-10-13
Conform VirtualPath to CustomStringConvertible

vbergae:
Hi! 

Just a PR to fix a really simple `FIXME`. Not sure if the `FIXME` will finally apply or keeping the attribute `name` is the desired solution. Any comment or suggestion to finish the PR is more than welcome :)

### DIFF
--- a/Sources/SwiftDriver/Utilities/VirtualPath.swift
+++ b/Sources/SwiftDriver/Utilities/VirtualPath.swift
@@ -40,23 +40,7 @@ public enum VirtualPath: Hashable {
   }
 
   /// The name of the path for presentation purposes.
-  ///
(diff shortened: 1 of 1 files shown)


harlanhaskins:
Thanks for doing this! I am torn between "we should use the standard set of protocols" and "we should have this semantic property `name` that's very clear at the call site.

What if we added `name` back, but just as

```swift
var name: String { description }
```

vbergae:
True, I felt something similar looking from callee point of view. Let's see if the current solution is a good compromise 😄

DougGregor:

shahmishal:
@swift-ci test

shahmishal:
@swift-ci Please test

PR #14, opened 2019-10-13
Diagnose -i arg

sjavora:
Addresses [this FIXME](https://github.com/apple/swift-driver/blob/1e1c30534eb6af42bf66763c5ed6d918e2f56a60/Sources/SwiftDriver/Driver/Driver.swift#L691). 

I created a new test for this as the assert didn't seem to fit into `testPrimaryOutputKinds`, but I'll of course change it if desired.

I'm not exactly sure if `DriverKind.usage` should be used here, but the values seem to correspond to [this code](https://github.com/apple/swift/blob/d930599394b27fcac77a90b63e23372c6b31cd7a/lib/Driver/Driver.cpp#L95) in the original driver.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -688,8 +688,7 @@ extension Driver {
         compilerOutputType = nil
 
       case .i:
-        // FIXME: diagnose this
-        break
+        diagnosticsEngine.emit(.error_i_mode(driverKind))
 
       case .repl, .deprecatedIntegratedRepl, .lldbRepl:
         compilerOutputType = nil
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -707,6 +706,17 @@ extension Driver {
   }
 }
 
+extension Diagnostic.Message {
+  public static func error_i_mode(_ driverKind: DriverKind) -> Diagnostic.Message {
+    .error(
+      """
+      the flag '-i' is no longer required and has been removed;
```
harlanhaskins: Can you add a `\` at the end of this line, so it doesn't introduce a newline into the diagnostic text?
```suggestion
      the flag '-i' is no longer required and has been removed; \
```

sjavora: Of course, I thought that's what the original code was doing, but my C++ knowledge is fairly limited.

harlanhaskins:
@swift-ci please test

harlanhaskins:
Thanks @sjavora!

PR #15, opened 2019-10-14
Multithreading tests and diagnostics

sjavora:
Addresses [this FIXME](https://github.com/sjavora/swift-driver/blob/2737f31ad7b598c8a3ed481cad3b96a0c778c5f3/Sources/SwiftDriver/Driver/Driver.swift#L739). Also adds tests for the `// Multithreading` extension.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -707,7 +707,7 @@ extension Driver {
 }
 
 extension Diagnostic.Message {
-  public static func error_i_mode(_ driverKind: DriverKind) -> Diagnostic.Message {
+  static func error_i_mode(_ driverKind: DriverKind) -> Diagnostic.Message {
     .error(
       """
       the flag '-i' is no longer required and has been removed; \
@@ -731,17 +731,14 @@ extension Driver {
 
     // Make sure we have a non-negative integer value.
     guard let numThreads = Int(numThreadsArg.asSingle), numThreads >= 0 else {
-      diagnosticsEngine.emit(Diagnostic.Message.error_invalid_arg_value(arg: .numThreads, value: numThreadsArg.asSingle))
+      diagnosticsEngine.emit(.error_invalid_arg_value(arg: .numThreads, value: numThreadsArg.asSingle))
       return 0
     }
 
-    #if false
-    // FIXME: Check for batch mode.
-    if false {
+    if case .batchCompile = compilerMode {
       diagnosticsEngine.emit(.warning_cannot_multithread_batch_mode)
       return 0
     }
-    #endif
 
     return numThreads
   }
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -466,7 +466,7 @@ extension Diagnostic.Message {
     .error("-driver-use-frontend-path requires a Swift compiler executable argument")
   }
 
-  static var warning_cannot_multithread_batch_mode: Diagnostic.Message {
+  public static var warning_cannot_multithread_batch_mode: Diagnostic.Message {
```
DougGregor: Can we leave these internal and test for the text string? It's repetitive, but we might want to shuffle these things around over time and would rather not commit to them being at all stable.

sjavora: Done. I left `error_invalid_arg_value` alone, though, because that was already public.

DougGregor: Thank you! We can update error_invalid_arg_value in another PR

DougGregor:
@swift-ci please test

sjavora:
`
/home/buildnode/jenkins/workspace/pr-swift-driver-linux/branch-master/swift-driver/Tests/SwiftDriverTests/SwiftDriverTests.swift:169: error: SwiftDriverTests.testMultithreading : XCTAssertEqual threw error "unableToFind(tool: "clang")" -
`

Not sure what's going on here - the tests don't do anything special 🤔

DougGregor:
Hmm. Perhaps you need to add an explicit `-target` to your `Driver` invocations, e.g., as in nhttps://github.com/apple/swift-driver/blob/master/Tests/SwiftDriverTests/SwiftDriverTests.swift#L232

harlanhaskins:
Every time I've run it on Linux, `clang` has been in my `PATH`. Is it not in the `PATH` on the bots?

Also, we really need to figure out toolchain-relative lookups :/

DougGregor:
> Every time I've run it on Linux, `clang` has been in my `PATH`. Is it not in the `PATH` on the bots?

> Also, we really need to figure out toolchain-relative lookups :/

Yes, definitely.

sjavora:
> https://github.com/apple/swift-driver/blob/master/Tests/SwiftDriverTests/SwiftDriverTests.swift#L232

@DougGregor so this should work? I only have access to a Mac at the moment so I can't really test this. Also, the test failed only on the last assert - should I add the target only there or to all the invocations?

sjavora:
Added the target argument to the one failing assert for now.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci test

sjavora:
@DougGregor now it failed on the other test - added target there as well, hopefully this will finally pass. I'm sorry about all the hassle for such small changes 😞

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
No worries; we need to figure out how best to find tools within our toolchain so these tests will become target-agnostic, but that's a separate issue.

sjavora:
@DougGregor thanks! Should I open a PR to make some other diagnostics internal? If so, which ones? 

Better yet, is there a place to discuss this project on the Swift forums? https://forums.swift.org/c/development/tooling?

DougGregor:
Yes, I think I'd like to make all of the diagnostics internal. For discussions, I recommend https://forums.swift.org/c/development/compiler

PR #16, opened 2019-10-15
[Driver] - Tokenization for response files

JhonnyBillM:
This PR tries to address the "Implement proper tokenization for response files" to-do item that is on the README.

**Now we can correctly parse:**

- Double slash comments.
- Quotes.
- Escaping characters.
- Espaces.
 
This matches the current behavior of the [LLVM tokenization method](https://llvm.org/doxygen/namespacellvm_1_1cl.html#ac14be339a38954dac57ac3352c948c1f) _(the Swift compiler is currently using the LLVM method)._

Also, after reading LLVM's [ExpandResponseFiles method](https://llvm.org/doxygen/namespacellvm_1_1cl.html#a80bb7b995361b8d2c104983f15cf06a0), I think our current implementation has all we need.

_Note:_ we need to update this implementation once we get Windows support. 

_For "documentation" purposes_
What are `Response files`? Files that contains a list of command line arguments.
For more info search for "response files" in [this document](https://buildmedia.readthedocs.org/media/pdf/argo/latest/argo.pdf)

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -306,14 +306,14 @@ extension Driver {
   /// 2. Backslash escaping.
   /// 3. Space character (U+0020 SPACE).
   ///
-  /// - Returns: A String ready to be used in the shell.
+  /// - Returns: One line String ready to be used in the shell, if any.
   ///
   /// - Complexity: O(*n*), where *n* is the length of the line.
-  private static func tokenizeResponseFileLine<S: StringProtocol>(_ line: S) -> String {
-    if line.isEmpty { return "" }
+  private static func tokenizeResponseFileLine<S: StringProtocol>(_ line: S) -> String? {
+    if line.isEmpty { return nil }
     
     // Support double dash comments only if they start at the beginning of a line.
-    if line.count > 1, line.first == "/", line[line.index(after: line.startIndex)] == "/" { return "" }
+    if line.first == "/", line.count > 1, line[line.index(after: line.startIndex)] == "/" { return nil }
     
     var result: String = ""
     /// Indicates if we just parsed an escaping backslash.
@@ -336,7 +336,7 @@ extension Driver {
       
       result.append(char)
     }
-    return result
+    return result.isEmpty ? nil : result
   }
 
   /// Tokenize each line of the response file, omitting empty lines.
@@ -344,9 +344,8 @@ extension Driver {
   /// - Parameter content: response file's content to be tokenized.
   private static func tokenizeResponseFile(_ content: String) -> [String] {
     return content
-      .split(separator: "\n")
-      .map { tokenizeResponseFileLine($0) }
-      .filter { !$0.isEmpty }
+      .split(separator: "\n").map { $0 }
+      .compactMap { tokenizeResponseFileLine($0) }
   }
   
   /// Recursively expands the response files.
@@ -361,7 +360,7 @@ extension Driver {
     // Go through each arg and add arguments from response files.
     for arg in args {
       if arg.first == "@", let responseFile = try? AbsolutePath(validating: String(arg.dropFirst())) {
-        // Guard against infinity parsing loop.
+        // Guard against infinite parsing loop.
         guard visitedResponseFiles.insert(responseFile).inserted else {
           diagnosticsEngine.emit(.warn_recursive_response_file(responseFile))
           continue
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -313,7 +313,7 @@ extension Driver {
     if line.isEmpty { return nil }
     
     // Support double dash comments only if they start at the beginning of a line.
-    if line.first == "/", line.count > 1, line[line.index(after: line.startIndex)] == "/" { return nil }
+    if line.hasPrefix("//") { return nil }
     
     var result: String = ""
     /// Indicates if we just parsed an escaping backslash.
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -344,7 +344,7 @@ extension Driver {
   /// - Parameter content: response file's content to be tokenized.
   private static func tokenizeResponseFile(_ content: String) -> [String] {
     return content
-      .split(separator: "\n").map { $0 }
+      .split(separator: "\n")
       .compactMap { tokenizeResponseFileLine($0) }
   }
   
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -297,26 +297,71 @@ public struct Driver {
...
+      if char.isNewline { return result }
+
+      // Backslash escapes to the next character.
+      if char == #"\"#, !isEscaping {
+        isEscaping = true
+        continue
+      } else if isEscaping {
+        // Disable escaping and keep parsing.
+        isEscaping = false
+      }
```
JhonnyBillM: I am not 100 % convinced of this method to parse escaping characters, but it was the most convenient without having to mess the code with index management, a while loop or anything related.

If there is a more efficient approach just let me know and I can update it. Sorry.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -297,26 +297,71 @@ public struct Driver {
...
 
+  /// Tokenize each line of the response file, omitting empty lines.
+  ///
+  /// - Parameter content: response file's content to be tokenized.
+  private static func tokenizeResponseFile(_ content: String) -> [String] {
+    return content
+      .split(separator: "\n")
+      .map { tokenizeResponseFileLine($0) }
+      .filter { !$0.isEmpty }
+  }
```
JhonnyBillM: Complexity: O(n + (m×l) + m)
Where:
n = content's length.
m = content's lines number.
l = line's length.

I didn't put it in the method comments since I have never seen a "stripped down" complexity in the header files.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -297,26 +297,71 @@ public struct Driver {
...
+  /// This method supports response files with:
+  /// 1. Double slash comments at the beginning of a line.
+  /// 2. Backslash escaping.
+  /// 3. Space character (U+0020 SPACE).
+  ///
+  /// - Returns: A String ready to be used in the shell.
+  ///
+  /// - Complexity: O(*n*), where *n* is the length of the line.
+  private static func tokenizeResponseFileLine<S: StringProtocol>(_ line: S) -> String {
+    if line.isEmpty { return "" }
```
harlanhaskins: Can you return an optional here?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -297,26 +297,71 @@ public struct Driver {
...
   }
 
+  /// Tokenize each line of the response file, omitting empty lines.
+  ///
+  /// - Parameter content: response file's content to be tokenized.
+  private static func tokenizeResponseFile(_ content: String) -> [String] {
+    return content
+      .split(separator: "\n")
+      .map { tokenizeResponseFileLine($0) }
+      .filter { !$0.isEmpty }
```
harlanhaskins: And turn this map-then-filter into a `compactMap`?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -297,26 +297,71 @@ public struct Driver {
...
     visitedResponseFiles: inout Set<AbsolutePath>
   ) throws -> [String] {
-    // FIXME: This is very very prelimary. Need to look at how Swift compiler expands response file.
-
     var result: [String] = []
 
     // Go through each arg and add arguments from response files.
     for arg in args {
       if arg.first == "@", let responseFile = try? AbsolutePath(validating: String(arg.dropFirst())) {
+        // Guard against infinity parsing loop.
```
harlanhaskins: Nitpick: `infinite`

Sources/SwiftDriver/Driver/Driver.swift
JhonnyBillM: **Done.**

Very good suggestion. Thanks!

Sources/SwiftDriver/Driver/Driver.swift
JhonnyBillM: Update with the `compactMap` change:  **O(n + (m×l))**

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -297,26 +297,70 @@ public struct Driver {
...
+  /// 3. Space character (U+0020 SPACE).
+  ///
+  /// - Returns: One line String ready to be used in the shell, if any.
+  ///
+  /// - Complexity: O(*n*), where *n* is the length of the line.
+  private static func tokenizeResponseFileLine<S: StringProtocol>(_ line: S) -> String? {
+    if line.isEmpty { return nil }
+    
+    // Support double dash comments only if they start at the beginning of a line.
+    if line.first == "/", line.count > 1, line[line.index(after: line.startIndex)] == "/" { return nil }
```
JhonnyBillM: Inverted the order of this validation to prevent doing a `count` operation if there is no chance to have a comment.

JhonnyBillM:
Thanks for the review @harlanhaskins 
I have pushed the changes.

Sources/SwiftDriver/Driver/Driver.swift
harlanhaskins: This might be best as `line.hasPrefix(“//“)`

JhonnyBillM: 🤯 Wonderful. **Done.**

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -297,26 +297,70 @@ public struct Driver {
...
+    }
+    return result.isEmpty ? nil : result
   }
 
+  /// Tokenize each line of the response file, omitting empty lines.
+  ///
+  /// - Parameter content: response file's content to be tokenized.
+  private static func tokenizeResponseFile(_ content: String) -> [String] {
+    return content
+      .split(separator: "\n").map { $0 }
```
harlanhaskins: Hm, does the result of Split not have a compactMap implementation?

JhonnyBillM: **Removed.**
Sorry, my bad! That was not supposed to be there and I entirely missed it out.
Thanks for calling that out.

harlanhaskins:
@swift-ci please test

DougGregor:
@swift-ci please test

JhonnyBillM:
Thanks! @DougGregor @harlanhaskins 

Can I check this bullet point in the README ?
image
Or is anything else to do in that regard? 
_(besides the work that we need to do once we have WindowsToolchain support)_

harlanhaskins:
Feel free to submit another PR for that, if you want the credit for the commit 😄

PR #17, opened 2019-10-17
[Docs] - mark as done tokenization in response files.

JhonnyBillM:
After a day without getting anything done, needed something to "lift" myself... so, here it is @harlanhaskins xD
Sorry!


harlanhaskins:
@swift-ci please test

PR #21, opened 2019-10-18
Add generate PCH job

cltnschlosser:
First PR. Wanted to know more about how the bridging PCH worked, so I started there.

Obviously need tests still, but does it look like I'm on the right track?


harlanhaskins:
Definitely looks like you're on the right track, thanks for the contribution @cltnschlosser! Once you add some tests I think this is good to go.

Sources/SwiftDriver/Jobs/GeneratePCHJob.swift
```diff
@@ -0,0 +1,43 @@
+//===--------------- GeneratePCHJob.swift - Swift Autolink Extract ----===//
```
broadwaylamb: Can you please replace `Swift Autolink Extract` with `Generate PCH Job`?

cltnschlosser:
Okay, tried to mimic the tests from the swift repo. Found some issues while I was doing it, fixed those up. Still not exactly the same as the original driver, but I've put FIXME comments where I know there is a difference. I think this is ready for a more full review now.

DougGregor:
@swift-ci please test

Sources/SwiftDriver/Jobs/CompileJob.swift
```diff
@@ -103,6 +103,18 @@ extension Driver {
     try addCommonFrontendOptions(commandLine: &commandLine)
     // FIXME: MSVC runtime flags
 
+    if let importedObjCHeader = importedObjCHeader {
+      commandLine.appendFlag(.importObjcHeader)
+      if !parsedOptions.contains(.pchOutputDir), let pch = bridgingPrecompiledHeader {
```
DougGregor: The `parsedOptions.contains(.pchOutputDir)` check seems a little suspicious, because it feels like we should have 3 branches here for "no PCH", "specific PCH", and "PCH emitted to the PCH output directory"?. Can you check what the current driver does?

Sources/SwiftDriver/Jobs/CompileJob.swift
```diff
@@ -103,6 +103,18 @@ extension Driver {
...
+        commandLine.appendPath(pch)
+      } else {
+        commandLine.appendPath(importedObjCHeader)
+        if compilerMode != .singleCompile {
+          commandLine.appendFlag(.pchDisableValidation)
```
DougGregor: Should this go into the first `then` branch, where we are building the PCH file?

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -153,11 +153,6 @@ extension Driver {
     try commandLine.appendAll(.Xllvm, from: &parsedOptions)
     try commandLine.appendAll(.Xcc, from: &parsedOptions)
 
-    if let importedObjCHeader = importedObjCHeader {
```
DougGregor: Hmmm. If we move this logic out of the "common" frontend options, it means we'll need to make sure it shows up in each of the jobs that might need it (some of which will actually want the precompiled bridging header instead). Is there a way to centralize all of the logic for "figure out whether/which bridging header/precompiled bridging header to add" here, rather than scattering it?

Sources/SwiftDriver/Jobs/GeneratePCHJob.swift
```diff
@@ -0,0 +1,79 @@
...
+    
+    var commandLine: [Job.ArgTemplate] = swiftCompilerPrefixArgs.map { Job.ArgTemplate.flag($0) }
+    
+    inputs.append(input)
+    commandLine.appendPath(input.file)
+    commandLine.appendFlag(.emitPch)
+    
+    try addCommonFrontendOptions(commandLine: &commandLine)
+
+    if let importedObjCHeader = importedObjCHeader {
```
DougGregor: FWIW, this is what I meant about when I referred to having to copy this same logic in a few different places

Sources/SwiftDriver/Jobs/CompileJob.swift
cltnschlosser: I think you're right, I think I do want to split this out a little more, because the else case for this originally handled both no pch and pch with output directory. But when I added the disable validation, that only makes sense for pch with output directory.

The current driver seems to have a different architecture, so PCH handling is a bit different (and a bit harder to follow in my opinion). I've found the tests the most helpful for making sure I'm hitting expected behavior and I tried to copy most of them over to XCTests the best I could.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
cltnschlosser: Yeah, the only job I know for sure that may want the PCH is the compile job and only under certain circumstances. I may be able to use the `bridgingPrecompiledHeader` field, because it's set near the end of the `generatePCHJob`. That felt a bit weird when I was doing it, but I didn't want to have to duplicate the logic for the output name, and that job needs to go through all the same checks, because different cases have different flags.

Do you know if the EmitModule and MergeModule jobs want the actually PCH (have same flags as Compile job)? Because, then this would be easy to move back into `addCommonFrontendOptions`

Sources/SwiftDriver/Jobs/CompileJob.swift
DougGregor: Yeah, I agree that the current driver's logic here is a bit hard to follow, and working back from test cases and thinking about how best to achieve those results is a good approach to finding a clean design for this new driver. Thanks for keeping at it!

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
DougGregor: Both `EmitModule` and `MergeModule` jobs need the PCH, because the sources and partial modules (respectively) that they're working with can reference the bridging header's contents. 

Regarding `bridgingPrecompiledHeader`, I had missed that previously. Please make this a `let` and compute its value as part of `Driver.init`. This is one of the things we're structuring differently in this new driver vs. the existing driver: all of the state that's derived from command-line arguments should be a `let` in the `Driver` so there is a single point of definition.

cltnschlosser: Okay, makes a lot of sense. I'll look into what I'll need to do to do that.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1332,13 +1338,39 @@ extension Driver {
       diagnosticEngine.emit(.error_bridging_header_module_interface)
     }
 
-    let objcHeaderPath = try VirtualPath(path: objcHeaderPathArg.asSingle)
-    // FIXME: Precompile bridging header if requested.
-    return objcHeaderPath
+    let objcHeader = try VirtualPath(path: objcHeaderPathArg.asSingle)
+    
+    if objcHeader.extension != FileType.objcHeader.rawValue {
+      diagnosticEngine.emit(.error_objc_header_not_header)
```
DougGregor: Technically, it's okay for the (now-poorly-named) imported Objective-C header to have an extension other than `.h`, so I don't think we should produce an error here. The existing driver will allow such a header and pass it through, but silently decides not to try to precompile it. I'd rather we go ahead and accept it and precompile it regardless of the extension, because the user told us explicitly it's an imported header.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1332,13 +1338,39 @@ extension Driver {
...
+  static func computeBridgingPrecompiledHeader(_ parsedOptions: inout ParsedOptions, importedObjCHeader: VirtualPath?) throws -> VirtualPath? {
+    guard let input = importedObjCHeader,
+      parsedOptions.hasFlag(positive: .enableBridgingPch, negative: .disableBridgingPch, default: true) else {
+        return nil
+    }
+    // FIXME: should have '-.*' at the end of the filename, similar to llvm::sys::fs::createTemporaryFile
+    let pchFileName = input.basenameWithoutExt.appendingFileTypeExtension(.pch)
+    let output: VirtualPath
+    if let outputDirectory = parsedOptions.getLastArgument(.pchOutputDir)?.asSingle {
+      let outputPath = (outputDirectory as NSString).appendingPathComponent(pchFileName)
```
DougGregor: Please use TSCBasic's path-manipulation utilities for this, e.g., by adding an `appending` method on `VirtualPath` to goes through `AbsolutePath` or `RelativePath`'s method of the same name. We shouldn't need to use Foundation or bridge to NSString for path manipulation.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1332,13 +1338,39 @@ extension Driver {
...
+      output = try VirtualPath(path: outputPath)
+    } else {
+      output = .temporary(RelativePath(pchFileName))
+    }
+    return output
   }
 }
 
 extension Diagnostic.Message {
+  static var error_objc_header_not_header: Diagnostic.Message {
```
DougGregor: As noted above, I don't think we should have this diagnostic.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -154,10 +155,21 @@ extension Driver {
     // Pass through any subsystem flags.
     try commandLine.appendAll(.Xllvm, from: &parsedOptions)
     try commandLine.appendAll(.Xcc, from: &parsedOptions)
-
+
```
DougGregor: Unnecessary whitespace addition?

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -154,10 +155,21 @@ extension Driver {
...
     try commandLine.appendAll(.Xcc, from: &parsedOptions)
-
+    
     if let importedObjCHeader = importedObjCHeader {
       commandLine.appendFlag(.importObjcHeader)
-      commandLine.appendPath(importedObjCHeader)
+      if requestPrecompiledObjCHeader, let pch = bridgingPrecompiledHeader {
+        if parsedOptions.contains(.pchOutputDir) {
+          commandLine.appendPath(importedObjCHeader)
+          if compilerMode != .singleCompile {
```
DougGregor: I wonder if this should be a `switch` so we update the logic if we ever add another compiler mode. Your change here uncovered a bug in the original C++ compiler driver (I went and filed it as https://bugs.swift.org/browse/SR-11986), where we didn't update an `if` when we added batch mode. Thanks! (and thanks for making the new driver more correct along the way!)

Sources/SwiftDriver/Jobs/GeneratePCHJob.swift
```diff
@@ -0,0 +1,69 @@
...
+    try addCommonFrontendOptions(commandLine: &commandLine, requestPrecompiledObjCHeader: false)
+    
+    try commandLine.appendLast(.indexStorePath, from: &parsedOptions)
+    
+    // TODO: Should this just be pch output with extension changed?
+    if parsedOptions.hasArgument(.serializeDiagnostics), let outputDirectory = parsedOptions.getLastArgument(.pchOutputDir)?.asSingle {
+      commandLine.appendFlag(.serializeDiagnosticsPath)
+      let path: VirtualPath
+      if let modulePath = parsedOptions.getLastArgument(.emitModulePath) {
+        var outputBase = (outputDirectory as NSString).appendingPathComponent(input.file.basenameWithoutExt)
```
DougGregor: As noted earlier, please use TSCBasic's Path facilities (directly or via additions to `VirtualPath`).

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1332,13 +1338,39 @@ extension Driver {
...
+  }
+  
+  /// Compute the path of the generated bridging PCH for the Objective-C header.
+  static func computeBridgingPrecompiledHeader(_ parsedOptions: inout ParsedOptions, importedObjCHeader: VirtualPath?) throws -> VirtualPath? {
+    guard let input = importedObjCHeader,
+      parsedOptions.hasFlag(positive: .enableBridgingPch, negative: .disableBridgingPch, default: true) else {
+        return nil
+    }
+    // FIXME: should have '-.*' at the end of the filename, similar to llvm::sys::fs::createTemporaryFile
+    let pchFileName = input.basenameWithoutExt.appendingFileTypeExtension(.pch)
```
DougGregor: I think you'll also need to look up this path in the `OutputFileMap`

Sources/SwiftDriver/Jobs/GeneratePCHJob.swift
```diff
@@ -0,0 +1,69 @@
...
+        // TODO: does this hash need to be persistent?
+        let code = UInt(bitPattern: modulePath.asSingle.hashValue)
+        outputBase.append(String(code, radix: 36))
+        path = try VirtualPath(path: outputBase.appendingFileTypeExtension(.diagnostics))
+      } else {
+        // FIXME: should have '-.*' at the end of the filename, similar to llvm::sys::fs::createTemporaryFile
+        path = .temporary(RelativePath(input.file.basenameWithoutExt.appendingFileTypeExtension(.diagnostics)))
+      }
+      commandLine.appendPath(path)
+      outputs.append(.init(file: path, type: .diagnostics))
```
DougGregor: This might be another case where we need to check the output file map, first. I recommend splitting the bridging PCH diagnostics path logic into a separate method. Maybe make another `let` constant for it in `Driver` to compute in the initializer of `Driver`?

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1199,6 +1200,234 @@ final class SwiftDriverTests: XCTestCase {
...
+      XCTAssert(plannedJobs[0].commandLine.contains(.path(try VirtualPath(path: "/pch"))))
+
+      XCTAssertEqual(plannedJobs[1].kind, .compile)
+      XCTAssertEqual(plannedJobs[1].inputs.count, 1)
+      XCTAssertEqual(plannedJobs[1].inputs[0].file, try VirtualPath(path: "foo.swift"))
+      XCTAssert(plannedJobs[1].commandLine.contains(.flag("-pch-disable-validation")))
+    }
+    
+    do {
+      var driver = try Driver(args: ["swiftc", "-c", "-embed-bitcode", "-import-objc-header", "TestInputHeader.h", "-pch-output-dir", "/pch", "foo.swift"])
```
DougGregor: FWIW, `-emit-bitcode` support isn't really implemented yet, so this particular test is going to break when we add it. I don't mind having it here, but a FIXME might be warranted to alert the person who trips over it.

cltnschlosser: I think this should be okay. This doesn't verify the output, instead there is this `// FIXME: Verify llvm-bc as output`

cltnschlosser:
@DougGregor Looking into `OutputFileMap` so I can understand what changes need to be made. Also haven't run this against a project, I'll look for a suitable project to test this. What's the best way to confirm that it's working correctly? Matching(Same commands and arguments) `json-compilation-database`?

Tests/SwiftDriverTests/SwiftDriverTests.swift
DougGregor: Ah, okay!

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -91,6 +91,24 @@ public enum VirtualPath: Hashable {
...
+  /// Returns the virtual path with an additional literal component appended.
+  ///
+  /// This should not be used with `.standardInput` or `.standardOutput`.
+  public func appending(component: String) -> VirtualPath {
+    switch self {
+    case .absolute(let path):
+      return .absolute(path.appending(component: component))
+    case .relative(let path):
+      // FIXME: TSCBasic should probably have RelativePath.appending(component:)
+      return .relative(RelativePath(path.pathString + "/" + component))
```
DougGregor: Huh, I wonder why that isn't available. This'll do for now.

DougGregor:
@swift-ci please test

DougGregor:
For testing a real project, I have a very ad hoc approach: I grab an iOS app ([MovieSwiftUI](https://github.com/Dimillian/MovieSwiftUI) is a favorite) and build it with the old driver; then I swap in the new driver, clean, and build again to see if the steps and command lines look right. It's not as scientific as comparing command lines exactly, but it establishes whether things seem to work end-to-end. There are other specific things to look for... for example, if the bridging header has code that produces a warning, does that show up in Xcode?

DougGregor:
@swift-ci please test

PR #22, opened 2019-10-19
Internalize diagnostics

sjavora:
Based on previous discussion, this makes all diagnostics declared in the project internal. Tests now compare the message strings directly.

Do we want to move the various extensions of `Diagnostic.Message` to `Diagnostics.swift`? Right now, they are sprinkled into various files based on relevance, but in `Driver.swift` there are now 5 separate extensions.


DougGregor:
It probably does make sense to put them all in one file. Originally I was thinking they would be defined closer to their use, but then there is no re-use

DougGregor:
@swift-ci please test

PR #24, opened 2019-10-21
[Parseable output] Use FileType's description for the file type.

DougGregor:
`FileType.description` provides the "external" name for each file type,
which is used for parsable output and output file maps, among other
things. Fix the description for serialized diagnostics files... so now
diagnostics show up properly in Xcode.


DougGregor:
@swift-ci please test

PR #25, opened 2019-10-23
[Toolchains] - look for tools relative to the executable path.

JhonnyBillM:
Once swift-driver is installed in the Toolchain, we can look for tools right next to it's executable path, this help us to be more platform agnostic while locating the toolchain.

This is the first step in this direction, we will need to make this more robust to get rid of xcrun if possible, and get a proper linux implementation.

---

The other options are:

1. Shim LLVM's [getMainExecutable](https://github.com/apple/swift-llvm/blob/5bccd3526fd93ceb9e9de4ece0ffa2c5e26a9a02/lib/Support/Unix/Path.inc#L174) method using `swift-lldb`, and expose the API from `lldbSwift` to `tools-support-core`, so other projects such as `SwiftPM` would benefit from this. 

2. Try to implement something like LLVM's [getMainExecutable](https://github.com/apple/swift-llvm/blob/5bccd3526fd93ceb9e9de4ece0ffa2c5e26a9a02/lib/Support/Unix/Path.inc#L174) method in `tools-support-core`. But I don't think this is 100 % feasible right now; this should be implemented in some low-level library such as `LLVM`, `swift-lldb` or `Foundation`, to maintain projects as `SwiftPM` and `swift-driver` platform agnostic as possible.


Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -74,4 +75,9 @@ extension Toolchain {
     ).spm_chomp()
     return Triple(triple)
   }
+  
+  /// Returns the `executablePath`'s directory.
+  public var executableDir: AbsolutePath {
+    AbsolutePath(Bundle.main.executablePath ?? "").parentDirectory
```
harlanhaskins: This seems like a good candidate for

```swift
guard let path = Bundle.main.executablePath else {
  fatalError("could not find executable path")
}
```

harlanhaskins:
Just wanna make sure this does the right thing falling back

@swift-ci please test

Sources/SwiftDriver/Toolchains/Toolchain.swift
JhonnyBillM: Agree! 
**Done**

harlanhaskins:
@swift-ci please test

harlanhaskins:
@swift-ci please test

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -20,38 +20,25 @@ fileprivate func envVarName(forExecutable toolName: String) -> String {
...
-    return AbsolutePath("/usr/bin/" + exec)
-  #endif
+  
+  public init(env: [String: String]) {
+    self.env = env
   }
-
+  
   /// Retrieve the absolute path for a given tool.
   public func getToolPath(_ tool: Tool) throws -> AbsolutePath {
```
DougGregor: There's a lot of redundancy between this and the `GenericUnixToolchain`. Can we sink the main implementation of `getToolPath(_:)` down into `Toolchain` and handle the differences some other way? We'd like the toolchains to behave similarly wherever we can.

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -137,4 +119,29 @@ public final class DarwinToolchain: Toolchain {
     \(isShared ? "_dynamic.dylib" : ".a")
     """
   }
+  
+  /// Looks for the executable in the `SWIFT_DRIVER_TOOLNAME_EXEC` enviroment variable, if found nothing,
```
DougGregor: Typo "enviroment"

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -137,4 +119,29 @@ public final class DarwinToolchain: Toolchain {
...
+  private func xcrunFind(exec: String) throws -> AbsolutePath {
+  #if os(macOS)
+    let path = try Process.checkNonZeroExit(
+      arguments: ["xcrun", "-sdk", "macosx", "--find", exec],
+      environment: env
+    ).spm_chomp()
+    return AbsolutePath(path)
+  #else
+    // This is a hack so our tests work on linux. We need a better way for looking up tools in general.
+    return AbsolutePath("/usr/bin/" + exec)
```
DougGregor: Can we search the `PATH` instead of hardcoding `/usr/bin`?

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -137,4 +119,29 @@ public final class DarwinToolchain: Toolchain {
...
+    } else if let path = lookupExecutablePath(filename: exec, searchPaths: [executableDir]) {
+      return path
+    }
+    return try xcrunFind(exec: exec)
+  }
+  
+  private func xcrunFind(exec: String) throws -> AbsolutePath {
+  #if os(macOS)
+    let path = try Process.checkNonZeroExit(
+      arguments: ["xcrun", "-sdk", "macosx", "--find", exec],
```
DougGregor: It's unfortunate to have host-specific behavior here. How strange would it be to search for `xcrun` in the path and use it if it's there?

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
JhonnyBillM: As Harlan mentioned, this work is being done in #26 I think we can keep working on this PR and merge it until we figure it out some things before completing the work in #26

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -137,4 +119,29 @@ public final class DarwinToolchain: Toolchain {
     \(isShared ? "_dynamic.dylib" : ".a")
     """
   }
+  
+  /// Looks for the executable in the `SWIFT_DRIVER_TOOLNAME_EXEC` enviroment variable, if found nothing,
+  /// looks in the executable path; finally, fallback to xcrunFind.
+  /// - Parameter exec: executable to look for [i.e. `swift`].
+  func lookup(exec: String) throws -> AbsolutePath {
```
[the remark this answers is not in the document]
JhonnyBillM: Agree! done.

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
JhonnyBillM: Thanks, fixed.

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
JhonnyBillM: I like this approach. Done.

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
JhonnyBillM: This is "mostly" what is going to happen with the generalized way to look for tools (see #26).
I think after working on both PRs the best would be to emit an error here if we don't find xcrun in the search paths (as we only have this method in the DarwinToolchain).

DougGregor:
Conflicts everywhere! It'll be good to get this up-to-date and merged now that #25 is in. Thank you!

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

JhonnyBillM:
`segmentation fault` on Linux without further details 🤔.

I saw some tests failing on Linux in the main swift project too. I think this can be a CI configuration issue.

DougGregor:
@swift-ci please test

PR #26, opened 2019-10-24
[Toolchains] - Generalize lookup routines as a Toolchain extension.

JhonnyBillM:
This generalizes`lookup(exec:)` and exposes `envVar(forExecutable:)`.

Due to the generalization, the following tests were failing in Linux:
1. `testDSYMGeneration`: unableToFind(tool: "dsymutil")
2. `testLinking`: unableToFind(tool: "libtool")
3. `testStandardCompileJobs`: unableToFind(tool: "dsymutil")

`testLinking` also failed in macOS when specifying the linux target.

So, I ifdef them out, but if we need to do anything else in the test suit, just please let me know.


Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -74,4 +81,61 @@ extension Toolchain {
...
+    return "SWIFT_DRIVER_\(toolName.uppercased())_EXEC"
+  }
+  
+  /// Looks for the executable in the `SWIFT_DRIVER_TOOLNAME_EXEC` enviroment variable, if found nothing,
+  /// looks in the executable path; finally, fallback to xcrunFind on macOS.
+  /// - Parameter exec: executable to look for [i.e. `swift`].
+  func lookup(exec: String) throws -> AbsolutePath {
+    if let overrideString = envVar(forExecutable: exec) {
+      return try AbsolutePath(validating: overrideString)
+    } else if let path = lookupExecutablePath(filename: exec, searchPaths: [executableDir]) {
```
harlanhaskins: Could this be `lookupExecutablePath(filename: exec, searchPaths: [executableDir] + searchPaths)`?

harlanhaskins:
We definitely want to un-`#if` those tests. Were you able to figure out why it couldn't find the tools on Linux?

Sources/SwiftDriver/Toolchains/Toolchain.swift
JhonnyBillM: Looking for `searchPaths`  in macOS was crashing. Later today I will dig into it to find the real reason and will provide a detailed response or the fix.

harlanhaskins:
@DougGregor some of your comments in #25 are addressed here.

Sources/SwiftDriver/Toolchains/Toolchain.swift
JhonnyBillM: The issue here is that after we find the swift compiler path in the searchPaths, we cannot find the resource directory under `/usr/lib/swift/`.

is macOS supposed to have a resource directory at `/usr/lib/swift/`, or what is the correct way to handle this?

harlanhaskins: Ah, the resource directory is in the Xcode toolchain on macOS, inside `/Applications/Xcode.app/Contents/Developer/Toolchains/<dir>.xctoolchain`

JhonnyBillM: Now we first look in the executable path, then go to xcrun, and finally look into the searchPaths.

This is the "best" approach right now because in macOS the toolchain lives under Xcode, not under the Swift executable in `~/usr/lib/swift/`.

JhonnyBillM:
1. Updated how we look for executables.
2. Tests are working on Linux, so I removed the `#if` clause.

---
**Why were those tests failing?**

They were failing in my Docker instance¹ because it doesn't have `dsymutil` nor `libtool`. 

I was looking to see if those tools are in GNU, and I thiiiink they are, but I am not sure about `dsymutil` (I kind-of saw in LLVM that dsymutil is only available in Darwin, but again, I am not 100 % sure). 

Tests are working because of [this hack](https://github.com/apple/swift-driver/blob/fda643c7a13a0dd205cb323dc2986f1f8bf931e4/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift#L36), and although this is outside the scope of this PR, in order to get rid of [the previously mentioned hack](https://github.com/apple/swift-driver/blob/fda643c7a13a0dd205cb323dc2986f1f8bf931e4/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift#L36) in a future PR, I would like to know:

• Does swift-driver requires `dsymutil` and `libtool` to be installed and in the PATH in linux?


¹ Configured following [this steps](https://github.com/apple/swift-driver#preparing-a-linux-docker-for-debug)

harlanhaskins:
`dsymutil`  and `libtool` _are_ Darwin-only tools, so we _should_ conditionalize checks for them.

JhonnyBillM:
Ok so, can we move forward with this PR and discuss those things in a future PR when trying to remove the aforementioned hack ?

harlanhaskins:
@swift-ci please test

harlanhaskins:
Hmm.

```
clang: error: unknown argument: '-print-target-triple'
```

Makes me wonder if we picked an old clang instead of the one next to swift? Can you add some tests that we find a clang path that’s relative to the swift path for a given toolchain?

JhonnyBillM:
Done. This was interesting, moved the hack to `lookup(exec:)`. Once this PR and #25 are merged the hack will depend on how we would like to handle the tests that needs `dsymutil` and `libtool `.  

---
This work will benefit from #25, should I keep that as a separate PR, or should we merge those changes into this PR and move forward with "1 big change"?

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
This is looking good; I'll go ahead and merge this and we can iterate on #25 separately.

JhonnyBillM:
Thank you @harlanhaskins for all your help in this PR. And thanks @DougGregor  for reviewing + merging.

PR #27, opened 2019-10-24
[Lint] Remove unused declarations and imports

jpsim:
Found using SwiftLint's unused declaration and unused import rules.

I'm not sure how maintainers feel about keeping unused code or imports around if it's expected to be used at some point in the future.

I'm happy to close this if everything found here is ok to keep around.

I was looking for some low hanging fruit to contribute to this project and cleaning up unused stuff seemed like a good start.

### DIFF
--- a/Sources/SwiftDriver/Execution/llbuild.swift
+++ b/Sources/SwiftDriver/Execution/llbuild.swift
@@ -12,12 +12,7 @@
 
 // FIXME: This is directly copied from SwiftPM, consider moving this to llbuild.
 
-// We either export the llbuildSwift shared library or the llbuild framework.
-#if canImport(llbuildSwift)
 @_exported import llbuildSwift
-#else
-@_exported import llbuild
-#endif
 
 import Foundation
 
--- a/Sources/SwiftDriver/Execution/llbuild.swift
+++ b/Sources/SwiftDriver/Execution/llbuild.swift
@@ -15,7 +15,6 @@
 // We either export the llbuildSwift shared library or the llbuild framework.
 #if canImport(llbuildSwift)
 @_exported import llbuildSwift
-@_exported import llbuild
 #else
 @_exported import llbuild
 #endif
--- a/Sources/SwiftDriver/Execution/ParsableOutput.swift
+++ b/Sources/SwiftDriver/Execution/ParsableOutput.swift
@@ -9,7 +9,6 @@
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
-import TSCBasic
 
 import Foundation
 
--- a/Sources/SwiftDriver/Incremental Compilation/IncrementalCompilation.swift
+++ b/Sources/SwiftDriver/Incremental Compilation/IncrementalCompilation.swift
@@ -10,7 +10,6 @@
 //
 //===----------------------------------------------------------------------===//
 import TSCBasic
-import TSCUtility
 import Foundation
 
 // FIXME: rename to something like IncrementalCompilationInitialState
--- a/Sources/SwiftDriver/Incremental Compilation/InputIInfoMap.swift
+++ b/Sources/SwiftDriver/Incremental Compilation/InputIInfoMap.swift
@@ -10,7 +10,6 @@
 //
 //===----------------------------------------------------------------------===//
 import TSCBasic
-import TSCUtility
 import Foundation
 @_implementationOnly import Yams
 
@@ -172,8 +171,5 @@ public extension InputInfoMap {
 extension Diagnostic.Message {
   static func remark_could_not_read_build_record(_ error: Error) -> Diagnostic.Message {
     .remark("Incremental compilation could not read build record: \(error.localizedDescription).")
-  }
-  static var remark_incremental_compilation_not_implemented_yet: Diagnostic.Message {
-    .remark("Incremental compilation not implemented yet")
   }
 }
--- a/Sources/SwiftDriver/Incremental Compilation/InputInfo.swift
+++ b/Sources/SwiftDriver/Incremental Compilation/InputInfo.swift
@@ -9,8 +9,6 @@
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
-import TSCBasic
-import TSCUtility
 import Foundation
 
 public struct InputInfo: Equatable {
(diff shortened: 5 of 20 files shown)


Sources/SwiftDriver/Execution/llbuild.swift
```diff
@@ -15,7 +15,6 @@
 // We either export the llbuildSwift shared library or the llbuild framework.
 #if canImport(llbuildSwift)
 @_exported import llbuildSwift
-@_exported import llbuild
 #else
 @_exported import llbuild
 #endif
```
aciidb0mb3r: Can you change this to:

```swift
#if canImport(llbuildSwift)
 @_exported import llbuildSwift
 #endif

 @_exported import llbuild
```

aciidb0mb3r:
@swift-ci smoke test

Sources/SwiftDriver/Execution/llbuild.swift
jpsim: This change is actually removing the import in the case where llbuildSwift can be imported since it won't be used.

What you're suggesting would always import llbuild.

Are you sure that's what you'd like me to do?

harlanhaskins:
@swift-ci please smoke test

Sources/SwiftDriver/Execution/llbuild.swift
aciidb0mb3r: You're right. This is entirely wrong because I blindly copied it  from SwiftPM 😁. We shouldn't be exporting llbuild from the driver and we'll never end up in the `else` case. So I guess this can be simplified to just:

```
import llbuildSwift
```

jpsim: I don't think that's quite right. This fails to build:

```diff
diff --git a/Sources/SwiftDriver/Execution/llbuild.swift b/Sources/SwiftDriver/Execution/llbuild.swift
index 0637ee6..c809b4d 100644
--- a/Sources/SwiftDriver/Execution/llbuild.swift
+++ b/Sources/SwiftDriver/Execution/llbuild.swift
@@ -12,12 +12,7 @@
 
 // FIXME: This is directly copied from SwiftPM, consider moving this to llbuild.
 
-// We either export the llbuildSwift shared library or the llbuild framework.
-#if canImport(llbuildSwift)
-@_exported import llbuildSwift
-#else
-@_exported import llbuild
-#endif
+import llbuildSwift
 
 import Foundation
```

Build output (truncated):

```
/Users/jsimard/Projects/swift-driver/Sources/SwiftDriver/Execution/JobExecutor.swift:203:38: error: use of undeclared type 'Key'
  func lookupRule(rule: String, key: Key) -> Rule {
                                     ^~~
/Users/jsimard/Projects/swift-driver/Sources/SwiftDriver/Execution/JobExecutor.swift:203:46: error: use of undeclared type 'Rule'
  func lookupRule(rule: String, key: Key) -> Rule {
                                             ^~~~
/Users/jsimard/Projects/swift-driver/Sources/SwiftDriver/Execution/JobExecutor.swift:247:15: error: use of undeclared type 'Key'
  init(_ key: Key) {
              ^~~
/Users/jsimard/Projects/swift-driver/Sources/SwiftDriver/Execution/JobExecutor.swift:259:45: error: use of undeclared type 'Value'
  override func isResultValid(_ priorValue: Value) -> Bool {
                                            ^~~~~
/Users/jsimard/Projects/swift-driver/Sources/SwiftDriver/Execution/JobExecutor.swift:259:17: error: method does not override any method from its superclass
  override func isResultValid(_ priorValue: Value) -> Bool {
  ~~~~~~~~      ^
```

aciidb0mb3r: I guess llbuildSwift is not exporting llbuild? Then how come the original diff suggested to remove `import llbuild`?

jpsim: Seems like llbuildSwift can be imported in both Darwin & Linux, based on my testing on macOS and Docker. Simplified in 8885c45.

jpsim:
Is CI still running? It was kicked off ~19 hours ago and still doesn't seem to have reported back 🤔

aciidb0mb3r:
@swift-ci please smoke test

aciidb0mb3r:
@swift-ci please test

aciidb0mb3r:
@swift-ci  test

aciidb0mb3r:
I see the [job running](https://ci.swift.org/job/pr-swift-driver-macos/38/) but the GH status is not updated. cc @shahmishal

aciidb0mb3r:
@swift-ci please test

shahmishal:
```
13:08:53 Setting status of 21c4c1b2f1d0511c76fcad2674355b1f6db6d0f0 to SUCCESS with url https://ci.swift.org/job/pr-swift-driver-macos/38/ and message: ' No test results found.'
```

jpsim:
Thanks for getting CI to report here folks. Can we merge this?

harlanhaskins:
Thanks, @jpsim!

jpsim:
Thanks for the reviews. I'm happy I could contribute a tiny bit.

By the way, there's quite a bit more that SwiftLint found as unused that I didn't remove here because it seemed useful to keep around for later. Sharing in case you're curious:

* [`BatchModeInfo.seed`](https://github.com/apple/swift-driver/blob/e1e26c3/Sources/SwiftDriver/Driver/CompilerMode.swift#L34)
* [`Triple.is32Bit`](https://github.com/apple/swift-driver/blob/e1e26c3/Sources/SwiftDriver/Utilities/Triple.swift#L594)
* [`Triple.is16Bit`](https://github.com/apple/swift-driver/blob/e1e26c3/Sources/SwiftDriver/Utilities/Triple.swift#L597)
* [`DriverBuildValue.kind`](https://github.com/apple/swift-driver/blob/e1e26c3/Sources/SwiftDriver/Execution/JobExecutor.swift#L225)
* [`GenericUnixToolchain.Error`](https://github.com/apple/swift-driver/blob/e1e26c3/Sources/SwiftDriver/Toolchains/GenericUnixToolchain.swift#L16) (compiled on macOS and Linux but just used on Linux)
* [`InputInfo.Status.identifier`](https://github.com/apple/swift-driver/blob/e1e26c3/Sources/SwiftDriver/Incremental%20Compilation/InputInfo.swift#L30)
* [`InputInfoMap.SectionName`](https://github.com/apple/swift-driver/blob/e1e26c3/Sources/SwiftDriver/Incremental%20Compilation/InputIInfoMap.swift#L37)

PR #28, opened 2019-10-26
Handle subcommands

sjavora:
This tries to implement the handling of subcommands based on the original driver. I must admit I'm a little out of my depth here so I'd appreciate some guidance:
- in the original driver, the subcommand check happens before the actual driver is even created ([here](https://github.com/apple/swift/blob/348fa2145e675ec4e396a1da54c2201a366a37da/tools/driver/driver.cpp#L234)) - I put the code in the init, because the FIXME was there, but should an init be invoking `exec`?
- related to the above, is `exec` the right function to use here?
- similarly, is `Process.findExecutable` the right method to use in place of `getExecutablePath`?
- there is a slight difference in behavior between `llvm::sys::path::stem` and `basenameWithoutExt` - namely `basenameWithoutExt` will return `.txt` for `/foo/.txt`, while `llvm::sys::path::stem` will return an empty string. Is this a problem? If so, should I implement something like `stem` on `VirtualPath`? 

Thank you very much for your time 🙂

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -43,6 +43,7 @@ public struct Driver {
   enum Error: Swift.Error {
     case invalidDriverName(String)
     case invalidInput(String)
+    case subcommandPassedToDriver
   }
   
   /// The set of environment variables that are visible to the driver and
@@ -193,11 +194,15 @@ public struct Driver {
     diagnosticsHandler: @escaping DiagnosticsEngine.DiagnosticsHandler = Driver.stderrDiagnosticsHandler
   ) throws {
     self.env = env
-    
-    // FIXME: Determine if we should run as subcommand.
 
     self.diagnosticEngine = DiagnosticsEngine(handlers: [diagnosticsHandler])
+
+    if case .subcommand = try Self.invocationRunMode(forArgs: args).mode {
+      throw Error.subcommandPassedToDriver
+    }
+
     var args = try Self.expandResponseFiles(args, diagnosticsEngine: self.diagnosticEngine)[...]
+
     self.driverKind = try Self.determineDriverKind(args: &args)
     self.optionTable = OptionTable()
     self.parsedOptions = try optionTable.parse(Array(args))
@@ -348,6 +353,54 @@ public struct Driver {
         compilerMode: compilerMode,
         outputFileMap: self.outputFileMap,
         moduleName: moduleName)
+  }
+}
+
+extension Driver {
+
+  public enum InvocationRunMode: Equatable {
+    case normal(isRepl: Bool)
+    case subcommand(String)
+  }
+
+  /// Determines whether the given arguments constitute a normal invocation,
+  /// or whether they invoke a subcommand.
+  ///
+  /// Returns the invocation mode along with the arguments modified for that mode.
+  public static func invocationRunMode(
+    forArgs args: [String]
+  ) throws -> (mode: InvocationRunMode, args: [String]) {
+
+    assert(!args.isEmpty)
+
+    let execName = try VirtualPath(path: args[0]).basenameWithoutExt
+
+    // If we are not run as 'swift' or there are no program arguments, always invoke as normal.
+    guard execName == "swift", args.count > 1 else { return (.normal(isRepl: false), args) }
+
+    // Otherwise, we have a program argument.
+    let firstArg = args[1]
+
+    // If it looks like an option or a path, then invoke in interactive mode with the arguments as given.
+    if firstArg.hasPrefix("-") || firstArg.hasPrefix("/") || firstArg.contains(".") {
+        return (.normal(isRepl: false), args)
+    }
+
+    // Otherwise, we should have some sort of subcommand.
+
+    var updatedArgs = args
+
+    // If it is the "built-in" 'repl', then use the normal driver.
+    if firstArg == "repl" {
+        updatedArgs.remove(at: 1)
+        return (.normal(isRepl: true), updatedArgs)
+    }
+
+    let subcommand = "swift-\(firstArg)"
+
+    updatedArgs.replaceSubrange(0...1, with: [subcommand])
+
+    return (.subcommand(subcommand), updatedArgs)
   }
 }
 
--- a/Sources/SwiftDriver/Driver/OutputFileMap.swift
+++ b/Sources/SwiftDriver/Driver/OutputFileMap.swift
@@ -31,21 +31,12 @@ public struct OutputFileMap: Equatable {
       return output
     }
 
-    // Create a temporary file
-    let baseName: String
-    switch inputFile {
-    case .absolute(let path):
-      baseName = path.basenameWithoutExt
-    case .relative(let path), .temporary(let path):
-      baseName = path.basenameWithoutExt
-    case .standardInput:
-      baseName = ""
-    case .standardOutput:
+    if inputFile == .standardOutput {
       fatalError("Standard output cannot be an input file")
     }
 
     // Form the virtual path.
-    return .temporary(RelativePath(baseName.appendingFileTypeExtension(outputType)))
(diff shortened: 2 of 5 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -193,11 +193,30 @@ public struct Driver {
...
+      }
+    case .subcommand(let subcommand):
+
+      args.remove(at: 1)
+
+      // We are running as a subcommand, try to find the subcommand
+      // adjacent to the executable we are running as.
+      let subcommandPath = Process.findExecutable(args[0])?.parentDirectory.appending(component: subcommand)
+                           // If we didn't find the tool there, let the OS search for it.
+                           // If there was an error, ignore it and just let the exec fail.
```
DougGregor: Hmm. I would rather fail with a specific diagnostic (e.g., `cannot find subcommand executable 'swift-blah'`) than have the exec below fail.

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -79,6 +79,18 @@ public enum VirtualPath: Hashable {
     default: return nil
     }
   }
+
+  /// Retrieve the basename of the path without the extension.
+  public var basenameWithoutExt: String {
```
DougGregor: `OutputFileMap.getOutput(inputFile:outputType:)` has almost exactly this code inlined; can you update that method to use this property, rather than having two copies of the code?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -193,11 +193,30 @@ public struct Driver {
...
+                           ?? Process.findExecutable(subcommand)
+
+      // Execute the subcommand.
+      try exec(path: subcommandPath?.pathString ?? "", args: Array(args))
```
DougGregor: We shouldn't `exec` immediately, because the job of `Driver` is to come up with the set of jobs that someone else will execute. Ideally, when we're going to execute a subcommand, we'd have `planBuild` produce a `SubcommandJob` capturing all of the command-line arguments. Off hand, I don't know of a good way to do that.

DougGregor: Happy to chat through ideas, though! We could possibly stash the command-line arguments somewhere, then replace them with some "safe" set of command-line arguments so that the rest of the initializer can complete.

sjavora: What about delaying this until after the init returns? Say, `planBuild` itself. Though by that point we only have access to `ParsedOptions`, not the raw arguments. Do you think this is a way to go?

harlanhaskins: Is this the job of the Driver library? Or should the `swift-driver` command line tool handle subcommands and only create a `Driver` if there's no subcommand to execute (or it's loading the frontend directly)?

DougGregor: I think the driver should understand these command lines (the Driver initializer could throw if it sees one), and export and API to categorize command lines as driver commands vs. sub commands. The Swift-driver executable can use that API to determine what to do

Sources/SwiftDriver/Driver/Driver.swift
sjavora: I moved this to `main.swift`. Since it's outside of the driver now, I just used a `fatalError` instead.

Sources/SwiftDriver/Utilities/VirtualPath.swift
sjavora: Done.

Sources/SwiftDriver/Driver/Driver.swift
sjavora: Ok, I moved this to `main.swift`. If `Driver` is initialized with a subcommand, it now produces a diagnostic.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -351,6 +355,48 @@ public struct Driver {
...
 }
 
+extension Driver {
+
+  public enum InvocationRunMode: Equatable {
+    case normal(isRepl: Bool)
+    case subcommand(String)
+  }
+
+  public static func invocationRunMode(forArgs args: [String]) throws -> InvocationRunMode {
```
DougGregor: Please add a documentation comment for this API.

Sources/SwiftDriver/Driver/OutputFileMap.swift
```diff
@@ -31,21 +31,12 @@ public struct OutputFileMap: Equatable {
...
-    case .standardInput:
-      baseName = ""
-    case .standardOutput:
+    if inputFile == .standardOutput {
       fatalError("Standard output cannot be an input file")
     }
 
     // Form the virtual path.
-    return .temporary(RelativePath(baseName.appendingFileTypeExtension(outputType)))
+    return .temporary(RelativePath(inputFile.basenameWithoutExt.appendingFileTypeExtension(outputType)))
```
DougGregor: Thanks!

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -351,6 +355,48 @@ public struct Driver {
   }
 }
 
+extension Driver {
+
+  public enum InvocationRunMode: Equatable {
+    case normal(isRepl: Bool)
+    case subcommand(String)
```
DougGregor: I like this factoring. Should we handle the basic transformation of the command line (e.g., removing `repl`, replacing the first argument with the subcommand executable name, etc.) so that the clients don't need to perform the conditional `arguments.remove(at: 1)` themselves? The contract feels very implicit.

Sources/swift-driver/main.swift
```diff
@@ -23,7 +23,34 @@ do {
     processSet.terminate()
   }
 
-  var driver = try Driver(args: CommandLine.arguments)
+  var arguments = CommandLine.arguments
+
+  switch try Driver.invocationRunMode(forArgs: arguments) {
```
DougGregor: This definitely feels like the right structure for the driver, thank you!

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -193,11 +193,15 @@ public struct Driver {
     diagnosticsHandler: @escaping DiagnosticsEngine.DiagnosticsHandler = Driver.stderrDiagnosticsHandler
   ) throws {
     self.env = env
-    
-    // FIXME: Determine if we should run as subcommand.
 
     self.diagnosticEngine = DiagnosticsEngine(handlers: [diagnosticsHandler])
+
+    if case .subcommand = try Self.invocationRunMode(forArgs: args) {
+      diagnosticEngine.emit(.error_subcommand_in_driver)
```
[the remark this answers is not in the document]
sjavora: Sure thing, but then we can't check the exact error type in the tests, since `Driver.Error` is internal - should it be made public?

Alternatively, we could add descriptions to the errors and match those, but I tried overriding `localizedDescription` and it was not called for some reason.

Sources/SwiftDriver/Driver/Driver.swift
sjavora: Do you mean that the method should return the adjusted arguments in addition to the mode? I suppose the arguments could also be passed as `inout` so they could be modified in place, but then the arguments would get modified twice - first before the driver is initialized and again in the init itself.

sjavora: Decided to return a tuple instead - it seems easier to reason about.

harlanhaskins:
@swift-ci please test

sjavora:
@DougGregor is there anything else I should change here?

Sources/SwiftDriver/Driver/Driver.swift
DougGregor: I like it, thanks!

PR #29, opened 2019-10-28
[Docs] Add links to existing driver documentation

sjavora:
Adds links to existing driver docs in the docs directory.

I just found these docs by accident while looking something up in the old driver. I think it would be helpful to add them here so that new contributors can get a better idea about how the driver works. It certainly helped me.


harlanhaskins:
Thanks! @swift-ci please test

DougGregor:
Thanks!

PR #30, opened 2019-10-28
[Parseable output] Implement SignalledMessage

sjavora:
Implements `SignalledMessage` as another kind of `ParsableMessage`.

### DIFF
--- a/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
+++ b/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
@@ -11,6 +11,17 @@
 //===----------------------------------------------------------------------===//
 import TSCBasic
 
+#if canImport(Darwin)
+import Darwin.C
+#elseif os(Windows)
+import MSVCRT
+import WinSDK
+#elseif canImport(Glibc)
+import Glibc
+#else
+#error("Missing libc or equivalent")
+#endif
+
 /// Delegate for printing execution information on the command-line.
 public struct ToolExecutionDelegate: JobExecutorDelegate {
   public enum Mode {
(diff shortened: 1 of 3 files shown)


Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
```diff
@@ -58,18 +58,19 @@ public struct ToolExecutionDelegate: JobExecutorDelegate {
...
-        let message = ParsableMessage.finishedMessage(name: job.kind.rawValue, msg: finishedMessage)
-        emit(message)
+        let finishedMessage = FinishedMessage(exitStatus: Int(code), pid: pid, output: output)
+        message = ParsableMessage.finishedMessage(name: job.kind.rawValue, msg: finishedMessage)
 
-      case .signalled:
-        // FIXME: Implement this.
-        break
+      case .signalled(let signal):
+        let signalledMessage = SignalledMessage(pid: pid, output: output, errorMessage: output ?? "", signal: Int(signal))
```
sjavora: Is this right? I tried looking through the existing driver, but got lost somewhere around [here](https://github.com/apple/swift/blob/632399e0f8345e9c28d03b26252b5a30078489fc/lib/Driver/Compilation.cpp#L1131). I couldn't find any other errors in the `result`.

DougGregor: It looks like the existing driver is using [`strsignal`](http://man7.org/linux/man-pages/man3/strsignal.3.html) to get a string representation of the error message.

sjavora: Ah I see it now in TaskQueue. Implemented! Hope those new imports are correct.

sjavora:
> have you tried this out through a consumer of parsable output, like Xcode?

I've gotten this output: `error: Illegal instruction: 4 (in target 'TSCLibc' from project 'swift-driver')`

It seems consistent with the existing driver.

DougGregor:
> > have you tried this out through a consumer of parsable output, like Xcode?
> 
> I've gotten this output: `error: Illegal instruction: 4 (in target 'TSCLibc' from project 'swift-driver')`
> 
> It seems consistent with the existing driver.

Great! Before your PR, it would silently fail in Xcode with no message.

DougGregor:
@swift-ci please test

PR #31, opened 2019-11-03
NFC: Add missing copyright headers.

rxwei:
Add missing copyright headers in Options/Options.swift and Utilities/PrefixTrie.swift.

### DIFF
--- a/Sources/SwiftDriver/Options/Options.swift
+++ b/Sources/SwiftDriver/Options/Options.swift
@@ -1,3 +1,14 @@
+//===--------------- Options.swift - Swift Driver Options -----------------===//
+//
+// This source file is part of the Swift.org open source project
+//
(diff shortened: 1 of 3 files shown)


Sources/SwiftDriver/Options/Options.swift
```diff
@@ -1,3 +1,14 @@
+//===--------------- Options.swift - Swift Driver Options -----------------===//
```
DougGregor: Ah, this is generated source. We'll need to update [`makeOptions.cpp`](https://github.com/apple/swift-driver/blob/master/Sources/makeOptions/makeOptions.cpp) as well

rxwei: Done.

DougGregor:
@swift-ci please test

PR #32, opened 2019-11-05
Replace 'ParsableMessage' factory methods with an initializer.

rxwei:
This PR uses an initializer `init(name:kind:)` to unify multiple factory methods.

### DIFF
--- a/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
+++ b/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
@@ -54,7 +54,7 @@ public struct ToolExecutionDelegate: JobExecutorDelegate {
         commandArguments: arguments[1...].map{ String($0) }
       )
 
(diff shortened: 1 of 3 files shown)


harlanhaskins:
What if we changed each of these methods to drop the redundant type name, move the `message:` parameter first, then drop the redundant label there too?

```
ParsableMessage.signalled(signalledMessage, named: job.name.rawValue)
```

rxwei:
I like that idea.  But after giving some more thoughts to this, I think it might be better to drop these static methods and just use initializers instead, since it isn't more complicated:
```swift
ParsableMessage(name: job.kind.rawValue, kind: .signalled(message))
```

What do you think?

harlanhaskins:
That sounds great to me.

rxwei:
Done.

harlanhaskins:
@swift-ci please test

rxwei:
@swift-ci please test

PR #34, opened 2019-11-07
[Driver] Clean up platform minimum flag generation code

anayini:
The DarwinPlatform code provides a few derived vars
to determine the library name suffixes and platform
names based on the type of DarwinPlatform.  This commit
moves some code to determine platform minimum version
flags into that type.


anayini:
Honestly not even sure this is preferred, but saw that there were some derived vars on `DarwinPlatform` already so moved this to use the same pattern.

Also just a small test PR to figure out contribution workflow

harlanhaskins:
Thanks for your contribution, @anayini! Unfortunately, I think I prefer it the way it was -- these flags are very specific to Apple's linker on Darwin, so it doesn't really make much sense to move those flags which won't be used anywhere else. Maybe it would be good to move these into a helper function?

PR #35, opened 2019-11-15
Handle all FileTypes in OutputFileMap

sjavora:
This replaces the specific properties in `Entry` with a dictionary keyed by `FileType` and adds a custom implementation of `Codable`.

Originally, I wanted to just use `FileType`'s `description` property, but some of the cases just return `rawValue`, which is not compatible in all cases ("o" vs "object", in particular). Instead I added the `name` property with values from [FileTypes.def](https://github.com/apple/swift/blob/master/include/swift/Basic/FileTypes.def). While there, I noticed that `FileType` doesn't currently have a case for `swiftsourceinfo`, so I added that as well.

### DIFF
--- a/Sources/SwiftDriver/Utilities/FileType.swift
+++ b/Sources/SwiftDriver/Utilities/FileType.swift
@@ -12,7 +12,7 @@
 
 /// Describes the kinds of files that the driver understands.
 ///
-/// The raw values for these enumerations describe the default extension for]
+/// The raw values for these enumerations describe the default extension for
 /// the file type.
 public enum FileType: String, Hashable, CaseIterable, Codable {
   /// Swift source file.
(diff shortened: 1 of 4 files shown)


sjavora:
Hey @DougGregor - could someone please take a look at this? I'd request a review, but the button doesn't seem to be there...

Sources/SwiftDriver/Utilities/FileType.swift
```diff
@@ -163,8 +166,81 @@ extension FileType {
...
+         .swiftDeps, .moduleTrace, .tbd, .optimizationRecord, .swiftInterface, .swiftSourceInfoFile:
       return false
     }
   }
 }
+
+extension FileType {
+
+  init?(name: String) {
+    guard let type = FileType.allCases.first(where: { $0.name == name }) else { return nil }
```
harlanhaskins: Is there a way we can make this not O(n)?

harlanhaskins:
The approach looks good here, I've got one small nitpick though.

Sources/SwiftDriver/Utilities/FileType.swift
sjavora: We could have something like `static let typesByName = Dictionary(uniqueKeysWithValues: FileType.allCases.map { ($0.name, $0) })` which would only enumerate once and then use the dictionary in the initializer. Would that be OK?

harlanhaskins: Sounds good.

sjavora: Ok, pushed.

harlanhaskins:
@swift-ci please test

harlanhaskins:
@swift-ci please test

harlanhaskins:
Thanks, @sjavora!

PR #36, opened 2019-12-14
Add a bare-bones implementation of immediate mode

owenv:
This is an initial implementation of immediate mode. The big missing piece right now is setting [DY]LD_LIBRARY_PATH and DYLD_FRAMEWORK_PATH when needed. Before implementing that I think it's worth figuring out how the driver should run single job build plans for immediate mode and the REPL. I'm not sure if we should bypass the JobExecutor or not when executing in-place. If anyone has a strong opinion I could implement that in this PR or a follow up. I haven't made any changes yet so right now this just spawns a new process.


DougGregor:
@swift-ci please test

PR #37, opened 2019-12-15
Shuffle batch mode partitions if a seed is provided

owenv:
Also, add a `PredictableRandomNumberGenerator` to ensure deterministic partitioning. `PredictableRandomNumberGenerator` should maybe move into tools-support-core at some point, but I think it's fine where it is for now. It's not a great random number generator, but it should work fine for the driver's use cases.


DougGregor:
@swift-ci please test

PR #38, opened 2019-12-15
Update options list to pick up new dependencies, sanitizer, etc. options

owenv:
based on apple/swift 6cc1337f71c8e90a60a8cfa7c106132fb9fee202

I added a few of them to the list of common frontend options. The only one I'm unsure about is `-sanitize-recover`. The c++ driver only adds the last `-sanitize=` argument, but right now the swift version adds all that appear. I'm not familiar enough with the sanitizers to know which to copy.


DougGregor:
@swift-ci please test

PR #40, opened 2019-12-16
Allow specifying custom environment variables for jobs

owenv:
... and use it to set the right ones in immediate mode.


DougGregor:
@swift-ci please test

PR #42, opened 2019-12-17
Fix direct invocation of non-compiler driver kinds

owenv:
This is enough to fix ~2/3 of the failures in the interpreter and stdlib lit test suites, making it easier to focus on the driver-related failures.


DougGregor:
@swift-ci please test

PR #44, opened 2019-12-19
Miscellaneous improvements to -help messages

owenv:
- Add "See also" messages from the c++ driver on master
- Hide options not accepted by the current driver kind
- ~~update isAccepted(by:) to reject no-driver options in batch & interactive mode~~

### DIFF
--- a/Sources/SwiftDriver/Options/OptionTable.swift
+++ b/Sources/SwiftDriver/Options/OptionTable.swift
@@ -26,11 +26,11 @@ public struct OptionTable {
 
 extension OptionTable {
   /// Print help information to the terminal.
-  public func printHelp(usage: String, title: String, includeHidden: Bool) {
+  public func printHelp(driverKind: DriverKind, includeHidden: Bool) {
     print("""
-      OVERVIEW: \(title)
+      OVERVIEW: \(driverKind.title)
 
-      USAGE: \(usage)
+      USAGE: \(driverKind.usage)
 
       OPTIONS:
       """)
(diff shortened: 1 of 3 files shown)


harlanhaskins:
@swift-ci please test

owenv:
Oops, I missed that failure somehow when running tests locally. It looks like `-emit-module-doc-path` is listed as a no-driver option right now. I might need to change that in the swift repo and then rerun `makeOptions` before this can land.

owenv:
I'm dropping the `isAccepted(by:)` change for now, it looks like it might require a larger refactor. This just includes the first two changes now and is passing tests for me locally.

harlanhaskins:
@swift-ci please test

Sources/SwiftDriver/Options/OptionTable.swift
```diff
@@ -26,18 +26,19 @@ public struct OptionTable {
...
-      USAGE: \(usage)
+      USAGE: \(driverKind.usage)
 
       OPTIONS:
       """)
 
     for option in options {
       if option.isAlias { continue }
       if option.isHelpHidden && !includeHidden { continue }
+      guard option.isAccepted(by: driverKind) else { continue }
```
harlanhaskins: Is this still meant to be in?

owenv: Yeah, I think this is ok. As far as I can tell the current version of the acceptance check only fails to reject some invalid options, it doesn’t reject any valid ones.

DougGregor:
@swift-ci please test

PR #45, opened 2019-12-22
Infer module output path from '-o' where applicable

owenv:
We're still not taking the file output map or primary outputs into consideration, but this is enough to get a few more integration tests up and running.


DougGregor:
@swift-ci please test

PR #46, opened 2019-12-29
Implement -verify-debug-info

owenv:
The relevant lit tests don't pass yet, but they should once module name inference is fully implemented.


DougGregor:
@swift-ci please test

PR #47, opened 2019-12-30
[WIP] Invoke the Swift frontend to prescan for dependencies.

DougGregor:
Add a new flag `-driver-prebuild-module-dependencies` to instruct the driver
to have the Swift frontend prescan for dependencies, which will allow the
driver to explicitly schedule builds for both Swift and C modules,
eliminating the need for the frontend to rebuild those modules when it
tries to import them. This allows for an overall more coordinated
build.

At present, this WIP invokes the Swift frontend and then parses the
JSON before dumping out the dependency graph, so it's not actually
useful for anything.


DougGregor:
This utilizes the fast dependency scanning work in the Swift compiler itself (https://github.com/apple/swift/pull/28515), which is also a work in progress.

DougGregor:
@artemcm is doing this properly in https://github.com/apple/swift-driver/pull/104, so I'll close out this pull request.

PR #48, opened 2020-01-02
Add support for Clang module emission via -emit-pcm

DougGregor:
Teach the driver to emit Clang modules based on the with `-emit-pcm` command-line argument, where the input is the module map.

### DIFF
--- a/Sources/SwiftDriver/Driver/CompilerMode.swift
+++ b/Sources/SwiftDriver/Driver/CompilerMode.swift
@@ -26,6 +26,9 @@ public enum CompilerMode: Equatable {
 
   /// Compile and execute the inputs immediately.
   case immediate
+
+  /// Compile a Clang module (.pcm).
+  case compilePcm
 }
 
 /// Information about batch mode, which is used to determine how to form
@@ -40,7 +43,7 @@ extension CompilerMode {
   /// Whether this compilation mode uses -primary-file to specify its inputs.
   public var usesPrimaryFileInputs: Bool {
     switch self {
-    case .immediate, .repl, .singleCompile:
+    case .immediate, .repl, .singleCompile, .compilePcm:
       return false
 
     case .standardCompile, .batchCompile:
@@ -54,7 +57,7 @@ extension CompilerMode {
     case .immediate, .repl, .standardCompile, .batchCompile:
       return false
 
-    case .singleCompile:
+    case .singleCompile, .compilePcm:
       return true
     }
   }
@@ -62,18 +65,19 @@ extension CompilerMode {
 
 extension CompilerMode: CustomStringConvertible {
     public var description: String {
-        switch self {
-
-        case .standardCompile:
-            return "standard compilation"
-        case .batchCompile:
-            return "batch compilation"
-        case .singleCompile:
-            return "whole module optimization"
-        case .repl:
-            return "read-eval-print-loop compilation"
-        case .immediate:
-            return "immediate compilation"
-        }
+      switch self {
+      case .standardCompile:
+        return "standard compilation"
+      case .batchCompile:
+        return "batch compilation"
+      case .singleCompile:
+        return "whole module optimization"
+      case .repl:
+        return "read-eval-print-loop compilation"
+      case .immediate:
+        return "immediate compilation"
+      case .compilePcm:
+        return "compile Clang module (.pcm)"
+      }
   }
 }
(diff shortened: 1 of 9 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@allevato you might be interested in this

DougGregor:
@cltnschlosser mind reviewing, since you seem to also be hacking on this now?

Sources/SwiftDriver/Driver/CompilerMode.swift
```diff
@@ -26,6 +26,9 @@ public enum CompilerMode: Equatable {
 
   /// Compile and execute the inputs immediately.
   case immediate
+
+  /// Compile a Clang module (.pcm).
+  case compilePcm
```
brentdax: Do we want this to be `compilePcm` or `compilePCM`? (I know we don’t have a choice on `emitPcm`, but we have one here.)

DougGregor: Ah, I suppose it should be uppercase `compilePCM`. Thanks!

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -21,13 +21,29 @@ fileprivate func shouldColorDiagnostics() -> Bool {
...
+  /// How the bridging header should be handled.
+  enum BridgingHeaderHandling {
+    /// Ignore the bridging header entirely.
+    case ignored
+
+    /// Parse the bridging header, even if other jobs will use a precompiled
+    /// bridging header.
+    ///
+    /// This is typically used only when precompiling the bridging header.
+    case parsed
```
cltnschlosser: On further inspection, the emit-pch job doesn't even use `-import-objc-header`. So parsed isn't needed. I'll have another PR soon to fix up issues I'm finding while doing integration testing.

I like this enum though.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -156,9 +172,11 @@ extension Driver {
...
-    if let importedObjCHeader = importedObjCHeader {
+    if let importedObjCHeader = importedObjCHeader,
+        bridgingHeaderHandling != .ignored {
       commandLine.appendFlag(.importObjcHeader)
-      if requestPrecompiledObjCHeader, let pch = bridgingPrecompiledHeader {
+      if bridgingHeaderHandling == .precompiled,
+          let pch = bridgingPrecompiledHeader {
         if parsedOptions.contains(.pchOutputDir) {
           commandLine.appendPath(importedObjCHeader)
           switch compilerMode {
```
cltnschlosser: Would this whole switch be better replaced with:
```
if !compilerMode.isSingleCompilation {
  commandLine.appendFlag(.pchDisableValidation)
}
```

Sources/SwiftDriver/Jobs/GeneratePCMJob.swift
```diff
@@ -0,0 +1,63 @@
...
+    commandLine.appendPath(input.file)
+
+    // Compute the output file.
+    let output: TypedVirtualPath
+    if let outputArg = parsedOptions.getLastArgument(.o) {
+      output = .init(file: try VirtualPath(path: outputArg.asSingle),
+                     type: .pcm)
+    } else {
+      output = .init(
+        file: try VirtualPath(path: moduleName + "." + FileType.pcm.name),
```
cltnschlosser: Could use this extension (I guess this is probably faster since it doesn't do an isEmpty check).
```
  /// Append the extension for the given file type to the string.
  func appendingFileTypeExtension(_ type: FileType) -> String {
    let ext = type.rawValue
    if ext.isEmpty { return self }

    return self + "." + ext
  }
```

Also you are using name, which happens to be the same as the extension in this case, but the rawValue seems to be more accurate for all FileTypes.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
DougGregor: Yes, thanks!

Sources/SwiftDriver/Jobs/GeneratePCMJob.swift
DougGregor: Good point, thanks!

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #50, opened 2020-01-02
Respect -driver-use-frontend-path throughout job creation

DougGregor:
A number of job-creation functions were asking the toolchain for the frontend path, rather than using the frontend path computed by the `Driver` instance. Fix them, and thread the Driver-computed Swift frontend path everywhere. We might want to change the way things are layered here in the future.

### DIFF
--- a/Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
@@ -17,6 +17,7 @@ extension Toolchain {
   func computeResourceDirPath(
     for triple: Triple,
     parsedOptions: inout ParsedOptions,
+    swiftCompiler: AbsolutePath,
     isShared: Bool
   ) throws -> AbsolutePath {
     // FIXME: This almost certainly won't be an absolute path in practice...
@@ -29,8 +30,7 @@ extension Toolchain {
         .appending(components: "usr", "lib",
                    isShared ? "swift" : "swift_static")
     } else {
-      // FIXME: Thread the driver's swiftCompiler path here.
-      resourceDirBase = try getToolPath(.swiftCompiler)
+      resourceDirBase = swiftCompiler
         .parentDirectory // remove /swift
         .parentDirectory // remove /bin
         .appending(components: "lib", isShared ? "swift" : "swift_static")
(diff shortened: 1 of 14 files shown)


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -66,17 +67,19 @@ public protocol Toolchain {
...
 }
 
 extension Toolchain {
   public var searchPaths: [AbsolutePath] {
     getEnvSearchPaths(pathString: env["PATH"], currentWorkingDirectory: localFileSystem.currentWorkingDirectory)
   }
   
-  public func swiftCompilerVersion() throws -> String {
+  public func swiftCompilerVersion(_ swiftCompiler: AbsolutePath) throws -> String {
+    // FIXME: Thread the driver's swiftCompiler path here.
```
cltnschlosser: Can delete this now.

Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
```diff
@@ -17,6 +17,7 @@ extension Toolchain {
   func computeResourceDirPath(
     for triple: Triple,
     parsedOptions: inout ParsedOptions,
+    swiftCompiler: AbsolutePath,
```
brentdax: Should these `Toolchain` extension methods still be on `Toolchain`? They used to call `getToolPath(.swiftCompiler)`, but now that they’re taking `swiftCompiler` as an input, I don’t think they’re using any requirements. But I could be missing something, and even if I’m not, it’s reasonable to wait for a different PR.

DougGregor: It seems plausible that different toolchains could have different locations for the resource directory, so I feel like `Toolchain` needs to be involved here... but I don't have a good sense for how this should be architected. Perhaps all tools should be overridable in some generalized manner and we either register that with the toolchain or pass through some dictionary of overrides?

DougGregor: I'd like to tackle the fix-the-architecture question a bit later, because this unblocks more testing.

brentdax: Overridable tools make sense to me, but what you've got here is severable from that!

Sources/SwiftDriver/Toolchains/Toolchain.swift
DougGregor: Good point, removed!

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test platform macOS

DougGregor:
@swift-ci please test platform macOS

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test macOS

PR #52, opened 2020-01-02
Parse endian-dependent BPF architecture.

sjavora:
Taken from [LLVM](https://github.com/apple/swift-llvm/blob/7ade6a3e01d02275bcc9ee5a39142b5c9269d108/lib/Support/Triple.cpp#L248).


Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -574,10 +574,12 @@ extension Triple {
     }
 
     private static func parseBPFArch<S: StringProtocol>(_ archName: S) -> Triple.Arch? {
+
+      let isLittleEndianHost = 1.littleEndian == 1
```
owenv: Depending on what stance the project takes on using private language features, we could use `#if _endian(little)` here. The driver already uses `@_exported` and `@_implementationOnly` elsewhere, but on the other hand it might make a bootstrapped build more difficult in the future.

DougGregor: We'd prefer to use the public language if we can, so let's stick with this.

DougGregor:
@swift-ci please test

PR #53, opened 2020-01-02
Move command line utilities from CompileJob.swift to CommandLineArguments.swift

owenv:
These are used across many different jobs now and generalize well, so I think it makes sense to put them in their own file.


DougGregor:
@swift-ci please test

PR #55, opened 2020-01-02
Improved ARM arch parsing

owenv:
A couple of notes:
- thumb handling still isn't very good, I left a FIXME in the source
- there might be some opportunities to further improve the test coverage, I'm not very familiar with all of the different manufacturer variants

With this change, I was able to fix a few issues in the Driver integration test suite in addition to the newly enabled Triple tests.


DougGregor:
@swift-ci please test

PR #58, opened 2020-01-11
Forward -print-target-info to the frontend, begin refactoring planning for simple driver invocations

owenv:
The `-print-target-info` implementation itself is pretty straightforward. The refactor is a little more interesting, I pulled the `planBuild` call out of `Driver.run`. Now, all clients are expected to call `planBuild` for a list of jobs they can hand off to `Driver.run` or schedule themselves. I also pulled `-version` and `-print-target-info` handling out of `run` and into the build planning stage. This gives clients more control over executing them in case they want to do something like capture the output. It's not yet clear if we want to do something similar for printing help or forwarding to `swift-indent`/`swift-autolink-extract`, I've left that logic alone for now.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -551,6 +551,35 @@ extension Driver {
     }
   }
 
+  /// Create a job if needed for simple requests that can be immediately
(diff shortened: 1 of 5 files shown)


DougGregor:
@swift-ci please test

PR #60, opened 2020-01-18
[Triple] Add SubArchs, add validation for thumb architectures

sjavora:
This fills the (currently empty) Triple.SubArch enum based on LLVM and resolves the thumb architectures validation FIXME.

This also meant commented lines in tests related to `subArch`s could now be uncommented.


Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -682,8 +793,174 @@ extension Triple {
...
+    case armSubArch_v8
+    case armSubArch_v8_1a
+    case armSubArch_v8_1m_mainline
+    case armSubArch_v8_2a
+    case armSubArch_v8_3a
+    case armSubArch_v8_4a
+    case armSubArch_v8_5a
+    case armSubArch_v8m_baseline
+    case armSubArch_v8m_mainline
+    case armSubArch_v8r
```
sjavora: Not sure if all of these are needed.

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -682,8 +793,174 @@ extension Triple {
...
+           .armSubArch_v4, .armSubArch_v4t, .armSubArch_v5, .armSubArch_v5e,
+           .armSubArch_v6, .armSubArch_v6k, .armSubArch_v6kz, .armSubArch_v6t2,
+           .armSubArch_v7s,
+           .kalimbaSubArch_v3, .kalimbaSubArch_v4, .kalimbaSubArch_v5,
+           .mipsSubArch_r6:
+        return nil
+      }
+    }
+
+    var armVersion: Int? {
```
sjavora: I'm not exactly thrilled about having these properties here, since they always return `nil` on non-ARM architectures. Would it make sense to have sub-archs nested in the archs enum as associated values? That way, each architecture could have its own sub-archs and these could be moved to the ARM enum.

sjavora:
`Triple.swift` is now over 1500 lines - would it make sense to extract some code to separate files? If so, which parts?

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -603,13 +603,41 @@ extension Triple {
...
+            self = .r
+          case .armSubArch_v7, .armSubArch_v7ve, .armSubArch_v7k,
+               .armSubArch_v8, .armSubArch_v8_1a, .armSubArch_v8_2a,
+               .armSubArch_v8_3a, .armSubArch_v8_4a, .armSubArch_v8_5a:
+            self = .a
+          case .armSubArch_v2, .armSubArch_v2a, .armSubArch_v3, .armSubArch_v3m,
+               .armSubArch_v4, .armSubArch_v4t, .armSubArch_v5, .armSubArch_v5e,
+               .armSubArch_v6, .armSubArch_v6k, .armSubArch_v6kz, .armSubArch_v6t2,
+               .armSubArch_v7s,
+               .kalimbaSubArch_v3, .kalimbaSubArch_v4, .kalimbaSubArch_v5,
```
owenv: This confused me for a moment because Kalimba and mips aren't ARM subarchs. Maybe those could go in their own case block with a comment?

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -682,8 +793,174 @@ extension Triple {
...
+        return .armSubArch_v8_4a
+      case "v8.5-a":
+        return .armSubArch_v8_5a
+      default:
+        return nil
+      }
+    }
+
+    var armProfile: Triple.Arch.ARMProfile? {
+      switch self {
```
owenv: Is there any way to share this switch with the `ARMProfile` intializer above? I think it'll be difficult to keep the two in sync.

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -682,8 +793,174 @@ extension Triple {
...
+    case armSubArch_v6m
+    case armSubArch_v6t2
+    case armSubArch_v7
+    case armSubArch_v7em
+    case armSubArch_v7k
+    case armSubArch_v7m
+    case armSubArch_v7r
+    case armSubArch_v7s
+    case armSubArch_v7ve
+    case armSubArch_v8
```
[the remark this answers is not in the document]
sjavora: Hmm, I used that exact file. As I understand it, on [line 46](https://github.com/llvm-mirror/llvm/blob/2c4ca6832fa6b306ee6a7010bfb80a3f2596f824/include/llvm/Support/ARMTargetParser.def#L46) the element definitions are given as
`ARM_ARCH(NAME, ID, CPU_ATTR, SUB_ARCH, ARCH_ATTR, ARCH_FPU, ARCH_BASE_EXT)`

So what I did was look at the first element (`NAME`) and then used the fourth element (`SUB_ARCH`) in the actual enum cases. For example v8 is taken from
```
ARM_ARCH("armv8-a", ARMV8A, "8-A", "v8", ARMBuildAttrs::CPUArch::v8_A,
         FK_CRYPTO_NEON_FP_ARMV8,
         (ARM::AEK_SEC | ARM::AEK_MP | ARM::AEK_VIRT | ARM::AEK_HWDIVARM |
          ARM::AEK_HWDIVTHUMB | ARM::AEK_DSP | ARM::AEK_CRC))
```
where `NAME` = `armv8-a` and `SUB_ARCH` = `v8`.

Was this approach incorrect? I'm unsure, because I'm not that familiar with the architectures.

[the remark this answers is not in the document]
owenv: Hmm, in that case I think this is ok then. I don’t have a ton of experience in this area either.

Sources/SwiftDriver/Utilities/Triple.swift
sjavora: I added nested enums to solve this.

Sources/SwiftDriver/Utilities/Triple.swift
sjavora: Thanks  - I didn't even realize I duplicated that - fixed.

sjavora:
@owenv could you please take another look when you have the time? I think I've managed to eliminate most of the nonsensical states.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #62, opened 2020-01-23
Move forwarding to non-compiler driver kinds into the planning stage

owenv:
This is a continuation of https://github.com/apple/swift-driver/pull/58 which moves forwarding into jobs created during build planning.

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -168,6 +168,20 @@ extension Driver {
   /// Create a job if needed for simple requests that can be immediately
   /// forwarded to the frontend.
   public mutating func immediateForwardingJob() throws -> Job? {
+
+    // We just need to invoke the corresponding tool if the kind isn't Swift compiler.
+    if !driverKind.isSwiftCompiler {
+      var commandLine: [Job.ArgTemplate] = driverKind.usageArgs
+      try commandLine.append(contentsOf: parsedOptions.allArguments())
+      return Job(kind: .forwarding,
+                 tool: .absolute(try toolchain.getToolPath(.swiftCompiler)),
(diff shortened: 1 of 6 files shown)


Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -168,6 +168,20 @@ extension Driver {
   /// Create a job if needed for simple requests that can be immediately
   /// forwarded to the frontend.
   public mutating func immediateForwardingJob() throws -> Job? {
+
+    // We just need to invoke the corresponding tool if the kind isn't Swift compiler.
+    if !driverKind.isSwiftCompiler {
+      var commandLine: [Job.ArgTemplate] = driverKind.usageArgs
+      try commandLine.append(contentsOf: parsedOptions.allArguments())
+      return Job(kind: .forwarding,
+                 tool: .absolute(try toolchain.getToolPath(.swiftCompiler)),
```
DougGregor: This is an interesting choice. In many of these cases, we're not actually executing the Swift compiler, but some other tool. Should we look up that tool by name, rather than having the "tool" effectively be ignored?

DougGregor:
@swift-ci please test

Sources/SwiftDriver/Jobs/Planning.swift
owenv: I think this is the way to go. It'll be a lot simpler once the compiler binary is renamed to `swift-frontend` though. That'll allow us to avoid needing to add `-frontend` or `-modulewrap` when invoking it with the right tool name. This PR isn't an important change, so I might just hold off on landing it until then.

DougGregor: Sounds like I really need to finish off that compiler binary rename. Thanks for the nudge ;)

PR #63, opened 2020-01-27
Allow relative paths in output file map

zienag:
This allows having relative paths on output file map json file. 
Also, I've made working directory to default to cwd if no `-working-directory` arg is supplied.
Also, there is small bug, if there is two identical inputs in output file map, driver would silently crash, as it uses `Dictionary(uniqueKeysWithValues:)` initializer, now it will choose last record (behaviour that usually seen in clang, for example), but is it better to have warning or error? There is some difficulty thought, as closure that handles duplicates doesn't have key as argument, so we need to find duplicates beforehand.

Also, it looks that there is a lot of dangling whitespaces, is it ok to remove them? May be e need some codestyle tool (swift-format?) that will make codestyle automated?

This is my first contribution to swift project, so any help would be very much appreciated :)

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -247,13 +247,19 @@ public struct Driver {
     let inputFiles = try Self.collectInputFiles(&self.parsedOptions)
     self.inputFiles = inputFiles
 
+    let outputFileMap: OutputFileMap?
     // Initialize an empty output file map, which will be populated when we start creating jobs.
     if let outputFileMapArg = parsedOptions.getLastArgument(.outputFileMap)?.asSingle {
       let path = try AbsolutePath(validating: outputFileMapArg)
-      self.outputFileMap = try .load(file: path, diagnosticEngine: diagnosticEngine)
+      outputFileMap = try .load(file: path, diagnosticEngine: diagnosticEngine)
+    } else {
+      outputFileMap = nil
     }
-    else {
-      self.outputFileMap = nil
+
+    if let workingDirectory = self.workingDirectory {
+      self.outputFileMap = outputFileMap?.resolveRelativePaths(relativeTo: workingDirectory)
+    } else {
+      self.outputFileMap = outputFileMap
     }
 
     // Determine the compilation mode.
--- a/Sources/SwiftDriver/Driver/OutputFileMap.swift
+++ b/Sources/SwiftDriver/Driver/OutputFileMap.swift
@@ -47,6 +47,25 @@ public struct OutputFileMap: Equatable {
     try! existingOutput(inputFile: VirtualPath(path: ""), outputType: outputType)
   }
 
+  public func resolveRelativePaths(relativeTo absPath: AbsolutePath) -> OutputFileMap {
+    let resolvedKeyValues: [(VirtualPath, [FileType : VirtualPath])] = entries.map {
+      let resolvedKey: VirtualPath
+      // Special case for single dependency record, leave it as is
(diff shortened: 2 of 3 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -46,7 +46,7 @@ public struct Driver {
     case subcommandPassedToDriver
     case relativeFrontendPath(String)
   }
-
```
owenv: Do you mind breaking these whitespace changes out into a separate PR? We can probably get those merged faster and they make this diff a little harder to read.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -233,10 +233,10 @@ public struct Driver {
     }
 
     // Compute the working directory.
+    let cwd = localFileSystem.currentWorkingDirectory
```
owenv: I'm not sure we necessarily want to set the cwd here if we're not overriding it. It doesn't look like the C++ driver applies the cwd to the inputs in that case, it just uses relative paths.

Sources/SwiftDriver/Driver/OutputFileMap.swift
```diff
@@ -186,3 +205,12 @@ extension String {
...
   }
 }
+
+extension VirtualPath {
+  fileprivate func resolvedRelativePath(base: AbsolutePath) -> VirtualPath {
+    guard case let .relative(relPath) = self else { return self }
+    return .absolute(.init(base, relPath))
+  }
+
+  fileprivate static let singleDep = VirtualPath.relative(.init(""))
```
owenv: This is only used once, so I think you can just inline it into the if statement above with a comment explaining what it means.

Sources/SwiftDriver/Driver/OutputFileMap.swift
```diff
@@ -156,9 +175,9 @@ fileprivate struct OutputFileMapJSON: Codable {
 
   /// Converts into virtual path entries.
   func toVirtualOutputFileMap() throws -> [VirtualPath : [FileType : VirtualPath]] {
-    Dictionary(uniqueKeysWithValues: try entries.map { input, entry in
+    Dictionary(try entries.map { input, entry in
       (try VirtualPath(path: input), try entry.paths.mapValues(VirtualPath.init(path:)))
-    })
+    }, uniquingKeysWith: { $1 })
```
owenv: This behavior looks good to me, and I think it's consistent with the C++ driver

Sources/SwiftDriver/Driver/Driver.swift
zienag: The thing that bothers me here, that you can override working directory with relative path, and if you override it with `.` like this `-working-directory=.` and not override it at all, code will take completely different paths.

Do you think that instead of making all arguments absolute we need to teach code to correctly handle relative paths?

Sources/SwiftDriver/Driver/Driver.swift
zienag: done https://github.com/apple/swift-driver/pull/70

Sources/SwiftDriver/Driver/Driver.swift
zienag: Removed this change for now, it works only when `-working-directory` supplied. To work relative to cwd, further changes needed to teach code to work with relative paths.

zienag:
@owenv thanks for reviewing!

DougGregor:
@swift-ci please test

PR #65, opened 2020-01-28
Diagnose when inputs are modified during the build

owenv:
The condition is checked in both cases, but only single-job build plans emit a diagnostic right now, I left a fixme in the multi-job test. For that, I need to decide if the `JobExecutor` should propagate errors through the build graph, or if the `Context`should just include a `DiagnosticEngine`.


DougGregor:
@swift-ci please test

PR #66, opened 2020-02-03
Introduce a new executable, swift-help, which prints help info for tools

owenv:
`swift-help` is extremely simple, but it allows -help handling to more easily fit into the driver's Job model. A new SwiftOptions module contains the option types and definitions which are shared with the driver.

Note that the driver-side test is `#if`-ed out right now because there's no good way for the driver to find `swift-help` or ensure it's built when running tests yet.

Based on the discussion from https://forums.swift.org/t/swift-driver-and-help-documentation-lookup/32881/9, I don't think we'll end up using this for documentation lookup in the future. However, since it's such a small patch I think it's worth it for the architectural benefits alone.


owenv:
@DougGregor No problem, looks like it was just an import conflict with one of my other PRs, should be fixed now. Thanks for the review!

Sources/swift-help/main.swift
```diff
@@ -0,0 +1,52 @@
...
+extension DriverKind: StringEnumArgument {
+  public static var completion: ShellCompletion { .none }
+}
+
+struct Options {
+  var driverKind: DriverKind = .interactive
+  var showHidden: Bool = false
+}
+
+var args = CommandLine.arguments.dropFirst()
```
cltnschlosser: This is unused.

owenv: good catch, I left this in by accident

owenv:
@swift-ci test

owenv:
Forgot to regenerate testing boilerplate

@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test macOS

PR #67, opened 2020-02-04
More accurate computation of compile job output paths

owenv:
With this change, we're down to 6 interpreter and 3 stdlib lit tests still failing. I'm not all that happy with this solution, it feels kind of fragile. I haven't come up with anything better though so I thought I'd put this up in case anyone had a better solution in mind.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -605,7 +605,7 @@ extension Driver {
       return
     }
 
-    if jobs.contains(where: { $0.requiresInPlaceExecution }) {
+    if jobs.contains(where: { $0.requiresInPlaceExecution }) || jobs.count == 1 {
       assert(jobs.count == 1, "Cannot execute in place for multi-job build plans")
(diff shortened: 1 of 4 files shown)


owenv:
There's probably still room for improvement here, but with a few updates we now can pass all but one stdlib test

cltnschlosser:
I don't fully understand the OutputFileMap, but could we add this to there instead of injecting a different path? or would that not work?

owenv:
@cltnschlosser it might work, I need to look into it some more. There are some tricky cases where the inputs that produce the specified outputs might be temporary files.

owenv:
Marking this as WIP until I have time to look over it again and come up with a better design

owenv:
I decided to move the output path computation down into compile job planning. This now aims to reproduce the logic from the cpp `getOutputFilename`, excluding anything that isn't applicable to compile jobs. This makes it significantly easier to understand, but could lead to duplication throughout the other jobs in the future. Using a top-of-tree swift checkout, it passes the entire stdlib suite, and all but 5 interpreter tests. I'm not sure my `isTopLevel` check is 100% correct though.

DougGregor:
@swift-ci please test

PR #68, opened 2020-02-06
Search for the default SDK path only if triple is macOSX

LucianoPAlmeida:
Addresses a minor fixme on getting default SDK Path only if triple is macosx.

cc @aciidb0mb3r


owenv:
Thanks for fixing this! Could you add a quick check for the `-sdk` flag if the triple is macOS in `testImmediateMode()` of `SwiftDriverTests`?

LucianoPAlmeida:
> Thanks for fixing this! Could you add a quick check for the `-sdk` flag if the triple is macOS in `testImmediateMode()` of `SwiftDriverTests`?

Sure @owenv Thanks 👍

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #69, opened 2020-02-06
Don't use a bridging pch in immediate mode

owenv:
This is consistent with the c++ driver and fixes `Interpreter/availability_host_os.swift`

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -291,6 +291,7 @@ public struct Driver {
 
     self.importedObjCHeader = try Self.computeImportedObjCHeader(&parsedOptions, compilerMode: compilerMode, diagnosticEngine: diagnosticEngine)
     self.bridgingPrecompiledHeader = try Self.computeBridgingPrecompiledHeader(&parsedOptions,
+                                                                               compilerMode: compilerMode,
                                                                                importedObjCHeader: importedObjCHeader,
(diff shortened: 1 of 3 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1384,9 +1385,11 @@ extension Driver {
 
   /// Compute the path of the generated bridging PCH for the Objective-C header.
   static func computeBridgingPrecompiledHeader(_ parsedOptions: inout ParsedOptions,
+                                               compilerMode: CompilerMode,
                                                importedObjCHeader: VirtualPath?,
                                                outputFileMap: OutputFileMap?) throws -> VirtualPath? {
-    guard let input = importedObjCHeader,
+    guard !(compilerMode == .immediate || compilerMode == .repl),
```
DougGregor: This might seem minor, but could you turn this into something involving a `switch` on the compiler mode, either in some separate function/property or an inline `switch`? I've found that checking for individual cases like this is error-prone when we add a new compiler mode, because we forget to update cases.

DougGregor:
@swift-ci please test

Sources/SwiftDriver/Driver/Driver.swift
owenv: 👍 , good point

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci test

PR #71, opened 2020-02-13
Make module naming consistent with the integrated driver when the output filename has 2+ extensions

owenv:
This caused a surprising number of tests to fail which wrote something like `-o a.out.optimized`. The C++ driver uses `llvm::sys::path::stem` to remove extensions, which would result in `a.out`, which is an invalid module name, so the fallback of `main` would be used. `swift-driver` used to remove both extensions, and would name the module `a`. This behavior is clearly better, but I think it's too late to change considering how many tests broke...


DougGregor:
@swift-ci please test

PR #73, opened 2020-02-14
Make swift driver friendlier to relative paths

zienag:
Add handling for relative paths in output file map path,
resonse file paths, recorded input modification dates.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -252,17 +252,15 @@ public struct Driver {
     self.inputFiles = inputFiles
     self.recordedInputModificationDates = .init(uniqueKeysWithValues:
       Set(inputFiles).compactMap {
-        if case .absolute(let absolutePath) = $0.file,
-          let modTime = try? localFileSystem.getFileInfo(absolutePath).modTime {
-          return ($0, modTime)
-        }
-        return nil
+        guard let modTime = try? localFileSystem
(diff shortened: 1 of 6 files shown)


owenv:
@swift-ci test

owenv:
@zienag sorry for taking so long to come back to this. It looks one of the response file tests is failing on CI

zienag:
> @zienag sorry for taking so long to come back to this. It looks one of the response file tests is failing on CI

Hmm, that is unfortunately non trivial bug: given arguments 
```
swiftc -Xlinker @SOMETHING
```
`expandResponseFiles` would think that `@SOMETHING` is response file passed to driver. Before my changes it tried to make absolute path from `SOMETHING`, failed and skipped that argument. Now it makes relative path from `SOMETHING` and tries to read its content. This behaviour is totally wrong, and likely deserves separate pr.

Do you know any code/place that works with `-Xblabla arg` arguments to ignore or filter them out in `expandResponseFiles`?

zienag:
Meanwhile, I'll remove changes associated with response file from this pr, as extensions I've added are needed by other prs

owenv:
@zienag unfortunately I don't have a good idea about how to handle -X... arguments with response files. This might cause problems when we eventually migrate to swift-argument-parser too. In the meantime, the rest of this change will still be really helpful though!

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
Looks like the checks are all passing now. Thanks @zienag !

PR #76, opened 2020-02-22
Fix batch mode one partition.

cltnschlosser:
Previously it would make n jobs (where n is number of inputs).


DougGregor:
@swift-ci please test

PR #77, opened 2020-02-24
Support VirtualPaths in ToolChain

cltnschlosser:
Fixes:
Swift(macosx-x86_64) :: Driver/driver-use-frontend.swift
Swift(macosx-x86_64) :: Driver/linker-rpath.swift

This should go in after #73 since it copies some extensions from there.


Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -211,6 +260,28 @@ extension VirtualPath {
...
+
+      switch name {
+      case "", ".":
+          return self
+      default:
+        return RelativePath(pathString + "/" + name)
+      }
+  }
+
+  /// Returns the absolute path with additional literal components appended.
```
zienag: I think there is a typo here, `Returns the _relative_ path`

cltnschlosser: Updated.

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -240,6 +311,22 @@ extension TSCBasic.FileSystem {
     }
   }
 
+  func exists(_ path: VirtualPath) -> Bool {
+    do {
+      return try resolvingVirtualPath(path, apply: exists)
```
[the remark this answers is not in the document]
cltnschlosser: Updated these to preconditionFailure for types other than absolute and relative.

cltnschlosser:
> Looks good!
> 
> Btw, I think it looks more and more obvious that `VirtualPath` is not a silver bullet for all path operations. Most of operations with them are correct only for absolute or relative path, I wonder if it can be made more robust, but introducing another layer like `VirtualPath.filesystemPath(.absolute(foobar))` is also not looking good. It is probably worth looking at use cases of stdin/out and temporary to see if there is some good way.
> Also, it's a bit unfortunate that `TCSBasic` doesn't have ways to do operations with relative paths without resolving it to absolute path beforehand :(

Going to take a look at stdin, stdout, and temporary usages. Maybe we introduce another type for the ToolChain that only has Absolute and Relative paths.

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -80,6 +80,36 @@ public enum VirtualPath: Hashable {
...
+  public var pathString: String {
+    switch self {
+    case .relative(let path):
+      return path.pathString
+    case .absolute(let path):
+      return path.pathString
+    case .standardInput, .standardOutput:
+      preconditionFailure("standardInput and standardOutput do not have a pathString")
+    case .temporary(let path):
+      return path.pathString
```
cltnschlosser, line 94: This probably isn't correct. Given that this relative path doesn't contain the temporary directory.

cltnschlosser:
@zienag Curious about your thoughts on https://github.com/apple/swift-driver/pull/83

Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
```diff
@@ -73,13 +72,13 @@ extension Toolchain {
...
     if let path = sdkPath {
-      let sdkPath = AbsolutePath(path)
+      let sdkPath = try VirtualPath(path: path)
       // If we added the secondary resource dir, we also need the iOSSupport directory.
       if secondaryResourceDir != nil {
         result.append(sdkPath.appending(components: "System", "iOSSupport", "usr", "lib", "swift"))
       }
 
-      result.append(sdkPath.appending(RelativePath("usr/lib/swift")))
+      result.append(try VirtualPath(path: path).appending(components: "usr", "lib", "swift"))
```
cltnschlosser, line 81:
```suggestion
      result.append(sdkPath.appending(components: "usr", "lib", "swift"))
```

shahmishal:
The Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please re-create pull request with `main` branch.

PR #78, opened 2020-03-01
5.2 Compatibility updates to reflect the toolchain symlink removals

owenv:
`swift-indent` hasn't been included in toolchains for a while now, so remove all references to it.

`swift-autolink-extract` was removed from macOS toolchains fairly recently, so #if-out the linux linker tests that rely on it on macOS.


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #79, opened 2020-03-01
Diagnose inputs modified during the build for multi-job plans

owenv:
The JobExecutor's context now holds a diagnostic engine which can be used to emit errors encountered when running jobs.


owenv:
@swift-ci test

PR #80, opened 2020-03-01
Windows triple normalization

sjavora:
Adds logic from `llvm::Triple::normalize`, which enables all the asserts from `testNormalizeWindows` as well as some other edge cases. 

Notably, `objectFormat` for some Windows triples still doesn't parse. As far as I can tell, this is because `Triple.EnvInfo` checks for both environment and object format in the fourth position, so in strings like `i686-pc-windows-msvc-elf`, the `elf` is just totally ignored.


Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -98,21 +98,64 @@ public struct Triple {
...
+      isMinGW32 = parser.componentsIndicateMinGW32
+
+      if
+      let parsedEnv = parsedEnv,
+      let environment = parsedEnv.value.environment,
+      environment == .android,
+      parsedEnv.substring.starts(with: "androideabi") {
+        let androidVersion = parsedEnv.substring.dropFirst("androideabi".count)
+
+        parser.components[3] = "android\(androidVersion)"
```
sjavora: There is no test for this - I didn't find one in `TripleTest.cpp` and I don't really know what a "valid" triple like this would look like.

DougGregor:
@swift-ci please test

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -98,21 +98,64 @@ public struct Triple {
...
-      // In particular, it doesn't correct Windows triples as well as
-      // llvm::Triple::normalize() does.
+      isCygwin = parser.componentsIndicateCygwin
+      isMinGW32 = parser.componentsIndicateMinGW32
+
+      if
+      let parsedEnv = parsedEnv,
+      let environment = parsedEnv.value.environment,
+      environment == .android,
+      parsedEnv.substring.starts(with: "androideabi") {
```
aciidb0mb3r: nit: I think you can simplify this if to:

```swift
if let parsedEnv = parsedEnv, parsedEnv.value.environment == .android, parsedEnv.hasPrefix("androideabi")
```

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -98,21 +98,64 @@ public struct Triple {
     var parsedOS = parser.match(OS.self, at: 2)
     var parsedEnv = parser.match(EnvInfo.self, at: 3)
 
+    var isCygwin = parser.componentsIndicateCygwin
+    var isMinGW32 = parser.componentsIndicateMinGW32
```
aciidb0mb3r: I think these are only used inside the `if normalizing` so you can make these `let` and declare the variables there.

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -178,6 +229,14 @@ fileprivate struct TripleParser {
   var components: [Substring]
   var isMatched: Set<Int> = []
 
+  var componentsIndicateCygwin: Bool {
+    components.count > 2 ? components[2].starts(with: "cygwin") : false
```
aciidb0mb3r: nit: use `hasPrefix` instead of `starts(with:`

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -178,6 +229,14 @@ fileprivate struct TripleParser {
...
+  }
+
+  var componentsIndicateMinGW32: Bool {
+    components.count > 2 ? components[2].starts(with: "mingw") : false
```
aciidb0mb3r: nit: use `hasPrefix` instead of `starts(with:`

Sources/SwiftDriver/Utilities/Triple.swift
sjavora: May I ask why? Both seem to be used, though `hasPrefix` clearly more so...

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -214,11 +273,11 @@ fileprivate struct TripleParser {
                  "Lost the parsed component somehow?")
 
     for i in unmatchedIndices {
-      guard let parsed = ParsedComponent(components[i], as: Value.self) else {
+      guard Value.valueIsValid(components[i]) else {
```
aciidb0mb3r: maybe I am missing something but what's the advantage of using valueIsValid over the previous code?

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -1558,3 +1631,15 @@ extension Triple {
     }
   }
 }
+
+extension Array {
```
aciidb0mb3r: Maybe make this extension fileprivate?

Sources/SwiftDriver/Utilities/Triple.swift
sjavora: For `Triple.OS`, value can be valid even if the component doesn't actually parse:
```swift
fileprivate static func valueIsValid(_ value: Substring) -> Bool {
  parse(value) != nil || value.hasPrefix("cygwin") || hasPrefix("mingw")
}
```

Sources/SwiftDriver/Utilities/Triple.swift
aciidb0mb3r: `starts(with:)` is a general purpose utility on Collection protocol and hasPrefix is for String. I don't think it matters in practice but hasPrefix is more idiomatic Swift (at least IMO 😄).

Sources/SwiftDriver/Utilities/Triple.swift
aciidb0mb3r: gotcha, saw that later :D

Sources/SwiftDriver/Utilities/Triple.swift
sjavora: OK, changed everywhere in this PR.

sjavora:
Updated.

aciidb0mb3r:
@swift-ci test

DougGregor:
@swift-ci test

PR #82, opened 2020-03-06
Relative rsp files

zienag:
Allow relative paths for rsp file, skip forwarded args in rsp expansion, fail on invalid rsp files.

Should go after https://github.com/apple/swift-driver/pull/73

The solution for skipping `-Xblabla @SOMETHING` is rather fragile, but I didn't find any better without major changes. One way to do it might be to add option attribute, something like "argument escape" (as -Xblabla is similar to escaping)

### DIFF
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -582,16 +582,19 @@ final class SwiftDriverTests: XCTestCase {
       let diags = DiagnosticsEngine()
       let fooPath = path.appending(component: "foo.rsp")
       let barPath = path.appending(component: "bar.rsp")
+      let notAFilePath = path.appending(component: "nonexistent.rsp")
       try localFileSystem.writeFileContents(fooPath) {
         $0 <<< "hello\nbye\nbye\\ to\\ you\n@\(barPath.pathString)"
       }
       try localFileSystem.writeFileContents(barPath) {
-        $0 <<< "from\nbar\n@\(fooPath.pathString)"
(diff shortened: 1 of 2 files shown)


Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -582,16 +582,19 @@ final class SwiftDriverTests: XCTestCase {
...
-        $0 <<< "from\nbar\n@\(fooPath.pathString)"
+        $0 <<< "from\nbar\n@\(fooPath.pathString)\n@\(notAFilePath.pathString)"
       }
       let args = try Driver.expandResponseFiles(["swift", "compiler", "-Xlinker", "@loader_path", "@" + fooPath.pathString, "something"], diagnosticsEngine: diags)
       XCTAssertEqual(args, ["swift", "compiler", "-Xlinker", "@loader_path", "hello", "bye", "bye to you", "from", "bar", "something"])
-      XCTAssertEqual(diags.diagnostics.count, 1)
+      XCTAssertEqual(diags.diagnostics.count, 2)
       XCTAssert(diags.diagnostics.first!.description.contains("is recursively expanded"))
+      // FIXME: Error message about nonexistent response file could be improved
+      XCTAssertEqual(diags.diagnostics[1].description, "noEntry")
```
zienag: Is it better to change it in tools-support-core or just transform it in swift-driver code?

shahmishal:
The Swift project moved the default branch to `main` and deleted `master` branch, so GitHub automatically closed the PR. Please re-create pull request with `main` branch.

PR #88, opened 2020-05-09
When bootstrapping SwiftPM, use relative paths for dependencies.

DougGregor:
When bootstrapping SwiftPM, SwiftPM uses relative paths to find
its dependencies (e.g., llbuild) rather than repository URLs.
Do the same thing for the driver when `SWIFTCI_USE_LOCAL_DEPS` is set in
the environment. This allows us to use the Swift driver when
bootstrapping SwiftPM.


DougGregor:
@swift-ci please test

PR #89, opened 2020-05-09
Set macOS minimum deployment target to 10.10.

DougGregor:
To match SwiftPM, set the macOS minimum deployment target to 10.10.


DougGregor:
@swift-ci please test

PR #90, opened 2020-05-10
Remove support for the integrated repl

owenv:
Now that it's been removed from the C++ driver, there's no reason to keep it around here.


owenv:
@swift-ci test

PR #92, opened 2020-05-11
Add a swift-argument-parser dependency and use it in swift-help

owenv:
swift-argument-parser needs some CMake fixes before this can land

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ endif()
 find_package(dispatch QUIET)
 find_package(Foundation QUIET)
 find_package(Yams CONFIG REQUIRED)
+find_package(ArgumentParser CONFIG REQUIRED)
 
 add_subdirectory(Sources)
 add_subdirectory(cmake/modules)
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,9 +5,21 @@
         "package": "swift-argument-parser",
         "repositoryURL": "https://github.com/apple/swift-argument-parser.git",
         "state": {
+<<<<<<< HEAD
           "branch": null,
           "revision": "15351c1cd009eba0b6e438bfef55ea9847a8dc4a",
           "version": "0.3.0"
+=======
+<<<<<<< HEAD
+          "branch": "master",
+          "revision": "15351c1cd009eba0b6e438bfef55ea9847a8dc4a",
+          "version": null
+=======
+          "branch": null,
+          "revision": "223d62adc52d51669ae2ee19bdb8b7d9fd6fcd9c",
+          "version": "0.0.6"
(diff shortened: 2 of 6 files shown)


owenv:
Tested a cmake build with the change in https://github.com/apple/swift-argument-parser/pull/138 and it seemed to work

DougGregor:
Good idea! To do this, we'll need to update (or add on to) Swift's list of repos to check out (see https://github.com/apple/swift/pull/31664) and the SwiftPM bootstrap script in https://github.com/apple/swift-package-manager/pull/2736.

DougGregor:
I'm pulling in swift-argument-parser as a dependency for CI purposes in https://github.com/apple/swift/pull/31664.

owenv:
@DougGregor Cool, I hadn't  seen that SwiftPM PR! I opened https://github.com/apple/swift-argument-parser/pull/139 to update swift-argument-parser cmake and https://github.com/apple/swift-package-manager/pull/2739 in SwiftPM based on your PR. They probably need a little additional work, but should be mostly on the right track.

natecook1000:
@owenv This is great! Merged your PR in the argument parser.

We'll want to follow up later to add support for querying help for the other swift tools using `swift-help`, so that writing e.g. `swift help package init` gives you the same help as `swift package init --help`.

DougGregor:
I think this is great! Now that the integrated driver is merged into SwiftPM, there are a few steps needed to make this work out:
* swift-argument-parser will need to tag (0.0.6 or whatever) that includes @owenv 's change
* update-checkout will need to switch to that tag
* SwiftPM's bootstrap script will need to build ArgumentParser and pass its location in when building the SwiftDriver

... and then we can merge this

natecook1000:
@DougGregor I've tagged an 0.0.6 release for swift-argument-parser that includes @owenv's CMake fixes.

owenv:
Thanks @natecook1000, I'll switch update-checkout to start using the new tag and then we should be all set to adopt ArgumentParser!

DougGregor:
@owenv apparently the source compatibility suite has another copy of the update-checkout information that will also need updating (ugh)

owenv:
@DougGregor Thanks for the heads up, I had no idea that had its own config. I just saw this commit: https://github.com/apple/swift-source-compat-suite/commit/8d478f0b8e31bd569ce0e1ae6f765f6f83f6ff1a , I'll copy that

owenv:
@swift-ci test

owenv:
Once the tests at https://github.com/apple/swift-package-manager/pull/2739 pass, we should be able to merge this 🎉

DougGregor (approved):

DougGregor:
@swift-ci test

PR #95, opened 2020-05-13
Parallelize task build

DougGregor:
Revert the change to avoid calling LLBuild's `taskIsComplete` from a thread, which should improve the parallelism in the build.


DougGregor:
@swift-ci please test

PR #96, opened 2020-05-13
Improve handling of multithreaded builds

cltnschlosser:
Fixes most of the issues with test `Driver/multi-threaded.swift`.

Future improvements required for all tests to pass:
1. Spacing of parseable-output json is different than c++ driver (use a regex in the test to allow optional space)
2. Support embed bitcode (I might look at this next)

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -651,7 +651,9 @@ extension Driver {
       return
     }
 
-    if jobs.contains(where: { $0.requiresInPlaceExecution }) || jobs.count == 1 {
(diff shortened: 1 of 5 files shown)


DougGregor:
@swift-ci please test

cltnschlosser:
@DougGregor Opened https://github.com/apple/swift/pull/31815 for the spaces. The parsable-output json has it's keys in a different order (swift-driver sorts, and c++ driver doesn't). Is there a way to support this in the test? or do we just accept this failure for now?

CodaFi:
I don't think we need to go as far as sorting the keys. If the idea is to share test infrastructure, it seems like it would be easier to just emulate the ordering of the legacy driver.

PR #99, opened 2020-05-14
Thread a filesystem throughout the Swift driver.

DougGregor:
The Swift driver library should never directly access the local file
system. Thread a file system parameter through the SwiftDriver library
so clients can provide an alternative file system if they want.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -77,6 +77,9 @@ public struct Driver {
   /// it should be identical to the real environment.
   public let env: [String: String]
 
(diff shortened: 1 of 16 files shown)


DougGregor:
@swift-ci test

ddunbar:
Does this unblock additional or perhaps better testing?

DougGregor:
@swift-ci test Linux

DougGregor:
> Does this unblock additional or perhaps better testing?

I'm planning to use it with a partial pass-through file system so we don't have to write as much to disk while doing build planning. That's causing some problems for me with SwiftPM and the integrated driver (hence, the need for this PR).

PR #100, opened 2020-05-14
Add support for reading fine-grained source file dependency graphs

owenv:
This just ports over the parts of the source file fine-grained dependency graph the driver cares about. I just called it `SourceFileDependencyGraph` because I assume we don't intend to support the older dependency system. I'm still learning how all this works in the C++ driver, so I'm open to feedback on the API and approach!


owenv:
@swift-ci test

owenv:
@swift-ci test

DougGregor:
@swift-ci test

davidungar:
Looks good! Thanks.

PR #101, opened 2020-05-15
Support -embed-bitcode and -driver-print-bindings

cltnschlosser:
- The moduleNameIsFallback fix for the linker fixes `Driver/verify-debug-info.swift` (dysm at wrong name)
- embed-bitcode tests in `Driver/multi-threaded.swift` are fixed (Still some tests there "failing" because parsable output json is in different order)
- `Driver/embed-bitcode.swift` tests are fixed except one. Driver jobs are printed in different order.
```
// RUN: %target-swiftc_driver -embed-bitcode -Xcc -DDEBUG -Xllvm -fake-llvm-option -c -emit-module %s 2>&1 -### | %FileCheck %s -check-prefix=CHECK-MODULE
// CHECK-MODULE: -frontend
// CHECK-MODULE: -emit-bc
// CHECK-MODULE-DAG: -Xcc -DDEBUG
// CHECK-MODULE-DAG: -Xllvm -fake-llvm-option
// CHECK-MODULE-DAG: -emit-module-path
// CHECK-MODULE: -frontend
// CHECK-MODULE: -emit-module
// CHECK-MODULE: -frontend
// CHECK-MODULE: -c
// CHECK-MODULE-NOT: -Xcc
// CHECK-MODULE-NOT: -DDEBUG
// CHECK-MODULE-NOT: -fake-llvm-option
// CHECK-MODULE-NOT: -emit-module-path
```
In this case the merge module job gets printed before the backend job. For all other tests the merge module job is last. Not sure what is effecting the print order in c++ driver.

### DIFF
--- a/Sources/SwiftDriver/Jobs/BackendJob.swift
+++ b/Sources/SwiftDriver/Jobs/BackendJob.swift
@@ -1,0 +1,77 @@
+//===--------------- BackendJob.swift - Swift Backend Job -------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+extension Driver {
+  /// Form a backend job.
+  mutating func backendJob(input: TypedVirtualPath,
+                           allOutputs: inout [TypedVirtualPath]) throws -> Job {
+    var commandLine: [Job.ArgTemplate] = swiftCompilerPrefixArgs.map { Job.ArgTemplate.flag($0) }
+    var inputs = [TypedVirtualPath]()
+    var outputs = [TypedVirtualPath]()
+
+    commandLine.appendFlag("-frontend")
+    addCompileModeOption(outputType: compilerOutputType, commandLine: &commandLine)
+
+    // Add input arguments.
+    commandLine.appendFlag(.primaryFile)
+    commandLine.appendPath(input.file)
+    inputs.append(input)
+
+    commandLine.appendFlag(.embedBitcode)
+
+    // -embed-bitcode only supports a restricted set of flags.
+    commandLine.appendFlag(.target)
+    commandLine.appendFlag(targetTriple.triple)
+
+    // Enable address top-byte ignored in the ARM64 backend.
+    if targetTriple.arch == .aarch64 {
+      commandLine.appendFlag(.Xllvm)
+      commandLine.appendFlag("-aarch64-use-tbi")
+    }
+
+    // Handle the CPU and its preferences.
+    try commandLine.appendLast(.targetCpu, from: &parsedOptions)
+
+    // Enable optimizations, but disable all LLVM-IR-level transformations.
+    try commandLine.appendLast(in: .O, from: &parsedOptions)
+    commandLine.appendFlag(.disableLlvmOptzns)
+
+    try commandLine.appendLast(.parseStdlib, from: &parsedOptions)
+
+    commandLine.appendFlag(.moduleName)
+    commandLine.appendFlag(moduleName)
+
+    // Add the output file argument if necessary.
+    if let compilerOutputType = compilerOutputType {
+      let output = computePrimaryOutput(for: input,
+                                        outputType: compilerOutputType,
+                                        isTopLevel: isTopLevelOutput(type: compilerOutputType))
+      commandLine.appendFlag(.o)
+      commandLine.appendPath(output.file)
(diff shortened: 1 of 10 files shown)


Sources/SwiftDriver/Jobs/BackendJob.swift
```diff
@@ -0,0 +1,77 @@
...
+    commandLine.appendFlag(targetTriple.triple)
+
+    // Enable address top-byte ignored in the ARM64 backend.
+    if targetTriple.arch == .aarch64 {
+      commandLine.appendFlag(.Xllvm)
+      commandLine.appendFlag("-aarch64-use-tbi")
+    }
+
+    // Handle the CPU and its preferences.
+    try commandLine.appendLast(.targetCpu, from: &parsedOptions)
```
DougGregor, line 44: A bunch of this is repeated from`addCommonFrontendOptions`. Can you factor some of it our so there's less duplication here?

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -110,10 +124,21 @@ extension Driver {
...
 
-        var jobOutputs: [TypedVirtualPath] = []
-        let job = try compileJob(primaryInputs: primaryInputs, outputType: compilerOutputType, allOutputs: &jobOutputs)
-        jobs.append(job)
-        addJobOutputs(jobOutputs)
+        if parsedOptions.hasArgument(.embedBitcode) {
+          var jobOutputs: [TypedVirtualPath] = []
+          let job = try compileJob(primaryInputs: primaryInputs, outputType: .llvmBitcode, allOutputs: &jobOutputs)
+          jobs.append(job)
+          for input in job.outputs.filter({ $0.type == .llvmBitcode }) {
```
DougGregor, line 131: This is really nice and clean, thanks!

Sources/SwiftDriver/Jobs/BackendJob.swift
cltnschlosser: So there is some duplication here, but I don't know how much I can factor out without changing the order of the arguments (That shouldn't really matter, but there are probably some lit tests that rely on the current ordering.

The interesting part from `addCommonFrontendOptions` is:
```swift
    switch compilerMode {
    case .standardCompile, .singleCompile, .batchCompile, .compilePCM:
      commandLine.appendFlag(.target)
      commandLine.appendFlag(targetTriple.triple)

    case .repl, .immediate:
      if parsedOptions.hasArgument(.target) {
        commandLine.appendFlag(.target)
        commandLine.appendFlag(targetTriple.triple)
      }
    }

    if let variant = parsedOptions.getLastArgument(.targetVariant)?.asSingle {
      commandLine.appendFlag(.targetVariant)
      commandLine.appendFlag(Triple(variant, normalizing: true).triple)
    }

    // Enable address top-byte ignored in the ARM64 backend.
    if (targetTriple.arch == .aarch64) {
      commandLine.appendFlag(.Xllvm)
      commandLine.appendFlag("-aarch64-use-tbi")
    }

    // Enable or disable ObjC interop appropriately for the platform
    if targetTriple.isDarwin {
      commandLine.appendFlag(.enableObjcInterop)
    } else {
      commandLine.appendFlag(.disableObjcInterop)
    }

    // Handle the CPU and its preferences.
    try commandLine.appendLast(.targetCpu, from: &parsedOptions)
```

The switch is fine, because we don't have backend jobs for repl and immediate builds, but would have to move these out of order.
```swift
    if let variant = parsedOptions.getLastArgument(.targetVariant)?.asSingle {
      commandLine.appendFlag(.targetVariant)
      commandLine.appendFlag(Triple(variant, normalizing: true).triple)
    }
```
```swift
    // Enable or disable ObjC interop appropriately for the platform
    if targetTriple.isDarwin {
      commandLine.appendFlag(.enableObjcInterop)
    } else {
      commandLine.appendFlag(.disableObjcInterop)
    }
```

Looking at this further. Is there any reason that the embed-bitcode job doesn't use the targetVariant parameter. Seems like it should, but the c++ driver doesn't.

DougGregor: Hmm. I see what you mean. We can keep it like this, and maybe there’s some future refactoring we can do. I worry about maintainability when we’re doing the same argument translation in several places. 

Regarding -target-variant, yes, I think we should pass it through for bitcode compilation. It gets embedded in the binary. This is probably a bug in the C++ driver

cltnschlosser: Should I add the -target-variant here?

DougGregor: Yes, please add the -target-variant

DougGregor:
@swift-ci test

DougGregor:
@cltnschlosser Mind if I go ahead and merge this? The `-target-variant` change can come in later

cltnschlosser:
> @cltnschlosser Mind if I go ahead and merge this? The `-target-variant` change can come in later

Yep, sounds good

PR #103, opened 2020-05-17
Driver performance improvements (profile test/Driver/response-file.swift)

cltnschlosser:
Big change is:
> Remove job from build keys. Saves encoding and decoding the jobs
> 
> The runtime of the following (in release build) went from 2.27 minutes to 2.29 seconds (notice the change in units!!!)
> // RUN: %{python} -c 'for i in range(500001): print("-DTEST5_" + str(i))' > %t.5.resp
> // RUN: not %target-swiftc_driver %s @%t.5.resp -Xfrontend -debug-crash-immediately 2>&1


owenv:
@swift-ci test

PR #105, opened 2020-05-19
Set LC_ALL=en_US.UTF-8 when running integration tests with lit.py

owenv:
I needed to make this change to successfully run the `driver/` integration tests inside Xcode, otherwise they would hang because the encoding was defaulting to ascii. It might be ok to just set LC_CTYPES instead, but I'm really not sure which is better.


DougGregor:
@swift-ci test

PR #107, opened 2020-05-20
Add BackendJob.swift to cmake build

owenv:
https://github.com/apple/swift-driver/pull/101 added this file, but it wasn't added to the CMake build which I think is causing SwiftPM to fail on CI.


owenv:
@swift-ci test

owenv:
@DougGregor no problem, I only happened to notice it because I had another commit in the failing run and saw the email! I think SwiftPM may also need an update to handle the new `Job.Kind.backend` enum case.

cltnschlosser:
Should we add something to the CI to do a build with CMake?

PR #109, opened 2020-05-23
Significantly improve prefix trie performance and memory usage

harlanhaskins:
The previous prefix trie made a new `Node` for each character along the
input string. In addition to being wasteful, this would introduce many
more hops to match a given string.

Instead, store common `Substring`s in the `Node`s, and use a sorted
array/binary search through children to avoid extra time spend hashing
and storing in hashed collections.

Doing this reduces the number of nodes in the trie for the driver's
options by an order of magnitude -- from 6601 nodes to 660.

(This is actually a port of LLVM's Trie, but cleaned up and made more
idiomatic in Swift)

### DIFF
--- a/Sources/SwiftOptions/PrefixTrie.swift
+++ b/Sources/SwiftOptions/PrefixTrie.swift
@@ -14,16 +14,32 @@
 /// It's a tree structure where each node in the tree corresponds to another
 /// element in the prefix of whatever you're looking for.
 ///
-/// For example, the strings 'Hello' and 'Help' would look like this:
+/// For example, if you entered these strings:
+///
+/// "Help", "Helper", "Helping", "Helicopters", "Helicoptering"
+///
+/// The resulting trie would look like this:
 ///
 /// ```
-///                      ┌───┐  ┌───┐
-///                 ┌───▶│ l │─▶│ o │
-/// ┌───┐  ┌───┐  ┌───┐  └───┘  └───┘
-/// │ H │─▶│ e │─▶│ l │
-/// └───┘  └───┘  └───┘  ┌───┐
-///                 └───▶│ p │
-///                      └───┘
+///                      ┌───┐
+///              ┌──────▶│ s │
+///              │       └───┘
+///         ┌─────────┐
+///    ┌───▶│ icopter │
+///    │    └─────────┘
+///    │         │       ┌───┐
+///    │         └──────▶│ing│
+/// ┌─────┐              └───┘
+/// │ Hel │
+/// └─────┘              ┌────┐
+///    │      ┌─────────▶│ er │
+///    │      │          └────┘
+///    │    ┌───┐
+///    └───▶│ p │
+///         └───┘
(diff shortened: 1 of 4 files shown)


harlanhaskins:
@swift-ci please test

harlanhaskins:
@swift-ci please test

harlanhaskins:
[Here's the old GraphViz graph](https://github.com/apple/swift-driver/files/4670999/graph.pdf)

[Here's the new GraphViz graph](https://github.com/apple/swift-driver/files/4670976/graphviz-3.pdf)

harlanhaskins:
@swift-ci please test

harlanhaskins:
It shaves off 3MB of memory, too. But that's not really that much.

DougGregor:
Woo this go into Swift-argument-parser too?

harlanhaskins:
I tried to work it in, but the swift-argument-parser doesn’t do the kind of ‘best prefix’ matching we need to do in the driver. I was planning to follow up with @nnnnnnnn about this specifically!

Sources/SwiftOptions/PrefixTrie.swift
```diff
@@ -32,114 +48,365 @@
...
+          current = existing
+          if let data = existing.data {
+            // If there's data associated with this node, though, keep track of
+            // it. We may end up with a later match that doesn't have associated
+            // data.
+            bestMatch = data
+          }
+        case .dontMatch:
+          fatalError("Impossible because we found a child with id \(id)")
+        default:
```
cltnschlosser, line 236: Explicitly declare the last 2 cases, so the compiler catches if more are added.

Sources/SwiftOptions/PrefixTrie.swift
```diff
@@ -32,114 +48,365 @@
...
-    init(value: Value?) {
-      self.value = value
+    /// The children of this node. This array is always in sorted order by
+    /// the node's `id`.
+    var children = [Node]()
+
+    /// Each node is identified in the child list by the first character in its
+    /// label.
+    var id: Character {
+      label.first!
```
cltnschlosser, line 90: I personally have always liked `label[0]` to avoid the implicitly unwrapped optional, but do you know if there are any performance differences between the 2? Maybe the optimize the same, I'll try it out.

harlanhaskins: In this case, the label is a `Substring` which doesn’t have an int subscript.

harlanhaskins:
I’ll get these in a follow-up, thanks @cltnschlosser!

Sources/SwiftOptions/PrefixTrie.swift
cltnschlosser: Ah, I guess you could use https://developer.apple.com/documentation/swift/substring/2894052-subscript

But `label[label.startIndex]` looks worse to me than `label.first!`.

Also, I tried tested it out and at least for an `[Int]` they are the same after optimizations.

PR #112, opened 2020-05-26
Add an end-to-end test for the option to print compile jobs for module dependencies

artemcm:
As discovered by invoking the fast dependency scanner from within the driver.

PR #113, opened 2020-06-01
Propagate Explicit Module Build command arguments to compile jobs.

artemcm:
This PR adds the compile command argument interface for Explicit Module Builds, as specified in: https://github.com/apple/swift/pull/32125

- `-disable-implicit-swift-modules` and `-Xcc -Xclang -Xcc -fno-implicit-modules` are passed to all compile jobs to prevent the frontend from building any modules implicitly and diagnose if they have to.
- When building a module, the compile commands are passed paths to the module's direct dependencies with `-Xcc -Xclang -Xcc -fmodule-map-file=` and `-Xcc -Xclang -Xcc -fmodule-file=` and `-swift-module-file=`.
- Add `-experimental-explicit-module-build` option to request that the driver must attempt to perform a full explicit-module-build compile. 

For now, I include a test that verifies that the commands are generated as expected w.r.t. their dependencies. Once the frontend puzzle piece is landed, we should be close to getting a full end-to-end test. 

This PR subsumes #112.

### DIFF
--- a/Sources/SwiftDriver/Dependency Scanning/ModuleDependencyBuildGeneration.swift
+++ b/Sources/SwiftDriver/Dependency Scanning/ModuleDependencyBuildGeneration.swift
@@ -26,14 +26,17 @@ extension Driver {
       }
       switch id {
         case .swift(let moduleName):
-          let swiftModuleBuildJob = try genSwiftModuleDependencyBuildJob(moduleInfo: moduleInfo,
-                                                                         moduleName: moduleName)
+          let swiftModuleBuildJob =
+            try genSwiftModuleDependencyBuildJob(moduleInfo: moduleInfo,
+                                                 moduleName: moduleName,
+                                                 dependencyGraph: dependencyGraph)
           jobs.append(swiftModuleBuildJob)
         case .clang(let moduleName):
-          let clangModuleBuildJob = try genClangModuleDependencyBuildJob(moduleInfo: moduleInfo,
-                                                                         moduleName: moduleName)
+          let clangModuleBuildJob =
+            try genClangModuleDependencyBuildJob(moduleInfo: moduleInfo,
+                                                 moduleName: moduleName,
+                                                 dependencyGraph: dependencyGraph)
           jobs.append(clangModuleBuildJob)
-
       }
     }
     return jobs
@@ -41,7 +44,9 @@ extension Driver {
 
   /// For a given swift module dependency, generate a build job
   mutating private func genSwiftModuleDependencyBuildJob(moduleInfo: ModuleInfo,
-                                                         moduleName: String) throws -> Job {
+                                                         moduleName: String,
+                                                         dependencyGraph: InterModuleDependencyGraph)
+  throws -> Job {
     // FIXIT: Needs more error handling
     guard case .swift(let swiftModuleDetails) = moduleInfo.details else {
       throw Error.malformedModuleDependency(moduleName, "no `details` object")
@@ -53,6 +58,11 @@ extension Driver {
     ]
     var commandLine: [Job.ArgTemplate] = swiftCompilerPrefixArgs.map { Job.ArgTemplate.flag($0) }
     commandLine.appendFlag("-frontend")
+
+    try addModuleDependencies(moduleInfo: moduleInfo,
+                              dependencyGraph: dependencyGraph,
+                              inputs: &inputs,
+                              commandLine: &commandLine)
 
     // Build the .swiftinterfaces file using a list of command line options specified in the
     // `details` field.
@@ -76,7 +86,9 @@ extension Driver {
 
   /// For a given clang module dependency, generate a build job
   mutating private func genClangModuleDependencyBuildJob(moduleInfo: ModuleInfo,
-                                                         moduleName: String) throws -> Job {
+                                                         moduleName: String,
+                                                         dependencyGraph: InterModuleDependencyGraph)
+  throws -> Job {
     // For clang modules, the Fast Dependency Scanner emits a list of source
     // files (with a .modulemap among them), and a list of compile command
     // options.
@@ -92,12 +104,18 @@ extension Driver {
     commandLine.appendFlag("-frontend")
     commandLine.appendFlags("-emit-pcm", "-module-name", moduleName)
 
+    try addModuleDependencies(moduleInfo: moduleInfo,
+                              dependencyGraph: dependencyGraph,
+                              inputs: &inputs,
+                              commandLine: &commandLine)
+
     // The only required input is the .modulemap for this module.
-    commandLine.append(Job.ArgTemplate.path(try VirtualPath(path: clangModuleDetails.moduleMapPath)))
+    commandLine.append(Job.ArgTemplate.path(
+                        try VirtualPath(path: clangModuleDetails.moduleMapPath)))
     inputs.append(TypedVirtualPath(file: try VirtualPath(path: clangModuleDetails.moduleMapPath),
                                    type: .clangModuleMap))
     try addCommonModuleOptions(commandLine: &commandLine, outputs: &outputs)
-    clangModuleDetails.commandLine?.forEach { commandLine.appendFlags("-Xcc", $0) }
+    clangModuleDetails.commandLine?.forEach { commandLine.appendFlags($0) }
 
     return Job(
       moduleName: moduleName,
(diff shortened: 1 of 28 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -200,6 +208,11 @@ public struct Driver {
   /// This will force the driver to first emit the module and then run compile jobs.
   public var forceEmitModuleInSingleInvocation: Bool = false
 
+  /// The module dependency graph, which is populated during the planning phase
+  /// only when all modules will be prebuilt and treated as explicit by the
+  /// various compilation jobs.
+  var interModuleDependencyGraph: InterModuleDependencyGraph? = nil
```
artemcm, line 214: I brought this back after we decided against having this as shared mutable state in the driver in #104, for two reasons:
- The dependency graph is now used in more places, we need access to it when generating *any* compile job, not just compile jobs for module dependencies.
- Being able to access this structure inside tests turned out to be very useful for ensuring that generated jobs are consistent with the dependency graph used to generate them.

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -69,6 +52,26 @@ extension Driver {
       }
     }
 
+    // If asked, add jobs to precompile module dependencies
+    if parsedOptions.contains(.driverExplicitModuleBuild) ||
+        parsedOptions.contains(.driverPrintModuleDependenciesJobs) {
```
artemcm, line 57: `driverPrintModuleDependenciesJobs` makes this messy, but I plan on removing it in a followup and making tweaks to `-print-driver-jobs` to make debugging this a bit easier.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -304,4 +304,49 @@ extension Driver {
...
+                                               type: .swiftModule)
+        commandLine.appendFlag("-swift-module-file=\(swiftModulePath.file.description)")
+        inputs.append(swiftModulePath)
+      case .clang(let clangDependencyDetails):
+        let clangModulePath = TypedVirtualPath(file: try VirtualPath(path: moduleInfo.modulePath),
+                                               type: .pcm)
+        let clangModuleMapPath = TypedVirtualPath(file: try VirtualPath(path: clangDependencyDetails.moduleMapPath),
+                                                  type: .pcm)
+        commandLine.appendFlag("-clang-module-file=\(clangModulePath.file.description)")
+        commandLine.appendFlag("-clang-module-map-file=\(clangModuleMapPath.file.description)")
```
nkcsgexi, line 325: Sorry that I've changed the compiler to not adding these two Swift flags since they are redundant of the underlying Clang ones. Could you change them to `-Xcc -Xclang -Xcc -fmodule-map-file=\(clangModuleMapPath.file.description)` and `-Xcc -Xclang -Xcc -fmodule-file=\(clangModuleMapPath.file.description)`

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -304,4 +304,49 @@ extension Driver {
...
+
+  /// Adds all dependecies required for an explicit module build
+  /// to inputs and comman line arguments of a compile job.
+  func addExplicitModuleBuildArguments(commandLine: inout [Job.ArgTemplate],
+                                       inputs: inout [TypedVirtualPath]) throws {
+    guard let dependencyGraph = interModuleDependencyGraph else {
+      fatalError("Inter Module Dependency Graph does not exist in explicit module build mode.")
+    }
+    // Prohibit the frontend from implicitly building textual modules into binary modules.
+    commandLine.appendFlags("-disable-implicit-swift-modules", "-disable-implicit-pcms")
```
nkcsgexi, line 339: Same here. I've dropped `-disable-implicit-pcms` in favor of `-Xcc -Xclang -Xcc -fno-implicit-modules`

nkcsgexi: Also, we need to pass down these flags for `-scan-dependencies` action as well. That way the dependencies canner can report them back in the build commands.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
artemcm: 👍 I adjusted the generated commands.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
artemcm: In this PR, the driver adds these options to the dependency module build commands explicitly, and adds all the module's dependency binary module files as inputs. 

Do you think it is better for the driver to keep doing that or to switch to relying on the dependency scanner to include those command line arguments in its output?

nkcsgexi: I prefer relying on the dependencies scanner to report these arguments back (we only need to specify them explicitly for the dependency scanning action). IMO, these arguments are no different from other arguments reported from the scanner.

artemcm: Okay. Making this change now will cause tests to fail, because the dependency scanner does not yet support or report these flags, but I will make the change as soon as it does, whether that be in this PR or a followup, timing-wise.

DougGregor: I agree with both of you :). It's better to have the dependency scanner report this, because it will (eventually) be able to canonicalize the command-line arguments. Let's make the switch once the dependency scanner reports this information, but for now it's better to make progress on end-to-end.

DougGregor:
@swift-ci test

Sources/SwiftDriver/Dependency Scanning/ModuleDependencyBuildGeneration.swift
```diff
@@ -108,4 +136,24 @@ extension Driver {
...
+  /// For the specified module, update its command line flags and inputs
+  /// to use explicitly-built module dependencies.
+  private func addModuleDependencies(moduleInfo: ModuleInfo,
+                                     dependencyGraph: InterModuleDependencyGraph,
+                                     inputs: inout [TypedVirtualPath],
+                                     commandLine: inout [Job.ArgTemplate]) throws {
+    // Prohibit the frontend from implicitly building textual modules into binary modules.
+    commandLine.appendFlags("-disable-implicit-swift-modules", "-Xcc", "-Xclang", "-Xcc",
+                            "-fno-implicit-modules")
+    for moduleId in moduleInfo.directDependencies {
```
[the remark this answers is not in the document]
nkcsgexi: hmm, my impression is direct dependencies alone are sufficient. How about we keep it as it is and later we can add transitive dependencies if they are also necessary.

shahmishal:
@swift-ci test

Sources/SwiftDriver/Dependency Scanning/ModuleDependencyBuildGeneration.swift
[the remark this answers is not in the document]
DougGregor: I disagree. The command line needs to identify all of the inputs needed for the compile, which includes transitive dependencies, so a distributed build system can realize them before executing the compiler. We should never have to search for a module.

DougGregor:
@swift-ci please test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

DougGregor:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
The test added in this PR is failing in CI because it needs to invoke the compiler (dependency scanner) and I suspect that this is due to the way the driver currently does path resolution when looking for the compiler executable. 
It attempts the following, in order, whichever method succeeds first:
  1. Try to find it with executable-specific environment variable
      (This is a typical development workflow, hence why "it works on my machine".)
  2. Try to find with `xcrun -sdk macosx --find <executable>`
  3. Try to search default search paths, which are just set to `PATH`
  4. Fallback to `/usr/bin/`.

Because CI points the driver to the latest toolchain with `PATH`, the driver ends up invoking the wrong compiler (system compiler found with `xcrun`, I suspect). 

@DougGregor , we can make this work by doing either:
1. Have CI export `SWIFT_DRIVER_SWIFT_EXEC` to point to the downloaded toolchain's `bin`.
2. Swap the order 2 <-> 3 above. 
Am I right in thinking that the path resolution order is what it is for a good reason and we should pursue option (1)?

DougGregor:
> @DougGregor , we can make this work by doing either:
> 
> 1. Have CI export `SWIFT_DRIVER_SWIFT_EXEC` to point to the downloaded toolchain's `bin`.
> 2. Swap the order 2 <-> 3 above.
>    Am I right in thinking that the path resolution order is what it is for a good reason and we should pursue option (1)?

I think we should do (1), because we might want to fiddle with path resolution order for other reasons, but `SWIFT_DRIVER_SWIFT_EXEC` will remain the first point of override regardless of what we do.

Thank you!

artemcm:
@swift-ci please test

Sources/SwiftDriver/Dependency Scanning/ModuleDependencyBuildGeneration.swift
```diff
@@ -16,8 +16,8 @@ import Foundation
 extension Driver {
   /// For the current moduleDependencyGraph, plan the order and generate jobs
   /// for explicitly building all dependency modules.
-  mutating func planExplicitModuleDependenciesCompile(dependencyGraph: InterModuleDependencyGraph)
-  throws -> [Job] {
+  mutating func planExplicitModuleDependenciesCompile(dependencyGraph: InterModuleDependencyGraph
```
harlanhaskins, line 19: Nitpick: can you bring the argument down onto its own line?

```suggestion
  mutating func planExplicitModuleDependenciesCompile(
    dependencyGraph: InterModuleDependencyGraph
  ) throws -> [Job] {
```

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

DougGregor:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #114, opened 2020-06-01
Update README to avoid recursive lookups

owenv:
This matches the workflow I've been using to avoid the issue where `swift-driver` keeps invoking itself instead of the frontend. Maybe there's a better way?

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,13 @@ ln -s /path/to/built/swift-driver $SOME_PATH/swift
 ln -s /path/to/built/swift-driver $SOME_PATH/swiftc
 ```
 
-Swift packages can be built with the new Swift driver by overriding `SWIFT_EXEC` to refer to the `swiftc` symbolic link created above, e.g.,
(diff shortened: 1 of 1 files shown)


owenv:
Also cc @cltnschlosser who I think ran into this problem at some point and may know of a better approach to using the swift driver with Xcode.

artemcm:
Thank you! It's the workaround I've been using and this would have saved me some headache when I started playing with the project.

cltnschlosser:
Yeah I think that’s what I ended up doing. Or makes more sense than whatever hack I had at the time.

@DougGregor 
What ever happened to https://github.com/apple/swift/pull/28003

owenv:
@swift-ci test

DougGregor:
> @DougGregor
> What ever happened to [apple/swift#28003](https://github.com/apple/swift/pull/28003)

I got busy and never got back to it. It's on my TODO list, which... doesn't say much. Hopefully I'll get to it in the next week or two, because it's important for getting swift-driver into the overall toolchain build.

owenv:
@swift-ci please test Linux

PR #117, opened 2020-06-05
Change -num-threads to only multithread for values > 1

keith:
Previously the 0 value of num-threads was the only one that meant don't
use single threading. Now 1 is treated the same way.


keith:
So FWIW this is also the behavior in Driver.cpp, so I'm kinda questioning both with this change (and will change it there too if this is what we think is right)

keith:
cc @eeckstein

eeckstein:
@Keith This behavior is intentional. Sometimes it's useful to test/analyze the multi-module IR generation, but within a single thread. This is purpose of "1".

PR #118, opened 2020-06-05
Move -num-threads tests to single function

keith:
This is just an organizational change to keep all the tests for
`-num-threads` behavior in one place.


keith:
@DougGregor I don't feel super strongly about this but wdyt?

DougGregor:
@swift-ci please test

PR #119, opened 2020-06-07
Extract debug info and module output properties from Driver.swift

sjavora:
I noticed that there were groups of properties in the driver that were returned as tuples and then assigned. It seemed natural to group these in structs.

### DIFF
--- a/Sources/SwiftDriver/Driver/ModuleOutputInfo.swift
+++ b/Sources/SwiftDriver/Driver/ModuleOutputInfo.swift
@@ -1,0 +1,51 @@
+//===--------------- ModuleOutputInfo.swift - Module output information ---===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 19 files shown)


Sources/SwiftDriver/Driver/ModuleOutputInfo.swift
```diff
@@ -0,0 +1,51 @@
...
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// How should the Swift module output be handled?
+public enum ModuleOutput: Equatable {
```
sjavora, line 14: Should this be nested in `ModuleOutputInfo`?

sjavora:
Thanks, fixed.

Sources/SwiftDriver/Driver/ModuleOutputInfo.swift
artemcm: Yeah, I think that would make sense.

sjavora: Done.

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
🚢

PR #121, opened 2020-06-08
Fix tests to match behaviour when when using specific Swift toolchain.

artemcm:
Having just switched CI to point to a specific (latest master snapshot) swift toolchain, instead of using the system toolchain (discussed in #113 ), a couple of test adjustments were required to get the test suite to pass:

- `testDOTFileEmission` to match latest toolchain's compiler executable name as `swiftc` vs `swift`. 
- `testToolchainClangPath` to skip the checking if the user overloaded swift executable path.


DougGregor:
@swift-ci test

PR #122, opened 2020-06-09
[Explicit Module Builds] Do not add the .modulemap file to the build command arguments

artemcm:
As these files are already specified in the relevant dependency scanner JSON output field.


artemcm:
@swift-ci please test

PR #123, opened 2020-06-09
[Explicit Module Builds] Add end-to-end explicit module build compile test

artemcm:
The test invokes the driver using `-experimental-explicit-module-build` and ensures an executable is produced without errors. 

The test runs for a long time because it must compile the `Swift` module, currently just testing how much overhead it will be with CI's toolchain.

### DIFF
--- a/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
+++ b/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
@@ -151,9 +151,6 @@ final class ExplicitModuleBuildTests: XCTestCase {
         $0 <<< "import G;"
       }
 
-      var pcmArgs9 = ["-Xcc","-target","-Xcc","x86_64-apple-macosx10.9"]
-      var pcmArgs15 = ["-Xcc","-target","-Xcc","x86_64-apple-macosx10.15"]
-
       let packageRootPath = URL(fileURLWithPath: #file).pathComponents
           .prefix(while: { $0 != "Tests" }).joined(separator: "/").dropFirst()
       let testInputsPath = packageRootPath + "/TestInputs"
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci please test

artemcm:
@swift-ci test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
The linux failure is due to the fact that the dependency scanning action is invoked with `-disable-objc-interop`, which causes the scanner to omit emitting a `SwiftShims` dependency for the `Swift` module.

DougGregor:
> The linux failure is due to the fact that the dependency scanning action is invoked with `-disable-objc-interop`, which causes the scanner to omit emitting a `SwiftShims` dependency for the `Swift` module.

Can you conditionality the test?

artemcm:
> > The linux failure is due to the fact that the dependency scanning action is invoked with `-disable-objc-interop`, which causes the scanner to omit emitting a `SwiftShims` dependency for the `Swift` module.
> 
> Can you conditionalize the test?

Do you mean make it run on macOS only?

DougGregor:
> Do you mean make it run on macOS only?

No, I mean either don't check for SwiftShims, or only check for SwiftShims on Darwin

artemcm:
I don't think I can change this at the test-level. The test compiles the `Swift` module, which will look for a `SwiftShims` module but will not find it (and error out) because 1. Implicit lookup is disabled, 2. `SwiftShims` is not provided explicitly because the scanner omitted it due to `disable_objc_interop` on Linux.

artemcm:
@DougGregor there's not a good way to get the dependency scanner working on Linux until Clang starts providing a by-name module lookup API. While that's being built, should we check this in for macOS only for now?

DougGregor:
I'm fine with checking only for macOS now while we await a Clang API to do by-name module lookup. That `.m` file with an `@import` is a really gross hack.

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #124, opened 2020-06-10
Remove @testable import from explicit module builds test

owenv:
`@testable` may make eventual toolchain integration more difficult (because we'd have to build once for testing and again for integrating), so remove it from the one place it's currently used. This requires making the module dependency graph structures public, which IMO is an ok tradeoff.

### DIFF
--- a/Sources/SwiftDriver/Dependency Scanning/InterModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/Dependency Scanning/InterModuleDependencyGraph.swift
@@ -12,11 +12,11 @@
 import Foundation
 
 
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci test

owenv:
I think `@_spi` could help, but I don't think we should consider using it until there's a version of Xcode shipping with swift 5.3. It's already required for all the tests to pass, but ideally people shouldn't need a snapshot to build the driver.

DougGregor:
I'm going to go ahead and merge this to wipe out the `@testable`. Thanks!

PR #126, opened 2020-06-13
[Explicit Module Builds] Add support for versioned PCMs

artemcm:
There exists a strict limitation that a compatible PCM must have the exact same target version with the loading module (either a Swift module or another PCM). This means that each PCM may have to be built multiple times, once for each unique (SDK, architecture, toolchain) tuple of some loading module.

This PR introduces support for generating versioned PCMs for each loading module.
Now there is more state that we needed to keep track of when generating jobs for module dependencies: a cache of which modules have already been built). This state, along with the Inter-Module Dependency Graph, now live in the `ExplicitModuleBuildHandler`, which is constructed on-demand by the driver in the explicit module build mode. The handler is responsible for generating dependency build jobs and for vending out dependency modules as inputs/command-line options whenever the driver creates a compile job in Explicit Module Build Mode.

This PR depends on Xi's work in: https://github.com/apple/swift/pull/32304
Resolves rdar://problem/63944640


artemcm:
Until we merge Xi's PR in apple/swift, the tests here will not pass. We can orchestrate the tandem submission once this review is complete.

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
```diff
@@ -0,0 +1,370 @@
...
+  /// 
+  /// Along the way, for any dependency module (direct or transitive) for which no build job
+  /// has yet been generated, generate one and cache it.
+  mutating private func resolveModuleDependencies(moduleId: ModuleDependencyId,
+                                                  pcmArgs: [String],
+                                                  inputs: inout [TypedVirtualPath],
+                                                  commandLine: inout [Job.ArgTemplate]) throws {
+    // Prohibit the frontend from implicitly building textual modules into binary modules.
+    commandLine.appendFlags("-disable-implicit-swift-modules", "-Xcc", "-Xclang", "-Xcc",
+                            "-fno-implicit-modules")
```
nkcsgexi, line 221: Do we need to append these flags here? As long as we have these flags in the scanning action, they will show up in each reported build commands from the JSON file.

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
```diff
@@ -0,0 +1,370 @@
...
+    try resolveModuleDependencies(moduleId: moduleId,
+                                  pcmArgs: dependencyGraph.swiftModulePCMArgs(of: moduleId),
+                                  inputs: &inputs, commandLine: &commandLine)
+
+    // Build the .swiftinterfaces file using a list of command line options specified in the
+    // `details` field.
+    guard let moduleInterfacePath = moduleDetails.moduleInterfacePath else {
+      throw Driver.Error.malformedModuleDependency(moduleId.moduleName,
+                                                   "no `moduleInterfacePath` object")
+    }
```
nkcsgexi, line 150: Assuming we always need to build a Swift module from its textual interface is good for now. But in the future, keep in mind that we need to handle available `.swiftmodule` files in the dependency graph, e.g. a `.swiftmodule` built from the source of another target. In this case, we don't need to schedule any actions but passing down the `.swiftmodule` itself to the `swiftModuleBuildCache`.

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
```diff
@@ -0,0 +1,370 @@
...
+    commandLine.appendFlags("-Xcc", "-Xclang", "-Xcc",
+                            "-fmodule-file=\(clangModulePath.file.description)")
+    inputs.append(clangModulePath)
+    inputs.append(clangModuleMapPath)
+
+    // Process all transitive dependencies as direct
+    for transitiveDependencyId in dependencyInfo.directDependencies {
+      try addClangModuleDependency(moduleId: moduleId, dependencyId: transitiveDependencyId,
+                                   pcmArgs: pcmArgs, inputs: &inputs, commandLine: &commandLine)
+    }
```
nkcsgexi, line 310: Can we lift this for loop to `addModuleDependency`?

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
```diff
@@ -0,0 +1,370 @@
...
+                                           type: .swiftModule)
+    commandLine.appendFlags("-swift-module-file")
+    commandLine.appendPath(swiftModulePath.file)
+    inputs.append(swiftModulePath)
+
+    // Process all transitive dependencies as direct
+    for transitiveDependencyId in dependencyInfo.directDependencies {
+      try addModuleDependency(moduleId: moduleId, dependencyId: transitiveDependencyId,
+                              pcmArgs: pcmArgs, inputs: &inputs, commandLine: &commandLine)
+    }
```
nkcsgexi, line 272: Can we lift this for loop to `addModuleDependency`?

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
```diff
@@ -0,0 +1,370 @@
...
+    let modifiedModulePath =
+      moduleInfo.modulePath.replacingOccurrences(of: plainModulePath.basenameWithoutExt,
+                                                 with: targetEncodedBaseName)
+    return try VirtualPath(path: modifiedModulePath)
+  }
+
+  /// Compute the name of a given Clang module, with the target triple encoded in the name.
+  static func targetEncodedClangModuleName(for moduleName: String,
+                                           pcmArgs: [String]) throws -> String {
+    return moduleName + "-" + pcmArgs.filter({ $0 != "-Xcc" && $0 != "-target"}).joined()
```
nkcsgexi, line 333: Can we change this to use a hash of all the strings in `pcmArgs`? That way if we have more `-target`-like arguments that need duplicate PCMs, the existing code should just work.

artemcm: Yes! Good suggestion, was just talking about this with Doug.

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
artemcm: The scanning action does not currently include these flags, but it can. 
This code is also used when the driver is generating build commands for the main module (outside of dependency build command generation) and there we don't really have a choice but have the driver add them.

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
artemcm: Yeah, I realized this recently as well. I will track getting pre-built `.swiftmodule` files working as a followup. 👍

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
nkcsgexi: I see. If that's the case, it probably won't provide much value by moving it to the scanning args because we have to add them explicitly any way.

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
nkcsgexi: That'll be awesome! We could potentially enable explicit PCMs first for now because it's much simpler and happens way more often than building `.swiftmodule`.

Sources/SwiftDriver/Explicit Module Builds/ClangModuleBuildJobCache.swift
```diff
@@ -0,0 +1,38 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Maps tuples (ModuleDependencyId, [String]), where [String] is an array of pcm build opitions,
+/// to the job used to build this module with these specified build options.
+struct ClangModuleBuildJobCache {
+
+  /// As tuples cannot be Hashable, wrap (ModuleDependencyId, [String]) into a struct pair
+  private struct ModuleArgumentPair : Hashable {
```
DougGregor, line 18: This makes me wonder whether there is a more fundamental problem here with the `ModuleDependencyId` abstraction. Right now, it assumes that the Clang modules are uniquely identified by their name within a Swift compiler job, but that's not true. Rather, they are uniquely identified by their name and by the extra build args needed to produce that variant of the model. How about pulling that into `ModuleDependencyId` directly:

```swift
enum ModuleDependencyId: Hashable {
  case swift(String)
  case clang(moduleName: String, buildArguments: [String])
}
```

then `ClangModuleBuildJobCache` would not have to exist and, more importantly, it wouldn't be possible to make the mistake of using the "wrong" Clang module or getting them confused, because `ModuleDependencyId` provides the right notion of identity. (This probably should be reflected in the JSON output from the dependency scanner, too).

Tests/SwiftDriverTests/Inputs/ExplicitModuleDependencyBuildInputs.swift
```diff
@@ -39,6 +39,12 @@ enum ModuleDependenciesInputs {
           ],
           "details": {
             "swift": {
+              "extraPcmArgs": [
+                "-Xcc",
+                "-target",
```
DougGregor, line 44: I think we should use `-target x86_64-apple-macosx10.15` here, as a Swift command-line argument, rather than avoiding Swift's option handling with `-Xcc`. Swift will translate that down to the appropriate `-target` for Clang automatically.

nkcsgexi: To explain why we need those `-Xcc`, the way Swift compiler emits those PCM build arguments is to solely reply on extra Clang arguments to instantiate a Clang instance rather than inferring any of them from Swift arguments. We add a new Swift front-end flag `-only-use-extra-clang-opts` to ensure no Swift arguments are translated. This is why we have to use `-Xcc` here.

Sources/SwiftDriver/Explicit Module Builds/ClangModuleBuildJobCache.swift
artemcm: Then the logic of "discovering" which clang module versions must be built would need to be moved to the dependency scanner. The dependency scanner would then need to either mutate the existing graph to have versioned PCMs or generate a versioned-PCM graph from the start, in a way that is similar to how "discovery" is done here.

DougGregor: It seems to me like the dependency scanner should be generating the versioned-PCM graph, because that's an accurate representation of the dependencies. Right now, the graph that comes out of the dependency scanner does not accurately represent the modules that need to be built.

Tests/SwiftDriverTests/Inputs/ExplicitModuleDependencyBuildInputs.swift
DougGregor: This makes me a little nervous, because a bunch of the setup logic for the Clang importer looks at the Swift options. Is all of that meant to be disabled? Should we change the way we emit PCM's from Swift entirely?

Sources/SwiftDriver/Explicit Module Builds/ClangModuleBuildJobCache.swift
nkcsgexi: I have experimented with constructing a graph from the compiler to include all versioned PCMs. The reason for me to not choose that alternative is that all transitive PCM dependencies have to be pulled into directive ones so that they can have unique command line arguments, which is not an accurate representation either.

DougGregor: The uniqued command line arguments are what make them unique; that's why I'm advocating for it to be part of module identity. Each module node in the dependency graph should correspond to exactly one module file that needs to be built. Anything else leaves too much interpretation to the client, and we'd rather keep the clients simple.

Tests/SwiftDriverTests/Inputs/ExplicitModuleDependencyBuildInputs.swift
nkcsgexi: > Is all of that meant to be disabled?

No, they will be preserved as long as these flags are specified in the scanning action. When we are collecting these Clang arguments, they are collecting the post-configuration arguments in the scanning action.

> Should we change the way we emit PCM's from Swift entirely?

hmm, It seems to me the existing mechanism makes more sense because generating Swift arguments always translate to the same set of Clang arguments is another layer of complexity.

nkcsgexi: Another reason for using Clang arguments here is that we are collecting them already in `ClangImporter.cpp`. Re-using these arguments directly seems to me to be less error-prone.

Sources/SwiftDriver/Explicit Module Builds/ClangModuleBuildJobCache.swift
DougGregor: Okay, I understand the "transitive dependency" issue now. This is fine.

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

DougGregor:
🚀

PR #127, opened 2020-06-16
Fix -debug-prefix-map handling

keith:
Previously only the values were passed through not the options as well.
This also adds the validation from Driver.cpp that they are correctly
formatted.


DougGregor:
@swift-ci test

PR #128, opened 2020-06-16
Add -coverage-prefix-map arg

keith:
This is analogous to this change in Driver.cpp https://github.com/apple/swift/pull/32175

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -296,6 +296,7 @@ public struct Driver {
     self.numParallelJobs = Self.determineNumParallelJobs(&parsedOptions, diagnosticsEngine: diagnosticEngine, env: env)
 
     try Self.validateWarningControlArgs(&parsedOptions)
+    Self.validateCoverageArgs(&parsedOptions, diagnosticsEngine: diagnosticEngine)
 
     // Compute debug information output.
     self.debugInfo = Self.computeDebugInfo(&parsedOptions, diagnosticsEngine: diagnosticEngine)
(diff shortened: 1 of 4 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -296,6 +296,7 @@ public struct Driver {
     self.numParallelJobs = Self.determineNumParallelJobs(&parsedOptions, diagnosticsEngine: diagnosticEngine, env: env)
 
     try Self.validateWarningControlArgs(&parsedOptions)
+    Self.validateCoverageArgs(&parsedOptions, diagnosticsEngine: diagnosticEngine)
```
keith, line 299: I didn't love creating a new function for this but there wasn't another obvious place to put it

Sources/SwiftOptions/Options.swift
```diff
@@ -31,6 +31,7 @@ extension Option {
   public static let colorDiagnostics: Option = Option("-color-diagnostics", .flag, attributes: [.frontend, .doesNotAffectIncrementalBuild], helpText: "Print diagnostics in color")
   public static let compileModuleFromInterface: Option = Option("-compile-module-from-interface", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Treat the (single) input as a swiftinterface and produce a module", group: .modes)
   public static let continueBuildingAfterErrors: Option = Option("-continue-building-after-errors", .flag, attributes: [.frontend, .doesNotAffectIncrementalBuild], helpText: "Continue building, even after errors are encountered")
+  public static let coveragePrefixMap: Option = Option("-coverage-prefix-map", .separate, attributes: [.frontend], helpText: "Remap source paths in coverage info")
```
keith, line 34: I did this manually because I couldn't get the option generation working, but I think I got the right stuff here

Sources/SwiftDriver/Driver/Driver.swift
DougGregor: Your comment made me think about an alternative approach. `validateCoverageArgs` could be `parseCoverageArgs` that parses the args, diagnoses failures, and then returns an array of `(original, remapped)` pairs. Then, we render that array to arguments in `addCommonFrontendOptions`. What you do you think about that approach?

keith: (I'm open to make any changes like that) Gut feeling is it might not be worth adding that much processing to these options considering we don't even handle them, it feels like it could increase the surface area for bugs around this

DougGregor: Sure. I'm fine with treating this as a "validate" call like you have it, but wanted to float the idea in case you preferred this alternative design. I don't feel strongly about it.

Sources/SwiftOptions/Options.swift
DougGregor: Looks fine. We can do a separate regeneration pass.

DougGregor:
@swift-ci please test

PR #129, opened 2020-06-18
Link runtime compatibility libs on darwin if needed

owenv:
This also fixes Interpreter/objc_class_properties_runtime from the integration tests


owenv:
@swift-ci test

owenv:
@DougGregor that's a good idea I hadn't thought of, it would be nice to avoid duplication between the driver & frontend. I'll look into integrating `-print-target-info` in a follow-up, I think it'll be easier now that there are a few unit tests to verify against.

PR #131, opened 2020-06-18
Do not force-unwrap Job displayInputs in Job description.

artemcm:
It seems that Job's clients are under no obligation to supply any inputs here.


artemcm:
@swift-ci please test

artemcm:
> Shouldn’t all those jobs have inputs? Seems like a fine change to me just for safety, but also seems like we shouldn’t hit this.

I ran into this with jobs that have `inputs`, but do not specify `displayInputs`.
The few examples in the driver:
- `generateDSYMJob`
- `generatePCHJob`
- `generatePCMJob`

We probably ought to fix those up too, for verbose output/debugging purposes, but I thought it made sense that the API does not require this to be present.

PR #132, opened 2020-06-19
[Build] Add a helper for integration into Swift's build-script.

DougGregor:
Allows building, testing, and installing swift-driver and swift-help
from Swift's build-script. Note that the installed swift-driver does
not yet replace `swift` or `swiftc`.


DougGregor:
@swift-ci please test

PR #134, opened 2020-06-20
Sketch out model for capturing JSON output from jobs

owenv:
This is an attempt to generalize the dependency scanning operation so we can model it as a `Job`. There's no real benefit at the moment, but I'd like to reuse this pattern to capture `-print-target-info` output for the reasons mentioned in https://github.com/apple/swift-driver/pull/129#pullrequestreview-432926884. I think it could also make delegating the job's execution to a library client easier if that's something we'd eventually like to do.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -24,11 +24,12 @@ public struct Driver {
     case subcommandPassedToDriver
     case integratedReplRemoved
     case conflictingOptions(Option, Option)
+    case failedToReadJobOutput
+    case jobFailedWithNonzeroExitCode(Int, String)
     // Explicit Module Build Failures
     case malformedModuleDependency(String, String)
     case missingPCMArguments(String)
     case missingModuleDependency(String)
-    case dependencyScanningFailure(Int, String)
 
     public var description: String {
       switch self {
@@ -54,8 +55,10 @@ public struct Driver {
         return "Missing extraPcmArgs to build Clang module: \(moduleName)"
       case .missingModuleDependency(let moduleName):
         return "Missing Module Dependency Info: \(moduleName)"
-      case .dependencyScanningFailure(let code, let error):
-        return "Module Dependency Scanner returned with non-zero exit status: \(code), \(error)"
+      case .jobFailedWithNonzeroExitCode(let code, let error):
+        return "job returned with non-zero exit status: \(code), \(error)"
+      case .failedToReadJobOutput:
+        return "failed to read output of job"
       }
     }
   }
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -697,6 +700,38 @@ extension Driver {
...
+
+    for (envVar, value) in job.extraEnvironment {
+      try ProcessEnv.setVar(envVar, value: value)
+    }
+
+    let process = try Process.launchProcess(arguments: arguments, env: env)
+    let result = try process.waitUntilExit()
+
+    if (result.exitStatus != .terminated(code: 0)) {
+      var returnCode = 0
```
cltnschlosser, line 719:
```suggestion
    if result.exitStatus != .terminated(code: EXIT_SUCCESS) {
      let returnCode: Int
```

owenv:
@swift-ci test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -697,6 +700,38 @@ extension Driver {
...
+  func execute<T: Decodable>(job: Job,
+                             capturingJSONOutputAs outputType: T.Type,
+                             resolver: ArgsResolver,
+                             forceResponseFiles: Bool) throws -> T {
+    let arguments: [String] =
+      try resolver.resolveArgumentList(for: job,
+                                       forceResponseFiles: forceResponseFiles)
+
+    for (envVar, value) in job.extraEnvironment {
+      try ProcessEnv.setVar(envVar, value: value)
```
DougGregor, line 712: We should be passing `job.extraEnvironment` down to `Process.launchProcess`, not setting it in the current process, right?

Sources/SwiftDriver/Explicit Module Builds/ModuleDependencyScanning.swift
```diff
@@ -40,24 +34,13 @@ extension Driver {
...
-        case .signalled(let signal):
-          returnCode = Int(signal)
-      }
-      throw Error.dependencyScanningFailure(returnCode, try result.utf8stderrOutput())
-    }
-    guard let outputData = try? Data(result.utf8Output().utf8) else {
-      return nil
-    }
-    return try JSONDecoder().decode(InterModuleDependencyGraph.self, from: outputData)
+    return Job(moduleName: moduleOutputInfo.name,
```
DougGregor, line 37: I definitely like this notion of making the scan-dependencies job a real Job...

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -222,10 +222,13 @@ extension Driver {
   /// Prescan the source files to produce a module dependency graph and turn it into a set
   /// of jobs required to build all dependencies.
   public mutating func generateExplicitModuleBuildJobs() throws -> [Job] {
-    let interModuleDependencyGraph = try computeModuleDependencyGraph()
-    guard let dependencyGraph = interModuleDependencyGraph else {
-      fatalError("Attempting to perform Explicit Module Build job generation, but the Inter Module Dependency Graph does not exist.")
-    }
+    let dependencyScannerJob = try dependencyScanningJob()
+    let forceResponseFiles = parsedOptions.hasArgument(.driverForceResponseFiles)
+    let resolver = try ArgsResolver(fileSystem: self.fileSystem)
```
DougGregor, line 227: The intent is that this is a big FIXME, right? We don't want to be creating an `ArgsResolver` within planning any more than we want to launch processes directly, but this is closer to the intended final form?

This is reinforcing to me that we have a layering issue in `SwiftDriver`, because the `Job` creation and planning stuff should really be completely separate from the job execution stuff, probably in a separate library entirely.

Sources/SwiftDriver/Driver/Driver.swift
owenv: yeah, that was a mistake

Sources/SwiftDriver/Jobs/Planning.swift
owenv: I added a comment to make that clear. At some point I think we'll want to delegate execution to library clients somehow, I don't have a specific design in mind though. I was thinking that separating the execution from job creation would make that eventual refactor easier

DougGregor: Yes, the part that's not totally clear to me is how execution of the scan-dependencies job can then create a number of other jobs.

tschuett: Maybe there are jobs that you need to execute during planning: getFrontendTargetInfo and scan-dependencies.
And there are jobs for building the modules. The client could provide a delegate to planning, which allows you to execute planning jobs.

owenv:
Closing in favor of https://github.com/apple/swift-driver/pull/143 which is a more comprehensive refactor

PR #135, opened 2020-06-20
Remove the localFileSystem default argument from ArgsResolver

owenv:
I noticed the default argument was accidentally being used in a couple places and IMO it's not obvious the resolver holds a FileSystem, so this change drops it and passes the localFileSystem explicitly in tests.


owenv:
@swift-ci test

owenv:
Converting to draft because I just realized this will break SPM

owenv:
False alarm, I was looking at a checkout from before https://github.com/apple/swift-package-manager/pull/2750/ :)

DougGregor:
Thanks for finishing this. I got lazy after apple/swift-package-manager#2750

PR #137, opened 2020-06-21
More module output path fixes

owenv:
If the swiftmodule is the primary output, we need to use the exact path provided by `-o` to match the behavior of the c++ driver.

Fixes Interpreter/SDK/objc_bridge.swift


owenv:
@swift-ci test

cltnschlosser:
Must only be a few failing interpreter tests left now? I know several of the validation tests fail just due to differences in json formatting.

owenv:
@cltnschlosser yeah, I think we're down to just one last failing interpreter test on macOS - `shebang-env.swift`. That failure seems to have something to do with how `xcrun` is found, but I haven't had any luck fixing it yet.

PR #138, opened 2020-06-22
Add missing output file map error

owenv:
- Add an error message matching the c++ driver for missing/malformed output file maps
- add a missing exit(EXIT_FAILURE) to main.swift (pretty sure it's my fault that was missing)

Fixes: Driver/warnings_control and Driver/missing-ofm


owenv:
@swift-ci test

PR #140, opened 2020-06-24
Use swift-frontend as the Swift frontend executable.

DougGregor:
Look for the swift-frontend executable for the Swift frontend, rather
than the "swift" executable. If there is no swift-frontend, fall back
to "swift".

The Swift compiler recently changed to make the primary (non-symlink)
executable "swift-frontend", although it contains both the driver and
the frontend logic. In this new scheme, "swift" and "swiftc" are
symlinks to swift-frontend (to use the existing C++ driver), but can
be replaced with symlinks to "swift-driver" to use this new driver.
This allows a much simpler drop-in replacement strategy.


DougGregor:
@swift-ci please test

cltnschlosser:
Awesome!

PR #143, opened 2020-06-27
Refactor process management & execution into a DriverExecutor protocol

owenv:
This is my attempt at moving all process management and argument resolution outside of `Driver`. Any time the driver needs to launch a process or run job(s), it delegates to its `DriverExecutor`. `DriverExecutor` has 3 requirements:

1. Running a single job and capturing the output. This covers the dependency scanning/-print-target-info use case
2. Running multiple jobs, with job lifetime information communicated via a `JobExecutionDelegate`. This supports executing the build graph after planning. By extending `JobExecutionDelegate` in the future, I think we can also add support for enqueuing new jobs as jobs finish to support incremental builds
3. Running a trivial command line, to support calls to `xcrun` and other miscellaneous toolchain utilities that don't make sense as Jobs.

In this PR, I've removed most ArgumentResolvers and use of exec(), Process.launchProcess, etc. from the driver. The remaining uses are:
- The driver still constructs an ArgumentResolver if -driver-print-jobs is passed
- The driver still calls exec() directly for frontend/autolink-extract/modulewrap command lines

I left those refactors out of this PR since it's already pretty large.


owenv:
@swift-ci test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -218,16 +218,23 @@ public struct Driver {
...
+    executor: DriverExecutor? = nil
   ) throws {
     self.env = env
     self.fileSystem = fileSystem
 
     self.diagnosticEngine = diagnosticsEngine
+    self.executor = try executor ?? SwiftDriverExecutor(diagnosticsEngine: diagnosticsEngine,
+                                                        processSet: ProcessSet(),
+                                                        fileSystem: fileSystem,
+                                                        env: env)
```
owenv, line 237: This is an unfortunate hack, but it avoids a lot of test breakage

Sources/SwiftDriver/Execution/ArgsResolver.swift
```diff
@@ -0,0 +1,92 @@
...
+      }
+      resolvedArguments = [resolvedArguments[0], "@\(responseFilePath.pathString)"]
+      return true
+    }
+    return false
+  }
+
+  /// Remove the temporary directory from disk.
+  public func removeTemporaryDirectory() throws {
+    try fileSystem.removeFileTree(temporaryDirectory)
```
owenv, line 90: The only change to ArgsResolver was replacing a use of FileManager with fileSystem here. I moved so it's clearer it's not only used by the JobExecutor

Sources/SwiftDriver/Execution/DriverExecutor.swift
```diff
@@ -0,0 +1,140 @@
...
+    return try JSONDecoder().decode(outputType, from: outputData)
+  }
+}
+
+public protocol JobExecutionDelegate {
+  /// Called when a job starts executing.
+  func jobStarted(job: Job, arguments: [String], pid: Int)
+  
+  /// Called when a job finished.
+  func jobFinished(job: Job, result: ProcessResult, pid: Int)
```
owenv, line 73: For incremental builds, I think we'll want some way to enqueue new jobs from here

Sources/SwiftDriver/Execution/MultiJobExecutor.swift
```diff
@@ -154,18 +56,22 @@ public final class JobExecutor {
     /// The diagnostics engine to use when reporting errors.
     let diagnosticsEngine: DiagnosticsEngine
 
+    /// The type to use when launching new processes. This mostly serves as an override for testing.
+    let processType: ProcessProtocol.Type
```
owenv, line 60: I'm not very happy with this hack. It replaces the launchProcess requirement of JobExecutionDelegate, which doesn't really make sense after this refactor

Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
```diff
@@ -0,0 +1,53 @@
...
+      commandLine += [.flag("-target-variant"), .flag(targetVariant.triple)]
+    }
+
+    if let sdkPath = sdkPath {
+      commandLine += [.flag("-sdk"), .path(sdkPath)]
+    }
+
+    if let resourceDirPath = resourceDirPath {
+      commandLine += [.flag("-resource-dir"), .path(resourceDirPath)]
+    }
```
cltnschlosser, line 39: Any reason this changed from the following?
```
try commandLine.appendLast(.target, from: &parsedOptions)
try commandLine.appendLast(.targetVariant, from: &parsedOptions)
try commandLine.appendLast(.sdk, from: &parsedOptions)
try commandLine.appendLast(.resourceDir, from: &parsedOptions)
```

Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    return Job(
+      moduleName: "",
```
cltnschlosser, line 42: I'm sure this doesn't matter, but it was `moduleOutputInfo.name` before.

Sources/SwiftDriver/Driver/Driver.swift
DougGregor: One thing we could do here would be to have an `@_spi(Testing)` overload of `init` that fills in `localFileSystem` and `executor`. That way, external clients have to provide appropriate values here but tests don't have to change.

Sources/SwiftDriver/Execution/DriverExecutor.swift
DougGregor: Yeah, and the scan-dependencies job will end up doing something similar.

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -110,42 +112,19 @@ extension Toolchain {
...
-        targetVariant.triple
-      ]
-    }
-
-    let resultString = try Process.checkNonZeroExit(arguments: args)
-    guard let resultData = resultString.data(using: .utf8) else {
-      throw Driver.Error.malformedSwiftTargetInfo(resultString)
-    }
-    return try JSONDecoder().decode(FrontendTargetInfo.self, from: resultData)
+    return try executor.execute(job: printTargetInfoJob(target: target, targetVariant: targetVariant),
```
DougGregor, line 124: This is really nice, thank you for the abstraction here.

DougGregor:
I'm going to opportunistically merge this.

Sources/SwiftDriver/Execution/MultiJobExecutor.swift
owenv: 👍 , I think with some additional refactoring we can get rid of this, it's used by a handful of tests right now.

Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
owenv: This just allows the job creation to also be used by the toolchain to get the host target info when cross-compiling, by ignoring the target passed via the command lne.

PR #83, opened 2020-04-11
Create FileSystemPath (Absolute or Relative path)

cltnschlosser:
Alternative approach to #77 

VirtualPath is used for Inputs/Outputs, while FileSystemPath is used by the Toolchain for locating tools/resources. I'm not sure if this actually makes sense, but it results in cleaner filesystem access with less fatalErrors and preconditions.

PR #144, opened 2020-06-28
File list support

cltnschlosser:
Major changes:
- VirtualPath enum case addition: `case fileList(RelativePath, FileList)`
- ArgsResolver handles `fileList` case.
- ArgsResolver is now reference type since it should only write each file list once.


Sources/SwiftDriver/Jobs/Job.swift
```diff
@@ -63,6 +63,9 @@ public struct Job: Codable, Equatable, Hashable {
   /// The outputs produced by the job.
   public var outputs: [TypedVirtualPath]
 
+  /// The fileLists used by the job.
+  public var fileLists: [FileList]
```
DougGregor, line 67: It seems unfortunate that we have to model file lists as part of the job, especially because we have other kinds of similar files (e.g., response files, the explicit module info that @artemcm is working on, etc.) that also do this create-a-temporary-file-to-pass-to-the-job thing. 

I have an idea and I'd love to hear your thoughts on it. What if we extend `VirtualPath` with a new case for a temporary file that knows its contents, e.g.,

```swift
case temporaryContents(path: RelativePath, contents: [UInt8])
```

Such a `VirtualPath` would need to be listed as an input to any jobs that need the file to exist. Then, the `ArgsResolver` would be responsible for ensuring that it writes the contents into that file before executing the job. The thing I love about this is that it takes the filesystem out of the `Driver` instances---and lets the `ArgsResolver` do all of that filesystem interaction itself.

Sources/SwiftDriver/Utilities/FileList.swift
```diff
@@ -0,0 +1,91 @@
+//
+//  FileList.swift
```
DougGregor, line 2: Please use the standard header for new files

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -330,13 +330,19 @@ extension Driver {
...
+  /// to inputs and command line arguments of a compile job.
   func addExplicitModuleBuildArguments(inputs: inout [TypedVirtualPath],
                                        commandLine: inout [Job.ArgTemplate]) throws {
     guard var handler = explicitModuleBuildHandler else {
       fatalError("No handler in Explicit Module Build mode.")
     }
     try handler.resolveMainModuleDependencies(inputs: &inputs, commandLine: &commandLine)
   }
+
+  func createTemporaryFileName(prefix: String, suffix: String?) -> String {
```
DougGregor, line 343: I'm a bit nervous about this API, for the reasons for FIXME implies. The randomness can cause collisions, and if we introduce this API with incorrect semantics, we're liable to run into hard-to-reproduce issues down the line as it grows more clients.

I think we're missing some kind of `Driver`-level API that vends unique `VirtualPath.temporary` instances. Perhaps one can specify part of a name so the actual file names produced are meaningful (e.g., pass it "input-file-list" and it will produce "input-file-list-<something>").

cltnschlosser: Thoughts on saving a static int that we just increment everytime this is called? Currently each driver (ArgsResolver) creates its own temporary directory, so at the file level it only needs to be unique for each driver instance.

Sources/SwiftDriver/Jobs/Job.swift
cltnschlosser: I haven't gotten to that part yet, but I was planning on having ArgsResolver do the FileList writing. This array is to keep track of the info that goes into each file list.

Can't do the temporaryContents thing because only the ArgsResolver will know the resolved paths that need to be put in the file lists.

If we are okay with it being a recursive enum, we could have a
```swift
case fileList(path: RelativePath, contents: [VirtualPath])
```

Also currently the FileList type is an enum to match what the c++ driver does, but I think it may end up being a simple struct:
```swift
struct FileList {
  let path: VirtualPath
  let contents: [VirtualPath]
}
```
Which I guess is basically the Virtual path enum case mentioned before.

The C++ driver doesn't track the contents, but instead figures out what they are at the time of writing the file. This seems to lead to duplicated logic, so I like the job specifying which paths should be in the file lists.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
cltnschlosser: I also thought about using UUIDs, but that requires Foundation.

Sources/SwiftDriver/Jobs/Job.swift
cltnschlosser: I do like embedding this information into VirtualPath the more I'm working with this code. I think that would increase the need for something like #83 A separate for focused implementation (than VirtualPath) for paths used by the toolchain.

cltnschlosser: Still working through this, but now that I'm farther I don't know if a VirtualPath case will work because we have the all sources file list which is shared between all compile jobs. 

I guess the ArgsResolver could keep track of FileLists and make sure to only write them once. Does that sound reasonable?

DougGregor: I was thinking that ArgsResolver would keep track of what it's already written, so it doesn't write the same file twice

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
DougGregor: A `static int` is probably okay as a temporary hack.

cltnschlosser:
Okay, still a bit of work todo. Need to refactor the FileList object and ArgsResolver (also add the new VirtualPath case). Then need to write tests for everything, there are a lot of little edge cases caught by the lit tests.

Also I want to pull out a few changes into other smaller PRs.

Newly passing:
Driver/batch_mode_with_supplementary_filelist.swift
Driver/batch_mode_with_supplementary_filelist_unicode.swift
Driver/emit-interface.swift
Driver/filelist-threshold.swift

Mostly working (Just temporary file handling, and some path formatting(I can open swift PR for this)):
Driver/filelists.swift

Issues with Driver/filelists.swift:
- Extra temporary directory layer
- Temporary files not removed on failure (C++ driver tracks all temporary files and deletes them after it's done. Unless -save-temps is passed, or the job crashes and the files are marked to be saved on signal)

cltnschlosser:
@DougGregor I've hit a snag in the ArgsResolver managing which filelists are written. In order to make the ArgsList actually keep track of this state it needs to be a class (or needs to be provided access to something else that tracks this state). Because two compile jobs can use the same filelist, the ArgsResolver then needs to use locking. Previously two jobs could resolve their arguments entirely in parallel.

I'm going to move forward with the locking to see how it works out, but it does seem less than ideal.

cltnschlosser:
Okay, rebased this. Most tests are failing locally due to `could not decode frontend target info; compiler driver and frontend executables may be incompatible`. I'm using the latest toolchain on swift.org for xcode.

DougGregor:
> Okay, rebased this. Most tests are failing locally due to `could not decode frontend target info; compiler driver and frontend executables may be incompatible`. I'm using the latest toolchain on swift.org for xcode.

Maybe you're picking up the Swift compiler/frontend executable from the default toolchain. If you haven't don't so yet, I suggest overriding the environment variable `SWIFT_DRIVER_SWIFT_FRONTEND_EXEC` to point at the `swift-frontend` executable in the toolchain you downloaded. (And make sure it's a `master` toolchain; these changes aren't in 5.3)

cltnschlosser:
> Maybe you're picking up the Swift compiler/frontend executable from the default toolchain

Yep, this was it. I'm hard coding it now for testing. Looks like `xcrun --find` only looks inside the default toolchain.

There might be some legitimate failures.

cltnschlosser:
Environment variable works as well. Using that now :)

cltnschlosser:
I think the only TODO that is left is to port some of the lit tests to XCTest, so we have coverage here.

cltnschlosser:
Okay, rebased. Resolved concerns from comments and added tests. Should be all good now :)

owenv:
@swift-ci please test

owenv (approved):
Everything here looks good to me, thanks again for working on this! 

I do think we'll want to rework the temporary file naming at some point to avoid collisions better. TSC has a withTemporaryFile utility for this, but I think it has the same problem as withTemporaryDirectory - it always uses the local file system. I'm fine with merging this as-is as long as @DougGregor doesn't have any remaining concerns!

DougGregor (approved):
This looks good to me, thank you for picking it up again!

cltnschlosser:
Okay, merge when ready

compnerd:
This is breaking the swift-package-manager build :-(

owenv:
Looks like a switch needs to be updated to support the new case added to VirtualPath:
```
/home/buildnode/jenkins/workspace/oss-swift-incremental-RA-linux-ubuntu-16_04/swiftpm/Sources/Build/ManifestBuilder.swift:855:9: error: switch must be exhaustive
10:12:41         switch file {
10:12:41         ^
```

PR #145, opened 2020-06-28
Make sure single job plans are executed in place by the DriverExecutor

owenv:
This fixes a small regression introduced by #143  (stdlib/FloatingPointIR was failing)


owenv:
@swift-ci test

cltnschlosser:
Maybe we should separate out execute(job:) and executeInPlace in `DriverExecutor`? Since executeInPlace has a different api. One returns `ProcessResult` and the other is `throws -> Never`

PR #146, opened 2020-06-28
[ArgsResolver] Allow overriding temporary directory

aciidb0mb3r:


aciidb0mb3r:
@swift-ci test

aciidb0mb3r:
Not happy with this but it unblocks progress in swiftpm-on-llbuild2...

Sources/SwiftDriver/Execution/ArgsResolver.swift
```diff
@@ -21,15 +21,22 @@ public struct ArgsResolver {
   private let fileSystem: FileSystem
 
   /// Path to the directory that will contain the temporary files.
-  private let temporaryDirectory: AbsolutePath
+  // FIXME: We probably need a dedicated type for this...
+  private let temporaryDirectory: VirtualPath
```
cltnschlosser, line 25: SPM won't have an AbsolutePath? Where is it getting its temporaryDirectory from?

Also ArgsResolver has the FileSystem, so it should be able to resolve Absolute and Relative paths to an AbsolutePath. Could resolve it in init.

Sources/SwiftDriver/Execution/ArgsResolver.swift
```diff
@@ -76,17 +83,25 @@ public struct ArgsResolver {
              "Platform does not support response files for job: \(job)")
       // Match the integrated driver's behavior, which uses response file names of the form "arguments-[0-9a-zA-Z].resp".
       let responseFilePath = temporaryDirectory.appending(component: "arguments-\(abs(job.hashValue)).resp")
-      try fileSystem.writeFileContents(responseFilePath) {
-        $0 <<< resolvedArguments[1...].map{ $0.spm_shellEscaped() }.joined(separator: "\n")
+
+      // FIXME: Need a way to support this for distributed build systems...
```
cltnschlosser, line 87: What needs to be done to solve this FIXME? What's the constraint caused by distributed build systems?

Sources/SwiftDriver/Execution/ArgsResolver.swift
aciidb0mb3r: This is related to the swiftpm-on-llbuild2 work we're exploring for distributed builds. The architecture completely separates the command-line construction with the action execution platform so it can be distributed to remote workers. This effectively means that we can't work with absolute paths during rule construction time as that will change depending on the remote worker. See: https://github.com/apple/swiftpm-on-llbuild2/pull/12/files#diff-533c9e67383e009e171f5731551d8a20R46-R62

Sources/SwiftDriver/Execution/ArgsResolver.swift
aciidb0mb3r: See my other comment https://github.com/apple/swift-driver/pull/146#discussion_r446701078 but the main issue is that we have to be careful about performing disk IO operations that be used during the build. In this case, I think we just need to abstract this into a protocol so swiftpm-on-llbuild2 can control how and when to write the response file. I think it'll be a while before we'll need response files though.

aciidb0mb3r:
@swift-ci test

Sources/SwiftDriver/Execution/ArgsResolver.swift
cltnschlosser: Oh okay, I think that makes sense. I’ve been working on file lists which will have similar issues. One is shared by all jobs and the rest are job specific. The all sources file list would need to be written on all workers.

aciidb0mb3r:
Ugh, I'll have to re-create the PR because I messed up my forks

PR #147, opened 2020-06-28
[Driver] Allow passing OutputFileMap directly to the driver

aciidb0mb3r:


aciidb0mb3r:
Ignore the temporary directory commit. @DougGregor what do you think about exposing this? It makes it much easier to integrate and we don't have to go through the file system... See usage here: https://github.com/apple/swiftpm-on-llbuild2/pull/13/files#diff-1fd23c67dae9e4e23bfcb0ba11fc59d6R50-R58

aciidb0mb3r:
sgtm!

PR #148, opened 2020-06-29
Use the frontend's -print-target-info more often

DougGregor:
The Swift frontend provides a wealth of information about target triples, runtime compatibility, paths, and so on via its `-print-target-info`. Query `-print-target-info` and use it for several things:
* Canonicalizing target triples
* Determining the version of the Swift runtime that's guaranteed to be available for a given target triple
* Determining when to link against the OS-provided Swift standard library via rpaths.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -77,14 +77,19 @@ public struct Driver {
   /// The executor the driver uses to run jobs.
   public let executor: DriverExecutor
 
-  /// The target triple.
-  public let targetTriple: Triple
-
-  /// The variant target triple.
-  public let targetVariantTriple: Triple?
-
   /// The toolchain to use for resolution.
   public let toolchain: Toolchain
+
+  /// Information about the target, as reported by the Swift frontend.
+  let frontendTargetInfo: FrontendTargetInfo
+
+  /// The target triple.
+  public var targetTriple: Triple { frontendTargetInfo.target.triple }
+
+  /// The variant target triple.
+  public var targetVariantTriple: Triple? {
+    frontendTargetInfo.targetVariant?.triple
+  }
 
   /// The kind of driver.
   public let driverKind: DriverKind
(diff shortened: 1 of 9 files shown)


DougGregor:
@swift-ci please test

Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
```diff
@@ -10,6 +10,77 @@
...
+    let librariesRequireRPath: Bool
+  }
+
+  struct Paths: Codable {
+    let runtimeLibraryPaths: [String]
+    let runtimeLibraryImportPaths: [String]
+    let runtimeResourcePath: String
+  }
+
+  var target: Target
```
aciidb0mb3r, line 80: What do you think about exposing these models? I need similar information in swiftpm-on-llbuild2 and it'll be nice if I can just re-use them from the driver...OTOH this means bigger API surface area for the driver.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -728,8 +735,8 @@ extension Driver {
     print(result)
   }
 
-  private func printBindings(_ job: Job) throws {
-    try stdoutStream <<< #"# ""# <<< toolchain.hostTargetTriple().triple
+  private func printBindings(_ job: Job) {
+    stdoutStream <<< #"# ""# <<< targetTriple.triple
```
cltnschlosser, line 739: Was this incorrect before? Since it was always the hosts triple even when cross compiling?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1427,8 +1435,7 @@ extension Driver {
     } else if let SDKROOT = env["SDKROOT"] {
       sdkPath = SDKROOT
     } else if compilerMode == .immediate || compilerMode == .repl {
-      let triple = try? toolchain.hostTargetTriple()
```
cltnschlosser, line 1430: Was this incorrect before? Since it was always the hosts triple even when cross compiling?

Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
DougGregor: I'd be okay with exposing these models. It's better than having `swiftpm(-on-llbuild2)?` duplicating logic.

Sources/SwiftDriver/Driver/Driver.swift
DougGregor: I believe it was incorrect before; we should be printing based on what we're targeting, because the host should be irrelevant.

Sources/SwiftDriver/Driver/Driver.swift
DougGregor: Immediate & REPL modes are only really relevant when host and target are the same, so this is at worst a no-op and at best a bug fix.

PR #151, opened 2020-06-29
Pass the SDK path and resource directory through to -print-target-info.

DougGregor:


DougGregor:
@swift-ci please test

PR #154, opened 2020-07-01
[Explicit Module Builds] Avoid adding duplicate module dependency options to compile commands.

artemcm:
When traversing transitive dependencies of some module, there may be many paths that encountered the same module.
We currently ensure we only generate a compile job for each module once, but do not ensure each dependency gets
added as an input to the build command once. This PR fixes that.

### DIFF
--- a/Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
+++ b/Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
@@ -241,7 +241,9 @@ public struct ExplicitModuleBuildHandler {
                             "-fno-implicit-modules")
     var swiftDependencyArtifacts: [SwiftModuleArtifactInfo] = []
     var clangDependencyArtifacts: [ClangModuleArtifactInfo] = []
+    var alreadyAddedDependencies = Set<ModuleDependencyId>()
     try addModuleDependencies(moduleId: moduleId, pcmArgs: pcmArgs,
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci please test

Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
```diff
@@ -281,22 +283,29 @@ public struct ExplicitModuleBuildHandler {
   /// Add a specific module dependency as an input and a corresponding command
   /// line flag. Dispatches to clang and swift-specific variants.
   mutating private func addModuleDependencies(moduleId: ModuleDependencyId, pcmArgs: [String],
+                                              addedDependenciesSet: inout Set<ModuleDependencyId>,
                                               clangDependencyArtifacts: inout [ClangModuleArtifactInfo],
                                               swiftDependencyArtifacts: inout [SwiftModuleArtifactInfo]
   ) throws {
     for dependencyId in try dependencyGraph.moduleInfo(of: moduleId).directDependencies {
+      guard !addedDependenciesSet.contains(dependencyId) else {
```
DougGregor, line 291: I recommend using:

```swift
guard !addedDependenciesSet.insert(dependencyId). inserted else { ... }
```

here so you can remove the "insert" below.

artemcm: Oh neat! Will do a followup.

PR #156, opened 2020-07-03
Delegate job printing to the DriverExecutor since it involves argument resolution

owenv:
This gets rid of the last two ArgsResolvers outside of the executor


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #158, opened 2020-07-04
Drop the executor default argument value in Driver's init

owenv:
We shouldn't merge this until SPM stops relying on it (https://github.com/apple/swift-package-manager/pull/2798).


owenv:
Now that spm no longer relies on this, it should be safe to merge

@swift-ci test

owenv:
@swift-ci test macOS

PR #159, opened 2020-07-04
Make SwiftVersion struct Comparable

owenv:
This enables a tiny cleanup when choosing which compatibility libs to link


owenv:
@swift-ci test

PR #163, opened 2020-07-05
[Explicit Modules] Handle compiledModulePaths output by the dependency scanner

owenv:
As of https://github.com/apple/swift/pull/32633 when the scanner finds an up-to-date, already-compiled module next to the interface or cached, it reports a compiledModulePath in lieu of the usual dependency info. Instead of trying to generate a job, we should use this path directly.

I think this should fix the recent CI failures


owenv:
@swift-ci test

PR #164, opened 2020-07-06
[Explicit Modules] Provide an API for querying whether a job is a main module job.

artemcm:
This will be useful for SwiftPM (and other build systems) to distinguish intermediate dependency artifacts versus the main module build, in order to correctly pass in cross-target dependencies.


artemcm:
@swift-ci please test

PR #165, opened 2020-07-08
Gather more information from -print-target-info

DougGregor:
Adopt improvements to `-print-target-info` from https://github.com/apple/swift/pull/32758:
* Use the compiler version provided by `-print-target-info` rather than invoking a second time with `-version`
* Use the compatibility libraries provided by `-print-target-info` rather than hard-coding the known set.

PR #167, opened 2020-07-09
[Darwin] Miscellaneous toolchain cleanups and modernizations

DougGregor:
A couple of minor changes on top of https://github.com/apple/swift-driver/pull/165:
* Sink some macOS-specific APIs down into tests; I kept thinking they needed generalization, but really they needed not to be APIs at all
* Lookup "clang" via the normal approach, rather than only using the one from the Xcode default toolchain
* Stop inferring simulator-ness based on architecture.


DougGregor:
@swift-ci test

PR #168, opened 2020-07-10
Add -jit-build mode

hamishknight:
This new compilation mode emits a (symlink to a) stub executable and invokes the frontend with `-jit-build`, which will eventually allow it to register itself as a provider that can JIT code for the executable.

### DIFF
--- a/Sources/SwiftDriver/Jobs/FilesystemJobs.swift
+++ b/Sources/SwiftDriver/Jobs/FilesystemJobs.swift
@@ -1,0 +1,40 @@
+//===--------------- FilesystemJobs.swift - Generic Filesystem Jobs -------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension Driver {
+  /// Create a symbolic link.
+  func symlinkJob(from input: TypedVirtualPath, to output: VirtualPath) throws -> Job {
+    let invocation = try toolchain.getSymlinkInvocation(from: input.file, to: output)
+    return Job(
+      moduleName: moduleOutputInfo.name,
+      kind: .symlink,
+      tool: invocation.command,
+      commandLine: invocation.args,
+      displayInputs: [input],
+      inputs: [input],
+      outputs: [.init(file: output, type: input.type)]
+    )
+  }
+
+  /// Create a directory.
+  func mkdirJob(for path: VirtualPath) throws -> Job {
+    let invocation = try toolchain.getMkdirInvocation(for: path)
+    return Job(
+      moduleName: moduleOutputInfo.name,
+      kind: .mkdir,
+      tool: invocation.command,
+      commandLine: invocation.args,
+      inputs: [],
+      outputs: []
+    )
+  }
+}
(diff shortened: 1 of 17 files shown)


hamishknight:
@swift-ci please test

Sources/SwiftDriver/Jobs/JITBuildJob.swift
```diff
@@ -0,0 +1,88 @@
...
+      outputs: []
+    )
+  }
+
+  /// Retrieve the jobs needed to generate a stub executable with a frontend
+  /// process registered to JIT code for it.
+  public mutating func generateJITCompileJobs() throws -> [Job] {
+    // For now, get the stub executable from an environment variable. Eventually
+    // it'll be part of the toolchain.
+    let stubEnvVar = "SWIFT_DRIVER_JIT_STUB_EXEC"
```
owenv, line 44: If you add the stub executable to the Tool enumeration and add something like the following to `lookupToolPath` you can get the envvar override behavior for free and it won't require changes once the stub is part of the toolchain.
```
case .jitStub:
      return try lookup(executable: "jit-stub")
```

hamishknight: Ah nice, didn't realise the toolchain lookup logic had that, thanks!

Sources/SwiftDriver/Jobs/FilesystemJobs.swift
```diff
@@ -0,0 +1,53 @@
...
+
+extension Driver {
+  /// Create a symbolic link.
+  func symlinkJob(input: TypedVirtualPath, output: VirtualPath) throws -> Job {
+    var commandLine = [Job.ArgTemplate]()
+
+    let linkTool = try toolchain.getToolPath(.symlinker)
+
+    // If the linking tool is ln, we need to specify -s for a symbolic link.
+    if linkTool.basename == "ln" {
```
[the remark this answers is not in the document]
hamishknight: Yeah I agree that's much nicer, in fact let's add a `CommandInvocation` type and have it return that

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
@owenv Thanks for the review! While this probably could replace immediate mode, that's indeed not the main motivation for it – we want to try leverage the request evaluator by having a long-running frontend process where we can invalidate certain requests and only re-compute the minimal amount of work needed :)

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

Sources/SwiftDriver/Jobs/JITBuildJob.swift
```diff
@@ -0,0 +1,75 @@
...
+    let stubPath = TypedVirtualPath(file: .absolute(rawStubPath), type: .image)
+
+    // Compute the output path for the stub executable. If the user has
+    // specified a path with -o, use that, or otherwise infer it from the
+    // module name.
+    let stubOutputPath: VirtualPath
+    if let output = parsedOptions.getLastArgument(.o) {
+      stubOutputPath = try VirtualPath(path: output.asSingle)
+    } else {
+      let outputName =
```
artemcm, line 53: Can use the `outputFileForImage` property here:
```swift
if let output = parsedOptions.getLastArgument(.o) {
  stubOutputPath = try VirtualPath(path: output.asSingle)
} else {
  stubOutputPath = outputFileForImage
}
```

hamishknight: Ah nice, let's also factor out the check for `-o` into it :)

hamishknight:
@swift-ci please test

hamishknight:
@swift-ci please test

hamishknight:
Closing as the LLVM changes required to support this aren't ready yet

PR #170, opened 2020-07-11
Fix and re-run makeOptions

owenv:
A new supplementary output option attribute was added in the main compiler at some point, so this PR adds it to makeOptions as well so it continues to compile. I also made a handful of additions to the common frontend options.


owenv:
@swift-ci test

PR #171, opened 2020-07-12
Treat -frontend and -modulewrap invocations as subcommands

owenv:
This eliminates the last case where the driver was launching processes directly. IMO it makes sense to treat these flags as subcommands since they're direct frontend tool invocations that don't involve any coordination by the driver. I also dropped references to swift-autolink-extract because it's now a direct symlink to swift-frontend alongside swift-symbolgraph-extract.

### DIFF
--- a/Sources/swift-driver/main.swift
+++ b/Sources/swift-driver/main.swift
@@ -37,7 +37,7 @@ do {
     }
 
     // Execute the subcommand.
-    try exec(path: subcommandPath?.pathString ?? "", args: Array(arguments.dropFirst()))
+    try exec(path: subcommandPath?.pathString ?? "", args: arguments)
   }
 
   let executor = try SwiftDriverExecutor(diagnosticsEngine: diagnosticsEngine,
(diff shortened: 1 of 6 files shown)


Sources/swift-driver/main.swift
```diff
@@ -37,7 +37,7 @@ do {
     }
 
     // Execute the subcommand.
-    try exec(path: subcommandPath?.pathString ?? "", args: Array(arguments.dropFirst()))
+    try exec(path: subcommandPath?.pathString ?? "", args: arguments)
```
owenv, line 40: I think the use of dropFirst here was a bug -- I had to remove it to ensure args[0] was set to the program name as expected

owenv:
@swift-ci test

Sources/swift-driver/main.swift
cltnschlosser: No that makes sense because you're dropping the subcommand flag and forwarding the rest of the flags. Ex: `swift help --flag` goes to `swift-help --flag` and not `swift-help help --flag`

owenv: I think we're already doing that correctly when computing the `InvocationRunMode`. I added some logging to this locally and it looks like we were converting `swift package update` to `swift-package update` correctly and then dropping the first argument and ending up with just `update`.

PR #175, opened 2020-07-18
Add -sanitize=scudo support

owenv:
This is enough to get all of sanitize_scudo.swift passing except for the Windows test cases, and one that relies on the sanitizer ordering. Rather than fix the ordering here, I'm going to change the integrated driver to match swift-driver.


owenv:
@swift-ci test

PR #176, opened 2020-07-19
Add support for remaining optimization record options

owenv:
- -save-optimization-record= is distinct from -save-optimization-record and has some unusual behavior to support different record types
- -save-optimization-record-passes just needs to be added to compile jobs

This change fixes Driver/opt-record.swift and Driver/opt-record-options.swift


owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
Recently it looks like Linux CI has been taking ~12 minutes to run `testUsingResponseFiles` 🤔 . I wonder if there's a JSONEncoder performance regression on Linux. I'm going to submit a PR to decrease the size of response files used in that test

PR #177, opened 2020-07-19
Validate Darwin deployment targets and target variants

owenv:
This fixes Driver/os-deployment.swift and slightly improves Driver/macabi-environment.swift. There's some remaining Darwin-specific argument validation left to be implemented, which I'll create a starter bug for

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -121,26 +121,34 @@ public final class DarwinToolchain: Toolchain {
     """
   }
 
-  enum ToolchainValidationError: Error, DiagnosticData {
+  public enum ToolchainValidationError: Error, DiagnosticData {
     case osVersionBelowMinimumDeploymentTarget(String)
     case iOSVersionAboveMaximumDeploymentTarget(Int)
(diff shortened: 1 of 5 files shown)


owenv:
@swift-ci test

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -119,4 +120,62 @@ public final class DarwinToolchain: Toolchain {
...
+    // TODO: Validating darwin unsupported -static-stdlib argument.
+    // TODO: If a C++ standard library is specified, it has to be libc++.
+  }
+
+  func validateDeploymentTarget(_ parsedOptions: inout ParsedOptions,
+                                targetTriple: Triple) throws {
+    // Check minimum supported OS versions.
+    if targetTriple.isMacOSX,
+       targetTriple.version(for: .macOS) < Triple.Version(10, 9, 0) {
+      throw ToolchainValidationError.osVersionBelowMinimumDeploymentTarget("OS X 10.9")
```
DougGregor, line 162: What's this "OS X" thing? :) We should switch to macOS both here and in the integrated driver

owenv: 👍 , this error message could definitely use an update

PR #178, opened 2020-07-19
Shrink response file test input size

owenv:
This test runs extremely slowly on Linux CI (~12 min.). Shrink the input size while the performance regression is investigated (I'm guessing it's a JSONEncoder issue or a missed optimization on `(1...500_000).map { "-DTEST_\($0)" }`).


owenv:
@swift-ci test

owenv:
`13:00:51 Test Case 'SwiftDriverTests.testUsingResponseFiles' passed (2.136 seconds)`

PR #180, opened 2020-07-23
RFC: New Job discovery mechanism

owenv:
~This isn't complete yet, but I figured I'd post it now to get some feedback.~ It's intended to support job discovery based on the output of preceding jobs for use in explicit module/incremental builds.

1. An executor can now have > 1 JobExecutionDelegate. This allows, e.g. one for parseable output and one for incremental build computation
2. JobExecutionDelegate.jobFinished can now return newly discovered jobs

~The problem I'm running into now is creating new ExecuteJobRule keys for discovered jobs because jobs are no longer serialized as part of the key, they're referenced by index instead. Undoing that change would solve the problem but introduce a pretty big performance regression.~

Also, if we use this approach, we'll probably want planBuild to return a list of "mandatory execution delegates" in addition to the list of jobs. If a client like SwiftPM wants to merge those jobs into a larger build graph it'll require some additional record keeping to ensure the right delegates get notified about the status of jobs and have a chance to schedule new work.

### DIFF
--- a/Sources/SwiftDriver/Execution/DriverExecutor.swift
+++ b/Sources/SwiftDriver/Execution/DriverExecutor.swift
@@ -72,8 +72,8 @@ public protocol JobExecutionDelegate {
   /// Called when a job starts executing.
   func jobStarted(job: Job, arguments: [String], pid: Int)
   
-  /// Called when a job finished.
-  func jobFinished(job: Job, result: ProcessResult, pid: Int)
+  /// Called when a job finished. May return newly discovered jobs.
+  func jobFinished(job: Job, result: ProcessResult, pid: Int) -> [Job]
 }
 
 public final class SwiftDriverExecutor: DriverExecutor {
--- a/Sources/SwiftDriver/Execution/MultiJobExecutor.swift
+++ b/Sources/SwiftDriver/Execution/MultiJobExecutor.swift
@@ -18,13 +18,13 @@ import Dispatch
 public final class MultiJobExecutor {
 
   /// The context required during job execution.
-  struct Context {
+  final class Context {
 
     /// This contains mapping from an output to the index(in the jobs array) of the job that produces that output.
     let producerMap: [VirtualPath: Int]
 
     /// All the jobs being executed.
-    let jobs: [Job]
+    var jobs: [Job]
 
     /// The resolver for argument template.
     let argsResolver: ArgsResolver
@@ -94,7 +94,7 @@ public final class MultiJobExecutor {
   /// The argument resolver.
   let argsResolver: ArgsResolver
 
-  /// The job executor delegate.
+  /// The job executor delegates.
   let executorDelegate: JobExecutionDelegate
 
   /// The number of jobs to run in parallel.
@@ -194,7 +194,7 @@ struct JobExecutorBuildDelegate: LLBuildEngineDelegate {
   func lookupRule(rule: String, key: Key) -> Rule {
     switch rule {
     case ExecuteAllJobsRule.ruleName:
-      return ExecuteAllJobsRule(key, jobs: context.jobs, fileSystem: context.fileSystem)
+      return ExecuteAllJobsRule(key, context: context)
     case ExecuteJobRule.ruleName:
       return ExecuteJobRule(key, context: context)
     default:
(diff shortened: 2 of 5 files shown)


cltnschlosser:
The key doesn’t have to be an index, we can use some other identifier. We just need to be able to get the job from that identifier. I haven’t looked at this too closely yet, so maybe this doesn’t apply, but we could have some dictionary or other lookup mechanism and give each job a UUID when created that gets used as a key. Not sure if that would solve the problem. But just wanted to point out that we have more options than index and full job.

owenv:
You were right @cltnschlosser, in the end I was able to get this working without changing the indexing much at all, I just needed to make the context mutable 👍. I think I have a workable approach now.

owenv:
@swift-ci test

owenv:
@swift-ci test

Sources/SwiftDriver/Execution/MultiJobExecutor.swift
```diff
@@ -18,13 +18,13 @@ import Dispatch
 public final class MultiJobExecutor {
 
   /// The context required during job execution.
-  struct Context {
+  final class Context {
```
cltnschlosser, line 21: Can it still be a struct? Since it’s only modified in the ExecuteAllJobsRule? Then we don’t have to worry about child jobs mutating it at the wrong time.

owenv: I think we need reference semantics here to ensure all the rules are working with the same set of jobs. If this was a struct, there'd be no way to propagate changes from ExecuteAllJobsRule up to the context in JobExecutorBuildDelegate so they could be used when constructing new ExecuteJobRules in lookupRule.

Sources/SwiftDriver/Execution/MultiJobExecutor.swift
```diff
@@ -346,7 +354,7 @@ class ExecuteJobRule: LLBuildRule {
 
       // Inform the delegate.
       context.delegateQueue.async {
-        context.executorDelegate.jobStarted(job: job, arguments: arguments, pid: pid)
+        context.executorDelegates.forEach { $0.jobStarted(job: job, arguments: arguments, pid: pid) }
```
DougGregor, line 357: I don't know if this is the only way in which multiple executor delegates will end up being used, where each delegate is independently invoked. Rather than carrying around an array of executor delegates that are meant to be independent, how about we leave the driver as taking only a single executor delegate---then we can provide a delegate wrapper type that composes multiple delegates together.

DougGregor:
I think this is generally the right direction for the Swift driver, but I'm concerned that SwiftPM cannot currently implement this delegate protocol. In my understanding, SwiftPM's build manifest is very "static" in nature, where we effectively need to build up the graph based on the `planBuild()` calls and then that graph gets written into the manifest. @artemcm has been dealing with this for explicit module builds, and the scheme we came up with there involves having SwiftPM carry information from a target to the targets that depend on it, which allows us to pre-construct all of the jobs. (It's a better scheme overall because it lets us plan further ahead, but we started down that path because we couldn't discover more module-build jobs later in the build).

I don't know what it would take to teach SwiftPM to support this delegate protocol fully. Perhaps @aciidb0mb3r could weigh in on that. Also, I wonder if SwiftPM-on-llbuild2 will address this anyway.

Now, we could certainly have two "modes" for the driver: one that plans everything ahead (useful for SwiftPM as it is today, `-###` output, and any other clients that want static build graphs), and one that does the more dynamic scheduling like you're proposing (and our incremental builds might need).

Sources/SwiftDriver/Execution/DriverExecutor.swift
```diff
@@ -120,14 +120,14 @@ public final class SwiftDriverExecutor: DriverExecutor {
   }
   
   public func execute(jobs: [Job],
-                      delegate: JobExecutionDelegate,
+                      delegates: [JobExecutionDelegate],
```
DougGregor, line 123: Longer comment below, but I think we should support multiple delegates through composition, not by passing in an array.

Sources/SwiftDriver/Execution/MultiJobExecutor.swift
owenv: 👍 that sounds good to me. I added a `ForwardingExecutionDelegate` for now which independently invokes a list of delegates. This is only used in tests atm, but we'll need it for incremental builds + parseable output.

owenv:
@DougGregor thanks for the feedback on this, I'm not all that familiar with the SwiftPM build process. I think we'll still need dynamic job discovery for incremental builds, but I don't know of any other cases where we can't plan the entire build in advance now that @artemcm has solved the problem for explicit modules, so maybe the only two modes we need are incremental/non-incremental.

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
This might not actually be the direction we want to go in for incremental builds, closing

PR #181, opened 2020-07-23
[Explicit Module Builds] Support explicit module builds with external dependencies.

artemcm:
This change introduces support for external dependencies in Explicit Module Build compilation mode.

Building a Swift module in Explicit Module Build mode requires passing all of its module dependencies as explicit arguments to the build command.

When the driver's clients (build systems, e.g. SwiftPM) are planning a build that involves multiple Swift modules, planning for each individual module may take place before its dependencies have been built. This means that the dependency scanning action will not be able to discover such modules. In such cases, the clients must provide the driver with information about such external dependencies, including the path to where their compiled `.swiftmodule` will be located, once built, and a full inter-module dependency graph for each such dependence.

The driver will pass down the information about such external dependencies to the scanning action, which will generate `placeholder` swift modules for them in the resulting dependency graph. The driver will then use the complete dependency graph provided by the client for each external dependency and use it to "resolve" the dependency's "placeholder" module.

This PR:
- Adds API for the clients to pass in external module dependency artifacts (module file path and dependency graph).
- Updates the Inter Module Dependency Graph to represent `swiftPlaceholder` dependencies.
- Updates the scanning action invocation to pass in information about external dependencies.
- Introduces the placeholder dependency resolution logic that updates the dependency graph by substituting in dependency graphs of external modules in place of placeholder modules.

This change works in tandem with the corresponding Dependency Scanner change in:
https://github.com/apple/swift/pull/33032
And a corresponding SwiftPM change:
https://github.com/apple/swift-package-manager/pull/2827

### DIFF
--- a/Sources/SwiftDriver/CMakeLists.txt
+++ b/Sources/SwiftDriver/CMakeLists.txt
@@ -8,7 +8,8 @@
 
 add_library(SwiftDriver
   "Explicit Module Builds/ClangModuleBuildJobCache.swift"
-  "Explicit Module Builds/ExplicitModuleBuildHandler.swift"   
+  "Explicit Module Builds/ExplicitModuleBuildHandler.swift"
+  "Explicit Module Builds/PlaceholderDependencyResolution.swift"
   "Explicit Module Builds/InterModuleDependencyGraph.swift"
   "Explicit Module Builds/ModuleDependencyScanning.swift"
   "Explicit Module Builds/ModuleArtifacts.swift"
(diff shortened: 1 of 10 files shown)


nkcsgexi:
Have we had a consensus about how to resolve cross-import overlays yet?

artemcm:
> Have we had a consensus about how to resolve cross-import overlays yet?

No, not yet, that needs further discussion.

nkcsgexi:
@DougGregor @brentdax What do you think of this plan: 

- In the `ExternalDependencyMap` passed down to the `-scan-dependency` action, `swift-driver` also specifies transitively imported Swift module names.
- Dependency scanner reads those transitive imports to resolve additional cross-import overlays

artemcm:
The toolchain snapshot build with the corresponding Dependency Scanner changes passed so the next created toolchain will allow this to be submitted. I expect it to become available tonight.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci test

PR #184, opened 2020-07-26
Validate C++ and ObjC related args for Darwin toolchain

Nikhil0487:
This commit  validates the following arguments to driver:
1. link-objc-runtime by checkin arc path,
2. experimentalCxxStdlib flag to limit it to libc++
3. -static-stdlib to flag that its not supported.

This commit is for SR-13255 https://bugs.swift.org/browse/SR-13255

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -123,8 +123,10 @@ public final class DarwinToolchain: Toolchain {
 
   public enum ToolchainValidationError: Error, DiagnosticData {
     case osVersionBelowMinimumDeploymentTarget(String)
+    case argumentNotSupported(String)
     case iOSVersionAboveMaximumDeploymentTarget(Int)
     case unsupportedTargetVariant(variant: Triple)
+    case darwinOnlySupportsLibCxx
 
     public var description: String {
       switch self {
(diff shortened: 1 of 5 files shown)


Nikhil0487:
@owenv

owenv:
@swift-ci test

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -134,24 +136,38 @@ public final class DarwinToolchain: Toolchain {
...
     try validateDeploymentTarget(&parsedOptions, targetTriple: targetTriple)
     if let targetVariantTriple = targetVariantTriple,
        !targetTriple.isValidForZipperingWithTriple(targetVariantTriple) {
       throw ToolchainValidationError.unsupportedTargetVariant(variant: targetVariantTriple)
     }
-
-    // TODO: Validating darwin unsupported -static-stdlib argument.
-    // TODO: If a C++ standard library is specified, it has to be libc++.
+    // Validating darwin unsupported -static-stdlib argument.
+    if parsedOptions.hasArgument(.staticStdlib) && targetTriple.isDarwin {
```
owenv, line 162: I don't think `targetTriple.isDarwin` is needed here since we've already selected a Darwin toolchain.

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1540,6 +1540,23 @@ final class SwiftDriverTests: XCTestCase {
         return
       }
     }
+    
+    XCTAssertThrowsError(try Driver(args: ["swiftc", "-c", "-static-stdlib",
```
owenv, line 1544: These tests might need to include something like `"-target", "x86_64-apple-macosx10.4"` so they select the right toolchain when run on Linux

Nikhil0487: Sure, will add the args for all the three tests. I thought since the test method was called `testDarwinToolchainArgumentValidation`i thought it wont be tested in Linux by default.

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
cltnschlosser: Darwin toolchain can build non Darwin targets, right? Distinction between the host and target.

owenv:
@swift-ci test

PR #186, opened 2020-07-29
[Explicit Module Builds] Pass down pre-built module candidates to the -compile-module-from-interface jobs

artemcm:
`-scan-dependencies` now emits ready-to-use binary module candidates that can be imported by the compiler, e.g. modules adjacent to .swiftinterface files or in prebuilt module dirs. 
When invoking `-compile-module-from-interface`, pass down these module candidates so the compiler can emit forwarding modules to them, saving the effort of type-checking/serializing binary modules at all.

Resolves rdar://problem/66243229

### DIFF
--- a/Sources/SwiftDriver/Explicit Module Builds/InterModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/Explicit Module Builds/InterModuleDependencyGraph.swift
@@ -58,7 +58,7 @@ extension ModuleDependencyId: Codable {
   public var moduleInterfacePath: String?
 
   /// The paths of potentially ready-to-use compiled modules for the interface.
-  var compiledModuleCandidates: [String]?
+  public var compiledModuleCandidates: [String]?
 
   /// The path to the already-compiled module.
   public var compiledModulePath: String?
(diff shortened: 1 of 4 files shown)


artemcm:
@swift-ci please test

Sources/SwiftDriver/Explicit Module Builds/InterModuleDependencyGraph.swift
```diff
@@ -58,7 +58,7 @@ extension ModuleDependencyId: Codable {
   public var moduleInterfacePath: String?
 
   /// The paths of potentially ready-to-use compiled modules for the interface.
-  var compiledModuleCandidates: [String]?
+  public var compiledModuleCandidates: [String]?
```
DougGregor, line 61: `@_spi(Testing)` ?

artemcm: Incorporated in:
https://github.com/apple/swift-driver/pull/181
Since it already heavily modifies the graph structure.

PR #189, opened 2020-07-31
Update tests to match latest toolchain expectations.

artemcm:
- Explicit Module Build test input .swiftinterface files to explicitly bring in SwiftONoneSupport
- Triple test to function when the toolchain uses fallback from `swift-driver` to `swift`


artemcm:
@swift-ci please test

PR #190, opened 2020-08-03
Support reading binary swiftdeps files

owenv:
A couple of notes:
- The yaml versions of the test inputs I checked in aren't needed to run the tests, I just found them useful to verify the tests were correct (I generated them from the binary versions using swift-dependency-tool)
- input info maps are now the only yaml the driver needs to read. Maybe we could switch those to JSON so we can drop the yaml dependency and simplify the build a little?

Also cc @belkadan who's the original author of the bitstream reader code added in the first commit. I've added you as a commit author, but let me know if you'd like a different form of attribution!


Sources/SwiftDriver/Utilities/Bits.swift
```diff
@@ -0,0 +1,107 @@
...
+    let byte = buffer[index / 8]
+    return (byte >> UInt8(index % 8)) & 1
+  }
+
+  func index(after i: Int) -> Int {
+    return i + 1
+  }
+  func index(before i: Int) -> Int {
+    return i - 1
+  }
```
belkadan, line 34: I believe `index(after:)` and `index(before:)` no longer need to be provided explicitly.

Sources/SwiftDriver/Utilities/Bits.swift
```diff
@@ -0,0 +1,107 @@
...
+
+  subscript(extractAt offset: Int, count count: Int) -> UInt64 {
```
belkadan, line 36: I don't know why I made this a subscript; it might as well just be a method.

Sources/SwiftDriver/Utilities/Bits.swift
```diff
@@ -0,0 +1,107 @@
...
+    precondition(offset >= 0)
+    precondition(offset &+ count >= offset)
+    precondition(offset &+ count <= self.endIndex)
+
+    return buffer.withUnsafeBytes { (bytes: UnsafeRawBufferPointer) in
+      let upperBound = offset &+ count
+      let topByteIndex = upperBound >> 3
+      var result: UInt64 = 0
+      if upperBound & 7 != 0 {
+        let mask: UInt8 = (1 << UInt8(upperBound & 7)) &- 1
```
belkadan, line 47: The cast to UInt8 shouldn't be necessary in modern Swift, which allows mixed-type casts. I don't *think* that has a performance impact either…

Sources/SwiftDriver/Utilities/Bits.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    mutating func read(bytes count: Int) throws -> Data {
+      precondition(count >= 0)
+      precondition(offset & 0b111 == 0)
+      let newOffset = offset &+ (count << 3)
+      precondition(newOffset >= offset)
+      if newOffset > buffer.count { throw Error.bufferOverflow }
+      defer { offset = newOffset }
+      return buffer.buffer.subdata(in: (offset >> 3)..<(newOffset >> 3))
```
belkadan, line 96: This should be a slice rather than a copy these days: `buffer.buffer.dropFirst(offset >> 3).prefix((newOffset - offset) >> 3)`

Sources/SwiftDriver/Utilities/Bits.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    mutating func advance(toBitAlignment align: Int) throws {
+      precondition(align > 0)
+      precondition(offset &+ (align&-1) >= offset)
+      if offset % align == 0 { return }
+      offset = ((offset / align) &+ 1) &* align
```
belkadan, line 103: It'd be good to use masking here instead, with a precondition that `align` is a power of 2: `align & (align&-1) == 0`.

Sources/SwiftDriver/Utilities/Bitstream.swift
```diff
@@ -0,0 +1,351 @@
...
+      throw Error.missingEndBlock(blockID: id)
+    }
+    return elements
+  }
+  
+  static let fakeTopLevelBlockID: UInt64 = ~0
+}
+
+extension Bitcode {
+  init(data: Data) throws {
```
belkadan, line 342: One downside of this approach is that it's going to read the entire file structure into memory at once, rather than processing entry by entry. That's probably fine for now, but it'll be something worth cleaning up later.

Sources/SwiftDriver/Utilities/Bitstream.swift
```diff
@@ -0,0 +1,351 @@
...
+    } else {
+      switch lastOperand {
+      case .array(let element):
+        let length = try cursor.readVBR(6)
+        var elements = [UInt64]()
+        for _ in 0..<length {
+          elements.append(try readSingleAbbreviatedRecordOperand(element))
+        }
+        if case .char6 = element {
+          payload = .char6String(String(elements.map { Character(UnicodeScalar(UInt8($0))) }))
```
belkadan, line 208: [Turns out going through UnicodeScalarView is easier, though only a little.](https://forums.swift.org/t/add-latin1-to-unicode-codecs/37978/10) Not that we use Char6Strings…pretty much ever.

Sources/SwiftDriver/Incremental Compilation/SourceFileDependencyGraph.swift
```diff
@@ -111,3 +111,109 @@ extension SourceFileDependencyGraph {
...
+    case malformedDependsOnDefinitionRecord
+    case malformedIdentifierRecord
+    case malformedSourceFileDepGraphNodeRecord
+    case unknownRecord
+    case unexpectedSubblock
+  }
+
+  public init(data: Data) throws {
+    let bitcode = try Bitcode(data: data)
+    guard bitcode.signature == 0x53_50_45_44 else { throw ReadError.badMagic } // DEPS
```
belkadan, line 137: If you want you can put `DEPS` in the source as a literal string: https://twitter.com/UINT_MIN/status/1229913969902538753

(Note that MacTypes FourCharCodes are big-endian while bitcode signatures are little-endian, so you'll want to reverse the logic I have there.)

Sources/SwiftDriver/Utilities/Bits.swift
owenv: yep, these were unnecessary

Sources/SwiftDriver/Utilities/Bits.swift
owenv: this turned out to be needed so that `7` gets inferred as `Int` instead of `UInt8` There might still be a simpler way to write this, but I couldn't find it

Sources/SwiftDriver/Utilities/Bitstream.swift
owenv: yeah, I added a fix to come back and improve this in a follow-up

Sources/SwiftDriver/Incremental Compilation/SourceFileDependencyGraph.swift
owenv: oh wow, that's some impressive constant folding! I added a `Bitcode.Signature` type which works similarly

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

belkadan:
> input info maps are now the only yaml the driver needs to read. Maybe we could switch those to JSON so we can drop the yaml dependency and simplify the build a little?

Driver options and inputs are "supposed" to be stable for external devs; in particular, I'm sure CMake is generating one of these by now. (It's even [documented](https://github.com/apple/swift/blob/master/docs/Driver.md#output-file-maps)! Because I wanted a place to point people when they kept trying to compile Swift code like C code!)

belkadan:
I only reviewed the parts I wrote originally but they look good to me. :-)

PR #193, opened 2020-08-06
[Build Script Helper] Setup environment variables for the specified toolchain's executables

artemcm:
If a toolchain is specified, set the search environment variable for each tool executable the driver cares about to ensure that the specified toolchain's specific version is used.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -9,6 +9,9 @@ import shutil
 import subprocess
 import sys
 
+driver_tools = ['swift', 'swift-frontend', 'ld', 'libtool', 'clang',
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci please test

artemcm:
I ran into this because on Ubuntu 16.04, even with a specified toolchain, the driver could not locate some executables. Curiously, this did not happen on Ubuntu 18.04 or 20.04.

I'm still trying to figure out why this specific version of Ubuntu breaks; but, regardless, I think this change is an improvement because it ensures that we do use the provided toolchain's executables, instead of relying on search paths and potentially accidentally using a wrong executable for any of these tools.

PR #194, opened 2020-08-08
Improve support for macOS 11/darwin 20 triple versioning

owenv:
This is a direct port of some of the updates to clang/llvm. I didn't port over canonicalization of the fake "macOS 10.16" triples because it looks like that's already been mostly removed elsewhere.


owenv:
@swift-ci test

PR #195, opened 2020-08-08
Report a better error message when retrieving target info from the frontend fails

owenv:
This should help make it a bit clearer what's going on when using the driver with an incompatible toolchain or configuring SWIFT_DRIVER_SWIFT_FRONTEND_EXEC incorrectly


owenv:
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
It looks like the 8/7 snapshot CI is using has some kind of code signing issue. I filed https://bugs.swift.org/browse/SR-13370 in case it doesn't resolve itself on its own tomorrow

DougGregor:
@swift-ci test macOS

DougGregor:
Hmm, still having trouble:

PR #197, opened 2020-08-08
Fix .swiftsourceinfo output path when the build system provides a Project/ dir

owenv:
This fixes the remaining source info integration test (Driver/sourceinfo_file.swift)

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -415,13 +415,17 @@ public struct Driver {
         compilerMode: compilerMode,
         outputFileMap: self.outputFileMap,
         moduleName: moduleOutputInfo.name)
+    let projectDirectory = Self.computeProjectDirectoryPath(
+      moduleOutputPath: self.moduleOutputInfo.output?.outputPath,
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1796,26 +1816,22 @@ extension Driver {
...
-      let path = try VirtualPath(path: outputPathArg.asSingle)
-
-      // If the compiler output is of this type, use the argument directly.
-      if type == compilerOutputType {
-        return path
-      }
-
-      return try path.replacingExtension(with: type)
-    }
-
```
owenv, line 1818: I _think_ this logic is no longer used due to improvements in the module output path computation since it was written. Removing it didn't introduce any regressions in the integration tests

owenv:
@swift-ci test Linux

Sources/SwiftDriver/Driver/Driver.swift
DougGregor: It looks like dead code to me, thanks!

DougGregor:
@swift-ci test macOS

PR #198, opened 2020-08-09
CODE_OF_CONDUCT

gaushikmr:


gaushikmr:
@owenv Ok I will change it

gaushikmr:
@owenv I have updated it pls check

owenv:
@gaushikmr thanks! It looks like the introduction paragraph may have gotten removed, could you update it to match the one from https://github.com/apple/swift/blob/master/CODE_OF_CONDUCT.md? Once that's done this should be good to go.

akyrtzi:
As @benlangmuir suggested in https://github.com/apple/sourcekit-lsp/pull/302, could we have a link to the swift project document instead of maintaining a separate copy?

DougGregor:
I agree with linking, so we don't have to update everywhere

gaushikmr:
@akyrtzi @DougGregor Ok I will update it

shahmishal:

PR #200, opened 2020-08-09
Fix build on Windows

egorzhdan:
* SQLite3 was recently added as a dependency, and the CMake build scripts needed an update
* `SwiftDriver` target imports `TSCUtility` (which has a dependency on `TSCBasic`) but only linked against `TSCBasic`
* SwiftPM uses `SwiftDriver.dll` & `SwiftOptions.dll`, so they should be installed into the toolchain bin directory
* signals are not used on Windows, and `strsignal` is not available


egorzhdan:
Updated this PR, now it depends on the changes in https://github.com/apple/swift-tools-support-core/pull/106.
cc @compnerd

CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ endif()
 find_package(dispatch QUIET)
 find_package(Foundation QUIET)
 find_package(Yams CONFIG REQUIRED)
+find_package(SQLite3 REQUIRED)
```
compnerd, line 45: This shouldn't be needed with apple/swift-tools-support-core#109

Sources/SwiftDriver/CMakeLists.txt
```diff
@@ -75,7 +75,7 @@ add_library(SwiftDriver
   Utilities/VirtualPath.swift)
 
 target_link_libraries(SwiftDriver PUBLIC
-  TSCBasic
+  TSCUtility
```
compnerd, line 78: What was the reason for this change?

Sources/SwiftDriver/Driver/OutputFileMap.swift
```diff
@@ -93,7 +93,7 @@ public struct OutputFileMap: Equatable {
   ) throws {
     let encoder = JSONEncoder()
 
-  #if os(Linux) || os(Android)
+  #if !canImport(ObjectiveC)
```
compnerd, line 96: Why does `withoutEscapingSlashes` require ObjectiveC?

owenv: When this was written, `withoutEscapingSlashes` wasn't implemented in corelibs-foundation yet. It looks like it's since been added in https://github.com/apple/swift-corelibs-foundation/commit/c3fd6ecd647a7dbc5b5c69f8f4966c5d5a1c897e, so I think this conditional block can be removed entirely.

Sources/SwiftDriver/CMakeLists.txt
egorzhdan: A linker error that was caused by the fact that `TSCBasic` doesn't link against `FoundationNetworking`.
Linking `TSCUtility` seems to be the correct option in any case, since it's imported from `SwiftDriver`.

compnerd: Interesting, I don't see the linker error.  However, fully specifying the link dependencies is better.  If there are references to TSCBasic, we should keep both, otherwise, good change :)

Sources/SwiftDriver/Driver/OutputFileMap.swift
egorzhdan: Thanks @owenv, turns out I was building with an outdated Foundation.
I'll drop that commit to keep the PR focused on Windows-related things.

Sources/SwiftDriver/CMakeLists.txt
egorzhdan: Both TSCBasic & TSCUtility are actually imported. Agreed, it's nicer to have both of them in the list, fixed.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@egorzhdan would you mind removing the SQLite3 dependency?  That's not a true dependency for this project AFAICT, and its better to fix that in t-s-c rather than add that dependency here.  (Or did I miss the import of SQLite in this project?)

egorzhdan:
@compnerd ok, removed it.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
CC: @DougGregor

tomerd:
lgtm, @DougGregor to approve

DougGregor:
LGTM, thank you!

PR #202, opened 2020-08-10
Make newly-added Frontend target info components optional.

artemcm:
This is done in an effort to still be able to parse the target info produced by the latest shipping Xcode. 
This will allow the target info to be used with current toolchains and, for example, distinguish between the toolchain that includes a compiler version (sufficiently new) and doesn't.


artemcm:
@swift-ci test

artemcm:
Abandoning in favour of relying on better error diagnostics produced in: https://github.com/apple/swift-driver/pull/195

PR #201, opened 2020-08-10
 Infer object format when it can't be determined from a triple's successfully parsed environment

owenv:
Builds on https://github.com/apple/swift-driver/pull/199, only the second commit is new.

This PR infers a triple's object format if the environment is successfully parsed but the format can't be determined from that environment. This can happen with triples like `x86_64-unknown-linux-gnu` and caused autolink-extract jobs not to be generated sometimes because the object format wasn't correctly inferred as ELF. Unfortunately I had to ifdef out some tests on macOS as a result since `swift-autolink-extract` isn't available there.

PR #203, opened 2020-08-11
Revert " Infer object format when it can't be determined from a triple's successfully parsed environment"

owenv:
Reverts apple/swift-driver#201

Linux CI builds with this change are failing with "unableToFind(tool: "swift-autolink-extract")" during SMP tests. I've been unable to find the cause, so reverting for now.

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple.swift
+++ b/Sources/SwiftDriver/Utilities/Triple.swift
@@ -167,8 +167,6 @@ public struct Triple {
     if let parsedEnv = parsedEnv {
       self.environment = parsedEnv.value.environment
       self.objectFormat = parsedEnv.value.objectFormat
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci test

owenv:
@DougGregor yeah, I'm pretty confused as to what happened here, I'll need to investigate this more before reapplying. Thanks for the approval!

artemcm:
> Thanks for the quick revert! Not sure how this got through swift-driver CI but failed in Swift CI.

Swift-driver CI tested against a toolchain that did not yet contain the change that would expose this failure. Also, the test that failed is a SwiftPM test, and I don't think those run in swift-driver CI at all. As SwiftPM and swift-driver get friendlier, perhaps we should also run SwiftPM tests with the integrated driver enabled as a part of swift-driver CI. It will add some overhead to our lightning-fast CI, but it will probably be worth it.

PR #204, opened 2020-08-12
[Explicit Module Builds] Add `isFramework` flag to the dependency graph and swift module info.

artemcm:
Ensure the flag is propagated from a given swift module's dependency graph details to its swift module info passed in as an explicit module map to frontend jobs.

This is a companion PR to: 
https://github.com/apple/swift/pull/33430

### DIFF
--- a/Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
+++ b/Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
@@ -342,11 +342,15 @@ public typealias ExternalDependencyArtifactMap =
     let dependencyInfo = try dependencyGraph.moduleInfo(of: dependencyId)
 
     let swiftModulePath: TypedVirtualPath
+    let isFramework: Bool
     if case .swift(let details) = dependencyInfo.details,
        let compiledModulePath = details.explicitCompiledModulePath {
       // If an already-compiled module is available, use it.
       swiftModulePath = .init(file: try VirtualPath(path: compiledModulePath),
                               type: .swiftModule)
+      // Since this module has already been built, it is no longer relevant
+      // whether it is a framework or not.
+      isFramework = false
     } else {
       // Generate a build job for the dependency module, if not already generated
       if swiftModuleBuildCache[dependencyId] == nil {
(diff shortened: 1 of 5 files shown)


Sources/SwiftDriver/Explicit Module Builds/ExplicitModuleBuildHandler.swift
```diff
@@ -351,11 +351,15 @@ public typealias ExternalDependencyArtifactMap =
...
     let swiftModulePath: TypedVirtualPath
+    let isFramework: Bool
     if case .swift(let details) = dependencyInfo.details,
        let compiledModulePath = details.explicitCompiledModulePath {
       // If an already-compiled module is available, use it.
       swiftModulePath = .init(file: try VirtualPath(path: compiledModulePath),
                               type: .swiftModule)
+      // Since this module has already been built, it is no longer relevant
+      // whether it is a framework or not.
+      isFramework = false
```
DougGregor, line 362: Is this comment true? An already-built `.swiftmodule` can be part of a `.framework`. This happens both with libraries built for distribution (the `.swiftmodule` and `.swiftinterface` are installed together) and with libraries not built for distribution (where it's just the `.swiftmodule`).

DougGregor: ... and even if we don't "need" to know when it's a framework, is it better to have that information for consistency reasons?

artemcm: The `explicitCompiledModulePath` is used only during placeholder dependency resolution, which tells us that this `.swiftmodule` is built as a part of an external dependency (separate from the `compiledModuleCandidates`). With that in mind, I guess my thought process was that it is the responsibility of the builder/planner of that external dependency to ensure that it is treated correctly w.r.t. whether or not it is a framework. Hence the "no longer relevant" part. 

I do agree that if we don't _need_ to know it's a framework, it is better to have that information; though, I am not yet sure how we can get this information. Perhaps we can have SwiftPM query it from the swift-driver after planning and propagate it to the dependees along with the `.swiftmodule` path and the dependency graph.

cltnschlosser: If it’s not needed you could change the variable name to something like `isNonPreCompiledFramework`. Then this false isn’t a lie 😅

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #205, opened 2020-08-13
[Build Script Helper] Improvements to build with a specified toolchain

artemcm:
- Only set environment variables for executables built by `swift` `build-script` as a part of a toolchain.
- Export `SDKROOT` on Darwin to ensure tests always have an SDK if one is not specified explicitly. (this mimics how SwiftPM sets up its environment for a bootstrap)


artemcm:
@swift-ci please test

PR #206, opened 2020-08-14
[Build Script Helper] Only export toolchain executable paths when the executable exist and bring back wrongly-removed tools.

artemcm:
Sorry for the back and forth, this seems to have been a bit of a chicken-and-egg between macOS and Linux.


artemcm:
@swift-ci please test

PR #207, opened 2020-08-15
Reapply "Infer object format when it can't be determined from a triple's successfully parsed environment"

owenv:
Reverts apple/swift-driver#203

I suspect https://github.com/apple/swift-driver/pull/206 fixed the failures this caused in the SPM CI tests.


owenv:
Running SPM tests @ https://github.com/apple/swift-package-manager/pull/2871

owenv:
Hmm, this is still failing on Linux with `unableToFind(tool: "swift-autolink-extract")`, I'll need to investigate more

PR #208, opened 2020-08-15
Reapply "Infer object format when it can't be determined from a triple's successfully parsed environment

owenv:
Testing another potential fix - If a tool can't be found relative to the driver library's client executable, look for it relative to the frontend instead


Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -128,6 +128,13 @@ extension Toolchain {
       return path
     } else if let path = try? xcrunFind(executable: executable) {
       return path
+    } else if !["swift-frontend", "swift"].contains(executable),
+              let parentDirectory = try? getToolPath(.swiftCompiler).parentDirectory,
+              parentDirectory != executableDir,
+              let path = lookupExecutablePath(filename: executable, searchPaths: [parentDirectory]) {
+      // If the driver library's client and the frontend are in different directories,
+      // try looking for tools next to the frontend.
+      return path
```
owenv, line 137: This is kind of a hack, not sure if there's a better wya to solve this that doesn't involve overriding the path to every toolchain tool individually

owenv:
With https://github.com/apple/swift-package-manager/pull/2871, this now passes SPM CI testing

@swift-ci test

owenv:
With the SPM change merged, this should be safe to reapply so long as swift-driver tests still pass

@swift-ci test

PR #209, opened 2020-08-16
Attempt to use SDKSettings.json on Darwin platforms to set -target-sdk-version

owenv:
This fixes the Driver/sdk-version.swift integration test, and it's also a prerequisite to begin modernizing linker deployment targets to use -platform_version instead of the old flags.

Unfortunately, SDKSettings.json doesn't work very well with Codable synthesis and the implementation of Codable for `Version` in TSC, so there's a lot of manual decoding. Also, this implementation sometimes includes a minor version of `0` where the integrated driver would not (for example -target-sdk-version 13.4.0 instead of 13.4); as far as I know this won't break anything, and it makes the version remapping much simpler.


owenv:
@swift-ci test

PR #211, opened 2020-08-17
Use platform_version to pass deployment target information to the linker

owenv:
Builds on #209, only the second commit is new.

This fixes Driver/macabi-environment.swift and begins revolving some of the failures in Driver/linker.swift. It replaces the older deployment target linking flags with `-platform_version`.

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple+Platforms.swift
+++ b/Sources/SwiftDriver/Utilities/Triple+Platforms.swift
@@ -102,6 +102,29 @@ public enum DarwinPlatform: Hashable {
     }
   }
 
+  /// The name the linker uses to identify this platform.
+  public var linkerPlatformName: String {
+    switch self {
+    case .macOS:
+      return "macos"
+    case .iOS(.device):
+      return "ios"
+    case .iOS(.simulator):
+      return "ios-simulator"
+    case .iOS(.catalyst):
+      return "mac-catalyst"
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci test

Sources/SwiftDriver/Utilities/Triple+Platforms.swift
```diff
@@ -184,6 +207,47 @@ extension Triple {
...
+    case .iOS(.device), .iOS(.simulator), .tvOS(_):
+      // The first deployment of arm64 simulators is iOS/tvOS 14.0;
+      // the linker doesn't want to see a deployment target before that.
+      if _isSimulatorEnvironment && _iOSVersion.major < 14 && arch == .aarch64 {
+        return Version(14, 0, 0)
+      }
+
+      return _iOSVersion
+    case .watchOS(_):
+      return osVersion
```
DougGregor, line 247: IIRC, there should also be a case for arm64 watchOS simulator targets being watchOS >= 7.0

owenv: Good catch, thanks!

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
Re-running tests before merge

@swift-ci test

PR #212, opened 2020-08-17
Describe generatePCM jobs with the module name, not the first display input.

artemcm:
We generate such jobs on a per-module basis anyways.
This will give slightly nicer status commands in SwiftPM.


artemcm:
@swift-ci please test

PR #213, opened 2020-08-18
Disable the REPL test when no LLDB is found in the toolchain.

artemcm:
We have build bots that build `swift` but do not build `lldb`, this ensures that this test does not fail in that environment. 

Example of such build bot:
https://ci.swift.org/job/oss-swift-incremental-RA-linux-ubuntu-16_04-long-test/6634/console


artemcm:
@swift-ci please test

PR #214, opened 2020-08-19
[Explicit Module Builds] Re-scan all clang modules against all targets against which they will be built.

artemcm:
- Factor out placeholder dependency resolution out of the ExplicitModuleHandler into the Driver
- Generate and post-process the dependency graph before passing it on to the ExplicitModuleHandler:
  - Run the Swift Fast Dependency Scanner.
  - Resolve placeholder dependencies, if any.
  - Re-scan and update Clang modules with a super-set of their dependencies against all compilation targets.

**Some background on this issue:**
When building a Swift module at a fixed target version, the module may have dependencies which specify a different target version. It is possible that in this case, multiple swift modules, with different targets, have a dependency on the same Clang module. There is a hard requirement that the target version of the Clang module compilation match exactly the target version of the depending Swift module compilation. 

**The problem this PR addresses:**
Dependency scanning results may vary depending on the target version specified on the dependency scanning action. If the scanning action is performed at a fixed target, and the scanned module is later compiled with a higher version target, miscomputation may occur due to dependencies present only at the higher version number and thus not detected by the dependency scanner. 

Uses the newly-added Clang dependency scanning mode in the swift-frontend, added here:
https://github.com/apple/swift/pull/33522

Resolves rdar://67079780

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -205,7 +205,7 @@ public struct Driver {
   /// A collection describing external dependencies for the current main module that may be invisible to
   /// the driver itself, but visible to its clients (e.g. build systems like SwiftPM). Along with the external dependencies'
   /// module dependency graphs.
-  internal var externalDependencyArtifactMap: ExternalDependencyArtifactMap? = nil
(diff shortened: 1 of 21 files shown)


artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #215, opened 2020-08-20
[Build Script Helper] Make cleaning up build directory a separate "clean" action.

artemcm:
Instead of doing it automatically every time.
This will allow callers (mostly `swift` `build-script`) decide/override whether or not cleanup should be done.
https://github.com/apple/swift/pull/33563 shows the use of this action in the compiler's `build-script`.


artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #216, opened 2020-08-21
Revert "[Explicit Module Builds] Re-scan all clang modules against all targets against which they will be built."

artemcm:
Reverts apple/swift-driver#214

This seems to be causing failures in `swift` CI with error:

```
/Users/buildnode/jenkins/workspace/swift-PR-osx-smoke-test/branch-master/swift-driver/Sources/SwiftDriver/Jobs/Planning.swift:266:9: error: cannot find 'resolveVersionedClangDependencies' in scope
18:17:44     try resolveVersionedClangDependencies(dependencyGraph: &dependencyGraph)
18:17:44         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
18:17:44 ninja: build stopped: subcommand failed.
```

PR #217, opened 2020-08-21
[Explicit Module Builds] Re-scan all clang modules against all targets against which they will be built.

artemcm:
Originally merged in #214, reverted in #216.

- Factor out placeholder dependency resolution out of the ExplicitModuleHandler into the Driver
- Generate and post-process the dependency graph before passing it on to the ExplicitModuleHandler:
  - Run the Swift Fast Dependency Scanner.
  - Resolve placeholder dependencies, if any.
  - Re-scan and update Clang modules with a super-set of their dependencies against all compilation targets.


artemcm:
In the original PR I'd forgotten to add the new file to the CMakeLists.txt. And our CI builds using SwiftPM so it didn't catch that. :(

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #219, opened 2020-08-22
Reapply "Use platform_version to pass deployment target info to the linker"

owenv:
I haven't been able to reproduce the issue that caused this to be reverted locally, so running on Swift CI to try and diagnose the problem


owenv:
@swift-ci test

owenv:
@swift-ci test

PR #220, opened 2020-08-22
Properly resolve paths to temporary response files when autolinking

owenv:
Previously, .autolink files were passed as flags throughout the driver so that they could be prefixed by "@". This PR adds a new Job.ArgTemplateKind, `responseFilePath` to make sure we resolve the underlying path properly. This fixes most of the Driver/tools_directory integration test (there's at least one other remaining unrelated issue preventing it from passing)


owenv:
@swift-ci test

PR #221, opened 2020-08-24
Fix the swift-driver build using CMake

artemcm:
Add missing file.

It seems that when I re-based https://github.com/apple/swift-driver/pull/217, I somehow managed to silently undo the CMakeLists change it meant to contain...


artemcm:
@swift-ci test

PR #225, opened 2020-08-27
[Explicit Module Builds] Use the new batch scanning mode for versioned PCM clang modele re-scan

artemcm:
https://github.com/apple/swift/pull/33569 introduced a batch dependency scanning mode to save up on the overhead of clang instance creation.
This patch switches `swift-driver` over to using the new batch mode instead of generating individual `-scan-clang-dependencies` jobs for each re-scan.

### DIFF
--- a/Sources/SwiftDriver/Explicit Module Builds/ModuleDependencyScanning.swift
+++ b/Sources/SwiftDriver/Explicit Module Builds/ModuleDependencyScanning.swift
@@ -24,9 +24,6 @@ extension Driver {
     var commandLine: [Job.ArgTemplate] = swiftCompilerPrefixArgs.map { Job.ArgTemplate.flag($0) }
     commandLine.appendFlag("-frontend")
     commandLine.appendFlag("-scan-dependencies")
-    if parsedOptions.hasArgument(.parseStdlib) {
-       commandLine.appendFlag(.disableObjcAttrRequiresFoundationModule)
-    }
     try addCommonFrontendOptions(commandLine: &commandLine, inputs: &inputs,
                                  bridgingHeaderHandling: .precompiled,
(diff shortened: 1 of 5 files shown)


artemcm:
The test that relies on the versioned PCM re-scanning is currently failing with this change, but the driver side of things looks okay to me. I'm investigating if there is something we are missing in the scanner.

Sources/SwiftDriver/Explicit Module Builds/ModuleDependencyScanning.swift
```diff
@@ -75,65 +75,108 @@ extension Driver {
...
     // Aggregate the fast dependency scanner arguments
     var commandLine: [Job.ArgTemplate] = swiftCompilerPrefixArgs.map { Job.ArgTemplate.flag($0) }
     commandLine.appendFlag("-frontend")
-    commandLine.appendFlag("-scan-clang-dependencies")
-
+    // The dependency scanner automatically operates in batch mode if -batch-scan-input-file
+    // is present.
+    commandLine.appendFlag("-scan-dependencies")
+    if parsedOptions.hasArgument(.parseStdlib) {
+       commandLine.appendFlag(.disableObjcAttrRequiresFoundationModule)
```
DougGregor, line 130: This is a surprise at this point in the code. Is the dependency scanner not providing this argument appropriately?

artemcm: Good catch, this does not need to be here (confirmed with testing). 
It was inherited from the `-scan-dependencies` job which originally inherited it from the `compileJob` code. Removed.

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #226, opened 2020-08-28
Use Yams 4.0.0 instead of master

neonichu:


neonichu:
@swift-ci test

owenv:
update-checkout is currently fetching version 3.0.1 of Yams, should that be updated as well?

DougGregor:
Good catch. Yes, we should keep them in sync

PR #228, opened 2020-08-29
Update the no inputs error to allow invocations like 'swiftc -v'

owenv:
This fixes the Driver/no-inputs.swift integration test, and by moving the check to the start of build planning, it also eliminates a couple of crashes where when we were creating compile jobs we assumed there was at least 1 input file.


owenv:
@swift-ci test

PR #229, opened 2020-08-29
Support adding a "joined" option with a path argument to a command line

owenv:
Previously, appending a parsed option like '-I=/some/path' to a job's command line was unimplemented. This change represents these internally as a new Job.ArgTemplate case, joinedOptionAndPath.

I noticed this when running the full swift integration test suite with swift-driver. It turns out this was the only thing keeping all of the PlaygroundTransform tests from passing.


owenv:
@swift-ci test

PR #230, opened 2020-08-29
Fixup linker input handling and use -add_ast_path when needed

owenv:
While debugging why the TypeDecoder tests were all failing when run with swift-driver, I discovered that for invocations like `swiftc -emit-executable foo.swift -g -emit-module`, we weren't prefixing the swiftmodule file with -add_ast_path in the linker job. This PR fixes ~20 assorted integration test failures, and makes some minor cleanups to linker input handling on Linux as well.


owenv:
@swift-ci test

PR #231, opened 2020-08-29
Add -require-explicit-availability and -require-explicit-availability-target to common frontend opts

owenv:
This corresponds to this fix from January in the C++ driver: https://github.com/apple/swift/pull/29183


owenv:
@swift-ci test

PR #232, opened 2020-08-30
Preserve relative ordering in ParsedOptions.arguments(for options:)

owenv:
Another small bugfix: according to https://github.com/apple/swift/pull/8134, the relative ordering of "-F" and "-Fsystem" search paths needs to be preserved in frontend invocations. Instead of introducing a special case, I just changed the behavior of ParsedOptions.arguments(for options:) since it shouldn't be that expensive to always sort the options and I'm guessing this isn't the only case where it could end up mattering for correctness.


owenv:
@swift-ci test

PR #234, opened 2020-09-01
Fix CMake Build

artemcm:
Correct missing quote mark in Sources/SwiftDriver/CMakeLists.txt.

Just as I was working on preventing this kind of thing...


artemcm:
@swift-ci please test

PR #235, opened 2020-09-02
Revert #225 and #234

artemcm:
They seem to have test-breaking interactions with `swift` trunk, causing CI failures.


artemcm:
@swift-ci please test

PR #236, opened 2020-09-02
[Explicit Module Builds] Use the new batch scanning mode for versioned PCM clang modele re-scan

artemcm:
Another attempt at landing #225. 

----------- Original Description -----------
https://github.com/apple/swift/pull/33569 introduced a batch dependency scanning mode to save up on the overhead of clang instance creation.
This patch switches `swift-driver` over to using the new batch mode instead of generating individual `-scan-clang-dependencies` jobs for each re-scan.

### DIFF
--- a/Sources/SwiftDriver/CMakeLists.txt
+++ b/Sources/SwiftDriver/CMakeLists.txt
@@ -13,7 +13,7 @@ add_library(SwiftDriver
   "Explicit Module Builds/ClangVersionedDependencyResolution.swift"
   "Explicit Module Builds/InterModuleDependencyGraph.swift"
   "Explicit Module Builds/ModuleDependencyScanning.swift"
-  "Explicit Module Builds/ModuleArtifacts.swift"
+  "Explicit Module Builds/SerializableModuleArtifacts.swift"
 
   Driver/CompilerMode.swift
   Driver/DebugInfo.swift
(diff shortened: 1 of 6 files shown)


artemcm:
I haven't been able to reproduce the `swift` CI failure locally yet so debugging that issue is still WIP.

artemcm:
Does not seem that the CI failure is a code-interaction issue, it just seems to intermittently happen.

artemcm:
Ah. The bug seems to be test-specific. Across different tests, instances of `Driver` share a temporary directory that we get from `TSCBasic`, because the path to the temporary directory is cached as a global variable. Tests also share the main module name (`main`). The input to the batch scanning is computed using the temporary directory and the main module name, hence we get collisions and why testing intermittently passes.

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

owenv:
> Across different tests, instances of Driver share a temporary directory that we get from TSCBasic, because the path to the temporary directory is cached as a global variable.

That's some impressive debugging 😄 . Could this cause a problem for SPM builds with multiple targets as well, since it creates multiple drivers? Maybe each driver should create its own subdirectory for temporary files.

artemcm:
> > Across different tests, instances of Driver share a temporary directory that we get from TSCBasic, because the path to the temporary directory is cached as a global variable.
> 
> That's some impressive debugging 😄 . Could this cause a problem for SPM builds with multiple targets as well, since it creates multiple drivers? Maybe each driver should create its own subdirectory for temporary files.

Yeah, I thought about this case as well, but I do not think we can run into this scenario with SPM because it does not allow targets with the same name, and we use the target name to create this file. Furthermore, we would need to be using these Driver instances concurrently to run into trouble. Still, I think it might be a good idea to have each driver instance have its own temporary subdirectory, yeah.

artemcm:
@swift-ci please test

DougGregor (approved):

PR #237, opened 2020-09-03
Enable cmake install target on all platforms

gmittert:
Summary: The CMake files look like they are copied over from the SwiftPM
CMake which does some custom installation when it bootstraps itself on
non Windows platforms. Swift-driver doesn't do this, so there's no need
to hide the install target on non Windows platforms.


gmittert:
@swift-ci please test

DougGregor (approved):
Glad we didn't need the hack, thanks!

gmittert:
@DougGregor can you merge this for me please? I don't have write access to swift-driver.

DougGregor:
> @DougGregor can you merge this for me please? I don't have write access to swift-driver.

Done!

PR #238, opened 2020-09-03
[Explicit Module Builds] Point module dependency modules to be put into the module cache

artemcm:
Preprocess the dependency graph, setting module paths to be relative to the module cache path, if one is present.
Only do so for modules that do not already specify a path.


artemcm:
One alternative I considered was to have the job generation code set the output paths, but that approach is not sufficient, because we also use the dependency graphs to communicate existence and location of module dependencies (for placeholder resolution), for which we would need to encode this information into the dependency graph anyway.

artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #239, opened 2020-09-04
Add a -driver-warn-unused-options flag for debugging

owenv:
This has been pretty useful for debugging some of the integration tests failures, so I think it makes sense to include, at least as an internal debugging option. Putting the check at the end of `Driver.run` seemed like the best way to avoid false positives, but I can move it somewhere else if anyone has a better idea.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -638,6 +638,12 @@ extension Driver {
   }
 }
 
+extension Diagnostic.Message {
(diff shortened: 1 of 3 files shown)


owenv:
@swift-ci test

DougGregor (approved):
Makes sense!

keith:
Tbh I'd be happy if something like this was always on. When do you want not to be informed of this?

owenv:
@keith that's a fair point, I think it's too early to turn this on by default though. It creates a lot of noise in the test suite because incremental builds & module trace support aren't implemented yet.

PR #242, opened 2020-09-06
Fix -debug-info-store-invocation logic on Darwin

owenv:
Fixes DebugInfo/compiler-flags-macosx.swift

With these updates, swift-driver should now match the behavior of the integrated driver and clang by passing -debug-info-store-invocation only when it's passed to the driver or RC_DEBUG_OPTIONS is set.


owenv:
@swift-ci test

DougGregor (approved):

PR #243, opened 2020-09-06
Implement additional argument validation steps and record multiple failures before exiting

owenv:
This PR adds validation for:
- conditional compilation arguments
- `-profile-generate` and `-profile-use`
- `-F` and `-Fsystem`

Additionally, if a validation step fails when initializing the driver, don't fail immediately. Instead, record any other validation failures and then exit before planning.

### DIFF
--- a/Sources/swift-driver/main.swift
+++ b/Sources/swift-driver/main.swift
@@ -47,6 +47,10 @@ do {
   var driver = try Driver(args: arguments,
                           diagnosticsEngine: diagnosticsEngine,
                           executor: executor)
+  // FIXME: The following check should be at the end of Driver.init, but current
+  // usage of the DiagnosticVerifier in tests makes this difficult.
+  guard !driver.diagnosticEngine.hasErrors else { throw Diagnostics.fatalError }
+
   let jobs = try driver.planBuild()
   try driver.run(jobs: jobs)
 
--- a/Tests/SwiftDriverTests/Helpers/AssertDiagnostics.swift
+++ b/Tests/SwiftDriverTests/Helpers/AssertDiagnostics.swift
@@ -13,7 +13,6 @@
 import XCTest
 import SwiftDriver
 import TSCBasic
-import TSCUtility
 
 func assertDriverDiagnostics(
   args: [String],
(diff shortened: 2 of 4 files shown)


owenv:
@swift-ci test

DougGregor (approved):
This looks great! Thank you

Sources/swift-driver/main.swift
```diff
@@ -47,6 +47,10 @@ do {
   var driver = try Driver(args: arguments,
                           diagnosticsEngine: diagnosticsEngine,
                           executor: executor)
+  // FIXME: The following check should be at the end of Driver.init, but current
+  // usage of the DiagnosticVerifier in tests makes this difficult.
+  guard !driver.diagnosticEngine.hasErrors else { throw Diagnostics.fatalError }
```
JhonnyBillM, line 52: Would suggest to avoid double negatives, they tend to be confusing.
```swift
if driver.diagnosticEngine.hasErrors { throw Diagnostics.fatalError }
```

owenv:
Looks like I made a small mistake with the original implementation - a couple of these diagnostics should be warnings. Thanks for the quick review @DougGregor!

owenv:
@swift-ci test

Tests/SwiftDriverTests/Helpers/AssertDiagnostics.swift
```diff
@@ -13,6 +13,7 @@
 import XCTest
 import SwiftDriver
 import TSCBasic
+import TSCUtility
```
JhonnyBillM, line 16: Is this import needed?

JhonnyBillM (approved):

Sources/swift-driver/main.swift
owenv: It might be needed eventually to import `Diagnostics.fatalError` when we address the FIXME in main.swift, but for now I've dropped it.

Tests/SwiftDriverTests/Helpers/AssertDiagnostics.swift
owenv: Not right now, but we might need it to import `Diagnostics.fatalError` when we address the main.swift FIXME. For now I've dropped it.

Sources/swift-driver/main.swift
owenv: I see your point, but I'd prefer to use guard here to enforce the early exit. In the past we've had bugs in this code because of inadvertent fallthrough out of error conditions.

owenv:
@swift-ci test

owenv:
@swift-ci test

PR #244, opened 2020-09-07
build: link Yams privately

compnerd:
Because Yams is used with `@_implementationOnly`, the clients of
`SwiftDriver` do not need to link against `Yams` or `CYaml`.  This was
uncovered when building SourceKit-LSP.


compnerd:
@swift-ci please test

DougGregor (approved):
Great. I really want the default to be @_inplementationOnly in the language

PR #246, opened 2020-09-11
[Build Script Helper] On macOS make install action use CMake and install libSwiftDriver

artemcm:
Previously only the x86 `swift-driver` and `swift-help` executables were copied over to the specified toolchain's `bin` directory.
With this change, `libSwiftDriver` shared library is also packaged and installed into the toolchain, as well as `libSwiftOptions`.
To do so, we build the driver and all of its dependencies using CMake for `arm64` and `x86_64` and create universal binaries, as documented [here](https://developer.apple.com/documentation/xcode/building_a_universal_macos_binary).

Because the driver's dependencies need to come along for the ride, the installed artifacts are as follows, within the specified toolchain:
#### Executables:
```
bin/swift-driver
bin/swift-help
```
#### Binary Modules and Shared Libraries 
```
lib/swift/macosx/libTSCBasic.dylib
lib/swift/macosx/libTSCLibc.dylib
lib/swift/macosx/libTSCUtility.dylib
lib/swift/macosx/libSwiftDriver.dylib
lib/swift/macosx/libSwiftOptions.dylib

lib/swift/macosx/TSCBasic.swiftmodule
lib/swift/macosx/TSCLibc.swiftmodule
lib/swift/macosx/TSCUtility.swiftmodule
lib/swift/macosx/SwiftDriver.swiftmodule
lib/swift/macosx/SwiftOptions.swiftmodule
```
#### C Module .modulemap and header files
```
include/swift/TSCclibc/*
```
#### Misc
- The `Yams` dependency is `implementationOnly` and is built as a static-library.
- `ArgumentParser.swiftmodule` and `libArgumentParser.dylib` must also be installed into the toolchain, but the code currently omits them, to avoid adding them to the toolchain until they are actually needed. 

Resolves rdar://69708243

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -116,6 +116,20 @@ def delete_rpath(rpath, binary, verbose):
   if verbose:
     print(stdout)
 
+def add_rpath(rpath, binary, verbose):
+  cmd = ['install_name_tool', '-add_rpath', rpath, binary]
+  if verbose:
+    print(' '.join(cmd))
+  installToolProcess = subprocess.Popen(cmd,
+                                        stdout=subprocess.PIPE,
+                                        stderr=subprocess.PIPE)
+  stdout, stderr = installToolProcess.communicate()
+  if installToolProcess.returncode != 0:
+    print('install_name_tool command failed: ')
+    print(stderr)
+  if verbose:
+    print(stdout)
+
 def should_test_parallel():
   if platform.system() == 'Linux':
     distro = platform.linux_distribution()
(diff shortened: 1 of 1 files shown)


artemcm:
Two blockers for this, currently:
- Building `swift-driver` with library-evolution enabled fails because the compiler prints unqualified types for type extensions to the interface file and we are extending `TSCBasic.Process`, which is ambiguous with `Foundation.Process`. Brent has a fix PR to make the compiler print qualified extension type names here: https://github.com/apple/swift/pull/32780, but just throwing this fix in I am seeing other failures. More investigation needed. 
- Without library-evolution, the installation works for me locally, but does not work in Swift CI (e.g. here: https://github.com/apple/swift/pull/33918) because when building `LLBuild`, my local build figures out to use Xcode SDK's version of libSQLite, whereas in CI, the build uses the one in `/usr/local/lib/`, which only has an x86_64 slice so the arm64 build fails.

nkcsgexi:
@artemcm Just discussed @DougGregor about this, we could first disable `-enable-library-evolution` for now and use the binary module format for submission. The build system can link against libSwiftDriver in the default toolchain so we could start experimenting with the explicit module feature. Multiplexing libSwiftDriver can be a concern we tackle some point later.

artemcm:
@shahmishal , For the second point in the above point,  

> @artemcm Just discussed @DougGregor about this, we could first disable `-enable-library-evolution` for now and use the binary module format for submission. The build system can link against libSwiftDriver in the default toolchain so we could start experimenting with the explicit module feature. Multiplexing libSwiftDriver can be a concern we tackle some point later.

Sounds good! I'll push through with resolving the remaining issues without using `-enable-library-evolution`.

nkcsgexi:
> > @artemcm Just discussed @DougGregor about this, we could first disable `-enable-library-evolution` for now and use the binary module format for submission. The build system can link against libSwiftDriver in the default toolchain so we could start experimenting with the explicit module feature. Multiplexing libSwiftDriver can be a concern we tackle some point later.
> 
> Sounds good! I'll push through with resolving the remaining issues without using `-enable-library-evolution`.

Thank you!

artemcm:
This is currently functional (I can build a simple swift program that imports and uses `SwiftDriver` by pointing library and include paths to the built toolchain). 
In our chat, @nkcsgexi brought up a good point that we should probably be able to refactor `SwiftDriver` into a client library for build systems that does not itself depend on `llbuild`. We can do this refactoring as either a follow-up or a pre-requisite to this change.

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
Note: This must be merged in tandem with: https://github.com/apple/swift/pull/33918

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

DougGregor (approved):
This all looks reasonable to me.  Thank you!

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #247, opened 2020-09-11
Update swift-argument-parser dependency

natecook1000:
The issue with `swift-argument-parser` version 0.3.1 is being resolved in https://github.com/apple/swift-package-manager/pull/2919, so we need to coordinate the dependency update here as well.


natecook1000:
@swift-ci Please test

DougGregor (approved):

PR #248, opened 2020-09-12
Decode paths as VirtualPaths from -print-target-info output

owenv:
This change decodes the paths in `-print-target-info` as `VirtualPath`s to simplify their use throughout the driver. This initial change is pretty small, but I think it'll make it easier to use the runtime library and resource dir paths reported by the frontend throughout the rest of the driver. 

I thought about modifying the frontend so it always resolves these paths to absolute, but I think that would just create problems for distributed build systems in the future.

### DIFF
--- a/Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
+++ b/Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
@@ -62,6 +62,7 @@ extension SwiftVersion: Codable {
 }
 
 /// Describes information about the target as provided by the Swift frontend.
+@dynamicMemberLookup
 @_spi(Testing) public struct FrontendTargetInfo: Codable {
   struct CompatibilityLibrary: Codable {
     enum Filter: String, Codable {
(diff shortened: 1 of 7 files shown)


owenv:
@swift-ci test

Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
```diff
@@ -98,10 +98,30 @@ public struct FrontendTargetInfo: Codable {
...
+    let sdkPath: VirtualPath?
+    let runtimeLibraryPaths: [VirtualPath]
+    let runtimeLibraryImportPaths: [VirtualPath]
+    let runtimeResourcePath: VirtualPath
+
+    enum CodingKeys: String, CodingKey {
+      case sdkPath, runtimeLibraryPaths, runtimeLibraryImportPaths, runtimeResourcePath
+    }
+
+    init(from decoder: Decoder) throws {
```
DougGregor, line 110: Somewhat sad to have to maintain these manually. Should we consider a VirtualPath wrapper of some sort that does Codable the way we want?

DougGregor (approved):
This does seem easier for clients of the target info, so +1 from me

Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
owenv: Good idea, I added a `TextualVirtualPath` wrapper for this.

owenv:
I made a couple changes in this version:
- `TextualVirtualPath` is a small wrapper to make decoding VirtualPaths from strings easier
- I added dynamic member lookup to avoid some unnecessary nesting when accessing members of FrontendTargetInfo.Paths
- rebased on the new file list changes

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci please test

owenv:
Oops, I forgot about this one. Let's see if the tests still pass

@swift-ci test

PR #249, opened 2020-09-12
Make `FeatureAvailability` more explicit

sjavora:
Looking at the `Triple.FeatureAvailability` initializers, I came across this comment:
```swift
/// Each version parameter is `Optional`; a `nil` value means the feature is		
/// not supported on any version of that platform. Use `Triple.Version.zero`		
/// for a feature that is available in all versions.
```
It stuck out to me because Swift is a very expressive language and can model this better than with a convention described in a comment (which was repeated in both initializers). So I did:

```swift
public enum Availability {
  case unavailable
  case available(since: Version)
  case availableInAllVersions
}
```

I'm not thrilled about `availableInAllVersions`, but it is explicit. Also, nesting `Availability` in `FeatureAvailability` is probably not great, but I couldn't think of a better name.

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple+Platforms.swift
+++ b/Sources/SwiftDriver/Utilities/Triple+Platforms.swift
@@ -310,27 +310,28 @@ extension Triple {
   ///
   /// - SeeAlso: `Triple.supports(_:)`
   public struct FeatureAvailability {
-    public let macOS: Triple.Version?
-    public let iOS: Triple.Version?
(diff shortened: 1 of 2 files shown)


owenv (approved):
Looks good to me, thanks!

owenv:
@swift-ci test

owenv:
It looks like there may be an issue with the macOS CI, once that's resolved we can run those tests and merge this

sjavora:
@owenv macOS passed.

PR #250, opened 2020-09-12
Remove FileSystem usage in ExplicitModuleBuildHandler

owenv:
Instead of writing temporary files directly, it now uses a new VirtualPath case, temporaryFileWithKnownContents, to record the encoded artifact info. ArgsResolver is then responsible for writing the contents to a temporary file. This works pretty similar to file lists, but the contents of the files in this case don't require any path resolution.

Builds on #248, but only to avoid merge conflicts. Only the second commit is new.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -182,7 +182,7 @@ public struct Driver {
       return cwd.map { AbsolutePath($0, path) }
     case nil:
       return nil
-    case .standardInput, .standardOutput, .temporary, .fileList:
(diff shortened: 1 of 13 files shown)


owenv:
Looks like this will require an SPM update too

DougGregor (approved):
Yes, I love this direction! Getting out of the business of writing temporarily files throughout the driver is 💯

artemcm (approved):
Awesome.
Thank you!

artemcm:
@swift-ci please test

cltnschlosser:
Long term the filelist case could be removed, since it's really just a temporary file with known contents as well.

shahmishal:

PR #252, opened 2020-09-13
[WIP] Implement WindowsToolchain

stevapple:
Implemented the Windows-specific toolchain.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -370,6 +370,9 @@ public struct Driver {
     self.numParallelJobs = Self.determineNumParallelJobs(&parsedOptions, diagnosticsEngine: diagnosticEngine, env: env)
 
     Self.validateWarningControlArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
+    Self.validateCRuntimeArgs(&parsedOptions,
+                              diagnosticEngine: diagnosticsEngine,
+                              targetTriple: self.frontendTargetInfo.target.triple)
     Self.validateProfilingArgs(&parsedOptions,
                                fileSystem: fileSystem,
                                workingDirectory: workingDirectory,
@@ -1657,6 +1660,14 @@ extension Driver {
     if parsedOptions.hasArgument(.suppressWarnings) &&
         parsedOptions.hasFlag(positive: .warningsAsErrors, negative: .noWarningsAsErrors, default: false) {
       diagnosticEngine.emit(Error.conflictingOptions(.warningsAsErrors, .suppressWarnings))
+    }
+  }
+
+  static func validateCRuntimeArgs(_ parsedOptions: inout ParsedOptions,
+                                   diagnosticEngine: DiagnosticsEngine,
+                                   targetTriple: Triple) {
+    if parsedOptions.hasArgument(.libc) && !targetTriple.isWindows {
+      diagnosticEngine.emit(.error_unsupported_opt_for_target(arg: "-libc", target: targetTriple))
     }
   }
 
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -373,8 +373,7 @@ public struct Driver {
     Self.validateProfilingArgs(&parsedOptions,
                                fileSystem: fileSystem,
                                workingDirectory: workingDirectory,
-                               diagnosticEngine: diagnosticEngine,
-                               targetTriple: self.frontendTargetInfo.target.triple)
+                               diagnosticEngine: diagnosticEngine)
     Self.validateCompilationConditionArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
     Self.validateFrameworkSearchPathArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
     Self.validateCoverageArgs(&parsedOptions, diagnosticsEngine: diagnosticEngine)
@@ -1300,6 +1299,11 @@ extension Driver {
         sanitizerSupported = false
       }
 
+      // Currently only ASAN is supported on Windows.
+      if sanitizer != .address && targetTriple.isWindows {
+        sanitizerSupported = false
+      }
+
       if !sanitizerSupported {
         diagnosticEngine.emit(
           .error_unsupported_opt_for_target(
@@ -1659,30 +1663,10 @@ extension Driver {
   static func validateProfilingArgs(_ parsedOptions: inout ParsedOptions,
                                     fileSystem: FileSystem,
                                     workingDirectory: AbsolutePath?,
-                                    diagnosticEngine: DiagnosticsEngine,
-                                    targetTriple: Triple) {
+                                    diagnosticEngine: DiagnosticsEngine) {
     if parsedOptions.hasArgument(.profileGenerate) &&
         parsedOptions.hasArgument(.profileUse) {
       diagnosticEngine.emit(Error.conflictingOptions(.profileGenerate, .profileUse))
-    }
-
-    // Windows executables should be profiled with ETW, whose support needs to be
-    // implemented before we can enable the option.
-    if targetTriple.isWindows {
-      if parsedOptions.hasArgument(.profileGenerate) {
-        diagnosticEngine.emit(
-          .error_unsupported_opt_for_target(
-            arg: "-profile-generate",
-            target: targetTriple)
-          )
-      }
-      if parsedOptions.hasArgument(.profileUse) {
-        diagnosticEngine.emit(
-          .error_unsupported_opt_for_target(
-            arg: "-profile-use=",
-            target: targetTriple)
-          )
-      }
     }
 
     if let profileArgs = parsedOptions.getLastArgument(.profileUse)?.asMultiple,
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -1702,7 +1702,7 @@ extension Triple {
     case .freeBSD, .haiku:
       return GenericUnixToolchain.self
     case .win32:
-      fatalError("Windows target not supported yet")
+      return WindowsToolchain.self
     default:
       diagnosticsEngine.emit(.error_unknown_target(triple))
       throw Diagnostics.fatalError
@@ -1715,7 +1715,7 @@ extension Driver {
   #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
   static let defaultToolchainType: Toolchain.Type = DarwinToolchain.self
   #elseif os(Windows)
-  static let defaultToolchainType: Toolchain.Type = { fatalError("Windows target not supported yet") }()
+  static let defaultToolchainType: Toolchain.Type = WindowsToolchain.self
   #else
   static let defaultToolchainType: Toolchain.Type = GenericUnixToolchain.self
   #endif
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -361,7 +361,8 @@ public struct Driver {
     Self.validateProfilingArgs(&parsedOptions,
                                fileSystem: fileSystem,
                                workingDirectory: workingDirectory,
-                               diagnosticEngine: diagnosticEngine)
+                               diagnosticEngine: diagnosticEngine,
+                               targetTriple: self.frontendTargetInfo.target.triple)
     Self.validateCompilationConditionArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
     Self.validateFrameworkSearchPathArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
     Self.validateCoverageArgs(&parsedOptions, diagnosticsEngine: diagnosticEngine)
@@ -1642,10 +1643,30 @@ extension Driver {
   static func validateProfilingArgs(_ parsedOptions: inout ParsedOptions,
                                     fileSystem: FileSystem,
                                     workingDirectory: AbsolutePath?,
-                                    diagnosticEngine: DiagnosticsEngine) {
+                                    diagnosticEngine: DiagnosticsEngine,
+                                    targetTriple: Triple) {
     if parsedOptions.hasArgument(.profileGenerate) &&
         parsedOptions.hasArgument(.profileUse) {
       diagnosticEngine.emit(Error.conflictingOptions(.profileGenerate, .profileUse))
+    }
+
+    // Windows executables should be profiled with ETW, whose support needs to be
+    // implemented before we can enable the option.
+    if targetTriple.isWindows {
+      if parsedOptions.hasArgument(.profileGenerate) {
+        diagnosticEngine.emit(
+          .error_unsupported_opt_for_target(
+            arg: "-profile-generate",
+            target: targetTriple)
+          )
+      }
+      if parsedOptions.hasArgument(.profileUse) {
+        diagnosticEngine.emit(
+          .error_unsupported_opt_for_target(
+            arg: "-profile-use=",
+            target: targetTriple)
+          )
+      }
     }
 
     if let profileArgs = parsedOptions.getLastArgument(.profileUse)?.asMultiple,
--- a/Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
@@ -119,13 +119,6 @@ extension Toolchain {
     parsedOptions: inout ParsedOptions,
     isShared: Bool
   ) throws -> Bool {
-    // Currently only ASAN is supported on Windows, but clang builds may
-    // include runtime libraries for unsupported sanitizers.  Manually
-    // disable unsupported sanitizers.
-    if targetTriple.isWindows && sanitizer != .address {
-      return false
-    }
-
     let runtimeName = try runtimeLibraryName(
       for: sanitizer,
       targetTriple: targetTriple,
--- a/Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
@@ -54,21 +54,6 @@ extension Toolchain {
     for triple: Triple,
     parsedOptions: inout ParsedOptions
   ) throws -> AbsolutePath {
-    /// Use the one in `VCTools` if exists on Windows.
-    if triple.isWindows,
-      let vctools = env["VCToolsInstallDir"],
-      let root = try? AbsolutePath(validating: vctools) {
-        let archName: String = {
-          switch triple.arch {
-          case .aarch64: return "arm64"
-          case .arm: return "arm"
-          case .x86: return "x86"
-          case nil, .x86_64: return "x64"
-          default: fatalError("unknown arch \(triple.archName) on Windows")
-          }
-        }()
-      return root.appending(components: "lib", archName)
-    }
     return try computeResourceDirPath(for: triple,
                                       parsedOptions: &parsedOptions,
                                       isShared: true)
--- a/Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
@@ -54,12 +54,20 @@ extension Toolchain {
     for triple: Triple,
     parsedOptions: inout ParsedOptions
   ) throws -> AbsolutePath {
-    return try computeResourceDirPath(for: triple,
-                                      parsedOptions: &parsedOptions,
-                                      isShared: true)
-      .parentDirectory // Remove platform name.
-      .appending(components: "clang", "lib",
-                 triple.platformName(conflatingDarwin: true)!)
+    #if os(Windows)
+      return try getToolPath(.swiftCompiler)
+        .parentDirectory // remove /swift
+        .parentDirectory // remove /bin
+        .appending(components: "lib", "swift", "clang", "lib",
+                   triple.platformName(conflatingDarwin: true)!)
+    #else
+      return try computeResourceDirPath(for: triple,
+                                        parsedOptions: &parsedOptions,
+                                        isShared: true)
+        .parentDirectory // Remove platform name.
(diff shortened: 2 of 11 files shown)


stevapple:
CC @compnerd , I hope to get SPM fully usable on Windows through `swift-driver`.

Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
```diff
@@ -48,6 +54,21 @@ extension Toolchain {
     for triple: Triple,
     parsedOptions: inout ParsedOptions
   ) throws -> AbsolutePath {
+    /// Use the one in `VCTools` if exists on Windows.
+    if triple.isWindows,
+      let vctools = env["VCToolsInstallDir"],
+      let root = try? AbsolutePath(validating: vctools) {
+        let archName: String = {
+          switch triple.arch {
+          case .aarch64, .aarch64_32: return "arm64"
```
compnerd, line 63: `.aarch64_32` makes no sense on Windows.  There is no 32-bit ABI for ARM64 on Windows that is distinct from ARMv7.  That should be a `fatalError`.

Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
```diff
@@ -48,6 +54,21 @@ extension Toolchain {
     for triple: Triple,
     parsedOptions: inout ParsedOptions
   ) throws -> AbsolutePath {
+    /// Use the one in `VCTools` if exists on Windows.
+    if triple.isWindows,
+      let vctools = env["VCToolsInstallDir"],
```
compnerd, line 59: I would rather not dig for clang resources from the `VCToolsInstallDir`.  You should use the path relative location for clang resources.

Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
```diff
@@ -258,3 +279,5 @@ extension DarwinToolchain {
   }
 
 }
+
+// TODO: See whether Windows needs `addArgsToLinkStdlib`.
```
compnerd, line 283: Yes, it will definitely needs arguments for that.  At the very least, you need to ensure that the ucrt link matches the stdlib.

Sources/SwiftDriver/Jobs/WindowsToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,219 @@
...
+    case .dynamicLibrary:
+      // Same options as an executable, just with '-shared'
+      commandLine.appendFlags("-parse-as-library", "-emit-library")
+      fallthrough
+    case .executable:
+      if !targetTriple.triple.isEmpty {
+        commandLine.appendFlag("-target")
+        commandLine.appendFlag(targetTriple.triple)
+      }
+        commandLine.appendFlag("-emit-executable")
```
compnerd, line 38: Can you duplicate rather than try to share the logic here?  You are always adding `-emit-executable` even in the shared case.

Nit: indentation is off.

Sources/SwiftDriver/Jobs/WindowsToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,219 @@
...
+
+        if hasRuntimeArgs {
+          // FIXME: We probably shouldn't be adding an rpath here unless we know
+          //        ahead of time the standard library won't be copied.
+          for path in runtimePaths {
+            commandLine.appendFlag(.Xlinker)
+            commandLine.appendFlag("-rpath")
+            commandLine.appendFlag(.Xlinker)
+            commandLine.appendPath(path)
+          }
```
compnerd, line 104: This stanza doesn't make sense.  Windows does not support `-rpath`.

Sources/SwiftDriver/Jobs/WindowsToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,219 @@
...
+          for: targetTriple,
+          parsedOptions: &parsedOptions,
+          isShared: false
+        )
+
+        if staticExecutable {
+          linkFilePath = linkFilePath?.appending(component: "static-executable-args.lnk")
+        } else if staticStdlib {
+          linkFilePath = linkFilePath?.appending(component: "static-stdlib-args.lnk")
+        } else {
```
compnerd, line 159: I don't think that we want to use these.  The response file support in GNU is different from MSVC.  Additionally, Windows is significantly friendlier - we can use linker directives to achieve this in a much more reasonable fashion.  I think that this should be something like `if !no_stdlib { commandLine.appendFlag("-lswiftCore") }`.

Sources/SwiftDriver/Jobs/WindowsToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,219 @@
...
+          linkFilePath = nil
+          commandLine.appendFlag("-lswiftCore")
+        }
+
+        if let linkFile = linkFilePath {
+          guard fileSystem.isFile(linkFile) else {
+            fatalError("\(linkFile.pathString) not found")
+          }
+          commandLine.append(.responseFilePath(.absolute(linkFile)))
+        }
```
compnerd, line 169: This can be entirely dropped.

Sources/SwiftDriver/Jobs/WindowsToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,219 @@
...
+
+        // Explicitly pass the target to the linker
+        commandLine.appendFlag("--target=\(targetTriple.triple)")
```
compnerd, line 172: Please use the split form.

Sources/SwiftDriver/Jobs/WindowsToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,219 @@
...
+            commandLine.appendFlag("-lBlocksRuntime")
+            commandLine.appendFlag("-ldispatch")
+          }
+        }
+
+        if parsedOptions.hasArgument(.profileGenerate) {
+          let libProfile = sharedResourceDirPath
+            .parentDirectory // remove platform name
+            .appending(components: "clang", "lib", targetTriple.osName,
+                                   "libclangrt_profile-\(targetTriple.archName).a")
```
compnerd, line 194: Please derive the prefix and suffix from the platform settings.

Sources/SwiftDriver/Jobs/WindowsToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,219 @@
...
+          commandLine.appendPath(libProfile)
+
+          // HACK: Hard-coded from llvm::getInstrProfRuntimeHookVarName()
+          commandLine.appendFlag("-u__llvm_profile_runtime")
```
compnerd, line 198: There is no `-u` flag that is part of the LINK specification.

Sources/SwiftDriver/Jobs/WindowsToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,219 @@
...
+        )
+        try commandLine.appendAllArguments(.Xlinker, from: &parsedOptions)
+        try commandLine.appendAllArguments(.XclangLinker, from: &parsedOptions)
+
+          // This should be the last option, for convenience in checking output.
+        commandLine.appendFlag(.o)
+        commandLine.appendPath(outputFile)
+        return clangPath
+    }
+
```
compnerd, line 217: I don't see `-use-ld` being forwarded here?

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -135,7 +135,7 @@ extension Toolchain {
       return path
     } else if let path = try? xcrunFind(executable: executable) {
       return path
-    } else if !["swift-frontend", "swift"].contains(executable),
+    } else if !["swift-frontend", "swift", "swift-frontend.exe", "swift.exe"].contains(executable),
```
compnerd, line 138: Can you derive the names from the platform instead?

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
```diff
@@ -0,0 +1,94 @@
...
+    self.env = env
+    self.executor = executor
+    self.fileSystem = fileSystem
+  }
+
+  public func makeLinkerOutputFilename(moduleName: String, type: LinkOutputType) -> String {
+    switch type {
+    case .executable: return "\(moduleName).exe"
+    case .dynamicLibrary: return "\(moduleName).dll"
+    case .staticLibrary: return "\(moduleName).lib"
```
compnerd, line 37: You should add a case for an import library

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
```diff
@@ -0,0 +1,94 @@
...
+
+  private func lookupToolPath(_ tool: Tool) throws -> AbsolutePath {
+    switch tool {
+    case .swiftCompiler:
+      return try lookup(executable: "swift-frontend.exe")
+    case .staticLinker:
+      return try lookup(executable: "llvm-lib.exe")
+    case .dynamicLinker:
+      // FIXME: This needs to look in the tools_directory first.
+      return try lookup(executable: "lld-link.exe")
```
compnerd, line 61: I think that we should derive this based on `-use-ld`.  The default should be `link.exe`, unless otherwise specified.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
```diff
@@ -0,0 +1,94 @@
...
+    case .clang:
+      return try lookup(executable: "clang.exe")
+    case .swiftAutolinkExtract:
+      return try lookup(executable: "swift-autolink-extract.exe")
+    case .dsymutil:
+      return try lookup(executable: "dsymutil.exe")
```
compnerd, line 67: There is no `dsymutil` support on Windows.  The PDB is generated by `link`.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
```diff
@@ -0,0 +1,94 @@
...
+      return try lookup(executable: "swift-frontend.exe")
+    case .staticLinker:
+      return try lookup(executable: "llvm-lib.exe")
+    case .dynamicLinker:
+      // FIXME: This needs to look in the tools_directory first.
+      return try lookup(executable: "lld-link.exe")
+    case .clang:
+      return try lookup(executable: "clang.exe")
+    case .swiftAutolinkExtract:
+      return try lookup(executable: "swift-autolink-extract.exe")
```
compnerd, line 65: `swift-autolink-extract` should never be used when targeting Windows.  When will this hit?

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
```diff
@@ -0,0 +1,94 @@
...
+    toolPaths[tool] = path
+    return path
+  }
+
+  private func lookupToolPath(_ tool: Tool) throws -> AbsolutePath {
+    switch tool {
+    case .swiftCompiler:
+      return try lookup(executable: "swift-frontend.exe")
+    case .staticLinker:
+      return try lookup(executable: "llvm-lib.exe")
```
compnerd, line 58: This should be controlled by `-use-ld`.  The default should be `link`, `lld` if `-use-ld=lld` is specified.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
```diff
@@ -0,0 +1,94 @@
...
+    toolPaths[tool] = path
+  }
+
+  public func defaultSDKPath(_ target: Triple?) throws -> AbsolutePath? {
+    return nil
+  }
+
+  public var shouldStoreInvocationInDebugInfo: Bool { false }
+
+  public func runtimeLibraryName(
```
compnerd, line 87: What is this used for?  I would imagine that the `runtimeLibraryName` should be the import library, not the DLL.

compnerd (changes requested):

Sources/SwiftDriver/Jobs/Toolchain+LinkerSupport.swift
stevapple: You mean search for clang resources in the whole `%PATH%`?

Sources/SwiftDriver/Jobs/WindowsToolchain+LinkerSupport.swift
stevapple: Sorry, this is a bug😢Will fix in the next commit

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
stevapple: So how should we handle here? Directly return `nil` or return something else?

I’ve found `dsymutils.exe` in the packaged toolchain, so...

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
stevapple: Probably not, but it exists so I put it here. Return `nil` instead? I think it’s okay to keep it.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
stevapple: On Linux it returns things like `clang_rt.asan-x86_64.a`. It’s a typo here (I intended to use `.lib` but...), and sorrily I didn’t find any `exp`s.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
owenv: Trying to lookup dsymutil can probably just call fatalError(), I'm not sure why we're not already doing that on Linux

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
owenv: Same here, this can be a fatalError because trying to look it up with a Windows toolchain is a programming error in the driver.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
stevapple: Do you have a Windows on ARM device to see what the `arch` part in `clang_rt` is for `arm` and `aarch64`? I assumed them to `arm` and `arm64`, but not for sure. Or is there any way to install ARM runtime support on `x64` devices?

owenv: This is used indirectly by `parseSanitizerArgValues` to determine whether sanitizers are supported based on whether the libraries can be found on disk. It's not necessarily something we want to keep long term though, there's a FIXME:
```
// FIXME: Should we do this? This prevents cross-compiling with sanitizers
//        enabled.
```

compnerd: They should be `armv7` and `aarch64`.  The only one to use `arm64` is Apple.

stevapple: Windows has a very weird directory structure...For `x86_64`, it’s parent name is `x64`; for `i386`, it is `x86`. There’s even no `x86` here in `clang_rt`. we probably need a real Surface Pro X to make it sure.

compnerd: I think that the Windows directory structure actually makes far more sense.  i386 doesn't make sense, it isn't i386, the modern derivates of it became 486, 586, 686, Pentium, and then Core.  These are all x86 derivates, but far different from i386.  x64 makes sense, it is a contraction of x86_64.  Correct, the ASAN libraries are not from MSVC or Microsoft, they are from the work that Google did, which is why its not surprising at all that the library does not exist there.  I don't think that the Surface Pro X is going to help as there is no VS on the device AFAIK.

stevapple: Fixed this part

stevapple:
I’m working on the linker part, and I’m now troubled by the following two question:

- Can we use things like `link.exe` and `lib.exe` in Cygwin and MingW? That is, is `link.exe` and `lib.exe` available to everyone using Swift for Windows?
- Do we also have `llvm-lib.exe` and `lldb-link.exe` in Cygwin and MingW toolchains?

It’s fairly easy to configure the MSVC linkers and their LLVM drop-in substitutes (I think it can be done today), but what about not using `msvc` here?

Also, we’re not using `cl.exe` here, right?

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
stevapple: Sadly, we cannot generate an import library for the moment. The binary of Swift code doesn’t contain the flags necessary for it, and we also have no `.def` file. `link.exe` needs at least [one of four marks](https://docs.microsoft.com/zh-cn/cpp/build/reference/export-exports-a-function?view=vs-2019) to generate an import library.

stevapple: If we can fill out one of those, please point out for me🙏thx a lot

stevapple:
Another question that I think only @compnerd is capable of answering... Are we going to include `clangLibrary` with Swift? If the answer is “yes”, then things will be much easier because we don’t need to search for clang libraries.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
stevapple: P.S. `link.exe` can automatically infer import library name from output file name (by default, it *will* generate an import library as long as the requirements are satisfied), so I think there’s no need to add a specific `OutputLinkType` for it.

compnerd:
The MinGW path should not use the WindowsToolchain but rather the GNUToolchain much Like the model in clang.  MinGW is substantially different in ABI that you cannot mix and match the behavior.  In fact there's a large number of other things that won't work - like linking because MinGW would require auto link extract that the Windows path does not use as the MINGW doesn't support linker directives.  The other thing is that binutils isn't supported by swift because of issues with relocation handling.  In general, MinGW isn't ready yet, and would require a lot more than just tweaking the driver.

No, there's no reason to include clangrt, that isn't used on windows.

stevapple:
clang_rt is required for sanitizer support.

```swift
    // Check that we're one of the known supported targets for sanitizers.
    if !(targetTriple.isWindows || targetTriple.isDarwin || targetTriple.os == .linux)
```

I’m seeing this line of code. It should mean that we are going to provide sanitizer support on Windows?

stevapple:
Finally gets a solution for sanitizer support. I’m going to filter out any sanitizer other than ASAN. This means if anyone builds his copy with `clang_rt`, he’ll be able to use ASAN. Other sanitizers will be blocked even if their libraries exist.

owenv (commented):
Thanks for updating this! It looks pretty good to me, but I'm not very familiar with Windows so I wasn't able to review it thoroughly.

It would be really great if we could include some basic tests alongside this change as well, at least for the platform-specific linking. I think a lot of the windows-related integration tests from the main Swift repo will continue to fail until MSVC runtime handling is added, I haven't looked into how complicated that is. Maybe a couple cases could be added to `testLinking` in `SwiftDriverTests.swift` that use the windows target?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1658,12 +1659,32 @@ extension Driver {
...
+                                    diagnosticEngine: DiagnosticsEngine,
+                                    targetTriple: Triple) {
     if parsedOptions.hasArgument(.profileGenerate) &&
         parsedOptions.hasArgument(.profileUse) {
       diagnosticEngine.emit(Error.conflictingOptions(.profileGenerate, .profileUse))
     }
 
+    // Windows executables should be profiled with ETW, whose support needs to be
+    // implemented before we can enable the option.
+    if targetTriple.isWindows {
```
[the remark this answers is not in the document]
stevapple: 👌I’ll take the advice

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -38,6 +38,8 @@ public protocol Toolchain {
 
   var executor: DriverExecutor { get }
 
+  var executableSuffix: String { get }
```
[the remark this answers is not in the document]
stevapple: I think so. Fixed👌

stevapple:
> It would be really great if we could include some basic tests alongside this change as well, at least for the platform-specific linking. I think a lot of the windows-related integration tests from the main Swift repo will continue to fail until MSVC runtime handling is added, I haven't looked into how complicated that is. Maybe a couple cases could be added to testLinking in SwiftDriverTests.swift that use the windows target?

Of course I will. However, there is still an unimplemented function (see the only unresolved review😂) which should be implemented before implementing tests.

stevapple:
Shall we use MD or MT as default for Windows? @compnerd

compnerd:
I don't think that we should have a default for the value.  It _must_ match the value for the runtime that you are building against.  I suppose that one possibility is to query the `SDKSettings.plist` for the correct value like I have done in swift-package-manager.  But, really, I think that we should have the user specify the correct value.

stevapple:
I think it necessary. It’s one of our goals to allow user run `swiftc` without any additional options, which means we should be able to infer every parameter we need from the context. We can have multiple ways to infer though, plist is one of them, but maybe we also need a hard-coded default.

compnerd:
If you are adamant that you must have a default, then you must serialize the compile options into the `Swift.swiftmodule` and then deserialize it to find the default at runtime.  Otherwise, whatever default you choose is the wrong one - my runtime will always be the inverse of whatever the default is, and therefore you have just built a binary that will arbitrarily crash.

stevapple:
I‘d like to know if `SDKSettings.plist` is built alongside with Windows SDK. If it is (which means every copy of Windows SDK contains `SDKSettings.plist`), we won’t need an extra default. That would be great and simple.

compnerd:
It is not, it is only something that is built on my builds of the toolchain; anyone building the toolchain manually will have to deal with that themselves.

shahmishal:

PR #254, opened 2020-09-14
Update isTopLevelOutput predicate

owenv:
- canonical sib, raw sib, imported modules, and index data are always top level (as far as I know)
- bitcode is top-level if it's the compiler output type (in other words, it's not being embedded)

This fixes:
- Frontend/sil-primary-file-with-sib.swift
- reproducible-builds/swiftc-emit-bc.swift
- reproducible-builds/swiftc-emit-sib.swift
- reproducible-builds/swiftc-emit-sibgen.swift


owenv:
@swift-ci test

DougGregor (approved):
These all look correct to me, thank you!

PR #255, opened 2020-09-14
[Integration Tests] Fixup the path to llvm's lit.py to reflect the monorepo structure

artemcm:
Matching the directory structure created by the `update-checkout` script.
With this change, we can run integration tests according to the instructions in the README.


DougGregor (approved):

artemcm:
@swift-ci please test

owenv (approved):
Thanks! I just realized I still have an `llvm` symlink from the transition, so I never noticed this :)

PR #259, opened 2020-09-16
Add a lit feature for the c++ driver

owenv:
Counterpart to https://github.com/apple/swift/pull/33961 which has the context.

This PR passes the swift_driver parameter to lit so that it disables the integrated_driver feature.


owenv:
@swift-ci test

DougGregor (approved):
The change here LGTM, but please see my terminology comment on the Swift compiler-side change

PR #261, opened 2020-09-17
Mark more types and Driver state as internal/@_spi(Testing)

artemcm:


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -122,101 +122,101 @@ public struct Driver {
...
-  public let inputFiles: [TypedVirtualPath]
+  @_spi(Testing) public let inputFiles: [TypedVirtualPath]
 
   /// The last time each input file was modified, recorded at the start of the build.
-  public let recordedInputModificationDates: [TypedVirtualPath: Date]
+  @_spi(Testing) public let recordedInputModificationDates: [TypedVirtualPath: Date]
 
   /// The mapping from input files to output files for each kind.
-  internal let outputFileMap: OutputFileMap?
+  @_spi(Testing) public let outputFileMap: OutputFileMap?
```
cltnschlosser, line 137: This increased visibility from internal to public/spi.

cltnschlosser (commented):
What's the difference/benefit of using spi over internal and `@testable`?

Sources/SwiftDriver/Driver/Driver.swift
artemcm: 🤦‍♂️ Thank you!

artemcm:
> What's the difference/benefit of using spi over internal and `@testable`?

One of the key benefits is that `@testable` is not compatible with textual interfaces and we would really like to build SwiftDriver with library-evolution quite soon. Still using `@testable` would mean building twice, once for testing and then again for distribution. 

A little bit more reasoning here:
https://forums.swift.org/t/the-future-of-testable/9655/14

artemcm:
@swift-ci please test

owenv (approved):
Nice!

artemcm:
@swift-ci please test

PR #262, opened 2020-09-19
Add support for -driver-time-compilation

owenv:
The PR sets up a basic stats reporter and uses it to implement -driver-time compilation. It's not thread safe, which isn't a problem now because JobExecutionDelegate callbacks all happen on the same queue, but might need to be reworked a bit when adding the driver performance counters. It also only works on macOS and Linux currently, but there's a stub so everything should keep compiling on other platforms.

### DIFF
--- a/Sources/SwiftDriver/Utilities/System.swift
+++ b/Sources/SwiftDriver/Utilities/System.swift
@@ -52,3 +52,61 @@ func commandLineFitsWithinSystemLimits(path: String, args: [String]) -> Bool {
   return true
 }
 #endif
+
+struct Microseconds {
+  var value: UInt64
+  private static let usecPerSec: UInt64 = 1_000_000
+
+  static func +(lhs: Microseconds, rhs: Microseconds) -> Microseconds {
+    .init(value: lhs.value + rhs.value)
+  }
+
+  static func -(lhs: Microseconds, rhs: Microseconds) -> Microseconds {
+    .init(value: lhs.value - rhs.value)
+  }
+
+  init(value: UInt64) {
+    self.value = value
+  }
+
+  #if os(macOS) || os(Linux)
+  init(timeVal: timeval) {
+    value = Self.usecPerSec * UInt64(timeVal.tv_sec) + UInt64(timeVal.tv_usec)
+  }
+  #endif
+
+  var seconds: Double {
+    Double(value) / 1_000_000
+  }
+}
+
+struct TimeValues {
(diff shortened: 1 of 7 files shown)


owenv:
@swift-ci please test

owenv:
@swift-ci please test Linux

owenv:
@swift-ci please test Linux

owenv:
@swift-ci please test

owenv:
@swift-ci please test macOS

owenv:
@swift-ci please test

owenv:
@swift-ci please test

owenv:
@swift-ci please test

owenv:
@swift-ci please test

Sources/SwiftDriver/Utilities/System.swift
```diff
@@ -52,3 +52,61 @@ func commandLineFitsWithinSystemLimits(path: String, args: [String]) -> Bool {
...
+  }
+}
+
+struct TimeValues {
+  let userTime: Microseconds
+  let systemTime: Microseconds
+  let wallTime: Microseconds
+}
+
+#if os(macOS) || os(Linux)
```
DougGregor, line 89: We can probably ask @compnerd about the appropriate Windows API, or dig it out of the llvm sources.

DougGregor (approved):
This looks useful! I suppose that based on your other PR to split out the execution bits of the driver, you would want the new delegate to live in SwiftDriverExecution?

Sources/SwiftDriver/Utilities/System.swift
owenv: yeah, it looks like Windows has some roughly equivalent API that LLVM uses: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getprocesstimes. I'll see if I can setup a Windows build to test this

owenv:
I was planning on keeping the execution delegates in the main SwiftDriver library, at least for now. It makes handling `-v`, `-parseable-output`, and `-driver-time-compilation` easier because we can just check them in `Driver.run` and configure the delegate appropriately. Once the delegate is extended to support the `-stats-output-dir` counters, I think we'll also want to be able to use those with SPM, which won't import the new library.

DougGregor:
Ah, I understand now why the delegates remain here. Thanks for clarifying!

shahmishal:

PR #263, opened 2020-09-20
Extract everything depending on LLBuild into a SwiftDriverExecution library

owenv:
- Moved llbuild.Swift and the SwiftDriverExecutor into the new library
- Avoid reexporting llbuildSwift through the new library (I don't know why it was using @_exported before, I switched it to an @_implementationOnly import)
- Update SPM and CMake builds
- SwiftDriver and SwiftDriverExecution share a test target, splitting it didn't seem worth the trouble

This is a WIP because I'm guessing I won't get the CMake changes right on the first try and I still need to update SPM. This should hopefully make https://github.com/apple/swift-driver/pull/246 and related changes simpler, although splitting the libraries is a bit of a pain in the short term.


owenv:
Just to make sure I didn't completely break anything...

@swift-ci please test

DougGregor (approved):
This is great! Much better layering to separate out the execution bits, which several clients (like SwiftPM) don't need. We should be able to simplify the bootstrap script in CMake with this

artemcm (approved):
This is really great.

owenv:
I've now tested the CMake build locally and everything seemed to work, so once https://github.com/apple/swift-package-manager/pull/2946 is merged I'll run the full test suite with this PR and then it should be mergeable.

DougGregor:
Very excited for this to land :)

owenv:
Full swift-ci passed (https://github.com/apple/swift/pull/34040), so this should be good to go!

@swift-ci test

DougGregor:
Fantastic, go for it once CI completes!

PR #264, opened 2020-09-20
Make joinedArguments helper public so it can be used by SPM

owenv:
This is a tiny change so that I can merge this, then update SPM to use its own executor instead of `SwiftDriverExecutor`, and then land https://github.com/apple/swift-driver/pull/263 without a cross-repository merge.


owenv:
@swift-ci test

DougGregor (approved):

PR #267, opened 2020-09-21
Update llbuild CMake directions

owenv:
I had to pass -DCMAKE_OSX_ARCHITECTURES=x86_64 today to get the build to work, the error was pretty vague but I think it must be related to macOS on arm64


owenv:
@swift-ci test

DougGregor (approved):
This is a bug we should fix, but until then I agree that it's best to document the procedure that works here. Thanks!

owenv:
Sounds good!

PR #270, opened 2020-09-23
Refactoring plan standard compile

davidungar:
Prior to work on incremental compilation, refactor planStandardCompile in order to clarify causality chains.
Does change the order of planned jobs in jobs, but I believe that isn't the scheduled order, anyway.

Passing some mutable state around via "add" functions to clarify that the callee is only adding to the collections in question.


davidungar:
@swift-ci please test

davidungar:
@DougGregor This is the same thing you already reviewed. It seems that I had to redo the PR after the master-main switch. Thanks!

DougGregor (approved):

PR #273, opened 2020-09-23
Set the line-width parameter on the Yams `serialize` routine to unlimited.

artemcm:
The default "preferred line-width" setting appears to be `80` characters. At that setting, encoding a filesystem path that is longer than `80` characters long and contains a whitespace at an index `> 80`, a line-break will be inserted at that index, which then results in serializing a malformed `outputFileMap` that contains invalid paths the frontend will not be able to read.


artemcm:
@swift-ci please test

cltnschlosser (approved):
I think the only thing I actually ended up using Yams for here was to quote and escape the paths for output file map serialization. And other use of Yaml seems to be being replaced, so long term we may want to replace this with something else. It was just the easiest way to escape the paths.

DougGregor (approved):
I'm fine with this change, although I do wonder if we can eliminate the use of Yams here in the future.

PR #274, opened 2020-09-24
[Incremental] Recreate https://github.com/apple/swift-driver/pull/180 w/ small changes

davidungar:
Essentially recreate https://github.com/apple/swift-driver/pull/180 to provide a hook for incremental compilation to schedule the 2nd wave of compile jobs.

### DIFF
--- a/Sources/SwiftDriver/Driver/CompundExecutionDelegate.swift
+++ b/Sources/SwiftDriver/Driver/CompundExecutionDelegate.swift
@@ -1,0 +1,38 @@
+//===------CompoundExecutionDelegate.swift --------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import TSCBasic
+
+/// A delegate compounds the effects of multiple delegates and collects discovered jobs.
+public struct CompoundExecutionDelegate: JobExecutionDelegate {
+  var delegates: [JobExecutionDelegate]
+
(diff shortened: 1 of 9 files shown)


davidungar:
@swift-ci please test

owenv (commented):
The code changes look good (I just added a couple minor comments), but please add the two new files to CMakeLists.txt, the CI testing in this repository doesn't currently test the CMake build.

It does seem like it's going to be difficult to integrate this dynamic job discovery into SPM's build graph, but I haven't come up with any good alternatives that could support `-incremental`, so I think we can figure that out later

davidungar:
Thank you for checking this! Really appreciate your care and effort.

I have been working to understand the llbuild architecture and the various strata: CoreBindings, LLbuild, etc. I was wondering about the thought process you went through.  For instance, is there some way to make incremental compilation work just with the right inputs and outputs, and the hook in llbuild that says an input has changed?
The link job could have as an input the driver graph step that verifies no more compilations are needed. That step would take an input which is the swiftdeps files of the 1st wave of compile jobs. The link job would have all the .o files as its input. If the driver dep graph step finds the need for additional compilations, it would tell llbuild that the .o inputs for the link job are invalid and llbuild would then schedule the needed compilations.
Is this crazy talk? Or is it a more conceptually economical way than the design in this PR?
(I might hold off on landing this, until I figure out that the graph-strategy is worse.)
+ @DougGregor

owenv:
To be honest, I'm really not familiar enough with llbuild to know if that would work or not, but it sounds promising! #180 was just the solution that seemed simplest at the time to get incremental builds up and running in the command line driver, but I didn't put a lot of thought into making it work with libSwiftDriver clients and eventually moved on to working on other things.

davidungar:
Hey, thanks a lot! It's very helpful to know what those who came here before me were thinking.￼￼

tschuett:
Isn't the problem you want to solve independent of llbuild. Some unknown user of the swift-driver could use his private job scheduler.
- You want to be informed when a `Job` finishes
- You want to schedule more `Job`s depending on information that you extracted from the finished job.

PR #277, opened 2020-09-27
[NFC] Use frontend target info to print compiler version in -v mode

owenv:
No functional changes, but it eliminates an unnecessary frontend invocation


owenv:
@swift-ci test

DougGregor (approved):
Makes sense!

cltnschlosser:
Nice!

PR #278, opened 2020-09-27
Add support for loaded module trace emission

owenv:
Since the docs for this aren't open source, I'm not sure my test cases are all that great, but all the relevant integration tests are passing now and the implementation wasn't as complicated as I expected. The use of a "claimed" bit on `Driver.loadedModuleTracePath` is definitely a bit of a hack though.

Fixes:
Driver/loaded_module_trace.swift
Driver/loaded_module_trace_append.swift
Driver/loaded_module_trace_env.swift
Driver/loaded_module_trace_foundation.swift
Driver/loaded_module_trace_header.swift
Driver/loaded_module_trace_multifile.swift
Driver/loaded_module_trace_spi.swift

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -467,13 +467,19 @@ public struct Driver {
         compilerMode: compilerMode,
         outputFileMap: self.outputFileMap,
         moduleName: moduleOutputInfo.name)
-    self.loadedModuleTracePath = try Self.computeSupplementaryOutputPath(
+
+    if let loadedModuleTraceEnvVar = env["SWIFT_LOADED_MODULE_TRACE_FILE"] {
+      self.loadedModuleTracePath = try VirtualPath(path: loadedModuleTraceEnvVar)
+    } else {
+      self.loadedModuleTracePath = try Self.computeSupplementaryOutputPath(
         &parsedOptions, type: .moduleTrace, isOutputOptions: [.emitLoadedModuleTrace],
         outputPath: .emitLoadedModuleTracePath,
         compilerOutputType: compilerOutputType,
         compilerMode: compilerMode,
         outputFileMap: self.outputFileMap,
         moduleName: moduleOutputInfo.name)
+    }
+
     self.tbdPath = try Self.computeSupplementaryOutputPath(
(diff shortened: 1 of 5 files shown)


owenv:
@swift-ci test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -212,8 +212,10 @@ public struct Driver {
   /// Path to the Objective-C generated header.
   let objcGeneratedHeaderPath: VirtualPath?
 
-  /// Path to the loaded module trace file.
-  let loadedModuleTracePath: VirtualPath?
+  /// Path to the loaded module trace file, and whether it has been claimed by
+  /// a frontend job. It's important to track this because only one frontend
+  /// job should be used for module trace emission.
+  var loadedModuleTracePath: (VirtualPath, claimed: Bool)?
```
DougGregor, line 218: Mutable state, hmmm. Is there some way we could ensure that only one fronted job gets assigned this task without having this but if mutable state? E.g., have planning treat the first batch as special?

DougGregor (approved):
One question about mutable state, but otherwise it's looking good. @varungandhi-apple might want to take a peek, but that can be post-merge review

Sources/SwiftDriver/Driver/Driver.swift
owenv: Passing that info through compile planning adds a little bit of boilerplate, but yeah, it's probably best to avoid mutable state here.

owenv:
@swift-ci test

DougGregor (approved):
Thank you!

varungandhi-apple:
Are the trace tests being run with the new driver as well? Or were they already being run and failing? I'm not sure how the testing is structured there, I don't see anything in this PR adding something which makes sure those tests are run.

cltnschlosser:
> Are the trace tests being run with the new driver as well? Or were they already being run and failing? I'm not sure how the testing is structured there, I don't see anything in this PR adding something which makes sure those tests are run.

They aren’t run in CI, but we have a mechanism to run them. So they were failing before.

owenv:
Yeah, `IntegrationTests.swift` is setup to run them if the path to a checkout of the main Swift repo is provided. As @cltnschlosser mentioned, none of the CI jobs are running these yet, so I just verified it locally

PR #279, opened 2020-09-29
[Explicit Module Builds] Keep track of PCMArgs at which clang modules have been scanned

artemcm:
- Add a new `InterModuleDependencyGraph` field in `ClangModuleDetails` to capture against which targets this module has been scanned. The `directDependencies` of this module then represent the super-set of dependencies at all of the specified targets. 
- When performing a batched, versioned-PCM re-scan, only perform the re-scan at sets of PCMArgs that are "new" to this module to potentially discover additional dependencies. 

On its own, this change will not yield much improvement, but it is a necessary piece of the puzzle for a follow-up change that will build on top of this to drastically reduce the amount of redundant re-scanning that is currently happening in multi-module build contexts.


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #280, opened 2020-09-29
[Explicit Module Builds] Change interface used by SwiftPM to communicate external dependencies.

artemcm:
Instead of SwiftPM passing in a set of external targets along with their dependency graphs, SwiftPM will now pass in:
- A set of external targets along with the location of their resulting `.swiftmodule` files, as before 
- a collection of *all* modules seen so far across all targets already planned in this build.

The second point replaces previous behaviour, where SwiftPM only passed in the dependency graphs of targets which are dependencies of the target swift-driver is currently planning the build for. Having access to *all* previously-seen modules will allow us to be more intelligent with knowing what modules do not need scanning as well as what clang modules require a re-scan, as each re-scan updates Clang module's details with the set of PCMArgs their dependencies currently capture and propagate it up to the depending modules' build planning actions. 

This PR produces an API change that is matched with a corresponding depending SwiftPM change: https://github.com/apple/swift-package-manager/pull/2959. 

When used together with #279 and the respective SwiftPM changes, this change results in a significant reduction in the amount of versioned batched re-scanning we perform. 

For example: 
When constructing the build manifest for building SwiftPM, this change reduces the total sum of batch sizes of all versioned PCM re-scans involved in the build from _2525_ to _303_ (**8.2x** improvement).

Further followup change will address redundant re-scanning of Swift modules by using the additional information SwiftPM now provides to create more placeholder modules in the dependency scanning action input.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -14,6 +14,10 @@ import TSCUtility
 import Foundation
 import SwiftOptions
 
+/// Temporary backwards-compatible API for SwiftPM
+public typealias ExternalDependencyArtifactMap =
+    [ModuleDependencyId: (AbsolutePath, InterModuleDependencyGraph)]
+
 /// The Swift driver.
 public struct Driver {
   public enum Error: Swift.Error, Equatable, DiagnosticData {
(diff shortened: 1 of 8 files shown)


artemcm:
@swift-ci please test

nkcsgexi (approved):

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
I have added a temporary adapter to preserve the old API so that this change can be submitted without breaking the SwiftPM build. This is necessary because SwiftPM CI does not currently support cross-repository testing with the swift-driver. Once the corresponding SwiftPM change to adapt to the new API has been submitted, I will submit a followup patch to remove the legacy API adapter.

PR #284, opened 2020-10-02
[Explicit Module Builds] Provide API for a common dependency graph operation performed by SwiftPM

artemcm:
It will be much better for SwiftPM to not have to know how to mutate the graph in this way when it has to, and to have this code be common across other clients.


DougGregor (approved):
Yes, thank you! This is a much nicer API for clients to use.

artemcm:
@swift-ci please test

PR #285, opened 2020-10-03
Add API so SPM doesn't need to switch over VirtualPath cases

owenv:
This PR adds new public computed properties `relativePath` and `temporaryFileName` to `VirtualPath`. The goal is to use these in SPM's ManifestBuilder to avoid switching over all of the VirtualPath cases. This should unblock https://github.com/apple/swift-driver/pull/250 which adds a new enum case, and might help avoid some issues once SwiftDriver is being built with library evolution enabled in some contexts but not others.


owenv:
@swift-ci test

DougGregor (approved):
Good idea. This will make it easier to evolve the codebases

PR #286, opened 2020-10-04
Pass -index-file-path as a primary file input

owenv:
`-index-file` uses single compiles, but the `-index-file-path` needs to be passed as a `-primary-file` input. It looks like all the other indexing flags were already being handled correctly.

This fixes the remaining index store tests:
Index/Store/driver-index-with-auxiliary-outputs.swift
Index/Store/driver-index.swift
Index/Store/driver-multiple-inputs.swift
Index/Store/unit-one-file-multi-file-invocation.swift

### DIFF
--- a/Sources/SwiftDriver/Jobs/CompileJob.swift
+++ b/Sources/SwiftDriver/Jobs/CompileJob.swift
@@ -79,6 +79,7 @@ extension Driver {
   /// Add the compiler inputs for a frontend compilation job, and return the
   /// corresponding primary set of outputs.
   mutating func addCompileInputs(primaryInputs: [TypedVirtualPath],
+                                 indexFilePath: TypedVirtualPath?,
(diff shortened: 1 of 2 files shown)


owenv:
@swift-ci test

owenv:
@swift-ci test

cltnschlosser:
I remember seeing some TODOs about handing index file paths when I was working on the file list stuff.

owenv:
@cltnschlosser I just saw the FIXME in addPlatformSpecificLinkerArgs for relation to -index-file-path, but I'm not sure we need special handling there since, as far as I know, we'll never be doing any linking when indexing. It also looks like the cpp driver has some special case handling of file lists, but so far it doesn't seem to be necessary here, I'm looking into that more to see if there are any behavior differences.

DougGregor (approved):

owenv:
@swift-ci test

PR #287, opened 2020-10-05
[Explicit Module Builds] Refactor the InterModuleDependencyGraph to distinguish prebuilt external dependencies.

artemcm:
This is a refactoring change intended to reduce ambiguity in what nodes in the dependency graph can represent. As dependency graph structure and individual modules are being communicated and merged among different components, such ambiguities in the graph introduce additional, unnecessary complexity to graph operations.

- Introduce a new module kind: `.swiftPrebuiltExternalDependency`.
This node kind is to be used for external target dependencies built by build system clients. Previously we were using `.swift` modules to represent both pre-compiled external targets and Swift module dependencies that require a build job to be created by the driver. This new structure means that every `.swift` module dependency must have a build job constructed for it, and every `.swiftPrebuiltExternal` module dependency does not require to be built.
- Factor out and unify common graph operations, so that placeholder resolution and external API for graph merging use common logic.

### DIFF
--- a/Sources/SwiftDriver/CMakeLists.txt
+++ b/Sources/SwiftDriver/CMakeLists.txt
@@ -8,11 +8,12 @@
 
 add_library(SwiftDriver
   "Explicit Module Builds/ClangModuleBuildJobCache.swift"
-  "Explicit Module Builds/ExplicitModuleBuildHandler.swift"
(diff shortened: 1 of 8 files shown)


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #283, opened 2020-10-01
Cargo-culted, but then Swift-ified and cleaned-up ModuleDependencyGraph

davidungar:
This is a bit of a big PR, sorry. It's the driver's dependency graph, along with the unit tests, which run!

PR #289, opened 2020-10-06
Revert "Cargo-culted, but then Swift-ified and cleaned-up ModuleDependencyGraph"

DougGregor:
Reverts apple/swift-driver#283. It broke the CMake build.


DougGregor:
@swift-ci please test

artemcm (approved):

PR #295, opened 2020-10-09
Implement -driver-print-actions for testing purposes

nkcsgexi:
The C++ driver uses -driver-print-actions in several tests. We should implement the same
action in the Swift driver so we can run the compiler-side tests by using the Swift driver.
Printing the exactly same results is difficult but they should be semantically equivalent.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -781,6 +781,14 @@ extension Driver {
       return
     }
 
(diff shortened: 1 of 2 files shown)


DougGregor (approved):

artemcm (approved):

compnerd:
This seems to be breaking the non-Darwin builds by use of an invalid type: `uint` is not a valid type, but it is a typedef on Darwin which is made available.

PR #298, opened 2020-10-10
[Incremental] Write build record & test

davidungar:
Adds `BuildRecordInfo` to hold information needed to read and write build records. Writes build records whether incremental or not. Reads when incremental. Tests reading and writing.


davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
```diff
@@ -0,0 +1,166 @@
...
+      argsHash: argsHash,
+      timeBeforeFirstJob: timeBeforeFirstJob)
+
+    let contents: String
+    do {  contents = try buildRecord.encode() }
+    catch let InputInfoMap.Errors.notAbsolutePath(p) {
+      diagnosticEngine.emit(
+        .warning_could_not_write_build_record_not_absolutePath(p))
+      return
+    }
```
owenv, line 119: Are there any build systems that rely on using a relative build record path in the output file path right now? (Bazel comes to mind as a possibility, but I'm not sure) Instead of failing here, we could try to resolve the path against the file system's cwd

Sources/SwiftDriver/Incremental Compilation/InputIInfoMap.swift
```diff
@@ -14,6 +14,7 @@ import Foundation
 @_implementationOnly import Yams
 
 /// Holds the info about inputs needed to plan incremenal compilation
+/// A.k.a. BuildRecord, but InputInfoMap is the legacy name
 public struct InputInfoMap {
```
owenv, line 18: Should we just rename this type to BuildRecord now? IMO it's a much better name, but I'm not as familiar with how this has changed over time.

owenv (commented):
Looks good! I added a handful of comments, nothing major.

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
```diff
@@ -0,0 +1,166 @@
...
+    // '~moduleonly', so that module-only mode doesn't mess up build-record
+    // file for full compilation.
+    return compilerOutputType == .swiftModule
+      ? partialBuildRecordPath.appendingToBaseName("~moduleonly")
+      : partialBuildRecordPath
+  }
+
+  /// Write out the build record.
+  /// \c Jobs must include all of the compilation jobs.
+  /// \c Inputs will hold all the primary inputs that were not compiled because of incremental compilation
```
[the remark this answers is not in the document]
davidungar: Thanks! Fixed.

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
```diff
@@ -0,0 +1,166 @@
...
+    else {
+      diagnosticEngine.emit(.warning_incremental_requires_build_record_entry)
+      return nil
+    }
+    // In 'emit-module' only mode, use build-record filename suffixed with
+    // '~moduleonly', so that module-only mode doesn't mess up build-record
+    // file for full compilation.
+    return compilerOutputType == .swiftModule
+      ? partialBuildRecordPath.appendingToBaseName("~moduleonly")
+      : partialBuildRecordPath
```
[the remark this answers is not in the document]
davidungar: Thanks again, nice catch!

Sources/SwiftDriver/Incremental Compilation/InputIInfoMap.swift
```diff
@@ -168,9 +144,99 @@ public extension InputInfoMap {
...
+  private static func encode(_ inputInfo: InputInfo) -> Yams.Node {
+    encode(inputInfo.previousModTime, tag: inputInfo.tag)
+  }
+
+}
+
+extension Diagnostic.Message {
+  static func warning_could_not_serialize_build_record(_ err: Error
+  ) -> Diagnostic.Message {
+    .warning("Next compile won't be incremental; Could not serialize build record: \(err.localizedDescription)")
```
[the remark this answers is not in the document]
davidungar: Thanks. I have fixed mine. (There are others lurking.)

Sources/SwiftDriver/Incremental Compilation/InputInfo.swift
```diff
@@ -23,42 +24,84 @@ public struct InputInfo: Equatable {
...
+extension InputInfo.Status {
+  public init?(identifier: String) {
+    switch identifier {
+    case "": self = .upToDate
+    case "!dirty": self = .needsCascadingBuild
+    case "!private": self = .needsNonCascadingBuild
+    default: return nil
+    }
+    assert(self.identifier == identifier, "Ensure reversibility")
+  }
```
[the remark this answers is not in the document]
davidungar: I'm not sure how to do that, given that there are two statuses with the same string: 
```
 case .needsCascadingBuild, .newlyAdded:
      return "!dirty"
```

Sources/SwiftDriver/Incremental Compilation/InputIInfoMap.swift
davidungar: I had thought of that, and you just pushed me over the top! Done!! (Thanks.)

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
davidungar: If I read the following code correctly (in Driver.swift):
```
 if let workingDirectory = self.workingDirectory {
      self.outputFileMap = outputFileMap?.resolveRelativePaths(relativeTo: workingDirectory)
    } else {
      self.outputFileMap = outputFileMap
    }
```
the warning will only happen if there is no working directory and the map contains a relative path.

davidungar:
@swift-ci please test

davidungar:
@swift-ci Thanks, @owenv , for your help. I *think* I've fixed the problems. Would you like to take another look?

davidungar:
@swift-ci please test

Sources/SwiftDriver/Incremental Compilation/InputInfo.swift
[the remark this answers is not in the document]
owenv: Oh yeah, I missed that, thanks! This looks good to me 👍

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
owenv: `Driver.workingDirectory` will only be non-nil if the user explicitly passes one with `-working-directory`, so we can still end up in this situation if the file system has one but the user didn't pass the command line option. That said, this is a pretty rare edge case and I don't think we need to handle it perfectly right now, so I'm fine with leaving this as-is

Sources/SwiftDriver/Incremental Compilation/BuildRecord.swift
```diff
@@ -0,0 +1,245 @@
+//===--------------- BuildRecord.swift - Swift Input File Info Map -------===//
```
owenv, line 1: The CMakeLists.txt entry for InputInfoMap will need to be renamed to BuildRecord

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
```diff
@@ -0,0 +1,148 @@
+//===--------------- BuildRecordInfo.swift --------------------------------===//
```
owenv, line 1: This will need a new entry in CMakeLists.txt

owenv (approved):
@davidungar Looks good to me, thanks! 

Sorry if I sound like a broken record about the CMake stuff, I've broken the build more times than I'd care to admit by forgetting to update it 🙂

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
davidungar: Thank you--yes, I remembered that too late, and I really appreciate the reminder. I will fix those.

davidungar:
@swift-ci please test

PR #301, opened 2020-10-10
Return the compiler resource directory as a VirtualPath and fixup clients

owenv:
In addition to the added tests, this also fixes Driver/linker-rpath.swift. It was kind of tricky to keep the scope of this change small, so this PR includes a few related changes.

- computeResourceDirPath now returns a VirtualPath and handles the case where the user overrides it to be relative. As a result, all the places we were computing paths based on the resource directory required minor, hopefully uncontroversial updates. We're still not using the resource dir from the frontend target info yet because it looks like there might be a few minor issues that need to be resolved first.
- Before this change, `-working-directory` was applied to arguments after frontend target info was fetched, now it is applied beforehand so that the combination of `-working-directory` and a relative `-resource-dir` works correctly
- `addCommonFrontendOptions` was adding `-resource-dir` twice if it was overridden on the command line, now it only appears once.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -323,13 +323,6 @@ public struct Driver {
     // Determine the compilation mode.
     self.compilerMode = try Self.computeCompilerMode(&parsedOptions, driverKind: driverKind, diagnosticsEngine: diagnosticEngine)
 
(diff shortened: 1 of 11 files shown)


owenv:
@swift-ci please test

owenv:
@swift-ci please test

DougGregor (approved):
All of this looks totally reasonably, thank you!

PR #303, opened 2020-10-12
Consider all inputs when deciding whether supplementary output file map will be used

nkcsgexi:
rdar://70221742

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -379,8 +379,9 @@ extension Driver {
                                       output: TypedVirtualPath(file: tracePath, type: .moduleTrace)))
     }
 
+    let allInputsCount = primaryInputs.count + flaggedInputOutputPairs.count
     // Question: outputs.count > fileListThreshold makes sense, but c++ does the following:
-    if primaryInputs.count * FileType.allCases.count > fileListThreshold {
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -379,8 +379,9 @@ extension Driver {
                                       output: TypedVirtualPath(file: tracePath, type: .moduleTrace)))
     }
 
+    let allInputs = primaryInputs.count + flaggedInputOutputPairs.count
     // Question: outputs.count > fileListThreshold makes sense, but c++ does the following:
-    if primaryInputs.count * FileType.allCases.count > fileListThreshold {
+    if allInputs * FileType.allCases.count > fileListThreshold {
```
artemcm, line 384: This is a complete nit: `allInputs` next to `primaryInputs` makes it seem like a collection. It would be nice if the name captured that it is a count as well.

artemcm (approved):

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
nkcsgexi: Makes sense. Fixed.

nkcsgexi:
@swift-ci please test

PR #306, opened 2020-10-14
[Build Script Helper] Avoid outputting install_name_tool failures

artemcm:
In incremental builds of the driver, these commands will fail, which is okay.

Failing to remove specified Rpaths means they are already missing, and failure to add specified Rpaths means they are already present. So we should ignore these failures and proceed with the installation, which we already do. Always outputting these errors pollutes build logs and leads people to mistakenly blame these errors for unrelated failures.


DougGregor (approved):

artemcm:
@swift-ci please test

PR #308, opened 2020-10-14
Honor working directory setting for implicit output paths

nkcsgexi:
rdar://70311690


nkcsgexi:
@swift-ci please test

artemcm (approved):
From a quick scan, I think the `moduleWrapJob` might need to be tweaked to output relative to the working directory, also?

The other jobs not covered here seem to rely on `moduleOutputInfo` which already gets resolved relative to the working dir.

owenv (approved):
Looks good to me! I'm not 100% sure, but I think the module wrapping job should be ok as-is because it always places its output next to its input, and the output is always an intermediate output.

nkcsgexi:
Intermediate outputs should be fine without this change because they always end up in a temp directory IIUC. I'll merge this as-is and keep an eye on future `moduleWrapJob` failures because of it and fix that case by then if necessary.

PR #311, opened 2020-10-16
[Incremental] Testable debugging outputs & tests

davidungar:
1. Centralize the `parsedOptions` interrogation for `-driver-show-job-lifecycle`
2. Use diagnostic remarks for `-driver-show-job-lifecycle` and `-driver-show-incremental`
    Why? Much easier to test, may not mess up Xcode's reading of the compiler's output.
    Why not? We'll have to fix some integration tests, as it differs from legacy front end.
3. Add tests for these outputs
4. Call back into `BuildRecordInfo` and `IncrementalCompilationState` when jobs finish.

Robert floated, and I like the idea of mimicking the legacy driver's messages. However, since we don't have the same sort of task queue (at the present), I'm not sure it's feasible.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -135,6 +135,9 @@ public struct Driver {
   /// The set of parsed options.
   var parsedOptions: ParsedOptions
 
+  /// Whether to print out extra info regarding jobs
+  let showJobLifecycle: Bool
+
   /// Extra command-line arguments to pass to the Swift compiler.
   let swiftCompilerPrefixArgs: [String]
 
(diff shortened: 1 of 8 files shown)


davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

owenv (commented):
This looks good to me once the Linux tests are fixed - It looks like the it's failing because Linux adds an autolink-extract job to one of the test cases.

I did a quick search and found 15 integration tests that could depend on the old-style debugging output (because they use -driver-show-job-lifecycle or -driver-show-incremental). Updating those seems manageable but might be a little more work than trying to accurately replicate the C++ output

davidungar:
> This looks good to me once the Linux tests are fixed - It looks like the it's failing because Linux adds an autolink-extract job to one of the test cases.
> 
> I did a quick search and found 15 integration tests that could depend on the old-style debugging output (because they use -driver-show-job-lifecycle or -driver-show-incremental). Updating those at some point seems manageable but might be a little more work than trying to accurately replicate the C++ output

Thanks, Owen, for looking at this. I'll take a whack at the Linux tests. Regarding the debugging output, I'm not sure whether we want a separate task queue, like the one in the legacy driver. Seems like it might be redundant with using `llbuild`. If we don't, then it might be undesirable to pretend there is one in the debugging output. I'm still wrestling with exactly how to get llbuild to work for us. The old driver batches up jobs on-demand as cores free up, depending on which jobs in the 2nd wave have already been scheduled.

davidungar:
@swift-ci please test

owenv (approved):

davidungar:
Thanks, @owenv !

PR #313, opened 2020-10-17
Warn when a target environment is inferred to be a simulator

owenv:
The frontend takes care of the actual inference, so here I just check if we're in a situation where the explicit triple does not have the simulator environment but the frontend target info one does.


owenv:
@swift-ci test

DougGregor (approved):
Thanks!

PR #314, opened 2020-10-17
Add the module interface input to the command line of a verification job

owenv:
This fixes the ModuleInterface/verify-module-interfaces.swift test. Previously, the module interface was correctly being added as a job input, but it wasn't added to the command line.


owenv:
@swift-ci please test

DougGregor (approved):

PR #316, opened 2020-10-17
-dump-ast overrides -wmo and -index-file

owenv:
This fixes the Driver/ast_dump_with_WMO.swift test, and expands the override to cover -index-file as well based on a FIXME in the C++ source. If both are specified, a warning is emitted: `ignoring '-wmo' because '-dump-ast' was also specified`


DougGregor (approved):

owenv:
@swift-ci please test

PR #317, opened 2020-10-18
Properly set up jobs when -lto is set

nkcsgexi:
rdar://70414259

### DIFF
--- a/Sources/SwiftDriver/Jobs/CompileJob.swift
+++ b/Sources/SwiftDriver/Jobs/CompileJob.swift
@@ -63,10 +63,17 @@ extension Driver {
       return true
     case .object:
       return (linkerOutputType == nil)
+    case .llvmBitcode:
+      if compilerOutputType != .llvmBitcode {
+        // The compiler output isn't bitcode, so bitcode isn't top-level (-embed-bitcode).
+        return false
+      } else {
+        // When -lto is set, .bc will be used for linking. Otherwise, .bc is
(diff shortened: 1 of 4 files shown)


nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/CompileJob.swift
```diff
@@ -65,12 +65,10 @@ extension Driver {
       return (linkerOutputType == nil)
     case .swiftModule:
       return compilerMode.isSingleCompilation && moduleOutputInfo.output?.isTopLevel ?? false
-    case .llvmBitcode:
-      return compilerOutputType == type
```
owenv, line 69: It looks like this change breaks the reproducible-builds/swiftc-emit-bc.swift test, because bitcode can be a top level output if the command line is something like `swiftc -emit-bc foo.swift -o foo.bc`.

nkcsgexi: I've fixed in the new iteration. Mind taking another look?

nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/CompileJob.swift
owenv: Looks good to me, thanks!

owenv (approved):

PR #319, opened 2020-10-19
BatchMode: log batch configurations when -driver-show-job-lifecycle is set. NFC

nkcsgexi:
This allows us to compare the output with that of the C++ driver


nkcsgexi:
@swift-ci please test

PR #324, opened 2020-10-21
[DNM, Incremental] Driver does incremental builds

davidungar:
But not as nicely as ultimately desired. Two fixed waves. Also, needs carving up into reviewable pieces.


davidungar:
@swift-ci please test

PR #325, opened 2020-10-22
Teach swift-driver to emit private interfaces

nkcsgexi:
rdar://70562001


nkcsgexi:
@swift-ci please test

owenv (approved):
👍

PR #326, opened 2020-10-22
Derive supplementary output paths using the full basename of .swiftmodule

nkcsgexi:
rdar://70584224

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2179,7 +2179,7 @@ extension Driver {
         parentPath = moduleOutputPath.parentDirectory
       }
 
-      return parentPath.appending(component: moduleName).replacingExtension(with: type)
+      return parentPath.appending(component: moduleOutputPath.basename).replacingExtension(with: type)
     }
 
     // If the output option was not provided, don't produce this output at all.
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -2529,6 +2529,21 @@ final class SwiftDriverTests: XCTestCase {
     }
   }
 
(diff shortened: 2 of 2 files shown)


nkcsgexi:
@swift-ci please test

artemcm (approved):

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2179,7 +2179,7 @@ extension Driver {
         parentPath = moduleOutputPath.parentDirectory
       }
 
-      return parentPath.appending(component: moduleName).replacingExtension(with: type)
+      return parentPath.appending(component: moduleOutputPath.basename).replacingExtension(with: type)
```
cltnschlosser, line 2182: This function doesn't need to take `moduleName` as a parameter anymore.

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -2529,6 +2529,21 @@ final class SwiftDriverTests: XCTestCase {
     }
   }
 
+  func testDeriveSwiftDocPath() throws {
+    var driver = try Driver(args: [
+      "swiftc", "-emit-module", "/tmp/main.swift", "-emit-module-path", "test-ios-macabi.swiftmodule"
```
cltnschlosser, line 2534: What is module name in this case? main? Just wondering how off it was.

nkcsgexi: I think `main` will be the default name if not given at all.

Sources/SwiftDriver/Driver/Driver.swift
nkcsgexi: It still needs `moduleName` in the last return statement

cltnschlosser: Oh oops, just looked up, thought this was the last one.

PR #329, opened 2020-10-22
[Incremental] Do incremental builds

davidungar:
EDIT: It has been redone. It returns all jobs from the planning method to preserve compatibility, but the `run` function uses a side door for incremental execution.

Here is a large PR (sorry!) that performs incremental builds (per our discussion the other day). It isn't optimal: it uses rebatching, and it uses llbuild in waves. It does not assume only two waves, though. Probably a bit rough in spots--all feedback welcome!
If you want to see it work, try `testIncremental`.


davidungar:
@swift-ci please test

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
```diff
@@ -159,110 +184,114 @@ extension Diagnostic.Message {
...
-    let scheduledInputs = Array(inputsRequiringCompilation.union(additions))
+    let firstWaveInputs = Array(definitelyRequiredFirstWaveInputs.union(speculativeFirstWaveInputs))
       .sorted {$0.file.name < $1.file.name}
 
     let skippedInputs = Set(buildRecordInfo.compilationInputModificationDates.keys)
-      .subtracting(scheduledInputs)
+      .subtracting(firstWaveInputs)
     for skippedInput in skippedInputs.sorted(by: {$0.file.name < $1.file.name})  {
-      reportIncrementalQueuing("Skipping:", skippedInput.file)
+      reportIncrementalDecision?("Skipping:", skippedInput)
```
cltnschlosser, line 245: Can do if let check for reportIncrementalDecision before the loop so the sort and iteration is skipped when it's nil.

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
```diff
@@ -321,18 +352,103 @@ extension IncrementalCompilationState {
...
+        }
+      }
+    }
+  }
+
+  /// Remember a job that runs after all compile jobs
+  func addPostCompileJobs(_ jobs: [Job]) {
+    for job in jobs {
+      for input in job.primaryInputs {
+        reportIncrementalDecision?("Delaying pending discovering delayed dependencies", input)
```
cltnschlosser, line 382: if let reportIncrementalDecision to skip loop if nil

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
```diff
@@ -321,18 +352,103 @@ extension IncrementalCompilationState {
...
-    // additional jobs
+    defer {
+      amHandlingJobCompletion = false
+    }
+    assert(!amHandlingJobCompletion, "was reentered, need to synchronize")
+    amHandlingJobCompletion = true
+
+    let secondWaveInputs = collectInputsToCompileIn2ndWave(job)
+    for input in secondWaveInputs {
+      reportIncrementalDecision?(
```
cltnschlosser, line 405: Can skip loop if nil (I wonder if the optimizer catches this?)

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
```diff
@@ -321,18 +352,103 @@ extension IncrementalCompilationState {
...
+      }
+      else {
+        reportIncrementalDecision?("Tried to schedule 2nd wave input again", input)
+      }
+    }
+  }
+
+  private func rememberJobFor2ndWave(_ j: Job) {
+    for input in j.primaryInputs {
+      reportIncrementalDecision?("Scheduling for 2nd wave", input)
```
cltnschlosser, line 441: Can skip loop if nil

davidungar:
@cltnschlosser I'm very grateful you put in the time to look at this large PR so quickly! Thank you. I'll get to your specific comments in a bit.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -814,9 +814,7 @@ extension Driver {
       return
     }
 
-    if jobs.contains(where: { $0.requiresInPlaceExecution })
-      // Only one job and no cleanup required
-      || (jobs.count == 1 && !parsedOptions.hasArgument(.parseableOutput)) {
+    if jobs.contains(where: { $0.requiresInPlaceExecution }) {
```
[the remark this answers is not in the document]
davidungar: Thanks. Yes, it does complicate testing. We could add a boolean to pass in when creating the driver to disable this, just for testing. Or could fix the lit tests. What's best?

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
```diff
@@ -128,15 +128,36 @@ import SwiftOptions
  }
 
 // TODO: Incremental too many names, buildRecord BuildRecord outofdatemap
-  func populateOutOfDateBuildRecord() -> BuildRecord? {
+  func populateOutOfDateBuildRecord(
+    inputFiles: [TypedVirtualPath],
+    failed: (String) -> Void
```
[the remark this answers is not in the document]
davidungar: I tried to ask folks about this issue, but missed you, sorry! The advantages of using diagnostics are that it is much easier to test, and it won't end up messing up the xcodebuild logs. The disadvantage I had thought of was the incompatibility. I hadn't though of localization. What do you think?

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
davidungar: Thanks! I'll fix it.

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
davidungar: Thanks, fixed!

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
davidungar: Thanks, fixed!

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
davidungar: Thanks fixed

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
davidungar: Yes, would be nice to just let the optimizer deal with it.

davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

davidungar:
> @cltnschlosser I'm very grateful you put in the time to look at this large PR so quickly! Thank you. I'll get to your specific comments in a bit.

OK, I now believe I've addressed the points that I can. Thank again!

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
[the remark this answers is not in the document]
cltnschlosser: Probably safe to just leave it as is for now, seeing as there is no plan for localization currently.

Sources/SwiftDriver/Driver/Driver.swift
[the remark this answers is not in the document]
cltnschlosser: Been a while since I've contributed to swift-driver, but I try to keep up on the changes. So I haven't run the lit tests in a while, but I think it's worth running them with and without this to see what the differences are first. @owenv Might have thoughts too as he has been continuing to update swift-driver to pass more lit tests.

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -170,38 +186,32 @@ extension Driver {
...
-    if parsedOptions.hasArgument(.driverShowJobLifecycle) {
-      for input in inputFiles {
-        if let idx = partitions?.assignment[input] {
-          stdoutStream.write("Adding {compile: \(input.file.basename)} to batch \(idx)\n")
-          stdoutStream.flush()
-        }
-      }
+    let swiftInputFiles = inputFiles.filter { inputFile in
+      inputFile.type.isPartOfSwiftCompilation
     }
```
owenv, line 196: This change seems to break using the driver as a linker, because object files passed on the command line are no longer added to the set of `linkerInputs`. It seems to be causing the `Driver/unknown-inputs.swift` lit test to fail. Other than that, I don't see any regressions in the lit tests

owenv (commented):
Sorry, I haven't had a chance to look over this in detail yet, I'll try to set aside some time this weekend. I added a couple initial comments. It's really exciting to see this working!

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -39,16 +39,30 @@ extension Driver {
...
                                debugInfo: debugInfo,
-                               addJob: addJob)
-    return jobs
+                               addJob: skippedCompileJobs.isEmpty
+                                ? addJob
+                                : addPostCompileJob)
+
+    incrementalCompilationState?.addSkippedCompileJobs(skippedCompileJobs)
+    incrementalCompilationState?.addPostCompileJobs(postCompileJobs)
+    return try formBatchedJobs(jobs)
```
owenv, line 65: I've been thinking about this a bit more, and I think we should try to avoid explicitly partitioning jobs into first wave, potentially-skippable and link/post-link during the planning stage. In the short term, it complicates the SPM implementation, which expects a complete, static build graph from `Driver.planBuild`, and in the long term I think it will make it harder to move more of the scheduling and synchronization out of `Driver.run` and into the llbuild-based executor.

What do you think about adding a `potentiallySkippable` flag to `Job` instead? Then, based on the new flag and `Job.kind`, `Driver.run` can do this partitioning just before execution. SPM (which always passes `-incremental` afaict) should keep working in that case because `planBuild` will continue to return the full set of jobs, and true incremental behavior would be 'opt-in' because clients would need to look at the `potentiallySkippable` flag and schedule jobs appropriately.

davidungar: Very interesting! Let me think about it for a while. At first glance, I'm not sure that that flag is appropriate for inclusion in the job. We could get the same effect by having the planning phase include all the jobs, and the `run` method could consult the `incrementalBuildState`. Does that sound reasonable?

Also, someday, we'll have to enhance the relationship between SPM and the driver, if we ever want SPM to have incremental capability. But that's a task for a later time, IMO.

I also need to go back into my notes to see how to run the `lit` tests, etc., on swift-driver.

I very much appreciate the time and thought you're putting in. Thank you!

owenv: > Very interesting! Let me think about it for a while. At first glance, I'm not sure that that flag is appropriate for inclusion in the job. We could get the same effect by having the planning phase include all the jobs, and the run method could consult the incrementalBuildState. Does that sound reasonable?

That's fair, and your proposed solution sounds good to me overall! I think as long as planBuild is returning all of the jobs in some form rather than only registering some of them with the incremental state, my concerns are mostly addressed.

Also, I haven't tested this change with SPM at all, but it might not be a bad idea to patch it so that it doesn't pass  `-incremental` when using the integrated driver before landing this change.

davidungar:
@swift-ci please test

Sources/SwiftDriver/Jobs/Planning.swift
davidungar: @owenv OK, I've addressed the issues. Would you be interested in taking another look? If you're interested in the current state of `-driver-show-incremental` or `-driver-show-job-lifecycle` check out `testIncremental()`.

owenv: This looks good to me, thanks! I'm probably not the best person to review the batching changes, but I didn't notice any issues except the one I mentioned above where something like `swiftc test.o` no longer produces any jobs.

davidungar: Thank you, @owenv . Nice catches!

davidungar:
@swift-ci please test

Sources/SwiftDriver/Jobs/Planning.swift
davidungar: OK, I think it's fixed now. Tx again.

davidungar:

davidungar:
@swift-ci please test linux

davidungar:

Sources/SwiftDriver/Incremental Compilation/SynchronizedQueue.swift
```diff
@@ -0,0 +1,64 @@
+//===--------------- JobQueue.swift - Incremental -----------===//
```
artemcm, line 1: This filename got stale.

Sources/SwiftDriver/Incremental Compilation/SynchronizedQueue.swift
```diff
@@ -0,0 +1,64 @@
...
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
```
artemcm, line 5: 2020 :)

artemcm (approved):
I'm happy with how this looks! 
Excited to see how it evolves and about making SwiftPM do something similar, in-tandem.

Sources/SwiftDriver/Incremental Compilation/SynchronizedQueue.swift
davidungar: Thank you, @artemcm, for both of these catches and most of all, for taking the time to look over this large PR!

davidungar:
@swift-ci please test

PR #330, opened 2020-10-23
Include prefix args from -driver-use-frontend-path in -print-target-info jobs

owenv:
This, along with https://github.com/apple/swift/pull/34405 will let us see real failures in the incremental dependency lit tests. Without these changes, nearly all the tests fail immediately because they rely on fake frontends and the -print-target-info decoding fails.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -1967,7 +1967,8 @@ extension Driver {
           target: explicitTarget, targetVariant: explicitTargetVariant,
           sdkPath: sdkPath, resourceDirPath: resourceDirPath,
           runtimeCompatibilityVersion:
-            parsedOptions.getLastArgument(.runtimeCompatibilityVersion)?.asSingle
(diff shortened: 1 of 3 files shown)


owenv:
@swift-ci please test

buttaface:
Could you guys start doing trunk snapshot tags to match all the other Swift repos? I'm working on an SPM pull with a trunk source snapshot from a couple weeks ago and I ran into an issue when this repo was checked out to the latest commit, ie a couple weeks ahead of SPM, and SPM wouldn't build. I'm now eyeballing it and checked out a commit from this repo from a couple weeks ago instead, but it would be better if there was an official tag.

owenv:
@buttaface I'd recommend filing a JIRA issue about that under "project infrastructure". As far as I know the infra for the nightly tags isn't publicly accessible so there's not much I can do on that front.

buttaface:
Alright, I figured it was something this repo's owners setup, which is why I asked here, I will file a JIRA instead.

artemcm (approved):
This'll be super useful, thank you.

owenv:
@swift-ci please test

PR #331, opened 2020-10-24
Don't pass a precompiled header to module merging jobs

owenv:
Fixes:
ClangImporter/pch-bridging-header-unittest-ok.swift
ClangImporter/pch-bridging-header-unittest-warn.swift
ClangImporter/pch-bridging-header-with-another-bridging-header.swift

The c++ driver uses the original header when creating module merging jobs, and trying to use a pch causes weird failures in the test suite. I really don't understand why some frontend actions support precompiled headers and others don't, so it's possible there's a better solution to be found.


owenv:
@swift-ci please test

DougGregor (approved):
I'm slightly surprised that we need to do this, but at this point it's better to match the existing C++ driver's behavior. If we fix merge-modules to accept a precompiled bridging header, we'll update this accordingly.

owenv:
Sounds good 👍

PR #332, opened 2020-10-26
Intro incremental build alternative to emit module files separately

xymus:
Incremental compilation currently relies on a merge-module phase. A typical
build plan for a module App composed of two source files (`A.swift` and `B.swift`)
involves four main jobs:

1. Compile `A.swift`, output `A.o` and `A~partial.swiftmodule`.
2. Compile `B.swift`, output `B.o` and `B~partial.swiftmodule`.
3. Link `A.o` and `B.o` in the final executable.
4. Merge-module: merging `A~partial.swiftmodule` and `B~partial.swiftmodule` to form
the final `App.swiftmodule`.

Jobs 3 and 4 must run after jobs 1 and 2 to use their outputs.

The flag `-experimental-emit-module-separately` replaces the merge-module job but
the emit-module job that can run in parallel to the other jobs and doesn’t need
the partial swiftmodules.

1. Compile `A.swift`, output `A.o`.
2. Compile `B.swift`, output `B.o`.
3. Link `A.o` and `B.o` in the final executable. (Only job 3 must run after jobs 1. and 2.)
4. Emit-module: Parse `A.swift` and `B.swift` to generate the final `App.swiftmodule`.

Early comparison with merge-module indicates similar performances and even a
slight gain with all of the optimizations applied here. However, the use of
`-experimental-skip-non-inlinable-function-bodies` may be too aggressive.

rdar://69330639

### DIFF
--- a/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
+++ b/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
@@ -33,7 +33,7 @@ extension Driver {
     addSupplementalOutput(path: objcGeneratedHeaderPath, flag: "-emit-objc-header-path", type: .objcHeader)
     addSupplementalOutput(path: tbdPath, flag: "-emit-tbd-path", type: .tbd)
 
-    if isMergeModule {
+    if isMergeModule || shouldCreateEmitModuleJob {
       return
     }
     // Add outputs that can't be merged
@@ -92,8 +92,7 @@ extension Driver {
 
   /// Returns true if the emit module job should be created.
   var shouldCreateEmitModuleJob: Bool {
-    return forceEmitModuleInSingleInvocation
-      && compilerOutputType != .swiftModule
-      && moduleOutputInfo.output != nil
+    return forceEmitModuleBeforeCompile
+      || parsedOptions.hasArgument(.emitModuleSeparately)
   }
 }
--- a/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
+++ b/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
@@ -67,6 +67,10 @@ extension Driver {
       inputs.append(input)
     }
 
+    if let pchPath = bridgingPrecompiledHeader {
+      inputs.append(TypedVirtualPath(file: pchPath, type: .pch))
+    }
+
     try addCommonFrontendOptions(commandLine: &commandLine, inputs: &inputs)
     // FIXME: Add MSVC runtime library flags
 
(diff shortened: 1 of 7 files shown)


xymus:
@swift-ci please test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -307,20 +307,22 @@ extension Driver {
           finalOutputPath: moduleSourceInfoPath,
           input: input,
           flag: "-emit-module-source-info-path")
```
xymus, line 309: @nkcsgexi Do we need `-emit-module-doc-path` and `-emit-module-source-info-path` for per-file compile jobs in incremental compilation? I don't get the impression they are getting merged but are they?

nkcsgexi: They should be generated per-file and later merged because doc-comments and source info in each file is only available when that file is part of the input.

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -242,7 +242,7 @@ extension Driver {
           let job = try backendJob(input: input, addJobOutputs: addJobOutputs)
           addJob(job)
         }
-      } else {
+      } else if !shouldCreateEmitModuleJob || compilerOutputType != .swiftModule {
```
artemcm, line 245: Took some help in figuring out the context for this to understand why we are skipping jobs here. :)
Please add a comment to clarify why we want to omit all per-file compile jobs.

artemcm (approved):
This looks good! A couple of small comments:

Sources/SwiftDriver/Jobs/EmitModuleJob.swift
```diff
@@ -56,7 +56,7 @@ extension Driver {
       TypedVirtualPath(file: moduleOutputPath, type: .swiftModule)
     ]
 
-    commandLine.appendFlags("-frontend", "-emit-module")
+    commandLine.appendFlags("-frontend", "-emit-module", "-experimental-skip-non-inlinable-function-bodies")
```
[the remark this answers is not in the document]
xymus: Yes but this is probably over aggressive as it may drop types used by LLDB. It allows emit-module to take a similar amount of time to the compile jobs in my test cases and thus not slow down the build process. I have to investigate the effect that it has on LLDB, and may have to add an alternative to `-experimental-skip-non-inlinable-function-bodies` for this specific use case before we can use emit-module as a default over merge-module.

Sources/SwiftDriver/Jobs/Planning.swift
xymus: I shuffled the condition and added a comment to make things more explicit.

xymus:
@swift-ci please test

xymus:
Rebased again because there was a conflict with the PR that recently got merged and reverted.

@swift-ci please test

xymus:
@swift-ci please test

xymus:
@swift-ci please test

PR #335, opened 2020-10-28
Common out utility to check if LLDB is present, to skip tests that rely on LLDB

artemcm:
And use it to also skip a part of `testNoInput`.

The driver very well may be built and tested in environments that do not have LLDB.
We should skip tests that rely on LLDB being present in such environments.

Although we've had `testNoInput` for some time, we have not been updating the set of tests that run on Linux (adding tests does not automatically make them run on linux, that requires the additional step of running: `swift test --generate-linuxmain`, which we have been forgetting to do for a while (And I wasn't aware of 😬)). Now, some recent PR updated this list so we've noticed it failing on some Linux build bots where there is no LLDB. 
e.g. https://ci.swift.org/job/oss-swift-incremental-RA-linux-ubuntu-16_04/14947/

### DIFF
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -18,6 +18,27 @@ import XCTest
 
 final class SwiftDriverTests: XCTestCase {
 
+  /// Determine if the test's execution environment has LLDB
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci please test

nkcsgexi (approved):
LGTM!

PR #338, opened 2020-10-31
Static linking fixes on Darwin

owenv:
A few related fixes for static linking on macOS:
- Don't pass options like -platform_version, -force_load, etc. to the static linker
- Don't use .swiftmodule files as static linker inputs
- Don't pass -lto_library to the static linker, and only pass it to the dynamic linker if we're actually performing LTO


owenv:
@swift-ci please test

DougGregor (approved):

PR #339, opened 2020-10-31
Fix a regression affecting command line linker inputs

owenv:
Object files and other linker inputs specified on the command line should be included in the linker job. Additionally, if a loaded module trace is emitted, it should be associated with the first input to a compile job, which isn't necessarily the first input.

I added a couple of regression tests, and this also fixes a large number of lit tests (it took me awhile to notice the regression because it didn't affect anything in the Driver folder, I'm not sure exactly when it happened)


owenv:
@swift-ci please test

DougGregor (approved):
Looks good. Thanks especially for the extra test coverage here

PR #341, opened 2020-11-01
Add section to README about testing with toolchains

cltnschlosser:


README.md
```diff
@@ -102,6 +102,13 @@ apple/swift-driver#208
 ```
 @swift-ci cross-repository testing facilities are described [here](https://github.com/apple/swift/blob/main/docs/ContinuousIntegration.md#cross-repository-testing).
 
+### Testing in XCode with custom toolchain
+After the toolchain is installed, xcode needs to be told to use it. This can mean two things, building the driver with the toolchain and telling the driver to use the toolchain when running.
+
+Building with the toolchain is easy, set the toolchain in xcode: Menu Bar > Xcode > Toolchains > select your toolchain
+
+Running the driver requires setting the TOOLCHAINS evironment variable. This tells xcrun which toolchain to use, on darwin xcrun is used to find tools. This variable is the name of the toolchain and not the path (ex: `Swift Development Snapshot`)
```
owenv, line 110: I think it might be worth mentioning here that TOOLCHAINS will only have an affect if:
1. A tool's path hasn't been overridden using one of the SWIFT_DRIVER_*_EXEC environment variables, and
2. The driver executable is not installed inside a toolchain

The second point is probably less common though so maybe isn't as important.

cltnschlosser: Yeah this is more for running the tests in Xcode, so the second point doesn’t really apply. But yeah I can add some more here. The frontend is probably the only tool that actually gets executed, but TOOLCHAINS is what I was using. It’s worth noting that the SWIFT_DRIVER_ environment variables override TOOLCHAINS.

README.md
```diff
@@ -102,6 +102,13 @@ apple/swift-driver#208
 ```
 @swift-ci cross-repository testing facilities are described [here](https://github.com/apple/swift/blob/main/docs/ContinuousIntegration.md#cross-repository-testing).
 
+### Testing in Xcode with custom toolchain
+After the toolchain is installed, Xcode needs to be told to use it. This can mean two things, building the driver with the toolchain and telling the driver to use the toolchain when running.
+
+Building with the toolchain is easy, set the toolchain in Xcode: Menu Bar > Xcode > Toolchains > select your toolchain
+
+Running the driver requires setting the TOOLCHAINS environment variable. This tells xcrun which toolchain to use, on darwin xcrun is used to find tools. This variable is the name of the toolchain and not the path (ex: `Swift Development Snapshot`). Important note: xcrun is lower priority than SWIFT_EXEC_*_EXEC family of environment variables, tools directory, and for anything in the same directory as the driver (This includes a driver installed in a toolchain). Even though TOOLCHAINS is not highest priority it's a convenient one change for running the xctest suite against a custom toolchain.
```
owenv, line 110:
```suggestion
Running the driver requires setting the TOOLCHAINS environment variable. This tells xcrun which toolchain to use (on darwin xcrun is used to find tools). This variable is the name of the toolchain and not the path (ex: `Swift Development Snapshot`). Important note: xcrun lookup is lower priority than the SWIFT_EXEC_*_EXEC family of environment variables, the tools directory, and any tools in the same directory as the driver (This includes a driver installed in a toolchain). Even though TOOLCHAINS is not highest priority it's a convenient way to run the xctest suite using a custom toolchain.
```

owenv (approved):
LGTM, thanks!

cltnschlosser:
@swift-ci please test

cltnschlosser:
Looks like my bot permissions haven't gone through yet. Someone else want to kick this off for me. Thanks!

DougGregor:
@swift-ci please test

PR #342, opened 2020-11-01
Fix using the driver as a linker

cltnschlosser:
This fixes 16 Interpreter tests and 4 stdlib tests which relied on this functionality.


cltnschlosser:
Looks like this may have been fixed already with #339

owenv:
I think https://github.com/apple/swift-driver/pull/339 which I merged this morning should have addressed this in a slightly different way.

cltnschlosser:
Yep, this is fixed, closing.

PR #343, opened 2020-11-02
Fix for simulator and linux -profile-generate linker args

cltnschlosser:
Need the `#if os(Linux)` tests to run in CI to verify they are correct because I wasn't able to run them. The swift docker images only seem to be for releases, and we require print-target-info changes from main.

### DIFF
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -116,12 +116,11 @@ extension DarwinToolchain {
     let clangPath = try clangLibraryPath(for: targetTriple,
                                          parsedOptions: &parsedOptions)
 
-    let runtime = targetTriple.darwinPlatform!.libraryNameSuffix
-
-    let clangRTPath = clangPath
-      .appending(component: "libclang_rt.profile_\(runtime).a")
-
-    commandLine.appendPath(clangRTPath)
+    for runtime in targetTriple.darwinPlatform!.profileLibraryNameSuffixes {
+      let clangRTPath = clangPath
+        .appending(component: "libclang_rt.profile_\(runtime).a")
+      commandLine.appendPath(clangRTPath)
+    }
   }
 
   private func addPlatformVersionArg(to commandLine: inout [Job.ArgTemplate],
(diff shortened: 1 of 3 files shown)


DougGregor (approved):
LGTM, thanks!

DougGregor:
@swift-ci please test

cltnschlosser:
```
Fatal error: /home/buildnode/jenkins/workspace/pr-swift-driver-linux/latest_toolchain/swift-DEVELOPMENT-SNAPSHOT-2020-10-29-a-ubuntu16.04/usr/lib/swift_static/wasi/static-executable-args.lnk not found: file SwiftDriver/WebAssemblyToolchain+LinkerSupport.swift, line 121
```
The wasi test failed. I'll see if I can debug this.

cltnschlosser:
```
      guard !parsedOptions.hasArgument(.profileGenerate) else {
        throw Error.profilingUnsupportedForTarget(targetTriple.triple)
      }
```
@owenv @MaxDesiatov Is profiling supported for wasi? This commit https://github.com/apple/swift/commit/23b1a6b87afee535296a786dcdd2a6d0e65d67ef added c++ driver tests for it, so not sure which is correct.

cltnschlosser:
This is actually the bit that failed though. Looks like `swift_static/wasi` might not exist, is this expected: `/home/buildnode/jenkins/workspace/pr-swift-driver-linux/latest_toolchain/swift-DEVELOPMENT-SNAPSHOT-2020-10-29-a-ubuntu16.04/usr/lib/swift_static/wasi/static-executable-args.lnk not found`
```
let linkFilePath: VirtualPath = resourceDirPath
        .appending(components: "static-executable-args.lnk")

      guard try fileSystem.exists(linkFilePath) else {
        fatalError("\(linkFilePath) not found")
      }
```

I'm going to remove this test from this PR, can add it back later.

owenv:
The failure to find the link file is expected, because most swift toolchains don't support cross-compilation to wasi. In the linking tests, I created a fake resource dir with the expected layout and used that instead of the one in the toolchain. 

As far as I know though, WebAssembly can't support the profiling runtime library currently, so I don't think there's any need to test the behavior beyond what's in testWebassemblyUnsupportedFeatures. I think the lit tests were directly copied from the Linux ones because the c++ driver uses a GenericUnixToolchain for wasi targets, which afaik isn't fully functional. I don't have all the context behind the change you linked though.

owenv:
@swift-ci please test

PR #344, opened 2020-11-02
Only link compiler_rt when dynamic linking

cltnschlosser:
Fixes the darwin part of Driver/static-archive.swift. This test didn't check for this specifically, but it was testing for `libtool -static` and libclang_rt was in between that. Looked further and found c++ driver only includes libclang_rt when linking dynamically.


artemcm (approved):
Good catch, thank you.

artemcm:
@swift-ci please test

PR #345, opened 2020-11-02
Use the standard SynchronizedQueue instead of my custom one.

davidungar:
Replace my custom-coded SynchronizedQueue with the one in the standard toolkit.


Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
```diff
@@ -409,12 +409,14 @@ extension IncrementalCompilationState {
...
-    group.forEach {schedule(preOrCompileJob: $0)}
+    schedule(preOrCompileJobs: group)
   }
   /// Put job in queue for execution
-  func schedule(preOrCompileJob job: Job) {
-    reportIncrementalDecision?("Queuing \(job.descriptionForLifecycle)", nil)
-    preOrCompileJobs.append(job)
+  func schedule(preOrCompileJobs jobs: [Job]) {
+    for job in jobs {
+      reportIncrementalDecision?("Queuing \(job.descriptionForLifecycle)", nil)
```
cltnschlosser, line 417: I’m back, with the same comment about putting if let around the loop to avoid iteration when this is nil :)

cltnschlosser (commented):
I’m guessing you’ll need to remove the deleted file from the cmake list. Also which lib provides this? Never seen it before and did a google search and it didn’t come up.

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
davidungar: Thank you for catching those! I *really* appreciate your help. Should be fixed now.

davidungar:
@swift-ci please test

artemcm (approved):

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
davidungar: > I’m guessing you’ll need to remove the deleted file from the cmake list. Also which lib provides this? Never seen it before and did a google search and it didn’t come up.

See `SynchronizedQueue.swift` in `swift-tools-support-core/Sources/TSCBasic`

davidungar:

davidungar:
Thanks, @artemcm . I'll be investigating the Linux failure.

davidungar:
@swift-ci please test

PR #349, opened 2020-11-03
Add a version number indicating swift-driver's source version

nkcsgexi:
When swift-driver is releases separately from the compiler, an independent version number could help.

### DIFF
--- a/Sources/SwiftDriver/Driver/DriverVersion.swift
+++ b/Sources/SwiftDriver/Driver/DriverVersion.swift
@@ -1,0 +1,18 @@
+//===------ DriverVersion.swift - Swift Driver Source Version--------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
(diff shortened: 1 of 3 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Driver/DriverVersion.swift
```diff
@@ -0,0 +1,18 @@
...
+// Copyright (c) 2014 - 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+extension Driver {
+#if SWIFT_DRIVER_VERSION_DEFINED
+  static let driverSourceVersion: String = SWIFT_DRIVER_VERSION
```
artemcm, line 14: If the driver is to expose its own version as a public API and a command-line query, what do you think about adopting something similar to what SwiftPM does and using:
`SwiftVersion` from: `TSCUtility/Versioning.swift`. This will give us a matching API and a matching semver scheme. 
It has some other niceties such as being able to append a specific build number.

nkcsgexi: The intention here is that the version number will be passed down automatically from the build environment.

artemcm: Ah, that makes sense then.

artemcm (approved):

PR #354, opened 2020-11-09
Remove `computeResourceDirPath` and rely on the one reported by `-print-target-info`

owenv:
In theory this is NFC, but I'll run full CI just in case (Edit: looks like it passed). In addition to replacing all uses of the old method with reference to the target info's `runtimeResourcePath`, I had to update the -print-target-info job in toolchain computation to ensure it returned the static resource directory when needed. I tried replacing all uses of runtime library paths with the -print-target-info version too but ran into some issues where the macosx paths weren't included for catalyst targets, I'll need to look into that more.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -126,6 +126,9 @@ public struct Driver {
     frontendTargetInfo.targetVariant?.triple
   }
 
+  /// `true` if the driver should use the static resource directory.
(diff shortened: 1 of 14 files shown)


owenv:
@swift-ci please test

cltnschlosser:
Does this fix something or is it just clean up?

owenv:
@cltnschlosser it's just cleanup, this shouldn't change any observable behavior

owenv:
@swift-ci please test

PR #356, opened 2020-11-11
[Explicit Module Builds] Introduce an InterModuleDependencyOracle abstraction for tracking and sharing dependency scanning results

artemcm:
This PR refactors the handling of inter-module dependencies to centralize their handling in a new, reference-type, oracle.

`InterModuleDependencyOracle` will now be responsible for responding to queries about dependencies of a given module.
The intent behind it is to abstract the actual mechanism used for tracking/computing/caching inter-module dependencies. A single oracle reference is intended to be passed around to multiple Driver invocations, corresponding to different targets, by build-system clients (e.g. SwiftPM), replacing today's use of `ModuleInfoMap`-based API, to communicate already-computed dependency-scan results.

As implemented, it is a simple store of `ModuleInfo` value objects, which a `Driver` invocation populates by invoking a `-scan-dependencies` job; not unlike passing around an accumulating `ModuleInfoMap`.
In the future, the oracle's internal implementation may be replaced with a direct dependency-scanner library access, without having to break the clients' API.

There will be a followup PR to cleanup some of the shims required to ensure that this does not break the API currently used by SwiftPM. 

Corresponding PR to adapt this API in SwiftPM: https://github.com/apple/swift-package-manager/pull/3042

Resolves rdar://71209183

### DIFF
--- a/Sources/SwiftDriver/Explicit Module Builds/Inter Module Dependencies/InterModuleDependencyOracle.swift
+++ b/Sources/SwiftDriver/Explicit Module Builds/Inter Module Dependencies/InterModuleDependencyOracle.swift
@@ -10,39 +10,69 @@
 //
 //===----------------------------------------------------------------------===//
 
-
+// An inter-module dependency oracle, responsible for responding to queries about
+// dependencies of a given module, caching already-discovered dependencies along the way.
+//
 // The oracle is currently implemented as a simple store of ModuleInfo nodes.
-// It is the currently responsibility of the swift-driver to populate and update
+// It is the responsibility of the Driver to populate and update
 // the store. It does so by invoking individual -scan-dependencies jobs and
-// accumulating resulting dependency graphs.
+// accumulating resulting dependency graphs into the oracle's store.
 //
 // The design of the oracle's public API is meant to abstract that away,
-// allowing us to, in the future, replace the underlying implementation with
+// allowing us to replace the underlying implementation in the future, with
 // a persistent-across-targets dependency scanning library.
 //
 /// An abstraction of a cache and query-engine of inter-module dependencies
 public class InterModuleDependencyOracle {
-  // MARK: Public API
   /// Query the ModuleInfo of a module with a given ID
-  public func getModuleInfo(of moduleId: ModuleDependencyId) -> ModuleInfo? {
+  @_spi(Testing) public func getModuleInfo(of moduleId: ModuleDependencyId) -> ModuleInfo? {
     return modules[moduleId]
   }
 
   /// Query the direct dependencies of a module with a given ID
-  public func getDependencies(of moduleId: ModuleDependencyId) -> [ModuleDependencyId]? {
+  @_spi(Testing) public func getDependencies(of moduleId: ModuleDependencyId)
+  -> [ModuleDependencyId]? {
     return modules[moduleId]?.directDependencies
   }
-
-  // MARK: Private Implementation
 
   // TODO: This will require a SwiftDriver entry-point for scanning a module given
   // a command invocation and a set of source-files. As-is, the driver itself is responsible
   // for executing individual module dependency-scanning actions and updating oracle state.
+  // (Implemented with InterModuleDependencyOracle::mergeModules extension)
+  //
+  // func getFullDependencies(inputs: [TypedVirtualPath],
+  //                          commandLine: [Job.ArgTemplate]) -> InterModuleDependencyGraph {}
+  //
 
   /// The complete set of modules discovered so far, spanning potentially multiple targets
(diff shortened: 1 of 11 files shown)


artemcm:
@swift-ci please test

Sources/SwiftDriver/Explicit Module Builds/PlaceholderDependencyResolution.swift
```diff
@@ -119,15 +119,15 @@ fileprivate extension InterModuleDependencyGraph {
...
       externalModuleId = swiftModuleId
-    } else if externalModuleInfoMap[swiftPrebuiltModuleId] != nil {
+    } else if dependencyOracle.getModuleInfo(of: swiftPrebuiltModuleId) != nil {
       externalModuleId = swiftPrebuiltModuleId
     } else {
       throw Driver.Error.missingExternalDependency(placeholderId.moduleName)
     }
 
-    let externalModuleInfo = externalModuleInfoMap[externalModuleId]!
+    let externalModuleInfo = dependencyOracle.getModuleInfo(of: externalModuleId)!
```
cltnschlosser, line 130:
```swift
    let externalModuleId: ModuleDependencyId
    let externalModuleInfo: ModuleInfo
    if let moduleInfo = dependencyOracle.getModuleInfo(of: swiftModuleId) {
      externalModuleId = swiftModuleId
      externalModuleInfo = moduleInfo
    } else if let prebuiltModuleInfo = dependencyOracle.getModuleInfo(of: swiftPrebuiltModuleId) {
      externalModuleId = swiftPrebuiltModuleId
      externalModuleInfo = prebuiltModuleInfo
    } else {
      throw Driver.Error.missingExternalDependency(placeholderId.moduleName)
    }
```
Prevents duplicate query. A private throwing function that returns a tuple might be more readable, but that's probably a preference.

cltnschlosser (commented):
Seems like a clear win for users of libSwiftDriver (SPM).

Any negative here when trying to do explicit module builds solely from the command line? Didn't seem like it, but I don't know enough about how this works to tell.

Sources/SwiftDriver/Explicit Module Builds/PlaceholderDependencyResolution.swift
artemcm: Much cleaner, thank you.

artemcm:
Thank you for taking a look, @cltnschlosser, and for your suggestions.

> Any negative here when trying to do explicit module builds solely from the command line? Didn't seem like it, but I don't know enough about how this works to tell.

This does not make a difference to command line builds today; but, it does open up a possibility for the future to no longer have the driver launch `-scan-dependencies` jobs, instead talking to the dependency scanner directly via a library that the oracle will wrap.

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@DougGregor @nkcsgexi ping for a review.
This PR includes a bunch of shims/adaptors to ensure that the API currently in-use by SwiftPM continues working, but it would be nice to get this in to help with the transition to:
https://github.com/apple/swift/pull/34786

artemcm:
@swift-ci please test

Sources/SwiftDriver/Explicit Module Builds/Inter Module Dependencies/InterModuleDependencyOracle.swift
```diff
@@ -0,0 +1,78 @@
...
+// accumulating resulting dependency graphs into the oracle's store.
+//
+// The design of the oracle's public API is meant to abstract that away,
+// allowing us to replace the underlying implementation in the future, with
+// a persistent-across-targets dependency scanning library.
+//
+/// An abstraction of a cache and query-engine of inter-module dependencies
+public class InterModuleDependencyOracle {
+  /// Query the ModuleInfo of a module with a given ID
+  @_spi(Testing) public func getModuleInfo(of moduleId: ModuleDependencyId) -> ModuleInfo? {
```
DougGregor, line 28: Should this class be made thread-safe? Clients may very well have multiple threads in which they create `Driver` instances that reference the same oracle...

DougGregor (approved):

Sources/SwiftDriver/Explicit Module Builds/Inter Module Dependencies/InterModuleDependencyOracle.swift
artemcm: Yes! I've added some basic thread-safety to it now, and will iterate on it as this type transitions to wrapping `libSwiftScan`.

artemcm:
@swift-ci please test

PR #357, opened 2020-11-11
Install swift-help for integration tests. Suppress .noDriver options from being emitted by swift-help.

cltnschlosser:
Fixes: Driver/help.swift

Also replaces deprecated `createSymlink` calls with `localFileSystem.createSymbolicLink`

### DIFF
--- a/Sources/SwiftOptions/OptionTable.swift
+++ b/Sources/SwiftOptions/OptionTable.swift
@@ -37,6 +37,7 @@ extension OptionTable {
 
     for option in options {
       if option.isAlias { continue }
+      if option.attributes.contains(.noDriver) { continue }
       if option.isHelpHidden && !includeHidden { continue }
       guard option.isAccepted(by: driverKind) else { continue }
       guard let helpText = option.helpText else { continue }
--- a/Sources/SwiftOptions/OptionTable.swift
+++ b/Sources/SwiftOptions/OptionTable.swift
@@ -37,7 +37,6 @@ extension OptionTable {
 
     for option in options {
       if option.isAlias { continue }
-      if option.attributes.contains(.noDriver) { continue }
       if option.isHelpHidden && !includeHidden { continue }
       guard option.isAccepted(by: driverKind) else { continue }
       guard let helpText = option.helpText else { continue }
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -1048,7 +1048,7 @@ extension Driver {
     // Some output flags affect the compiler mode.
     if let outputOption = parsedOptions.getLast(in: .modes) {
       switch outputOption.option {
-      case .emitPch, .emitImportedModules:
+      case .emitImportedModules:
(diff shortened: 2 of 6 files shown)


Sources/SwiftOptions/OptionTable.swift
```diff
@@ -37,6 +37,7 @@ extension OptionTable {
 
     for option in options {
       if option.isAlias { continue }
+      if option.attributes.contains(.noDriver) { continue }
```
owenv, line 40: Assuming no options have been incorrectly marked noDriver, it might be good to move this check into isAccepted(by:)

owenv (approved):
Thanks!

DougGregor (approved):

artemcm (approved):

DougGregor:
@swift-ci test

Sources/SwiftOptions/OptionTable.swift
cltnschlosser: Did this, let me know what you think.

owenv (approved):
LGTM!

cltnschlosser:
@swift-ci test

owenv:
@swift-ci test

cltnschlosser:
@owenv `-emit-module-doc-path`, `-index-ignore-stdlib`, `-emit-module-source-info` are (seemingly incorrectly) marked `.noDriver`. Those are just the ones caught by the unit tests. I'll see if I can do a scan for any other `.noDriver` options that are used in the driver. Best solution is to update the options in c++ driver?

cltnschlosser:
Okay, found 3 categories of `.noDriver` options that are used.

Definitely checked for and used:
```
.emitDependenciesPath
.emitModuleDocPath
.emitModuleDoc
.emitModuleSourceInfo
.enableCxxInterop
```

Technically supported, sets the compiler output type to pch, but I don't know if it actually works:
```
.emitPch
```

Erased when `parsedOptions.hasArgument(.indexFile) && parsedOptions.hasArgument(.dumpAst)`, but never actually passed through or used, so maybe that code can be removed:
```
.indexIgnoreStdlib
.indexSystemModules
```

cltnschlosser:
Okay, actually I have to do a bit more digging here. C++ driver claims these are unsupported, so swift-driver probably should too `<unknown>:0: error: unknown argument: '-emit-pch'`.

This happens for all these parameters. First I'm going to find where this check happens in the c++ driver, then I'm going to look at the unit tests we have that are using these parameters.

cltnschlosser:
Okay, I've updated the driver to emit an `unknown argument` error for .noDriver args (this matches c++ driver). Also made a few changes where the swift-driver was expecting these arguments. Including unit tests that used them.

Re-requesting reviews since the diff has changed quite a bit.

Also I need someone to kick off CI :)

DougGregor (approved):

DougGregor:
@swift-ci please test

owenv (approved):
LGTM! I'm surprised this many options were affected, the C++ driver definitely has some dead code that's checking for these...

PR #358, opened 2020-11-12
[Explicit Module Builds] Use a stable hash to encode build arguments into PCM filenames

artemcm:
When generating build jobs for explicit module dependencies, we must distinguish multiple versions of the same Clang module/pcm by encoding the arguments used to build it (e.g. target triple) into the pcm filename.

So far, we have been using an unstable hash, which means that these filenames were not stable across invocations and were being needlessly re-built in some scenarios (e.g. many-target builds).

Switching to a stable hash (SHA256) ensures that the same Clang module, built with the same command line, will produce the same .pcm filename. This will allow the client build system (e.g. SwiftPM) to deduplicate identical compilation jobs because their output filenames will now also match.

### DIFF
--- a/Sources/SwiftDriver/Explicit Module Builds/ExplicitDependencyBuildPlanner.swift
+++ b/Sources/SwiftDriver/Explicit Module Builds/ExplicitDependencyBuildPlanner.swift
@@ -449,9 +449,18 @@ extension ExplicitDependencyBuildPlanner {
   /// is to be constructed with.
   public static func targetEncodedClangModuleName(for moduleName: String,
                                                   pcmArgs: [String]) throws -> String {
-    var hasher = Hasher()
-    pcmArgs.forEach { hasher.combine($0) }
-    return moduleName + String(hasher.finalize())
+    let hashInput = pcmArgs.sorted().joined()
+    let hashedArguments: String
+    #if os(macOS)
+    if #available(macOS 10.15, iOS 13, *) {
+      hashedArguments = CryptoKitSHA256().hash(hashInput).hexadecimalRepresentation
+    } else {
+      hashedArguments = SHA256().hash(hashInput).hexadecimalRepresentation
+    }
+    #else
+    hashedArguments = SHA256().hash(hashInput).hexadecimalRepresentation
+    #endif
+    return moduleName + hashedArguments
   }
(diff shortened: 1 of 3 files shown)


artemcm:
@swift-ci please test

artemcm:
😬 Linux build failing with:
`Fatal error: not supported on this platform: file TSCBasic/HashAlgorithms.swift, line 189`

Sources/SwiftDriver/Explicit Module Builds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -449,9 +449,14 @@ extension ExplicitDependencyBuildPlanner {
   /// is to be constructed with.
   public static func targetEncodedClangModuleName(for moduleName: String,
                                                   pcmArgs: [String]) throws -> String {
-    var hasher = Hasher()
-    pcmArgs.forEach { hasher.combine($0) }
-    return moduleName + String(hasher.finalize())
+    let hashInput = pcmArgs.sorted().joined()
+    let hashedArguments: String
+    if #available(macOS 10.15, iOS 13, *) {
+      hashedArguments = CryptoKitSHA256().hash(hashInput).hexadecimalRepresentation
```
owenv, line 455: It looks like this is failing on Linux because the true branch of the #available condition is taken. I'm not sure if there's a better way to write this than something like:
```swift
#if os(macOS)
if #available(macOS 10.15, iOS 13, *) {
        hashedArguments = CryptoKitSHA256().hash(hashInput).hexadecimalRepresentation
} else {
       hashedArguments = SHA256().hash(hashInput).hexadecimalRepresentation
}
#else 
hashedArguments = SHA256().hash(hashInput).hexadecimalRepresentation
#endif

artemcm: Right, the wildcard...
I can't think of a better way either.
Thanks, @owenv.

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

owenv (approved):

artemcm:
@swift-ci please test

PR #359, opened 2020-11-12
Put the output of the (Linux) autolink step in the same place as the .o files

davidungar:
Following the legacy driver, put the output of the autolink step in the same place as the .o files, instead of in a temp directory. That way, we could possibly omit this step in the future in an incremental build.

rdar://70961296

### DIFF
--- a/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
+++ b/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
@@ -190,17 +190,18 @@ final class NonincrementalCompilationTests: XCTestCase {
 
     XCTAssertEqual(try! buildRecord.inputInfos[VirtualPath(path: file2 )]!.status,
                    .needsCascadingBuild)
-    XCTAssert(try! isCloseEnough(buildRecord.inputInfos[VirtualPath(path: file2 )]!
-                                  .previousModTime.legacyDriverSecsAndNanos,
-                                 [1570318778, 0]))
-    XCTAssertEqual(try! buildRecord.inputInfos[VirtualPath(path: gazorp)]!.status,
+    XCTAssert(try! isCloseEnough(
+                XCTUnwrap(buildRecord.inputInfos[VirtualPath(path: file2 )])
+                  .previousModTime.legacyDriverSecsAndNanos,
+                [1570318778, 0]))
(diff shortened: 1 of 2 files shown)


davidungar:
@swift-ci please test

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
```diff
@@ -545,6 +545,30 @@ final class IncrementalCompilationTests: XCTestCase {
...
+    env["SWIFT_DRIVER_DSYMUTIL_EXEC"] = "/garbage/dsymutil"
+
+    var driver = try! Driver(
+      args: args
+        + ["-emit-library", "-target", "x86_64-unknown-linux"],
+      env: env)
+    let plannedJobs = try! driver.planBuild()
+    let autolinkExtractJob = plannedJobs
+      .filter { $0.kind == .autolinkExtract }
+      .first!
```
harlanhaskins, line 564: Tiny nit: Can you use `try XCTUnwrap(...)` instead? Just so that if this ends up nil, it fails the test rather than crashing.

harlanhaskins (approved):

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
davidungar: Thanks! Great catch, didn't think about XCTUnwrap.

davidungar:
@swift-ci please test

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
```diff
@@ -545,6 +546,31 @@ final class IncrementalCompilationTests: XCTestCase {
...
+
+    var driver = try! Driver(
+      args: args
+        + ["-emit-library", "-target", "x86_64-unknown-linux"],
+      env: env)
+    let plannedJobs = try! driver.planBuild()
+    let autolinkExtractJob = try! XCTUnwrap(
+      plannedJobs
+        .filter { $0.kind == .autolinkExtract }
+        .first)
```
cltnschlosser, line 566: This is test code so it doesn't really matter, but I like when people point things like this out to me.

You can use `first(where:)` https://developer.apple.com/documentation/swift/array/1848165-first to avoid iterating the whole array in cases like this, instead of doing filter then first.

PR #360, opened 2020-11-13
Disable a linux test temporarily

davidungar:
For some reason, the test that checks incremental diagnostics is failing on Linux.
Temporarily disable it; I'll look at it tomorrow.
Also, I had forgotten to regenerate the Linux test list, so this adds in the test for the autolink output path.


davidungar:
@swift-ci please test

CodaFi (approved):
Please file a radar to re-enable this test on Linux.

PR #362, opened 2020-11-13
Remove @_spi's that were preventing compilation

davidungar:
The latest Swift compiler is less tolerant of @_spi's than before. Remove the offenders.
rdar://71381579


davidungar:
@swift-ci please test

artemcm (approved):
It is unfortunate to have `testable` return to the code-base. 
This, plus the recent discovery that in the absence of library-evolution, `implementationOnly` is also somewhat broken, are good motivators to get around to resolving issues with building swift-driver with library evolution.

PR #365, opened 2020-11-14
[DNM] Audit of failing integration tests

cltnschlosser:
Started as an exploration of what to look into next, but figured it was worth doing this for all the currently failing tests and sharing for anyone else that might find it useful.

https://github.com/apple/swift/pull/34405 Will unblock most of the tests, but I assume that a lot of the dependencies tests will still fail after that.

There are 5 tests that looks like they could be fixed by updates to the tests (2 categories here):
- JSON formatting
  - (c++ `key: value`, swift-driver(JSONEncoder) `key : value`), can fix with `key{{ ?}}: value`
  - key/value pair ordering, can fix with `-DAG`
- Input file path -> output file path, swift-driver removes `./` when it's redundant. Can fix with `{{(\.\/)?}}`, tests that also support windows paths might need different regex.

### DIFF
--- a/test-fails.txt
+++ b/test-fails.txt
@@ -1,9 +1,5 @@
-Audit of Failing Driver Integration Tests
+Audit of Failing Driver Integration Tests (Errors may be out of date, but failures should be accurate to commit date)
 
-test/Interpreter
-  Swift(macosx-x86_64) :: Interpreter/vtables_multifile_testable.swift
-    Fatal error: multiple producers for output vtables_multifile_testable_helper.swiftmodule
-  
 validation-test/Driver:
   Swift(macosx-x86_64) :: Driver/Dependencies/rdar23148987.swift
     Required changes to test for spaces and json ordering. There is warning and skipped jobs aren't being skipped.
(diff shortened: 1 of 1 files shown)


artemcm:
Wow, thank you so much for this, @cltnschlosser .

test-fails.txt
```diff
@@ -0,0 +1,175 @@
...
+    Missing -emit-remap-file-path
+
+  Swift(macosx-x86_64) :: Driver/SourceRanges/range-incremental-no-build-record.swift
+    Same error messsage is worded differently
+  
+  Swift(macosx-x86_64) :: Driver/temp-files.swift
+    swift-driver has no -save-temps support
+
+  Swift(macosx-x86_64) :: Driver/batch_mode.swift
+    Linking failed
```
owenv, line 148:
```suggestion
    Linking failed (`-driver-skip-execution should prevent it from attempting the link in the first place, but isn't implemented yet`)
```

test-fails.txt
```diff
@@ -0,0 +1,175 @@
...
+
+  Swift(macosx-x86_64) :: Driver/emit-module-from-sib.swift
+    Output is missing some types
+
+  Swift(macosx-x86_64) :: Driver/options-interpreter.swift
+    Input validation order difference. swift-driver does SDKROOT validation first, c++ driver does objc runtime library validation first.
+    swift-driver needs sdk root early because it uses it to get target info.
+
+  Swift(macosx-x86_64) :: Driver/pipe_round_robin.swift.gyb
+    name 'NumDriverPipeReads' is not defined
```
owenv, line 172:
```suggestion
    name 'NumDriverPipeReads' is not defined (This test depends on the driver's performance counters, which aren't implemented yet)
```

test-fails.txt
```diff
@@ -0,0 +1,175 @@
...
+    ^ https://github.com/apple/swift-driver/pull/370
+
+  Swift(macosx-x86_64) :: Driver/driver_migration.swift
+    Missing -emit-remap-file-path
+
+  Swift(macosx-x86_64) :: Driver/SourceRanges/range-incremental-no-build-record.swift
+    Same error messsage is worded differently
+  
+  Swift(macosx-x86_64) :: Driver/temp-files.swift
+    swift-driver has no -save-temps support
```
owenv, line 145:
```suggestion
    swift-driver has no -save-temps support, and if an environment variable like `TMP` is set, swift-driver places temporary files in a subdirectory instead of immediately inside the specified one
```

test-fails.txt
```diff
@@ -0,0 +1,156 @@
...
+
+  Swift(macosx-x86_64) :: Driver/pipe_round_robin.swift.gyb
+    name 'NumDriverPipeReads' is not defined (This test depends on the driver's performance counters, which aren't implemented yet)
+
+  Swift(macosx-x86_64) :: Driver/response-file.swift
+    Temporary directory additional level, `TMPDIR` vs `TMPDIR/TemporaryDirectory.#####/`
+    
+  Swift(macosx-x86_64) :: Driver/help.swift
+    Expecting `SEE ALSO: swift build, swift run, swift package, swift test`
+    Broken with: https://github.com/apple/swift-driver/pull/433
```
cltnschlosser, line 156: @owenv Not sure if this is a regression or a purposeful change

owenv: This was intentional, I think there's a lot of room for improvement with the current help messages. It should be OK to XFAIL when running with the new driver, if we don't have one already I can add an XCTest for the help output

cltnschlosser:
A couple new failures to look into, but all the dependencies tests are passing now that they are running correctly 🎉 

https://github.com/apple/swift-driver/pull/365/commits/d1939e8a1bf7668ab902faafd5af2ab3167db601

PR #366, opened 2020-11-14
Option parsing fixes when args have unexpected suffixes

owenv:
1. If a flag has an unexpected suffix and is not supported by the current driver mode, diagnose it as an unsupported option instead of being available in another mode
2. Reject 'separate' options if they have an unexpected suffix


owenv:
@swift-ci please test

cltnschlosser (approved):

DougGregor (approved):

PR #370, opened 2020-11-15
Add Android linker -target conversion

cltnschlosser:
Fixes: Driver/android-link.swift

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -43,6 +43,27 @@ extension GenericUnixToolchain {
     return triple.archName
   }
 
+  private func getTargetForLinker(targetTriple: Triple) -> String {
+    if targetTriple.isAndroid, let arch = targetTriple.arch {
+      switch arch {
+      case .arm, .thumb:
+        // Current Android NDK versions only support ARMv7+.  Always assume ARMv7+
+        // for the arm/thumb target.
+        return "armv7-unknown-linux-androideabi";
+      case .aarch64:
+        return "aarch64-unknown-linux-android";
+      case .x86:
+        return "i686-unknown-linux-android";
+      case .x86_64:
+        return "x86_64-unknown-linux-android";
+      default:
+        // FIXME: we should just abort on an unsupported target
+        return targetTriple.triple
+      }
+    }
+    return targetTriple.triple
+  }
+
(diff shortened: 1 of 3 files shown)


Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
```diff
@@ -43,6 +43,27 @@ extension GenericUnixToolchain {
...
+        return "armv7-unknown-linux-androideabi";
+      case .aarch64:
+        return "aarch64-unknown-linux-android";
+      case .x86:
+        return "i686-unknown-linux-android";
+      case .x86_64:
+        return "x86_64-unknown-linux-android";
+      default:
+        // FIXME: we should just abort on an unsupported target
+        return targetTriple.triple
```
owenv, line 61: We can probably just throw an error like `unsupported target: 'xyz'` instead here

owenv (approved):
I'm not very familiar with the Android support but this looks like a straightforward port from the C++ version! Android is so similar to regular Linux that it probably doesn't make sense to add it as a separate toolchain, I think the only other difference is the sanitizer library names

cltnschlosser:
Yeah c++ subclasses the Unix one. We currently have that toolchain as a final class. It didn’t seem worth changing that just for this. I’ll look for the sanitizer stuff in the c++ driver.

buttaface:
Looked into this now, I think that entire code portion from the C++ driver is a mistake and unused. The Unix C++ driver sets the target for clang twice in [two](https://github.com/apple/swift/blob/f196c16532cd38f5b5ce4cf80fe9ab122d5d808b/lib/Driver/UnixToolChains.cpp#L152) [different](https://github.com/apple/swift/blob/f196c16532cd38f5b5ce4cf80fe9ab122d5d808b/lib/Driver/UnixToolChains.cpp#L323) places, with the second just overriding the first and this Android function. [The second one was added more than four years ago](https://github.com/apple/swift/blob/f196c16532cd38f5b5ce4cf80fe9ab122d5d808b/lib/Driver/UnixToolChains.cpp#L323) and was likely a mistake, as the first one was already there, but is now the only one used by clang. This repo simply copies over both flag invocations.

I suggest instead you simply remove the first `-target` flag from this repo and forget about these Android triples, and instead simply bring over [the sanitizer change I added last year](https://github.com/apple/swift/blob/f196c16532cd38f5b5ce4cf80fe9ab122d5d808b/lib/Driver/UnixToolChains.cpp#L46).

PR #371, opened 2020-11-16
Handle -continue-building-after-errors and add mechanism for stopping build on failure

cltnschlosser:
Fixes Driver/continue-building-after-errors.swift

Long term we probably want to add more to llbuildSwift, so that we can cancel the build there instead of going through all the jobs. We could suspend the jobQueue so it doesn't start anymore. It also feels like llbuild should have a mechanism for throttling parallel builds, so maybe with that the queue could be replaced.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -183,6 +183,9 @@ public struct Driver {
   /// The specified maximum number of parallel jobs to execute.
   @_spi(Testing) public let numParallelJobs: Int?
 
+  /// Whether jobs should continue to be executed after a failure.
+  @_spi(Testing) public let continueBuildingAfterErrors: Bool
+
   /// The set of sanitizers that were requested
   let enabledSanitizers: Set<Sanitizer>
 
@@ -407,6 +410,8 @@ public struct Driver {
     // Multithreading.
     self.numThreads = Self.determineNumThreads(&parsedOptions, compilerMode: compilerMode, diagnosticsEngine: diagnosticEngine)
     self.numParallelJobs = Self.determineNumParallelJobs(&parsedOptions, diagnosticsEngine: diagnosticEngine, env: env)
+
+    self.continueBuildingAfterErrors = Self.computeContinueBuildingAfterErrors(&parsedOptions, compilerMode: compilerMode)
 
     Self.validateWarningControlArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
     Self.validateProfilingArgs(&parsedOptions,
@@ -896,6 +901,7 @@ extension Driver {
       workload: .init(allJobs, incrementalCompilationState),
       delegate: createToolExecutionDelegate(),
       numParallelJobs: numParallelJobs ?? 1,
+      continueBuildingAfterErrors: continueBuildingAfterErrors,
       forceResponseFiles: forceResponseFiles,
       recordedInputModificationDates: recordedInputModificationDates)
   }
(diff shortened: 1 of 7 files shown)


Sources/SwiftDriver/Execution/DriverExecutor.swift
```diff
@@ -27,6 +27,7 @@ public protocol DriverExecutor {
   func execute(workload: DriverExecutorWorkload,
                delegate: JobExecutionDelegate,
                numParallelJobs: Int,
+               continueBuildingAfterErrors: Bool,
                forceResponseFiles: Bool,
```
owenv, line 31: This will require a corresponding update to SPM's executor

Tests/SwiftDriverTests/JobExecutorTests.swift
```diff
@@ -306,14 +306,98 @@ final class JobExecutorTests: XCTestCase {
           $0 <<< "let bar = 3"
         }
 
-        // FIXME: It's unfortunate we diagnose this twice, once for each job which uses the input.
-        verifier.expect(.error("input file '\(other.description)' was modified during the build"))
```
owenv, line 310: Nice!

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -183,6 +183,9 @@ public struct Driver {
   /// The specified maximum number of parallel jobs to execute.
   @_spi(Testing) public let numParallelJobs: Int?
 
+  /// Whether jobs should continue to be executed after a failure.
+  @_spi(Testing) public let continueBuildingAfterErrors: Bool
```
owenv, line 187: We'll need to drop te @_spi here until the related build issues get sorted out

owenv (commented):
This approach seems like it should work fine as far as I can tell, so these changes LGTM once one usage of _spi is temporarily dropped and we update SPM's executor to match the changed protocol requirement.

Sources/SwiftDriver/Driver/Driver.swift
cltnschlosser: It's being used a lot in this file. Not sure what the issue is, but it's still being used in a lot of places.

Sources/SwiftDriver/Execution/DriverExecutor.swift
davidungar: That's why in my version, I put this bit in the workload parameter. Seemed like one could defend that the decision to run jobs *after* an error was a workload parameter.

Tests/SwiftDriverTests/JobExecutorTests.swift
davidungar: Is this still needed if -driver-continue-building... is set? Or can we 86 it?

Sources/SwiftDriver/Driver/Driver.swift
davidungar: I have a set of changes I have not pushed in a PR (yet) that comments out all the spi's so we can use the older compilers. Not sure what to do about this issue.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -428,11 +446,14 @@ class ExecuteJobRule: LLBuildRule {
...
       return engine.taskIsComplete(DriverBuildValue.jobExecution(success: false))
     }
 
     context.jobQueue.addOperation {
+      guard !self.context.isCancelled else {
+        return engine.taskIsComplete(DriverBuildValue.jobExecution(success: false))
+      }
       self.executeJob(engine)
     }
   }
```
davidungar, line 459: If these changes work, this part of the PR is more elegant than mine.

davidungar:
@cltnschlosser What do you think? I like the workload integration in my PR, but I like the way your code stops the build better in your PR. What are your thoughts?

Tests/SwiftDriverTests/JobExecutorTests.swift
cltnschlosser: Nope not needed. It was getting emitted twice because the second job was running even though the first one failed.

cltnschlosser: Oh @davidungar I missunderstood your question. Yes, if continue building after errors is set, then we would get 2 errors, but since it's not and the driver defaults to 1 job at a time, it only outputs once.

Is this not causing a test failure in #387 ?

cltnschlosser:
Closing this in favor of #387

Tests/SwiftDriverTests/JobExecutorTests.swift
davidungar: Thank you, @cltnschlosser . Apparently not causing a failure there. As soon as I've gotten my PRs in, I'll push forward on the legacy tests and see what pops up.

PR #373, opened 2020-11-17
[build-script-helper] Use separate module cache for builds

edymtt:
This will ensure that multiple build jobs running in parallel on the
same machine using different Swift compilers (like it could be the case in CI) do not interfere with
each other.

Addresses rdar://71373849


artemcm (approved):
Thank you, @edymtt.
And sorry, a recent PR introduced a merge conflict to this change.

edymtt:
@swift-ci please test

PR #374, opened 2020-11-17
Ensure temp dir is different between testIncremental and testIncrementalDiagnostics to see if that fixes the test breakage in CI

davidungar:
Best theory to explain a mysterious test failure on CI Linux. Land it and if test starts failing, disable the test again.

### DIFF
--- a/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
+++ b/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
@@ -358,10 +358,8 @@ final class IncrementalCompilationTests: XCTestCase {
 
   // FIXME: why does it fail on Linux in CI?
   func testIncrementalDiagnostics() throws {
-    #if !os(Linux)
     prepare()
     try testIncremental(checkDiagnostics: true)
-    #endif
   }
 
   func testIncremental() throws {
--- a/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
+++ b/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
@@ -289,7 +289,7 @@ final class NonincrementalCompilationTests: XCTestCase {
 
 final class IncrementalCompilationTests: XCTestCase {
 
-  var tempDir: AbsolutePath = AbsolutePath("/tmp")
+  let tempDir: AbsolutePath
 
   let module = "theModule"
   var OFM: AbsolutePath {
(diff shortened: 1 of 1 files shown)


davidungar:
@swift-ci please test

artemcm (commented):
I think there is a better way to do this that avoids having to use `#line` and remembering to call `prepare` on new tests:

We can keep the `setUp` method in place and use the `self.name` `XCTestCase` property for the prefix on `withTemporaryDirectory`. This `name` property will have the format that, for a given test, looks like this:
`"-[IncrementalCompilationTests testIncrementalDiagnostics]"`

`self.name` docs:
https://developer.apple.com/documentation/xctest/xctest/1500990-name

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
```diff
@@ -357,12 +358,12 @@ final class IncrementalCompilationTests: XCTestCase {
 
   // FIXME: why does it fail on Linux in CI?
```
artemcm, line 359: Stale comment.

cltnschlosser:
What about removing the setup all together and wrapping the tests in withTemporaryDirectory. Has the additional benefit of deleting the folder after the test completes.

davidungar:
@swift-ci please test

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
davidungar: Thanks so much for reviewing so quickly. Nice catch!

davidungar:
> I think there is a better way to do this that avoids having to use `#line` and remembering to call `prepare` on new tests:
> 
> We can keep the `setUp` method in place and use the `self.name` `XCTestCase` property for the prefix on `withTemporaryDirectory`. This `name` property will have the format that, for a given test, looks like this:
> `"-[IncrementalCompilationTests testIncrementalDiagnostics]"`
> 
> `self.name` docs:
> https://developer.apple.com/documentation/xctest/xctest/1500990-name

I like your idea! I found a refinement, of overriding the constructor, so that tempDir can be constant. Thanks.

davidungar:
> What about removing the setup all together and wrapping the tests in withTemporaryDirectory. Has the additional benefit of deleting the folder after the test completes.

Thanks for looking at this. Even if setup is removed, all that stuff has to go somewhere. Somehow it seems easier to factor this way. And there is a deinit to rm the temp dir.

cltnschlosser:
Okay, yeah these tests you would have to pass the temp directory through quite a few layers, so this approach is definitely easier. Normally I think the closure approach is superior. We use it in other places.
```swift
  func testInputModifiedDuringSingleJobBuild() throws {
    try withTemporaryDirectory { path in
      let main = path.appending(component: "main.swift")
      try localFileSystem.writeFileContents(main) {
        $0 <<< "let foo = 1"
      }

      var driver = try Driver(args: ["swift", main.pathString])
      XCTAssertTrue(jobs.count == 1 && jobs[0].requiresInPlaceExecution)

      // Change the file
      try localFileSystem.writeFileContents(main) {
        $0 <<< "let foo = 1"
      }

      XCTAssertThrowsError(try driver.run(jobs: jobs)) {
        XCTAssertEqual($0 as? Job.InputError,
                       .inputUnexpectedlyModified(TypedVirtualPath(file: .absolute(main), type: .swift)))
      }

    }
  }
```

artemcm (approved):

davidungar:
@swift-ci please test

davidungar:
Thanks, @artemcm

PR #377, opened 2020-11-20
Don't print new line after the driver version

nkcsgexi:
This allows us to see the compiler version with the driver version in the same line.


nkcsgexi:
@swift-ci please test

artemcm (approved):

PR #380, opened 2020-11-28
Ensure uniqueness of output filenames when merging modules

owenv:
This fixes one of causes of the "multiple producers for output" assertion when building. For jobs that include a merge-modules step, it ensures that the intermediate swiftmodule don't have the same name as the top-level module output.

I've been thinking about how we could improve `createTemporaryFilename` too so it isn't a global function, but I haven't come up with any good ideas yet.

Fixes:
Driver/batch_mode_aux_file_order.swift
Driver/batch_mode_with_WMO_or_index.swift
ModuleInterface/linking-to-module.swift
multifile/objc_closure_escape/main.swift
multifile/typealias/one-module/main.swift

### DIFF
--- a/Sources/SwiftDriver/Jobs/CompileJob.swift
+++ b/Sources/SwiftDriver/Jobs/CompileJob.swift
@@ -51,8 +51,11 @@ extension Driver {
     }
 
     if !isTopLevel {
-      return TypedVirtualPath(file:VirtualPath.temporary(.init(baseName.appendingFileTypeExtension(outputType))),
-                              type: outputType)
+      return TypedVirtualPath(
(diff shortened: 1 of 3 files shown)


owenv:
@swift-ci please test

owenv:
Hmm, this breaks a lot of tests, I might need to reconsider the approach a bit

DougGregor (approved):

cltnschlosser:
> Hmm, this breaks a lot of tests, I might need to reconsider the approach a bit

This is probably the right approach, or atleast a step in the right direction. The tests likely just need to be updated to pattern match the output file names instead of equality checking.

owenv:
@swift-ci please test

artemcm (approved):

PR #381, opened 2020-12-02
Set up a client-side C library to talk with C APIs exposed from the compiler

nkcsgexi:
We could use functions like dlopen and dlsym in this client-side library to talk with
compiler-side C APIs.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -13,6 +13,7 @@ import TSCBasic
 import TSCUtility
 import Foundation
 import SwiftOptions
+@_implementationOnly import _CSwiftDriver
 
 /// The Swift driver.
 public struct Driver {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ cmake_minimum_required(VERSION 3.15.1)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules)
 
-project(SwiftDriver LANGUAGES Swift)
+project(SwiftDriver LANGUAGES Swift C)
 
 set(SWIFT_VERSION 5)
 set(CMAKE_Swift_LANGUAGE_VERSION ${SWIFT_VERSION})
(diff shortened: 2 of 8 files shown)


nkcsgexi:
@swift-ci please test

cltnschlosser:
Is the idea to use this for llvm?

artemcm (commented):
When I try to build this with CMake, I get:
```
CMake Error at Sources/CMakeLists.txt:9 (add_subdirectory):                                                                                                                                                                              The source directory                                                                                                                                                                                                                                                                                                                                                                                                                                                            /Volumes/Data/GHWorkspace1/swift-driver/Sources/_CSwiftDriver/src

  does not contain a CMakeLists.txt file.
```
Is there perhaps a file missing from this PR?
This target's `CMakeLists` should define it as a library, and we need to take care to link it in statically when building `libSwiftDriver` to avoid having to install another shared library into the toolchain.

artemcm:
> Is the idea to use this for llvm?

Something like this will be an example of where this is useful:
https://github.com/apple/swift/pull/34786

nkcsgexi:
@cltnschlosser, this is to allow libSwiftDriver to talk with Swift compiler from different toolchains. In that case, we cannot statically link against libraries vendored from a compiler. We need to rely on C APIs like `dlopen` and `dlsym` to load dylibs dynamically.

nkcsgexi:
@artemcm hmm, will take a look at the cmake issue.

tschuett:
sourcekit-lsp does this already:
https://github.com/apple/sourcekit-lsp/blob/eaea937ee4f5aea0a055bd5f682f31da0fa7f82f/Sources/SourceKitD/SourceKitDImpl.swift#L64
and
https://github.com/apple/sourcekit-lsp/blob/eaea937ee4f5aea0a055bd5f682f31da0fa7f82f/Sources/SourceKitD/sourcekitd_functions.swift#L16

nkcsgexi:
@tschuett yeah, refactoring those pieces seem to me a good idea. Regardless of whether we need to invoke `dlopen` in C, we still need to set up this C target to host those compiler headers for the dependencies scanning library.

tschuett:
No worries. You are missing some newlines at the end of files in a couple of places.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -13,6 +13,7 @@ import TSCBasic
 import TSCUtility
 import Foundation
 import SwiftOptions
+import _CSwiftDriver
```
artemcm, line 16:
```suggestion
@_implementationOnly import _CSwiftDriver
```
Because we do not wish clients of `SwiftDriver` to have to go looking for this module, I think it should be used with `@_implementationOnly`. Otherwise even `SwiftDriverExecution` target will not build with CMake because it will go looking for the modulemap of this new C target and not know where to find it.

nkcsgexi: That's reasonable. Fixed.

nkcsgexi:
@swift-ci please test

artemcm (approved):

PR #383, opened 2020-12-03
[NFC] Encapsulate local outputFileMap to forestall bugs

davidungar:
Since the constructor uses a local `outputFileMap` that is different from `self.outputFileMap`, wrap the local in a block so later uses in this long method cannot accidentally use the wrong one.


davidungar:
@swift-ci please test

artemcm (approved):
Oh my goodness, thank you.

davidungar:
My pleasure, and thank you.

PR #385, opened 2020-12-03
[Incremental] Allow build record to reside at a relative path

davidungar:
Allow build record to reside at a relative path on the way to making legacy lit tests work.

### DIFF
--- a/Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
+++ b/Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
@@ -96,7 +96,9 @@ import SwiftOptions
       diagnosticEngine.emit(.warning_incremental_requires_build_record_entry)
       return nil
     }
-    return partialBuildRecordPath.resolveRelativePath(workingDirectory)
+    return workingDirectory
+      .map(partialBuildRecordPath.resolvedRelativePath(base:))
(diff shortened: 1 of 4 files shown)


davidungar:
@swift-ci please test

davidungar:
Part of rdar://71869992 (Swift Driver must handle relative paths in build records and other places in order to support legacy Driver/Dependencies tests)

Sources/SwiftDriver/Incremental Compilation/BuildRecordInfo.swift
```diff
@@ -175,3 +178,12 @@ import SwiftOptions
     }
   }
 }
+
+extension VirtualPath {
```
artemcm, line 182: We already have:
```
extension VirtualPath {
  fileprivate func resolvedRelativePath(base: AbsolutePath) -> VirtualPath {
    guard case let .relative(relPath) = self else { return self }
    return .absolute(.init(base, relPath))
  }
}
```
In `OutputFileMap.swift`. Should we common these two out?

davidungar: Thanks for catching this! Will do!!

artemcm (approved):

davidungar:
@swift-ci please test

PR #387, opened 2020-12-04
[Incremental] stop if no -driver-continue-building-after-errors

davidungar:
Without being aware of @cltnschlosser 's similar PR ( https://github.com/apple/swift-driver/pull/371/files ), I wrote this code last week. I suspect the two PRs do the same thing. So, not being sure how to proceed, I thought I would put this up and we might review each other's work and come up with a nice synthesis, (hence the "DNM").

Rather than disturb the interface that goes across to swiftpm, I included the extra parameter in the workload structure.

BTW, some version of this functionality is needed for the legacy lit tests.

Have incorporated several great ideas from @cltnschlosser  's PR.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -859,12 +859,14 @@ extension Driver {
                            recordedInputModificationDates: recordedInputModificationDates)
       return
     }
-
-    try performTheBuild(allJobs: jobs, forceResponseFiles: forceResponseFiles)
-
-    buildRecordInfo?.writeBuildRecord(
-      jobs,
-      incrementalCompilationState?.skippedCompilationInputs)
+    do {
+      defer {
+        buildRecordInfo?.writeBuildRecord(
+          jobs,
+          incrementalCompilationState?.skippedCompilationInputs)
+      }
+      try performTheBuild(allJobs: jobs, forceResponseFiles: forceResponseFiles)
+    }
 
     // If requested, warn for options that weren't used by the driver after the build is finished.
     if parsedOptions.hasArgument(.driverWarnUnusedOptions) {
@@ -896,8 +898,11 @@ extension Driver {
     allJobs: [Job],
     forceResponseFiles: Bool
   ) throws {
+    let continueBuildingAfterErrors = parsedOptions.contains(.continueBuildingAfterErrors)
     try executor.execute(
-      workload: .init(allJobs, incrementalCompilationState),
+      workload: .init(allJobs,
+                      incrementalCompilationState,
+                      continueBuildingAfterErrors: continueBuildingAfterErrors),
       delegate: createToolExecutionDelegate(),
       numParallelJobs: numParallelJobs ?? 1,
       forceResponseFiles: forceResponseFiles,
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -898,7 +898,7 @@ extension Driver {
     allJobs: [Job],
     forceResponseFiles: Bool
   ) throws {
-    let continueBuildingAfterErrors = parsedOptions.contains(.continueBuildingAfterErrors)
+    let continueBuildingAfterErrors = computeContinueBuildingAfterErrors()
     try executor.execute(
       workload: .init(allJobs,
                       incrementalCompilationState,
@@ -1419,7 +1419,24 @@ extension Driver {
 
     return numJobs
   }
+
+  private mutating func computeContinueBuildingAfterErrors() -> Bool {
+    // Note: Batch mode handling of serialized diagnostics requires that all
+    // batches get to run, in order to make sure that all diagnostics emitted
+    // during the compilation end up in at least one serialized diagnostic file.
+    // Therefore, treat batch mode as implying -continue-building-after-errors.
+    // (This behavior could be limited to only when serialized diagnostics are
+    // being emitted, but this seems more consistent and less surprising for
+    // users.)
+    // FIXME: We don't really need (or want) a full ContinueBuildingAfterErrors.
+    // If we fail to precompile a bridging header, for example, there's no need
+    // to go on to compilation of source files, and if compilation of source files
+    // fails, we shouldn't try to link. Instead, we'd want to let all jobs finish
+    // but not schedule any new ones.
+    return compilerMode.isBatchCompile || parsedOptions.contains(.continueBuildingAfterErrors)
+  }
 }
+
 
 extension Diagnostic.Message {
   static func remark_max_determinism_overriding(_ option: Option) -> Diagnostic.Message {
--- a/Sources/SwiftDriver/Execution/DriverExecutor.swift
+++ b/Sources/SwiftDriver/Execution/DriverExecutor.swift
@@ -47,12 +47,26 @@ public protocol DriverExecutor {
   func description(of job: Job, forceResponseFiles: Bool) throws -> String
 }
 
-public enum DriverExecutorWorkload {
-  case all([Job])
-  case incremental(IncrementalCompilationState)
+public struct DriverExecutorWorkload {
+  public let continueBuildingAfterErrors: Bool
+  public enum Kind {
+    case all([Job])
+    case incremental(IncrementalCompilationState)
+  }
+  public let kind: Kind
 
-  init(_ allJobs: [Job], _ incrementalCompilationState: IncrementalCompilationState?) {
-    self = incrementalCompilationState.map {.incremental($0)} ?? .all(allJobs)
+  public init(_ allJobs: [Job],
+       _ incrementalCompilationState: IncrementalCompilationState?,
+       continueBuildingAfterErrors: Bool
+  ) {
+      self.continueBuildingAfterErrors = continueBuildingAfterErrors
+      self.kind = incrementalCompilationState
+        .map {.incremental($0)}
+        ?? .all(allJobs)
+  }
+
+  static public func all(_ jobs: [Job]) -> Self {
+    .init(jobs, nil, continueBuildingAfterErrors: false)
   }
 }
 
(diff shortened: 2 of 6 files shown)


davidungar:
@swift-ci please test

cltnschlosser:
I’ll take a look at this tomorrow! Was a few weeks ago when I posted that PR, but I do remember not being completely satisfied with it.

davidungar:
> I’ll take a look at this tomorrow! Was a few weeks ago when I posted that PR, but I do remember not being completely satisfied with it.

Thank you, @cltnschlosser . I'll look forward to your thoughts.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -896,8 +898,11 @@ extension Driver {
     allJobs: [Job],
     forceResponseFiles: Bool
   ) throws {
+    let continueBuildingAfterErrors = parsedOptions.contains(.continueBuildingAfterErrors)
```
cltnschlosser, line 901: You'll see in my PR I made a function for computing this and copied over a comment from the c++ driver. ContineuBuildingAfterErrors is set to true when compiling in batch mode.

Sources/SwiftDriver/Execution/DriverExecutor.swift
```diff
@@ -47,12 +47,26 @@ public protocol DriverExecutor {
   func description(of job: Job, forceResponseFiles: Bool) throws -> String
 }
 
-public enum DriverExecutorWorkload {
-  case all([Job])
-  case incremental(IncrementalCompilationState)
+public struct DriverExecutorWorkload {
```
cltnschlosser, line 50: I do like this better, from a SPM compatibility perspective.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -87,6 +87,12 @@ public final class MultiJobExecutor {
     /// any newly-required job. Set only once.
     private(set) var executeAllJobsTaskBuildEngine: LLTaskBuildEngine? = nil
 
+    /// If a job fails, the driver needs to stop running jobs.
+    private(set) var shouldStopRunningJobs = false
```
cltnschlosser, line 91: naming: `shouldStopRunningNewJobs`, because it doesn't stop any actively running jobs, just scheduled ones.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -472,6 +494,10 @@ class ExecuteJobRule: LLBuildRule {
     let value: DriverBuildValue
     var pid = 0
     do {
+      if context.shouldStopRunningJobs {
+        engine.taskIsComplete(DriverBuildValue.jobExecution(success: false))
+        return
+      }
```
cltnschlosser, line 500: You mentioned you liked my implementation better here. I'm fine with either. They should be logically equivalent, I was stopping before the call to executeJob and you're stopping near the beginning. I do think this should either be moved outside of executeJob or moved all the way to the top, because skips any job startup work that might occur. Environment variable merging in this case (Even though that's probably near 0 overhead anyway).

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -225,6 +234,19 @@ public final class MultiJobExecutor {
         executeAllJobsTaskBuildEngine!.taskNeedsInput(key, inputID: index)
       }
     }
+
+    fileprivate func decideIfShouldStopRunningJobs(_ result: ProcessResult) {
+      switch (result.exitStatus, continueBuildingAfterErrors) {
+      case (.terminated(let code), false) where code != EXIT_SUCCESS:
+         shouldStopRunningJobs = true
+       #if !os(Windows)
+       case (.signalled, _):
```
cltnschlosser, line 243: I'm guessing this is more correct than in my solution where I used result.success as the flag and never considered if the failure was due to a signal or not.

cltnschlosser (commented):
I like this, although I'm surprised `testInputModifiedDuringMultiJobBuild` didn't fail. Also I think we should bring over the test that I included in my PR.

davidungar:
@cltnschlosser Thank you for your thoughtful review. I'll work on incorporating your suggestions.

davidungar:
> I like this, although I'm surprised `testInputModifiedDuringMultiJobBuild` didn't fail. Also I think we should bring over the test that I included in my PR.

Yes! I'll bring over that test.

Sources/SwiftDriver/Driver/Driver.swift
davidungar: Yes. Good idea! Since I'm not storing this value as a driver instance variable, the function doesn't have to be static, though the test is a little more complicated. The test will require my other PR to land to support `-driver-use-frontend-path`. But I do like the idea of the test.

Sources/SwiftDriver/Execution/DriverExecutor.swift
davidungar: Thanks!

Sources/SwiftDriverExecution/MultiJobExecutor.swift
davidungar: Yes, good catch!

davidungar: How about `isBuildCancelled` ? I like the specificity of including the object.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
davidungar: Thank you. Yes, I've gone with your design, tightened up a bit.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
davidungar: Thanks!

Sources/SwiftDriver/Driver/Driver.swift
cltnschlosser: Yeah, I think I stored it specifically for test, but it's also not that important of a test.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
cltnschlosser: That name seems fine :)

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1471,6 +1471,31 @@ final class SwiftDriverTests: XCTestCase {
...
+                   numParallelJobs: Int,
+                   forceResponseFiles: Bool,
+                   recordedInputModificationDates: [TypedVirtualPath : Date]) throws {
+        XCTAssert(workload.continueBuildingAfterErrors)
+      }
+      func checkNonZeroExit(args: String..., environment: [String : String]) throws -> String {
+        fatalError()
+      }
+      func description(of job: Job, forceResponseFiles: Bool) throws -> String {
+        fatalError()
```
cltnschlosser, line 1492: Maybe XCTFail instead of crashing

davidungar: Yes, that's better. Done!

Sources/SwiftDriver/Driver/Driver.swift
davidungar: Thanks! I know at least one legacy will fail of the driver *doesn't* cancel when it needs to, but I like the idea of testing what  your test does, so I put one in.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
davidungar: Thanks!

davidungar:
@cltnschlosser If this looks good at this point, it would be productive to have your official approval.

cltnschlosser (approved):
I just realized that there might be a data race on the isBuildCancelled property because the operationQueue can run concurrent jobs, but that can be fixed in a follow up, and it's worth considering the thread safety of the whole context.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -426,9 +448,9 @@ class ExecuteJobRule: LLBuildRule {
     rememberIfInputSucceeded(engine, value: value)
   }
 
+  /// Called when the build engine thinks all inputs are available in order to run the job.
   override func inputsAvailable(_ engine: LLTaskBuildEngine) {
-    // Return early any of the input failed.
-    guard allInputsSucceeded else {
+    guard allInputsSucceeded, !context.isBuildCancelled else {
       return engine.taskIsComplete(DriverBuildValue.jobExecution(success: false))
     }
```
cltnschlosser, line 456: Oh @davidungar Looks like you are missing https://github.com/apple/swift-driver/pull/371/files/a00c657b31853c25f2544f0cef6d836f517c2179#diff-b2d6d242100a7556a15175d3576106d4a413a2b7b5ffa3bbb52f74210b7408e3R454-R456 Right below this. It's needed to stop jobs which have been scheduled(inputs are available) but haven't started yet(due to throttling the number of concurrent jobs). This is why you don't see the test failure I was talking about in my recent comment on #371

davidungar: Ah, yes, I see. Given the issue, I think it means the other way is better, moving the test to the top of `executeJob`.
Thanks a lot for pointing it out!

davidungar:
@swift-ci please test

davidungar:
> I just realized that there might be a data race on the isBuildCancelled property because the operationQueue can run concurrent jobs, but that can be fixed in a follow up, and it's worth considering the thread safety of the whole context.

Great point! And thanks for the approval.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
cltnschlosser: Yeah, I had it in both spots, inside the queue and before it, but inside only should be fine.

davidungar:
Fixes rdar://71869814

davidungar:
@swift-ci please test

PR #388, opened 2020-12-04
[Incremental] Honor -driver-use-frontend-path

davidungar:
In order to use legacy lit dependency tests, the driver must honor `-driver-use-fronttend-path`. In that use-case, there is no point in querying the frontend for target info, so add a computation for dummy values for this.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2055,6 +2055,9 @@ extension Driver {
     } else {
       swiftCompilerPrefixArgs = []
     }
+    var hasFrontendBeenRedirectedForTesting: Bool {
+      return !swiftCompilerPrefixArgs.isEmpty
+    }
 
     // Find the SDK, if any.
     let sdkPath: VirtualPath? = Self.computeSDKPath(
@@ -2063,8 +2066,11 @@ extension Driver {
       diagnosticsEngine: diagnosticsEngine, env: env)
 
     // Query the frontend for target information.
+    // If there's a dummy frontend, don't query it.
     do {
-      var info = try executor.execute(
+      var info = hasFrontendBeenRedirectedForTesting
+        ? FrontendTargetInfo.dummyForTesting(toolchain)
+        : try executor.execute(
         job: toolchain.printTargetInfoJob(
           target: explicitTarget, targetVariant: explicitTargetVariant,
           sdkPath: sdkPath, resourceDirPath: resourceDirPath,
--- a/Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
+++ b/Sources/SwiftDriver/Jobs/PrintTargetInfoJob.swift
@@ -95,6 +95,17 @@ extension SwiftVersion: Codable {
     /// Whether the Swift libraries need to be referenced in their system
     /// location (/usr/lib/swift) via rpath.
     let librariesRequireRPath: Bool
+
+    static func dummyForTesting(_ toolchain: Toolchain) -> Self {
+      let dummyForTestingTriple = Triple.dummyForTesting(toolchain)
+      return Self(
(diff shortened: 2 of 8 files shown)


davidungar:
@swift-ci please test

owenv:
I think I'd prefer to revive https://github.com/apple/swift/pull/34405 (which I'd forgotten about, oops) instead of working around this in the driver, just in case we want to write target-specific tests at some point. What do you think?

cltnschlosser:
Is that flag only meant to be used for testing? Or is it considered acceptable to use it for normal compilations? @owenv worked around this issue by changing the mock frontends: https://github.com/apple/swift/pull/34405

davidungar:
Thanks, @owenv and @cltnschlosser . I'll have a look tomorrow.

davidungar:
> I think I'd prefer to revive [apple/swift#34405](https://github.com/apple/swift/pull/34405) (which I'd forgotten about, oops) instead of working around this in the driver, just in case we want to write target-specific tests at some point. What do you think?

It looks to me that  apple/swift#34405  does not supplant this PR. (I haven't actually tried it, so please correct me if I'm wrong.) `-driver-use-frontend-path` is used in lots of tests Driver/Dependencies where the test is checking out the dependency mechanism, for instance `malformed-fine.swift`. So, I don't see how 34405 would help in these cases.

davidungar:
> Is that flag only meant to be used for testing? Or is it considered acceptable to use it for normal compilations? @owenv worked around this issue by changing the mock frontends: [apple/swift#34405](https://github.com/apple/swift/pull/34405)

Because in `Options.td`, I see ```
def driver_use_frontend_path : Separate<["-"], "driver-use-frontend-path">,
  InternalDebugOpt,
  HelpText<"Use the given executable to perform compilations. Arguments can be passed as a ';' separated list">;
```
The `InternalDebugOpt` suggests to me that it is only for testing. Once this (and others to come) fixes have been made, we'll be running swift-driver against all our legacy driver tests.

davidungar:
@cltnschlosser What do you think? This PR (or the equivalent) will need to land before the added test in https://github.com/apple/swift-driver/pull/387 will work. TIA.

cltnschlosser:
I think mocking the target info is fine if this flag is only for testing.

End users have the environment variables they can use if they want to specify a specific frontend, so there will be a workaround for anyone that was using this for something other than testing.

davidungar:
> I think mocking the target info is fine if this flag is only for testing.
> 
> End users have the environment variables they can use if they want to specify a specific frontend, so there will be a workaround for anyone that was using this for something other than testing.

Thanks! Now, if @owenv can weigh in, it would be great to move this forward. TIA

davidungar:
Fixes rdar://71998439

davidungar:
Fixes rdar://71869951

owenv (approved):
Sorry for the holdup!

> It looks to me that apple/swift#34405 does not supplant this PR. (I haven't actually tried it, so please correct me if I'm wrong.) -driver-use-frontend-path is used in lots of tests Driver/Dependencies where the test is checking out the dependency mechanism, for instance malformed-fine.swift. So, I don't see how 34405 would help in these cases.

My understanding of how these tests work is that a test like malformed-fine.swift is effectively using "python update-dependencies.py" as a stand in for the frontend.  apple/swift#34405 was intended to update that script so that in addition to the existing behavior, it could respond to -print-target-info invocations by forwarding them to the real compiler.

That said, I think this solution works too, though IMO we should rework the -driver-use-frontend-path option to separate its two use cases better once we no longer need to worry about compatibility with the old driver.

davidungar:
Thank you very much! No need to apologize. I get eager to move forward, sometimes.
Yes, the python scripts do stand in, you are right. But AFAICT, the dummying up in this PR is needed. Not a bad idea to do the rework someday, though.

PR #391, opened 2020-12-04
[Incremental] When tracing uses, include the implicit use of implementation when interface changes

davidungar:
Dependency graphs contain an implicit link from a given interface to its implementation. Take that into account when tracing uses of defs.


davidungar:
@swift-ci please test

davidungar:
Fixes rdar://71998246

CodaFi (approved):

davidungar:
Thanks, @CodaFi !

PR #393, opened 2020-12-05
[Incremental] Enable testBatchModeContinueAfterErrors & skip frontend query even if no extra arguments to -driver-use-frontend-path

davidungar:
Enable testBatchModeContinueAfterErrors & skip frontend query even if no extra arguments to -driver-use-frontend-path.
This test uncovered a bug, so add the test and and the fix.
Could not enable the test earlier because it depended on another PR.


davidungar:
@swift-ci please test

davidungar:
@swift-ci  please test

davidungar:
@swift-ci please test

PR #397, opened 2020-12-05
Add test for date accuracy.

davidungar:
Required because legacy driver does an exact check of input mod time vs time reconstructed from build record.

### DIFF
--- a/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
+++ b/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
@@ -369,8 +369,8 @@ final class IncrementalCompilationTests: XCTestCase {
   /// Ensure that the mod date of the input comes back exactly the same via the build-record.
   /// Otherwise the up-to-date calculation in `IncrementalCompilationState` will fail.
   func testBuildRecordDateAccuracy() throws {
+    tryInitial(false)
     (1...10).forEach { n in
-      tryInitial(false)
       tryNoChange(true)
     }
   }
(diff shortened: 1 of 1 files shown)


davidungar:
@swift-ci please test

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
```diff
@@ -366,6 +366,16 @@ final class IncrementalCompilationTests: XCTestCase {
     try testIncremental(checkDiagnostics: false)
   }
 
+  /// Ensure that the mod date of the input comes back exactly the same via the build-record.
+  /// Otherwise the up-to-date calculation in `IncrementalCompilationState` will fail.
+  func testBuildRecordDateAccuracy() throws {
+    (1...10) .forEach { n in
```
CodaFi, line 372: Nit: No space please.

CodaFi:
Is there something special about rapidly running clean builds here? It seems like we would want to run one clean build then many incremental null builds to ensure the mod times don't change.

davidungar:
> Is there something special about rapidly running clean builds here? It seems like we would want to run one clean build then many incremental null builds to ensure the mod times don't change.

Can't think of anything, so I changed it. Good point!

davidungar:
@swift-ci please test

davidungar:
Thank you for reviewing this, @CodaFi .

davidungar:
@CodaFi I think I've addressed your concerns. Care to have a quick look?

CodaFi (approved):

PR #399, opened 2020-12-08
Input file list itself should be part of inputs of the compilation job

nkcsgexi:
rdar://71942829

PR #401, opened 2020-12-08
Walk-around a dangling pointer issue

nkcsgexi:
This issue happens in B&I and only reproduces in release builds. It suggests that we should set up testing swift-driver with ASAN turned on.

rdar://71942829


nkcsgexi:
@swift-ci please test

DougGregor (approved):

DougGregor:
Can you link to the PR this is reverting?

artemcm (approved):

nkcsgexi:
This PR also reverts a previous PR: https://github.com/apple/swift-driver/pull/399, because the previous PR doesn't seem to fix the actual issue.

PR #402, opened 2020-12-08
[Incremental] Use same up-to-date test as legacy driver

davidungar:
Legacy driver insists that dates match exactly.  So do that here for compatibility. Required for some lit tests.


davidungar:
@swift-ci please test

davidungar:
@artemcm or @CodaFi : Care to look at this one? I know that the up-to-date test also has to consider the output when there is no input (per the legacy driver), but I'll tackle that in a later PR.

artemcm (approved):

davidungar:
Thank you, @artemcm !

PR #404, opened 2020-12-10
[Incremental] Another attempt to get the driver to honor -driver-use-frontend-path

davidungar:
This one only subverts the frontend target check if the new frontend path is a Python script. That way, the option still should work with swiftpm. Don't merge until the swiftpm unit tests pass with this.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2062,23 +2062,24 @@ extension Driver {
 
     // Find the Swift compiler executable.
     let swiftCompilerPrefixArgs: [String]
+    let hasFrontendBeenRedirectedForTesting: Bool
     if let frontendPath = parsedOptions.getLastArgument(.driverUseFrontendPath){
       var frontendCommandLine =
         frontendPath.asSingle.split(separator: ";").map { String($0) }
       if frontendCommandLine.isEmpty {
         diagnosticsEngine.emit(.error_no_swift_frontend)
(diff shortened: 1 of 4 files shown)


davidungar:
@artemcm What do you think about this approach? @DougGregor Do you want to weigh in? Is there a better alternative?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2062,23 +2062,24 @@ extension Driver {
...
+        let frontendPath = try AbsolutePath(validating: frontendPathString)
+        toolchain.overrideToolPath(.swiftCompiler, path: frontendPath)
         swiftCompilerPrefixArgs = frontendCommandLine
+        hasFrontendBeenRedirectedForTesting = FileType.isFrontendExtensionForTesting(frontendPath.extension)
       }
     } else {
       swiftCompilerPrefixArgs = []
-    }
-    var hasFrontendBeenRedirectedForTesting: Bool {
-      return !swiftCompilerPrefixArgs.isEmpty
```
artemcm, line 2081: I find this to be quite confusing and the new approach is much more concrete so I think it is an improvement.

artemcm (approved):

Sources/SwiftDriver/Driver/Driver.swift
davidungar: Thank you, @artem. Glad you think it's better.

davidungar:
@swift-ci please test

PR #413, opened 2020-12-15
Allow argsHash to be nil when testing. Use dummy value for compilerVersion when testing.

davidungar:
Although both are transmitted through the build record, they are different:
- argsHash for the current run can be determined, but not easily consed up in a test
- compilerVersion cannot be determined (when the frontend path is set to Python for testing), but the tests have no trouble finding one.

So, just as the legacy driver does, allow the "options" field to be omitted from the incoming build record without triggering a mismatch.

And when testing, let the driver set the compilerVersion to a dummy value, which can be recognized when testing.


davidungar:
@swift-ci please test

PR #414, opened 2020-12-15
[Incremental] Changes to pass: `dependencies-preservation-fine.swift`

davidungar:
Two major commits; recommend reviewing each in turn:

1. Allow argsHash to be nil when testing. Use dummy value for compilerVersion when testing

2. Build `IncrementalCompilationState` *after* jobs have been constructed, in order to check output existence.

*ArgHash* and *compilerVersion*:

Although both are transmitted through the build record, they are different:

argsHash for the current run can be determined, but not easily consed up in a test
compilerVersion cannot be determined (when the frontend path is set to Python for testing), but the tests have no trouble finding one.
So, just as the legacy driver does, allow the "options" field to be omitted from the incoming build record without triggering a mismatch.

And when testing, let the driver set the compilerVersion to a dummy value, which can be recognized when testing.


davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
```diff
@@ -20,140 +20,251 @@ public class IncrementalCompilationState {
...
-  public private (set) var mandatoryPreOrCompileJobsInOrder = [Job]()
+  /// All of the pre-compile or compilation job (groups) known to be required (i.e. in 1st wave).
+  /// Already batched, and in order of input files.
+  public let mandatoryJobsInOrder: [Job]
 
-  /// All the  pre- or compilation job (groups) known to be required, which have not finished yet.
-  /// (Changes as jobs complete.)
-  private var unfinishedMandatoryJobs = Set<Job>()
+  /// Sadly, has to be `var` for formBatchedJobs
+  private var driver: Driver
```
artemcm, line 28: I think this makes for an awkward relationship between `Driver` and `IncrementalCompilationState`. 

Looking at the usage here, we query:
`driver.shouldAttemptIncrementalCompilation()` 
`driver.computeReportIncrementalDecision()`
`driver.getBuildInfo()`
In the initializer.
And elsewhere we use the driver's properties such as `fileSystem`, `inputFiles`, `parsedOptions` and `compilerMode`. 

It seems to me that we should be able to package the above into an incrementalCompilationState context type of some kind. 

The one piece of functionality this doesn't account for is `formBatchedJobs` which is now needed both here and in the driver, but it looks like it can be factored out into a standalone utility used by both? 

Another thought is that `IncrementalCompilationState` does two things, it tracks state and performs scheduling. Perhaps separating the two concerns would make for a cleaner distinction of which pieces of the driver are required for each? 

I'm curious what you think.

davidungar:
@swift-ci please test

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
artemcm: David and I had a chat about this and the remedy to this structure is a bit broader in scope with the overall driver design. For now, this is much needed progress in getting us to incremental build parity.

artemcm (approved):

davidungar:
@swift-ci please test

davidungar:
Thank you, @artemcm !

davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

Sources/SwiftDriver/Incremental Compilation/IncrementalCompilationState.swift
davidungar: Thanks, @artemcm! Yes, the big push right now is for parity via passing the legacy tests.

davidungar:
@swift-ci please test

davidungar:
rdar://72508532

davidungar:
rdar://71869992

davidungar:
rdar://72157018

davidungar:
rdar://72235335

davidungar:
rdar://72508634

davidungar:
rdar://72508725

davidungar:
rdar://72508756

PR #416, opened 2020-12-15
Remove redundant clang '-target' flag from Unix Toolchain

buttaface:
This is a port of the change to the C++ Driver from apple/swift#34978 to Swift.

All the tests in this repo passed with this pull when run on linux x86_64. I don't know if this Driver is periodically run against the upstream testsuite, in which case the upstream pull will have to be merged first.

@cltnschlosser, this pull invalidates much of #370, maybe you can repurpose that by reworking the tests and adding the Android sanitizer change.

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -61,11 +61,6 @@ extension GenericUnixToolchain {
       commandLine.appendFlag("-shared")
       fallthrough
     case .executable:
-      if !targetTriple.triple.isEmpty {
-        commandLine.appendFlag("-target")
-        commandLine.appendFlag(targetTriple.triple)
-      }
-
         // Select the linker to use.
       var linker: String?
       if let arg = parsedOptions.getLastArgument(.useLd) {


cltnschlosser:
Okay, I’ll close that PR for now

artemcm (approved):

artemcm:
@swift-ci please test

artemcm:
> I don't know if this Driver is periodically run against the upstream testsuite, in which case the upstream pull will have to be merged first.

Yes, we track the upstream test suite failures, albeit somewhat informally. The upstream change should go in first.

buttaface:
Actually, thinking about it some more, this Swift Driver currently fails the tests in the upstream pull, as this Driver doesn't have the logic from #370, so this pull won't change those test results. If however, you don't want to change the behavior of passing the `-target` flag twice without first changing it upstream in the C++ driver, even though that is a mistake, you'd have to still wait for the upstream pull.

buttaface:
The upstream pull was just merged. Note that if you merge this before the 5.4 branch, this repo's 5.4 branch and the upstream C++ driver will not do the same for this flag. I don't think that matters since the clang flag we're removing is currently ignored both upstream and here, but just letting you know.

artemcm:
@swift-ci please test

buttaface:
Alright, release/5.4 was forked, this should be good to go into trunk, as it gets this driver to match the trunk C++ driver.

artemcm:
@swift-ci please test

PR #417, opened 2020-12-16
[Incremental] Fix race on `finishedJobResults`

davidungar:
The build engine can exit before all jobs have registered their results in `finishedJobResults`. Then when the build record is written, the read of these results races with the writes. Protect these with a semaphore.


davidungar:
rdar://problem/72341819

davidungar:
@swift-ci please test

nkcsgexi:
This fixes the data race. Does it guarantee that build records will be written after all Job finishes?

davidungar:
> This fixes the data race. Does it guarantee that build records will be written after all Job finishes?

Not sure I follow the question. Are you worried that the build record might never be written?

davidungar:
The driver passes the XC tests, some of which count on the build record getting written, so I think it should work in that respect. Are you seeing something I'm not seeing?

PR #418, opened 2020-12-18
Match the legacy driver's output file map behavior when there are no outputs

owenv:
- Adjust the supplementary output file map calculation to better match the C++ driver's WMO behavior
- Always include either primary inputs or the first input in the supplementary output file map, even if there aren't any corresponding outputs

Fixes Driver/supplementary-filelist-empty.swift and Driver/modulewrap.swift

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -378,15 +378,28 @@ extension Driver {
                                       output: TypedVirtualPath(file: tracePath, type: .moduleTrace)))
     }
 
-    let allInputsCount = primaryInputs.count + flaggedInputOutputPairs.count
+    // This calculation treats all the files in a WMO compile as a single input
+    let allInputsCount = max(primaryInputs.count, 1)
     // Question: outputs.count > fileListThreshold makes sense, but c++ does the following:
     if allInputsCount * FileType.allCases.count > fileListThreshold {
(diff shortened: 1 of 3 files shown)


owenv:
@swift-ci please test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -378,15 +378,28 @@ extension Driver {
                                       output: TypedVirtualPath(file: tracePath, type: .moduleTrace)))
     }
 
-    let allInputsCount = primaryInputs.count + flaggedInputOutputPairs.count
+    // This calculation treats all the files in a WMO compile as a single input
+    let allInputsCount = max(primaryInputs.count, 1)
```
cltnschlosser, line 382: This seems weird, because it will reduce how often we hit the fileListThreshold. But if there were no regressions in lit tests from this change, I guess it's fine.

cltnschlosser (approved):
Looks good

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
owenv: yeah, I was just able to verify this doesn't cause any regressions thankfully. I think it's a little closer to the calculation used in the old driver, which counts the InputActions of the compile job. That way, by multiplying by the number of file types, we approximate the maximum number of outputs in the worst case, even though we could estimate this better if we weren't worried about compatibility.

owenv:
@swift-ci please test

PR #421, opened 2020-12-20
[Explicit Modules] Support using -experimental-explicit-module-build with immediate mode

owenv:
For example, `swift -experimental-explicit-module-build hello.swift` will now prebuild module dependencies, and then run the interpret job in-place at the end. On its own this isn't that useful, but I'm using it to explore the idea of `@package` imports in scripts a bit more, and I think the generalization here is pretty harmless

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -871,32 +871,50 @@ extension Driver {
       return
     }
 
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci please test

artemcm (approved):
The generalization here is tidy, indeed. LGTM!

owenv:
@swift-ci please test

PR #424, opened 2020-12-31
Add support for the -update-code flag used to run the migrator

owenv:
This doesn't pass the corresponding lit test yet because the flags are in a different order, but everything else seems to match.

It's worth pointing out that I deleted the `-update-code` check in `determinePrimaryOutputs` because the condition would never be met (-update-code isn't in the modes option group and the same problem occurs in the old driver). This seems consistent with the intent of the original driver's implementation which treats the migration as a parallel output to the compile, but might be worth double checking for correctness because I don't think the migrator is used much outside of Xcode.

Also, please feel free to ignore this until after the holidays! I'm just posting it early to run CI tests

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -1335,10 +1335,6 @@ extension Driver {
       case .indexFile:
         compilerOutputType = .indexData
 
-      case .updateCode:
-        compilerOutputType = .remap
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci please test

artemcm (approved):
This looks good, thank you!

owenv:
@swift-ci please test

PR #426, opened 2021-01-07
Inform finished job only when the job has been started

nkcsgexi:
Otherwise the build system may complain about malformed output.

rdar://72830153


nkcsgexi:
@swift-ci please test

artemcm (approved):

cltnschlosser:
Is it possible for pids to be reused? Or is that not a concern?

artemcm:
> Is it possible for pids to be reused? Or is that not a concern?

Good point, I think that's a valid concern. Perhaps we should instead treat this set as `knownInProgress` and remove entries when we emit the jobFinished signal.

PR #427, opened 2021-01-08
Rely on `libSwiftScan` for dependency scanning actions. 

artemcm:
This PR changes the dependency scanning action to rely on the newly-introduced `libSwiftScan` library, instead of making calls to `swift-frontend -scan-dependencies` and parsing back their JSON output.

- Adds target `CSwiftScan` that contains the C header which captures all the required symbols from `libSwiftScan`.
    - Adds a `SwiftScan` subdirectory in the `SwiftDriver` target which contains code responsible for interactiosn with `libSwiftScan`.
- Adds a  library instance wrapper (`SwiftScan`) which takes care of initialization, shutdown, and dispatching dependency scanning queries to a library instance.
- Adds a collection of utilities (in `DependencyGraphBuilder.swift`) to translate dependency scanning query inputs into the C library expected format and translate the dependency scanning binary output format into the format the driver understands (`InterModuleDependencyGraph`).

Merging this is dependent on https://github.com/apple/swift/pull/34786 being merged and making it into a snapshot toolchain.

Additional cleanup will follow this once clients (SwiftPM) switch to the new API. For example, the notion of a placeholder module will go away completely so all of the placeholder resolution code will become unnecessary.


Sources/SwiftDriver/Execution/ArgsResolver.swift
```diff
@@ -93,59 +97,61 @@ public final class ArgsResolver {
...
         for path in contents {
-          try! out <<< unsafeResolve(path: path) <<< "\n"
+          try! out <<< unsafeResolve(path: path, quotePaths: quotePaths) <<< "\n"
         }
       }
     }
   }
 
-  private func createFileList(path: VirtualPath, outputFileMap: OutputFileMap) throws {
+  private func createFileList(path: VirtualPath, outputFileMap: OutputFileMap, quotePaths: Bool)
```
cltnschlosser, line 129: This function probably doesn't want `quotePaths` because `quoteAndEscape` already adds quotes. Wouldn't want double quoting.

This line below should probably be changed to just `quotePaths: false`
```swift
let inputNode = Node.scalar(Node.Scalar(try! unsafeResolve(path: path, quotePaths: quotePaths),
                                            Tag(.str), .doubleQuoted))
```

artemcm: Good catch, thank you.

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

nkcsgexi:
So excited to see libSwiftScanner is used here to enhance build performance! Is it difficult to also preserve the old interaction form of JSON files? One benefit of that is we can get a peek inside the scanning process for debugging purposes.

artemcm:
> So excited to see libSwiftScanner is used here to enhance build performance! Is it difficult to also preserve the old interaction form of JSON files? One benefit of that is we can get a peek inside the scanning process for debugging purposes.

No, shouldn't be difficult. I'll add an option for this.

artemcm:
@swift-ci please test

artemcm:
@swift-ci  please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@nkcsgexi ping

nkcsgexi (approved):

artemcm:
@swift-ci please test

nkcsgexi:
🥳

PR #428, opened 2021-01-08
[ExplicitModule] when using driver as an executable, differentiate PCM dependency names for each target

nkcsgexi:
We need this to enable explicit modules in the driver-as-executable mode. For instance,
we have two Swift targets A and B, where A depends on X.pcm which in turn depends on Y.pcm, and
B only depends on Y.pcm. In the driver-as-executable mode, the build system isn't aware
of the shared dependency of Y.pcm so it will be generated multiple times. If all these Y.pcm
share the same name, X.pcm may fail to be loaded because its dependency Y.pcm could have a changed mod time.

We only differentiate these PCM names in the non-integrated mode due to the lacking of
inter-module planning.

rdar://72830153

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -104,6 +104,9 @@ public struct Driver {
   /// it should be identical to the real environment.
   public let env: [String: String]
 
+  /// Whether we are using the driver as the integrated driver via libSwiftDriver
+  public let integratedDriver: Bool
+
   /// The file system which we should interact with.
   let fileSystem: FileSystem
 
@@ -322,6 +325,7 @@ public struct Driver {
     diagnosticsEngine: DiagnosticsEngine = DiagnosticsEngine(handlers: [Driver.stderrDiagnosticsHandler]),
     fileSystem: FileSystem = localFileSystem,
     executor: DriverExecutor,
+    integratedDriver: Bool = true,
     // FIXME: Duplication with externalBuildArtifacts and externalTargetModulePathMap
     // is a temporary backwards-compatibility shim to help transition SwiftPM to the new API
     externalBuildArtifacts: ExternalBuildArtifacts? = nil,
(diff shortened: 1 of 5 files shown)


nkcsgexi:
@artemcm IIUC, this won't regress the build performance of explicit modules in SwiftPM. Could you confirm?

nkcsgexi:
@swift-ci please test

artemcm:
> @artemcm IIUC, this won't regress the build performance of explicit modules in SwiftPM. Could you confirm?

I suspect that it will. Because currently we will come up with an ahead-of-time plan to build Y.pcm several times, and then when SwiftPM lowers our `Job`s into LLBuild rules, it will de-duplicate them, resulting in Y.pcm being built once (versioned compilation aside). But if we give them different names, then it will not be able to de-duplicate them and we will actually repeat an identical build action several times just to have these modules with different names. We could do an experiment to confirm, though.

nkcsgexi:
hmm, Will the flag `integratedDriver` be true in the SwiftPM case? Notice we only differentiate those names when `integratedDriver` is false, that's when running swift-driver as a standalone executable.

artemcm:
> hmm, Will the flag `integratedDriver` be true in the SwiftPM case? Notice we only differentiate those names when `integratedDriver` is false, that's when running swift-driver as a standalone executable.

Good point, yeah then this should not affect performance in this use-case. 👍

nkcsgexi:
> > hmm, Will the flag `integratedDriver` be true in the SwiftPM case? Notice we only differentiate those names when `integratedDriver` is false, that's when running swift-driver as a standalone executable.
> 
> Good point, yeah then this should not affect performance in this use-case. 👍

Thanks for confirming! When enabling explicit modules for swift-driver-as-executable, changing the PCMs names won't affect build performance because these PCMs are built multiple times disregardingly IIUC.

artemcm (approved):

nkcsgexi:
@swift-ci please test

PR #429, opened 2021-01-11
Delete temporaries after running unless -save-temps is passed

owenv:
Unlike the old driver, all of the temporaries are created in a subdirectory, so we can just delete that instead of tracking each temporary file individually. We should also save temps if any of the jobs signals - I left that as a todo in this PR because it will require a bit more refactoring to make that information available to the driver.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -167,7 +167,7 @@ public struct Driver {
   let shouldUseInputFileList: Bool
 
   /// VirtualPath for shared all sources file list. `nil` if unused.
-  let allSourcesFileList: VirtualPath?
+  @_spi(Testing) public let allSourcesFileList: VirtualPath?
 
   /// The mode in which the compiler will execute.
   @_spi(Testing) public let compilerMode: CompilerMode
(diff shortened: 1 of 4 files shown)


owenv:
@swift-ci please test

owenv:
@swift-ci please test

cltnschlosser:
This is an improvement, but I think the old driver had some weird logic where only certain temporaries were saved when a signal was received. Maybe it doesn’t matter and keeping everything is fine.

owenv:
I finally got around to revisiting this, and it's now setup to preserve temporary files if any of the jobs crashes. @cltnschlosser I decided to just keep everything in the event of a signal, at least for now. Based on https://github.com/apple/swift/commit/a875d8c7c77b092a4a3ba9c80b1099f342d7ae6c it looks like the old driver's implementation was intended to save disk space by always removing larger outputs. Compiler crashes are a lot rarer now than when that was written ~5 years ago, so I think this approach is _probably_ fine.

owenv:
@swift-ci please test

cltnschlosser (approved):

artemcm (approved):
This looks good, thank you @owenv!

owenv:
@swift-ci please test

PR #431, opened 2021-01-13
[Build Script Helper] Build for host platform only, by default.

artemcm:
Not all toolchains assume that we're also targeting arm64 on macOS (e.g. S4TF). 
Building for the host platform only will also match what we do in some other, similar components.
Previous behavior of building universal binaries on Darwin is still available using:
`--cross-compile-hosts x86_64-apple-macosx10.15 arm64-apple-macosx10.15`

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -17,7 +17,11 @@ if platform.system() == 'Darwin':
 else:
     shared_lib_ext = '.so'
 macos_deployment_target = '10.15'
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci please test

artemcm:
Cross-repository tests in https://github.com/apple/swift/pull/35419 are green.

nkcsgexi (approved):

artemcm:
Thanks @nkcsgexi.

PR #433, opened 2021-01-17
Improve support for subcommands help in swift-help

owenv:
This PR has a couple of changes to the help tool:
- The subcommand listings are reformatted a bit, and I added a note to "Use "swift help subcommand" for more information about a subcommand". This is inspired by Go's help page, which also supports more general help topics. I think we could do something similar in a follow-up to add the kind of driver documentation requested in https://bugs.swift.org/browse/SR-7704.
- "swift help build/test/package/run" now forward to the help pages for each subcommand

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -69,7 +69,7 @@ let package = Package(
     /// The help executable.
     .target(
       name: "swift-help",
-      dependencies: ["SwiftOptions", "ArgumentParser"]),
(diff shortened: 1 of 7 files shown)


owenv:
@swift-ci please test

artemcm (approved):
This is a nice improvement, thank you @owenv.

PR #434, opened 2021-01-19
OutputFileMap: use .swift file path as the key for .o file when .bc file is an intermeidate output

nkcsgexi:
The build system always uses .swift file for all output kinds. This causes an issue when we are using
the .bc file path as the key to get the .o file path.

rdar://73358362


nkcsgexi:
@swift-ci please test

artemcm (approved):
LGTM!

PR #435, opened 2021-01-19
[Explicit Module Builds] Refactor and simplify explicit dependency job-generation code

artemcm:
The existing implementation of this, while correct, is a bit more complex than it needs to be.
It is also recursive, which is not ideal because we'd like to be able to handle arbitrarily-large build graphs here.
At the core of this process, currently, is a way to compute transitive closure for the dependency graph that is done via DFS from every node (module) in the graph.
The acyclic property of the dependency graph lets us do better.
The new algorithm for computing transitive closure on the dependency graph is:
```
for each v ∈ V {
    T(v) = { v }
}
for v ∈ V in reverse topological order {
    for each (v, w) ∈ E {
         T(v) = T(v) ∪ T(w)
    }
}
```
This PR:
- Adds a `reachabilityMap` data-structure to `ExplicitDependencyBuildPlanner` computed using this algorithm at the planner construction.
- Refactors `generateExplicitModuleDependenciesBuildJobs` to use the `reachabilityMap` to get a module's explicit dependencies when generating jobs, which can now be generated without recursing.
- Eliminates all mutable state in the `ExplicitModuleBuildPlanner` (`swiftModuleBuildCache` and `clangTargetModuleBuildCache` are no longer necessary).


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #437, opened 2021-01-20
Decode path fields in `InterModuleDependencyGraph` as VirtualPaths

artemcm:
In the process of generating explicit dependency jobs, we end up spending a lot of time constructing `VirtualPath` objects from strings contained in the dependency graph, (It involves querying whether or not this string constitutes an absolute path, and breaking it down into components) which all add up to a large amount of string manipulation.

Instead, decode these strings into `VirtualPath` values directly, using the `TextualVirtualPath` wrapper.


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #438, opened 2021-01-21
Exit with 0 when swift-driver invocation is interrupted

nkcsgexi:
The build system may interrupt swift-driver invocation if some errors are hit in other targets.
If these interrupted swift-driver processes return a non-zero value, the build
log will show a bunch of red-herring error-messages.

rdar://73456559


nkcsgexi:
@swift-ci please test

artemcm (approved):

PR #439, opened 2021-01-23
[Incremental] Move subtraction to outer function to cope with multiple input jobs.

davidungar:
Fix assertion failure by moving subtraction of already-compiled inputs to the calling function so that if job compiling x.swift requires y.swift be compiled, and job compiling y.swift requires x.swift be compiled, x.swift and y.swift are subtracted out from jobs needing to be compiled. The bug shows up when testing with -j1.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -565,6 +565,7 @@ extension IncrementalCompilationState {
           return Array(skippedCompileGroups.keys)
         }
       )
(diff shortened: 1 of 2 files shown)


davidungar:
@swift-ci please test

cltnschlosser (approved):
Can you write an xctest case that uses -j1 and would catch this bug?

davidungar:
> Can you write an xctest case that uses -j1 and would catch this bug?

Thank you, Colton! It's a good suggestion, but I haven't figured out how to do that, yet. I'll think about it for a bit and try to do so.

davidungar:
I'm going to go ahead and merge. There's an assertion there, although a test would be even better.

PR #443, opened 2021-01-24
[NFC] Simplify Parts of the Change Computation's Interface

CodaFi:
- Add a new type representing changed inputs
- Break down the `forEach`-style APIs into explicit sequences and clean up their call sites.
- Break down the karnaugh map in `computeSpeculativeInputs`

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -192,13 +192,13 @@ extension ModuleDependencyGraph {
 
   /*@_spi(Testing)*/ public func untracedDependents(
     of externalSwiftDeps: ExternalDependency
-  ) -> [(ModuleDependencyGraph.Node, ModuleDependencyGraph.SwiftDeps)] {
+  ) -> [ModuleDependencyGraph.Node] {
     // These nodes will depend on the *interface* of the external Decl.
     let key = DependencyKey(interfaceFor: externalSwiftDeps)
     let node = Node(key: key, fingerprint: nil, swiftDeps: nil)
     return nodeFinder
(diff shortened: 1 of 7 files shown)


CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -190,18 +190,15 @@ extension ModuleDependencyGraph {
...
     let key = DependencyKey(interfaceFor: externalSwiftDeps)
     let node = Node(key: key, fingerprint: nil, swiftDeps: nil)
-    nodeFinder.forEachUseInOrder(of: node) { use, useSwiftDeps in
-      if isUntraced(use) {
-        fn(useSwiftDeps)
-      }
-    }
+    return nodeFinder
+      .orderedUses(of: node)
+      .filter({ use in isUntraced(use) })
```
cltnschlosser, line 201: Any significant benefit to reworking this to filter before the sort? Or does this not scale too large to really matter?

cltnschlosser (approved):

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
CodaFi: Good question. In my experience, tracing sets for medium-sized projects are on the order of 10-20 nodes.

CodaFi: Being traced is also an unlikely predicate for a given node among that set. I would expect we wouldn’t see much benefit from pruning the search space up front with that in mind, but performance is always something one wants numbers and not opinions for.

PR #444, opened 2021-01-25
[NFC, Incremental] Use let's instead of var's for the ChangedInput struct.

davidungar:
@CodaFi Introduced a nice new type, but it doesn't need to be mutable, so use `let`s.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -342,12 +342,12 @@ extension IncrementalCompilationState {
   /// changed in a way that requires an incremental rebuild.
   struct ChangedInput {
     /// The path to the input file.
-    var filePath: TypedVirtualPath
+    let filePath: TypedVirtualPath
     /// The status of the input file.
-    var status: InputInfo.Status
+    let status: InputInfo.Status
     /// If `true`, the modification time of this input matches the modification
     /// time recorded from the prior build in the build record.
-    var datesMatch: Bool
+    let datesMatch: Bool
   }
 
   /// Find the inputs that have changed since last compilation, or were marked as needed a build


davidungar:
@swift-ci please test

CodaFi:
> But it doesn't need to be mutable

That's not what `var` and `let` signal primarily in this language. In particularly, `let` is semantically stronger than something like `const`. It is an anti-pattern to use `let` for struct fields that are not semantically constants in most cases. All it does is make it harder to use - and in particular initialize - instances of this value in a way that does not go through the memberwise initializer. I am neutral on its removal given that we do not have code mutating through these members, but for the future this kind of refactoring is not always a positive.

CodaFi:
`let` in structs is almost an attractive nuisance. An assignment to a struct field declared as a `var` behaves semantically identically to a complete re-creation of the struct and its fields but with the point-mutation applied, which is what this kind of refactoring inspires users to type explicitly when they inevitably discover they don't have access to the setter. You don't really... win anything.

CodaFi (approved):

CodaFi:
⛵

davidungar:
Thank you, @CodaFi .

CodaFi:
If some nice contributor would like to take up the charge of refactoring this type to hide its memberwise init, I would greatly appreciate it.  Ideally, stat'ing inputs and comparing them against the build record would be a service provided by some aggregate with access to the file system that would be the only thing that could construct instances of this value.

@owenv @cltnschlosser what do you think?

cltnschlosser:
> If some nice contributor would like to take up the charge of refactoring this type to hide its memberwise init, I would greatly appreciate it. Ideally, stat'ing inputs and comparing them against the build record would be a service provided by some aggregate with access to the file system that would be the only thing that could construct instances of this value.
> 
> @owenv @cltnschlosser what do you think?

Is the idea just to make an explicit fileprivate initializer?

PR #445, opened 2021-01-26
Propagate Experimental Cross Module Incremental Build Flag

CodaFi:
Match the legacy driver and propagate this flag to the merge-modules job so embedded incremental information winds up in swift modules.

-------

Cross Module Incremental Builds Tasks

- [x] Teach the Driver to propagate the flag
- [ ] Integrate swiftdeps from swift modules
- [ ] Propagate bitstream writer infrastructure into Driver
- [ ] Write out the module dependency graph
- [ ] Integrate module dependency graph instead of swiftdeps for priors

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
@@ -139,6 +139,18 @@ public struct DependencyKey: Hashable, CustomStringConvertible {
     /// The `context` field corresponds to the mangled name of the type. The
     /// `name` field corresponds to the *unmangled* name of the member.
     case member(context: String, name: String)
+    /// A dependency that resides outside of the module being built, but which
+    /// has integrated dependency information in a format the driver
+    /// understands.
+    ///
+    /// These dependencies correspond to Swift modules that were built with
+    /// `-enable-experimental-cross-module-incremental-build`. These modules
+    /// contain a special section with swiftdeps information for the module
+    /// in it.
+    ///
+    /// The full path to the external dependency as seen by the frontend is
+    /// available from this node.
+    case incrementalExternalDependency(ExternalDependency)
 
     var externalDependency: ExternalDependency? {
       switch self {
@@ -164,6 +176,8 @@ public struct DependencyKey: Hashable, CustomStringConvertible {
         return "module '\(externalDependency)'"
       case let .sourceFileProvide(name: name):
         return "source file \((try? VirtualPath(path: name).basename) ?? name)"
+      case let .incrementalExternalDependency(externalDependency):
+        return "incremental module '\(externalDependency)'"
       }
     }
   }
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -202,7 +202,7 @@ extension ModuleDependencyGraph {
   }
 }
 fileprivate extension DependencyKey {
-  init(interfaceFor dep: ExternalDependency ) {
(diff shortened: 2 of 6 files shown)


CodaFi:
@swift-ci test

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
```diff
@@ -139,6 +139,18 @@ public struct DependencyKey: Hashable, CustomStringConvertible {
...
+    /// understands.
+    ///
+    /// These dependencies correspond to Swift modules that were built with
+    /// `-enable-experimental-cross-module-incremental-build`. These modules
+    /// contain a special section with swiftdeps information for the module
+    /// in it.
+    ///
+    /// The full path to the external dependency as seen by the frontend is
+    /// available from this node.
+    case incrementalExternalDependency(ExternalDependency)
```
davidungar, line 153: Nice explanation

davidungar: Here's a thought for a possible future design: Suppose every dependency has three components, not two:
a module, a context, and a name. Module can be empty (nil) if the def is in this module, context can be empty (nil) if the name is top-level. And we don't need multiple dependency types.... Hmm....... Should an inner type have N names, for the path into that type? Is the similar to the ideas you were floating a while back?
(Just a thought for the future...)

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
```diff
@@ -164,6 +176,8 @@ public struct DependencyKey: Hashable, CustomStringConvertible {
         return "module '\(externalDependency)'"
       case let .sourceFileProvide(name: name):
         return "source file \((try? VirtualPath(path: name).basename) ?? name)"
+      case let .incrementalExternalDependency(externalDependency):
+        return "incremental module '\(externalDependency)'"
```
davidungar, line 180: I understand how the name came about, but seeing this stirs up a thought that it might be misconstrued. It reads as if the *module* is incremental. Would "externalIncrementalDependency" be clearer?

davidungar (commented):
I'm torn with respect to approving this change:
On one hand, it's good enough.
On the other hand, I'd feel significantly better if this PR included some unit test additions to test out the new kind of dependency.

(As an aside, something bothers me about the name "incremental external dependency": All the other dependencies are incremental, so "incremental" seems to not belong, except as an historical note. But future maintainers won't be thinking that way. But I don't have a better alternative, so maybe it's fine. Now would be the time to change it. On the other hand, maybe someday we'll get rid of these dependency types, and just use stable paths. On the other other hand, sometimes stuff can live forever in these code bases...)

davidungar (approved):
Seeing that my non-approval would block you, I decided to approve. But please, consider adding a unit test in a subsequent PR.

CodaFi:
⛵

CodaFi:
> I'd feel significantly better if this PR included some unit test additions to test out the new kind of dependency.

There _are_ tests for what functionality we can observe at this time, which is the propagation of this flag. Without the integration code (which requires the bitstream writer I have just submitted to TSC) I cannot unit test any further behaviors of the system because... well, there aren't many other new behaviors here.

davidungar:
> > I'd feel significantly better if this PR included some unit test additions to test out the new kind of dependency.
> 
> There _are_ tests for what functionality we can observe at this time, which is the propagation of this flag. Without the integration code (which requires the bitstream writer I have just submitted to TSC) I cannot unit test any further behaviors of the system because... well, there aren't many other new behaviors here.

Thanks for your answer. Good point. Maybe as the new behaviors are added there will be tests that could be added to `FineGrainedDependencyTests.cpp`. Those tests were very helpful to me in the past for finding bugs in the code I was writing.

PR #447, opened 2021-01-28
Add Round-Trip Serialization For ModuleDependencyGraph

CodaFi:
Use bitstream to read and write ModuleDependencyGraph instances out to
disk. The idea is that eventually the Driver can emit its dependency
graph as a supplemental output (possibly adjacent to the build record)
and recover the state it was last in by reading that file. This is as
opposed to the status quo where O(N) swiftdeps files are read and
integrated at startup. Whether this is a net win has yet to be seen.

Also, we're using bitstream because it's convenient. It would be prudent
to explore alternative serialization formats at some point.

rdar://73517080

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -24,7 +24,7 @@
         "repositoryURL": "https://github.com/apple/swift-tools-support-core.git",
         "state": {
           "branch": "main",
-          "revision": "c98cfc216d22798dce7ce7b9cc171565371e967e",
+          "revision": "395e7fb3d6149dc7a999342f23cfefa1ada56264",
           "version": null
         }
       },
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 import Foundation
 import TSCBasic
+import TSCUtility
 import SwiftOptions
 
 
@@ -175,6 +176,7 @@ extension ModuleDependencyGraph {
                          fileSystem: fileSystem)
       .map {
         findSwiftDepsToRecompileWhenNodesChange($0)
+          .map { sourceSwiftDepsMap[$0] }
       }
   }
 }
@@ -185,6 +187,7 @@ extension ModuleDependencyGraph {
   /*@_spi(Testing)*/ public func findSwiftDepsToRecompileWhenNodesChange<Nodes: Sequence>(
     _ nodes: Nodes
   ) -> Set<SwiftDeps>
+    where Nodes.Element == Node
   {
     let affectedNodes = Tracer.findPreviouslyUntracedUsesOf(
       defs: nodes,
@@ -223,6 +226,7 @@ extension ModuleDependencyGraph {
     tracedNodes.insert(n)
   }
   func ensureGraphWillRetraceDependents<Nodes: Sequence>(of nodes: Nodes)
+    where Nodes.Element == Node
   {
     nodes.forEach { tracedNodes.remove($0) }
   }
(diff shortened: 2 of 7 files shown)


CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,499 @@ extension ModuleDependencyGraph {
...
+
+    private func writeMetadata() {
+      self.stream.writeRecord(self.abbreviations[.metadata]!, {
+        $0.append(RecordID.metadata)
+        // Major version
+        $0.append(1 as UInt32)
+        // Minor version
+        $0.append(0 as UInt32)
+      },
+      blob: "") // FIXME: Plumb the frontend compiler version here.
```
CodaFi, line 512: @davidungar What's the best way to get this data?

cltnschlosser: Probably `FrontendTargetInfo.compilerVersion` which is stored as an instance on Driver. It also looks like the value is propagated to BuildRecordInfo as `actualSwiftVersion`. Not sure if that's unrelated to this or not, haven't looked at this PR.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,513 @@ extension ModuleDependencyGraph {
...
+        reporter: IncrementalCompilationState.Reporter?
+      ) {
+        self.graph = ModuleDependencyGraph(diagnosticEngine: diagnosticEngine,
+                                           reporter: reporter,
+                                           emitDependencyDotFileAfterEveryImport: false,
+                                           verifyDependencyGraphAfterEveryImport: false)
+      }
+
+      func validate(signature: Bitcode.Signature) throws {
+        guard signature == .init(string: ModuleDependencyGraph.signature) else { throw ReadError.badMagic }
```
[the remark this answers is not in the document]
davidungar: But maybe the filename is supplied elsewhere, where the exception is caught?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,513 @@ extension ModuleDependencyGraph {
...
+  ///
+  /// - Parameters:
+  ///   - path: The absolute path to the file to be read.
+  ///   - fileSystem: The file system on which to search.
+  ///   - diagnosticEngine: The diagnostics engine.
+  ///   - reporter: An optional reporter used to log information about
+  /// - Throws: An error describing any failures to read the graph from the given file.
+  /// - Returns: A fully deserialized ModuleDependencyGraph.
+  static func read(
+    from path: AbsolutePath,
```
davidungar, line 303: I'm not sure the path should always be absolute. I've run into this elsewhere in the driver, where I had to make changes to accommodate relative paths for testing. Not a blocker to landing this PR; we can always fix it if need be. Just FYI.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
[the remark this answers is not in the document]
davidungar: In the case of a bad magic number, I would include both the desired and the actual magic numbers in the exception.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,513 @@ extension ModuleDependencyGraph {
...
+    on fileSystem: FileSystem,
+    diagnosticEngine: DiagnosticsEngine,
+    reporter: IncrementalCompilationState.Reporter?
+  ) throws -> ModuleDependencyGraph {
+    let data = try fileSystem.readFileContents(path)
+
+    struct Visitor: BitstreamVisitor {
```
davidungar, line 310: This `Visitor` struct is so long that it interrupts the flow of reading the `read` function. I realize that it is only used by this function and so wants to be scoped inside of the function. I guess my eyes would have an easier time were it outside the function. Just a thought for you to consider.

davidungar: Also, if moved outside, a more intension-revealing name might be nice. Like `ModuleDependencyGraphStreamVisitor` (I know that's a long name).

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,513 @@ extension ModuleDependencyGraph {
...
+          // If we've already read metadata, this is an unexpected duplicate.
+          guard self.majorVersion == nil, self.minorVersion == nil, self.compilerVersionString == nil else {
+            throw ReadError.unexpectedMetadataRecord
+          }
+          guard record.fields.count == 2,
+                case .blob(let compilerVersionBlob) = record.payload,
+                let compilerVersionString = String(data: compilerVersionBlob, encoding: .utf8)
+          else { throw ReadError.malformedMetadataRecord }
+
+          self.majorVersion = record.fields[0]
```
davidungar, line 366: These field indices--should they be named constants somewhere--raw values of an enum?
Then could they be shared with the code to write out the graph in the driver? (to ensure consistency)

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,513 @@ extension ModuleDependencyGraph {
...
+      // if possible. There's no real reason to use `Data` in this API.
+      let baseAddr = UnsafeMutableRawPointer(mutating: buf.baseAddress!)
+      let data = Data(bytesNoCopy: baseAddr, count: buf.count, deallocator: .none)
+      var visitor = Visitor(diagnosticEngine: diagnosticEngine,
+                            reporter: reporter)
+      try Bitcode.read(stream: data, using: &visitor)
+      guard let major = visitor.majorVersion,
+            let minor = visitor.minorVersion,
+            visitor.compilerVersionString != nil,
+            (major, minor) == (1, 0)
```
davidungar, line 425: same with these numbers, someday ought to be named constants

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,513 @@ extension ModuleDependencyGraph {
...
+
+      self.stream.writeRecord(Bitstream.BlockInfoCode.blockName) { buffer in
+        buffer.append(name)
+      }
+    }
+
+    private func emitRecordID(_ id: RecordID, named name: String) {
+      self.stream.writeRecord(Bitstream.BlockInfoCode.setRecordName) {
+        $0.append(id)
+        $0.append(name)
```
davidungar, line 501: I see why the constants above aren't named and shared. Would be nice someday to use a functional interface to ensure consistency.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,513 @@ extension ModuleDependencyGraph {
...
+      }
+    }
+
+    private func writeBlockInfoBlock() {
+      self.stream.writeBlockInfoBlock {
+        self.emitBlockID(.firstApplicationID, "RECORD_BLOCK")
+        self.emitRecordID(.metadata, named: "METADATA")
+        self.emitRecordID(.moduleDepGraphNode, named: "MODULE_DEP_GRAPH_NODE")
+        self.emitRecordID(.fingerprintNode, named: "FINGERPRINT_NODE")
+        self.emitRecordID(.identifierNode, named: "IDENTIFIER_NODE")
```
davidungar, line 511: Same thought about these strings--should someday be pulled out of the code here.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,513 @@ extension ModuleDependencyGraph {
...
+          // dependency decl aspect discriminator
+          .fixed(bitWidth: 1),
+          // dependency context
+          .vbr(chunkBitWidth: 13),
+          // dependency name
+          .vbr(chunkBitWidth: 13),
+          // swiftdeps?
+          .fixed(bitWidth: 1),
+          // swiftdeps path
+          .vbr(chunkBitWidth: 13),
```
davidungar, line 603: I'd love someday to see checks that things fit in the widths.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -262,3 +263,513 @@ extension ModuleDependencyGraph {
...
+          // identifier data
+          .blob
+        ])
+
+        serializer.writeMetadata()
+
+        serializer.writeStrings(in: graph)
+
+        graph.nodeFinder.forEachNode { node in
+          serializer.stream.writeRecord(serializer.abbreviations[.moduleDepGraphNode]!) {
```
davidungar, line 621: someday throw errors instead of crashing

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
```diff
@@ -45,7 +45,7 @@ extension ModuleDependencyGraph {
     /// frontend creates an interface node,
     /// it adds a dependency to it from the implementation source file node (which
     /// has the intefaceHash as its fingerprint).
-    let fingerprint: String?
+    var fingerprint: String?
```
davidungar, line 48: I see how, when deserializing it is convenient to assign to this, but would the same invariant argument apply here? Does a `let` capture the intension better? This variable is in a class, which might make it even more worthwhile to be a `let`.

Tests/SwiftDriverTests/DependencyGraphSerializationTests.swift
```diff
@@ -0,0 +1,226 @@
...
+        .load(index: 0, nodes: [.topLevel: ["a->", "b->"]]),
+        .load(index: 1, nodes: [.nominal: ["c->", "d->"]]),
+        .load(index: 2, nodes: [.topLevel: ["e", "f"]]),
+        .load(index: 3, nodes: [.nominal: ["g", "h"]]),
+        .load(index: 4, nodes: [.dynamicLookup: ["i", "j"]]),
+        .load(index: 5, nodes: [.dynamicLookup: ["k->", "l->"]]),
+        .load(index: 6, nodes: [.member: ["m,mm", "n,nn"]]),
+        .load(index: 7, nodes: [.member: ["o,oo->", "p,pp->"]]),
+        .load(index: 8, nodes: [.externalDepend: ["/foo->", "/bar->"]]),
+        .load(index: 9, nodes: [
```
davidungar, line 64: Don't know if there's a good way around it, but these integer constants may make the test more brittle.

davidungar (approved):
It will be good to have this code in the system for us to play with. In the long run, you'll see I made some comments about factoring out constants, etc, where, if you had had more time, the code could have been more robust.
The one that most concerns me is the `var` for the node fingerprint instead of the `let`, especially because the node is a class, and it would not be clear to a future maintainer that we count on the fingerprint remaining the same. That's probably a small enough change--if you agree with the premise--to be worth doing before merging.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: @cltnschlosser is right. The nuance is: `FrontendTargetInfo` is what is read from the frontend, and `actualSwiftVersion` is what is saved in the build record in order to permit incrementality. For example, if we ever move to a compound compiler version, in which the major part determines incrementality but we remain incremental across the minor part, then `actualSwiftVersion` should only include the major part. So after thinking about it out loud, maybe the `actualSwiftVersion` is the one to use here?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
CodaFi: It really does here. We could probably rework the serialization format to place a fingerprint blob at the end of the format. That would let us deserialize the entire node in one go and we could leave this alone. As a string, it's not going to do much good since the hash is expected to be fairly unique so we can't get any sharing out of the strings table.

Package.resolved
```diff
@@ -24,7 +24,7 @@
         "repositoryURL": "https://github.com/apple/swift-tools-support-core.git",
         "state": {
           "branch": "main",
-          "revision": "c98cfc216d22798dce7ce7b9cc171565371e967e",
+          "revision": "395e7fb3d6149dc7a999342f23cfefa1ada56264",
           "version": null
         }
```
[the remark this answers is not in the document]
CodaFi: I think we need this one. The Package.resolved file keeps all of us in sync with respect to the dependencies in the package file.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
CodaFi: Yeah, I somewhat regret that the bitstream writer's API is very very low level. I opined about how bitstream isn't really amenable to a Codable wrapper in the implementation in TSC...

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
CodaFi: I'd prefer we keep the force unwrap. If we ever fail to register an abbreviation before we use it, this is guaranteed to catch the mistake.

CodaFi:
Going to a address your comments in a follow-up. Will merge this to unblock you.

CodaFi:
⛵

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

Package.resolved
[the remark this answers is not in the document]
davidungar: Ah, OK. I thought this was just generated automatically, but sounds like there's something I'm not thinking about here.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: Yeah, someday...

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: Catching the mistake is good. I'm thinking that if this ever happens in the field, a force-unwrap will yield an error that is more work to understand than something that prints out the missing abbreviation. But it's fine for now.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
davidungar: > Going to a address your comments in a follow-up. Will merge this to unblock you.

Thanks! It will be great to have this in.

PR #451, opened 2021-01-31
Ensure REPL arguments are processed by the ArgsResolver

owenv:
Introduce a new Job.ArgTemplate case which represents "squashed" arguments, so we can represent the structure of LLDB's --repl= argument and resolve the args which will be used to configure the frontend before concatenating them. The old behavior wasn't causing too many problems in practice, but this should make it easier to eventually launch the REPL as the last step in a multi-job build plan.

Also, this PR shell escapes arguments when printing jobs with -###, matching the behavior of the old driver


owenv:
@swift-ci please test

artemcm (approved):
This looks good to me, thank you @owenv .

owenv:
@swift-ci please test

PR #453, opened 2021-02-01
Add argument to ensure B&W diagnostics when checking diagnostics

davidungar:


davidungar:
@swift-ci please test

CodaFi (approved):
If we can find a way to do this through the testing environment I'd be 100% okay with this. But let's not let that stop us.

PR #463, opened 2021-02-04
[Incremental] fix-path-in-guard-not-usable

davidungar:
At some point, the compiler decided to make an unexpected choice: ```
<path is defined>
guard if let path = path, ...
else {
  <use path>
}
<use path>
```
the `path` in the else used to be bound to the definition before the guard, but now this code won't compile.
So rename things.

rdar://problem/73984428

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -655,12 +655,12 @@ extension IncrementalCompilationState {
     ///   - message: The message to emit in the remark.
     ///   - path: If non-nil, the path of some file. If the output for an incremental job, will print out the
     ///           source and object files.
-    func report(_ message: String, _ path: TypedVirtualPath?) {
-       guard let path = path,
+    func report(_ message: String, _ pathIfPresent: TypedVirtualPath?) {
+       guard let path = pathIfPresent,
             let outputFileMap = outputFileMap,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by davidungar** [comment]
@swift-ci please test

**Comment by artemcm** [comment]
I think this will also need a similar change for:
```
/Users/buildnode/jenkins/workspace/oss-swift-package-macos/swift-driver/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/DependencySource.swift:35:17: error: definition conflicts with previous value
      guard let type = type else {
                ^
/Users/buildnode/jenkins/workspace/oss-swift-package-macos/swift-driver/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/DependencySource.swift:31:11: note: previous definition of 'type' is here
      let type =
          ^
```

artemcm:
> I think this will also need a similar change for:
> 
> ```
> /Users/buildnode/jenkins/workspace/oss-swift-package-macos/swift-driver/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/DependencySource.swift:35:17: error: definition conflicts with previous value
>       guard let type = type else {
>                 ^
> /Users/buildnode/jenkins/workspace/oss-swift-package-macos/swift-driver/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/DependencySource.swift:31:11: note: previous definition of 'type' is here
>       let type =
>           ^
> ```

From the full error messages here:
https://ci.swift.org/job/swift-main-source-compat-suite/5709/console

PR #467, opened 2021-02-04
Read and Write The Driver's Dependency Graph For Cross-Module Builds

CodaFi:
The refactoring of the incremental compilation state class is in three pieces:

1) Formalize the type of the InitialState computed by the initializer
2) Internalize computeModuleDependencyGraph and instead make a unified computeInitialState entrypoint (we'll need this for testing later)
3) Divide the status quo and cross-module initial state computations. There is no change to the status quo outside of the breakup of Driver.getBuildInfo

The cross-module path is slightly different than the status quo. For one, the incremental build need not fail entirely when either the build record or dependency graph cannot be read. In fact, it is a requirement that the incremental build proceed but schedule every single input file to rebuild, otherwise we won't have a module dependency graph to serialize!

As an aside, we ought to consider unifying the build record with the module dependency graph state we're writing here. They're read and written in lock-step when cross-module builds are turned on.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -206,15 +206,15 @@ public struct Driver {
   ///
   /// FIXME: This is a little ridiculous. We could probably just replace the
   /// build record outright with a serialized format.
-  var driverDependencyGraphPath: AbsolutePath? {
-    guard
-      let recordInfo = self.buildRecordInfo,
-      let recordPath = recordInfo.buildRecordPath.absolutePath
-    else {
+  var driverDependencyGraphPath: VirtualPath? {
+    guard let recordInfo = self.buildRecordInfo else {
       return nil
     }
-    let filename = recordPath.basenameWithoutExt
-    return recordPath.parentDirectory.appending(component: filename + ".priors")
+    let filename = recordInfo.buildRecordPath.basenameWithoutExt
+    return recordInfo
+      .buildRecordPath
+      .parentDirectory
+      .appending(component: filename + ".priors")
   }
   
   /// Code & data for incremental compilation. Nil if not running in incremental mode.
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -201,6 +201,22 @@ public struct Driver {
   /// Only used for reading when compiling incrementally.
   let buildRecordInfo: BuildRecordInfo?
 
+  /// A build-record-relative path to the location of a serialized copy of the
+  /// driver's dependency graph.
+  ///
+  /// FIXME: This is a little ridiculous. We could probably just replace the
+  /// build record outright with a serialized format.
+  var driverDependencyGraphPath: AbsolutePath? {
+    guard
+      let recordInfo = self.buildRecordInfo,
+      let recordPath = recordInfo.buildRecordPath.absolutePath
+    else {
+      return nil
+    }
+    let filename = recordPath.basenameWithoutExt
+    return recordPath.parentDirectory.appending(component: filename + ".priors")
+  }
+  
   /// Code & data for incremental compilation. Nil if not running in incremental mode.
   /// Set during planning because needs the jobs to look at outputs.
   @_spi(Testing) public private(set) var incrementalCompilationState: IncrementalCompilationState? = nil
@@ -910,6 +926,7 @@ extension Driver {
     if !childJobs.isEmpty {
       do {
         defer {
+          self.incrementalCompilationState?.writeDependencyGraph()
           buildRecordInfo?.writeBuildRecord(
             jobs,
             incrementalCompilationState?.skippedCompilationInputs)
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -63,60 +63,215 @@ public class IncrementalCompilationState {
       "\(options.contains(.enableCrossModuleIncrementalBuild) ? "Enabling" : "Disabling") incremental cross-module building")
 
 
-    guard let (outputFileMap, buildRecordInfo, outOfDateBuildRecord)
-            = try driver.getBuildInfo(self.reporter)
+    guard let outputFileMap = driver.outputFileMap else {
+      driver.diagnosticEngine.emit(.warning_incremental_requires_output_file_map)
+      return nil
+    }
+
+    guard let buildRecordInfo = driver.buildRecordInfo else {
+      reporter?.reportDisablingIncrementalBuild("no build record path")
+      return nil
+    }
+
+    // FIXME: This should work without an output file map. We should have
+    // another way to specify a build record and where to put intermediates.
+    let maybeBuildRecord = buildRecordInfo.populateOutOfDateBuildRecord(
+            inputFiles: driver.inputFiles, reporter: reporter)
+
+
+    guard
+      let initial = try Self.computeInitialState(options,
+                                                 jobsInPhases,
+                                                 outputFileMap,
+                                                 buildRecordInfo,
+                                                 maybeBuildRecord,
+                                                 &driver,
+                                                 self.reporter)
     else {
       return nil
     }
 
-    guard let (
-      moduleDependencyGraph,
-      inputsHavingMalformedDependencySources: inputsHavingMalformedDependencySources
-    ) = Self.computeModuleDependencyGraph(
-      options,
-      buildRecordInfo,
-      outOfDateBuildRecord,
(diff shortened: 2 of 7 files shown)


CodaFi:
@swift-ci test

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -63,60 +63,190 @@ public class IncrementalCompilationState {
...
-  )
-  -> (ModuleDependencyGraph,
-      inputsHavingMalformedDependencySources: [TypedVirtualPath])?
-  {
+  ) throws -> InitialState? {
+    precondition(options.contains(.enableCrossModuleIncrementalBuild))
     let diagnosticEngine = driver.diagnosticEngine
+    // If we cannot compute this path, it means the build record path was not
+    // provided to the driver, which our caller detects for us.
+    let absPath = driver.driverDependencyGraphPath!
```
davidungar, line 161: In the past, we've run into trouble with disallowing relative paths.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -201,6 +201,22 @@ public struct Driver {
   /// Only used for reading when compiling incrementally.
   let buildRecordInfo: BuildRecordInfo?
 
+  /// A build-record-relative path to the location of a serialized copy of the
+  /// driver's dependency graph.
+  ///
+  /// FIXME: This is a little ridiculous. We could probably just replace the
+  /// build record outright with a serialized format.
+  var driverDependencyGraphPath: AbsolutePath? {
```
davidungar, line 209: I suspect that this has to be a VirtualPath for the tests to run.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -63,60 +63,190 @@ public class IncrementalCompilationState {
...
+
+    let compileGroups =
+      Dictionary(uniqueKeysWithValues:
+                    jobsInPhases.compileGroups.map { ($0.primaryInput, $0) })
```
davidungar, line 165: Would be nice to move this computation to just before using `compileGroups`

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -63,60 +63,190 @@ public class IncrementalCompilationState {
...
+          from: absPath,
+          on: driver.fileSystem,
+          diagnosticEngine: diagnosticEngine,
+          reporter: reporter,
+          options: options),
+      let outOfDateBuildRecord = buildRecordInfo.populateOutOfDateBuildRecord(
+            inputFiles: driver.inputFiles, reporter: reporter)
+    else {
+      // We failed to read the components we need for the incremental build -
+      // just schedule every file to build!
```
davidungar, line 180: Is there some reason not to try a build reading the swiftDeps?
This raises another question: as it is, the code has two paths:
1. status quo - don't try to read the saved graph, don't try to exploit incrementalExternalDeps
2. cross module - do both

In other words, maybe there are four logical possibilities--I haven't thought about if it's best to collapse it to these two.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -63,60 +63,190 @@ public class IncrementalCompilationState {
...
+
+      let mandatoryCompileGroupsInOrder = driver.inputFiles.compactMap { input -> CompileJobGroup? in
+        compileGroups[input]
+      }
+
+      let mandatoryJobsInOrder = try
+        jobsInPhases.beforeCompiles +
+        driver.formBatchedJobs(
+          mandatoryCompileGroupsInOrder.flatMap {$0.allJobs()},
+          showJobLifecycle: driver.showJobLifecycle)
```
davidungar, line 197: Because this computation is submerged in the other case, I'm not sure about seeing it here; need to think on it. Oh, I see, this is in an error-recovery path.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -63,60 +63,190 @@ public class IncrementalCompilationState {
...
+
+      return InitialState(graph: emptyGraph,
+                          skippedCompileGroups: [:],
+                          mandatoryJobsInOrder: mandatoryJobsInOrder)
+    }
+
+    // Success! Let's try to go through the same path as the status quo.
```
davidungar, line 204: There's a difference, & I need to check to see of the same path does the right thing:
In the status quo, external deps get added, in this case, they are already present.
Any non-incremental ones, need to check file mod dates.
Any incremental ones, probably need to re-import the swift-module files.
Not sure that logic is in this PR--haven't gotten there yet.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -63,60 +63,190 @@ public class IncrementalCompilationState {
...
+    guard missingInputs.isEmpty else {
+      if let reporter = reporter {
+        reporter.report(
+          "Incremental compilation has been disabled, " +
+          " because  the following inputs were used in the previous compilation but not in this one: "
+            + missingInputs.map {$0.basename} .joined(separator: ", "))
+      }
+      return nil
+    }
+
```
davidungar, line 249: As I wrote above, I'm not sure about this factoring. I'd be tempted to:
Read the serialized Mod dep graph in any case,
and go from there. But maybe that's too much instability??

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -63,60 +63,190 @@ public class IncrementalCompilationState {
...
+    // record.
+    guard
+      let deserializedGraph = try? ModuleDependencyGraph.read(
+          from: absPath,
+          on: driver.fileSystem,
+          diagnosticEngine: diagnosticEngine,
+          reporter: reporter,
+          options: options),
+      let outOfDateBuildRecord = buildRecordInfo.populateOutOfDateBuildRecord(
+            inputFiles: driver.inputFiles, reporter: reporter)
```
davidungar, line 177: This same computation and the aborting incremental building appears in both the statusQuo and crossModule code. Feels like it should be in a caller and passed in if non-null.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -735,3 +849,28 @@ extension IncrementalCompilationState {
     public static let enableCrossModuleIncrementalBuild      = Options(rawValue: 1 << 4)
   }
 }
+
+// MARK: - Serialization
+
+extension IncrementalCompilationState {
+  @_spi(Testing) public func writeDependencyGraph() {
```
davidungar, line 856: In other places the _spi is commented out. We went through some kerfuffle wrt testable vs _spi. I'm not sure what the current state is.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -735,3 +849,28 @@ extension IncrementalCompilationState {
...
+    // If the cross-module build is not enabled, the status quo dictates we
+    // not emit this file.
+    guard self.options.contains(.enableCrossModuleIncrementalBuild) else {
+      return
+    }
```
davidungar, line 861: I'd be tempted to do so anyway, speeding up the non-cross-module case.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -56,14 +56,43 @@ import SwiftOptions
...
+        continue
+      }
+
+      assert(swiftDepsFile.extension == FileType.swiftDeps.rawValue)
+      let typedSwiftDepsFile = TypedVirtualPath(file: swiftDepsFile, type: .swiftDeps)
+      let dependencySource = ModuleDependencyGraph.DependencySource(typedSwiftDepsFile)
+      emptyGraph.inputDependencySourceMap[input] = dependencySource
+    }
+    return emptyGraph
+  }
```
davidungar, line 86: I don't think we ought to need to do this; just fall back to the `buildInitialGraph` path. Am I missing something?

davidungar (commented):
I'm not sure about some of the premises here, in particular the coupling of cross-module with ModuleDepGraph serialization. Want to have a chance to discuss before it lands.

davidungar (approved):
LGTM

CodaFi:
@swift-ci test

PR #468, opened 2021-02-04
[Incremental] Next step towards cross-module incremental dependencies

davidungar:
Lots of restructuring; passes the lit tests.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -42,17 +42,20 @@ import SwiftOptions
 
   let options: IncrementalCompilationState.Options
   let reporter: IncrementalCompilationState.Reporter?
+  let fileSystem: FileSystem
   
   private let diagnosticEngine: DiagnosticsEngine
   
   public init(
     diagnosticEngine: DiagnosticsEngine,
     reporter: IncrementalCompilationState.Reporter?,
+    fileSystem: FileSystem,
     options: IncrementalCompilationState.Options
   ) {
     self.reporter = reporter
     self.diagnosticEngine = diagnosticEngine
     self.options = options
+    self.fileSystem = fileSystem
   }
 
   var isCrossModuleIncrementalBuildEnabled: Bool {
@@ -83,6 +86,7 @@ extension ModuleDependencyGraph {
     let graph = Self(
       diagnosticEngine: diagnosticEngine,
       reporter: reporter,
+	  fileSystem: fileSystem,
       options: options)
 
     let inputsAndSwiftdeps = inputs.map { input in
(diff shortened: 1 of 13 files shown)


davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/SourceFileDependencyGraph.swift
```diff
@@ -250,13 +270,18 @@ extension SourceFileDependencyGraph {
...
+      try data.contents.withUnsafeBytes { buf in
+        // SAFETY: The bitcode reader does not mutate the data stream we give it.
+        // FIXME: Let's avoid this altogether and traffic in ByteString/[UInt8]
+        // if possible. There's no real reason to use `Data` in this API.
+        let baseAddr = UnsafeMutableRawPointer(mutating: buf.baseAddress!)
+        let data = Data(bytesNoCopy: baseAddr, count: buf.count, deallocator: .none)
+        try Bitcode.read(stream: data, using: &visitor)
+      }
+    }
+    catch ReadError.swiftModuleHasNoDependencies {
```
CodaFi, line 283: Shouldn't we catch every kind of error here? Like, if more than the signature were to be corrupted and the visitor were to detect it, we should just say the graph is busted and start over.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -105,84 +106,96 @@ extension ModuleDependencyGraph {
...
+    guard let sourceGraph = dependencySource.read(
+            in: fileSystem,
+            reporter: reporter)
+    else {
+      return nil
+    }
+    return integrate(sourceGraph: sourceGraph)
+  }
+
+  public func integrate(sourceGraph: SourceFileDependencyGraph) -> Set<Node>? {
```
CodaFi, line 140: This is a fantastic cleanup

CodaFi (approved):
Just one question, but it's not enough to hold back integrating this PR.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: thanks!

Sources/SwiftDriver/IncrementalCompilation/SourceFileDependencyGraph.swift
davidungar: I don't think so. This covers the special case of getting an external-dep with a fingerprint, but in the meantime the vendor has been compiled without the flag. But maybe I don't follow; we can discuss offline.

davidungar:

CodaFi:
⛵

PR #469, opened 2021-02-05
Revert "[Incremental] Next step towards cross-module incremental dependencies"

rintaro:
Reverts apple/swift-driver#468


CodaFi:
I will fix this error on our end. We just need to probably update the cmake.

PR #473, opened 2021-02-07
XCTFail a dependency scanner test which relies on libSwiftScan if it can't be found

owenv:
Otherwise, this test crashes when it calls `getDependencies`. With this change I can run the whole test suite using the frontend fallback.


owenv:
@swift-ci please test

artemcm:
Ah, this was an oversight on my part, thank you @owenv .

artemcm (approved):

PR #478, opened 2021-02-09
Indexing: include index file entry in supplementary output file map

nkcsgexi:
This is the behavior of the C++ driver and we should preserve it.

rdar://74127445

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -250,7 +250,8 @@ extension Driver {
                                                         primaryInputs: [TypedVirtualPath],
                                                         inputsGeneratingCodeCount: Int,
                                                         inputOutputMap: [TypedVirtualPath: TypedVirtualPath],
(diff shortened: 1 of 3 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -417,6 +418,11 @@ extension Driver {
       for flaggedPair in flaggedInputOutputPairs {
         addEntry(&entries, input: flaggedPair.input, output: flaggedPair.output)
       }
+      // To match the legacy driver behavior, make sure we add an entry for the
+      // file under indexing and the primary output file path.
+      if let indexFilePath = indexFilePath, let idxOutput = inputOutputMap[indexFilePath] {
```
artemcm, line 423: Ah, I guess after 3 input source files we are going to exceed:
```
if inputsGeneratingCodeCount * FileType.allCases.count > fileListThreshold {
```
Makes sense to me!

artemcm (approved):

cltnschlosser (approved):

PR #480, opened 2021-02-10
[RFC] Implement -driver-skip-execution

cltnschlosser:
`-driver-skip-execution` in c++ driver replaces the TaskQueue with DummyTaskQueue. DummyTaskQueue honors the number of parallel tasks but otherwise mocks out process execution and has it return 0 and output `Output placeholder\n`.
This flag is most useful for testing in combination with `-parseable-output` when you don't want the jobs to actually run.

First complication: swift-driver calls out to the frontend for more than just jobs, so the executor can't be completely mocked out. Also the executor handles quite a bit more work than the c++ TaskQueue, so a mock version would require much more duplicated logic.

Second complication: The swift-driver's executor code lives in `SwiftDriverExecution`, and providing it new information requires making API changes to `DriverExecutor` this is a source break to SPM (without tacky shims put in place).

Looking at the state of `-driver-skip-execution` usage in lit tests:
- 12 total usages in 8 files
- 3 are used with `-driver-print-jobs` which makes `-driver-skip-execution` ignored. That leaves 6 being used with `-parseable-output` (2 (in emit-dependencies.swift) could have test rewritten to use `-driver-print-jobs`) and 3 with `-driver-show-job-lifecycle`.

Questions: Does this kind of process mock provide value in swift-driver? If no, how do we handle the lit test failures?

My thoughts on that question:
This level of mocking and testing can easily be done in XCTest, if we were confident in coverage, we could `REQUIRES: cplusplus_driver`.
The only test this PR actually fixes is `emit-dependencies.swift` which can be fixed in the test by changing it to `-driver-print-jobs`. The reasons for other test not being fixed are:
- Parseable output: This tests rely on the ordering of JSON output by the c++ driver. The swift-driver uses JSONEncoder to format the JSON output, which doesn't match the c++ driver.
- `-driver-show-job-lifecycle`: These 3 tests are for batch mode, I believe the behavior is correct here, but the output isn't 100% the same and some of the ordering is different. Also these tests run even when the compile isn't mocked out (just slower).

TLDR: I think support for this flag is of little value given that most of the tests that rely on it are Parseable output and fail anyway, but it is a flag supported by the C++ driver, so maybe we want to continue supporting it.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -1014,6 +1014,7 @@ extension Driver {
       delegate: jobExecutionDelegate,
       numParallelJobs: numParallelJobs ?? 1,
       forceResponseFiles: forceResponseFiles,
+      skipProcessExecution: parsedOptions.contains(.driverSkipExecution),
       recordedInputModificationDates: recordedInputModificationDates)
   }
 
--- a/Sources/SwiftDriver/Execution/DriverExecutor.swift
+++ b/Sources/SwiftDriver/Execution/DriverExecutor.swift
@@ -28,6 +28,15 @@ public protocol DriverExecutor {
                delegate: JobExecutionDelegate,
                numParallelJobs: Int,
                forceResponseFiles: Bool,
+               skipProcessExecution: Bool,
+               recordedInputModificationDates: [TypedVirtualPath: Date]
+  ) throws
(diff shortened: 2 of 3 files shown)


owenv (approved):
I'm sort of inclined to just XFAIL the tests that use this and port them if needed, but if we do decide to keep the option this LGTM

PR #481, opened 2021-02-10
Clean Up the MultiDictionary Interface

CodaFi:
- Clean up subscripting so accessing one-past-the-end indices is no longer possible
- Make the interface for Multidictionary more closely align to either Dictionary or Set
- Document everything

This PR should be NFC, but we'll see if anything was relying on the old subscript order.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -889,7 +889,7 @@ extension ModuleDependencyGraph {
             $0.append(serializer.lookupIdentifierCode(
                         for: key.designator.name ?? ""))
           }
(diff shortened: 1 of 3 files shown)


CodaFi:
@swift-ci test

CodaFi:
@davidungar A quick question: The old interface is not consistent about its removal-based operations erasing keys when there are no associated values remaining. What's the intended behavior here?

davidungar (approved):
Fantastic! Thank you.

CodaFi:
⛵

davidungar:
> @davidungar A quick question: The old interface is not consistent about its removal-based operations erasing keys when there are no associated values remaining. What's the intended behavior here?

Good question--not something I have thought about. Looking at how it is used, I think my expectation was that removing a non-existent key would be a noop.

PR #482, opened 2021-02-11
Make the driver aware of all the flags the selected toolchain's `swift-frontend` supports.

artemcm:
Building on https://github.com/apple/swift/pull/35882, make the driver aware of which swift-frontend flags its toolchain's compiler supports. 

- Add `supportedFrontendFlags` property to the Driver
- This property is computed using either:
      - A `libSwiftScan` query if an instance of the library is available and supports executing such query
      - A fallback mechanism that invokes `swift-frontend -emit-supported-features` 

For `libSwiftDriver`, this is critical for multiplexing library instances, and for the library's compatibility with different versions of the compiler.

Resolves rdar://73631930

### DIFF
--- a/Sources/SwiftDriver/SwiftScan/DependencyGraphBuilder.swift
+++ b/Sources/SwiftDriver/SwiftScan/DependencyGraphBuilder.swift
@@ -222,7 +222,7 @@ private extension SwiftScan {
   }
 }
 
-private extension SwiftScan {
+internal extension SwiftScan {
   /// Convert a `swiftscan_string_ref_t` reference to a Swfit `String`, assuming the reference is to a valid string
   /// (non-null)
   func toSwiftString(_ string_ref: swiftscan_string_ref_t) throws -> String {
@@ -246,6 +246,18 @@ private extension SwiftScan {
                                                     count: Int(string_set.count)))
     for stringRef in stringRefArrray {
       result.append(try toSwiftString(stringRef))
+    }
+    return result
+  }
+
+  /// Convert a `swiftscan_string_set_t` reference to a Swfit `Set<String>`, assuming the individual string references
+  /// are to a valid strings (non-null)
+  func toSwiftStringSet(_ string_set: swiftscan_string_set_t) throws -> Set<String> {
+    var result = Set<String>()
+    let stringRefArrray = Array(UnsafeBufferPointer(start: string_set.strings,
+                                                    count: Int(string_set.count)))
+    for stringRef in stringRefArrray {
+      result.insert(try toSwiftString(stringRef))
     }
     return result
   }
--- a/Sources/CSwiftScan/include/swiftscan_header.h
+++ b/Sources/CSwiftScan/include/swiftscan_header.h
@@ -177,6 +177,8 @@ typedef struct {
 
   //=== Cleanup Functions ---------------------------------------------------===//
   void
+  (*swiftscan_string_set_dispose)(swiftscan_string_set_t *);
+  void
   (*swiftscan_dependency_graph_dispose)(swiftscan_dependency_graph_t);
   void
   (*swiftscan_import_set_dispose)(swiftscan_import_set_t);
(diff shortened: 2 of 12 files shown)


artemcm:
@swift-ci please test

akyrtzi:
> - A libSwiftScan query if an instance of the library is available and supports executing such query
> - A fallback mechanism that invokes swift-frontend -emit-supported-features

Have you considered just reading the json file? https://github.com/apple/swift/blob/main/lib/Option/features.json

artemcm:
> > * A libSwiftScan query if an instance of the library is available and supports executing such query
> > * A fallback mechanism that invokes swift-frontend -emit-supported-features
> 
> Have you considered just reading the json file? https://github.com/apple/swift/blob/main/lib/Option/features.json

Thanks for pointing me to this file, I wasn't aware this json file existed. As implemented, this only handles compiler flags, because I didn't know that we already have something like this to track which features are supported. 

Both approaches (library query and frontend invocation) just use the [compiler's list of flags](https://github.com/apple/swift/blob/main/tools/libSwiftScan/libSwiftScan.cpp#L481) that is generated with tablegen. Similarly, they can read the compiler's list of features from this JSON file.

Are suggesting that instead of relying on the presence of the compiler/library we should instead ship JSON files in the toolchain for both flags and features and just have the driver read those?

akyrtzi:
> we should instead ship JSON files in the toolchain for both flags and features and just have the driver read those?

Yes, this JSON file already ships with the toolchain. Reading a JSON file is way simpler.

Note the premise of this file is not to be a comprehensive list of *every* option that exists today, but to selectively add newly added features that clients needs to check for the presence of (while different toolchains take time to get new versions) and modify their behavior accordingly.

nkcsgexi (approved):
LGTM! I like the idea of calling the c API to collect the full list of supported compiler arguments. It never requires manual maintenance of an external JSON file.

artemcm:
> > we should instead ship JSON files in the toolchain for both flags and features and just have the driver read those?
> 
> Yes, this JSON file already ships with the toolchain. Reading a JSON file is way simpler.
> 
> Note the premise of this file is not to be a comprehensive list of _every_ option that exists today, but to selectively add newly added features that clients needs to check for the presence of (while different toolchains take time to get new versions) and modify their behavior accordingly.

Having a comprehensive collection of all options helps automate a part of this process, I think, because we don't have to depend on something being declared a feature to be able to be careful about relying on using it from the driver. And to keep that list in a toolchain would require building automation that ensures this file is always up-to-date when people added new options. 

For separately tracking selective "Features" vs. all available flags though, if this JSON file is already used as **the** source of truth in other parts of the toolchain (like tablegen files are used for options in the compiler), then we should be able to use that here as well, I agree.

akyrtzi:
I mainly wanted to bring awareness to that approach so the choice is a more informed one (no objections from me).

nkcsgexi:
Yeah, we already have this established communication channel of calling C APIs into the libSwiftScanner.dylib. So I don't find it appealing to not use it but deserializing an installed JSON file instead.

artemcm:
> I mainly wanted to bring awareness to that approach so the choice is a more informed one (no objections from me).

Thank you, @akyrtzi !

Sources/SwiftDriver/SwiftScan/DependencyGraphBuilder.swift
```diff
@@ -249,6 +249,18 @@ private extension SwiftScan {
     }
     return result
   }
+
+  /// Convert a `swiftscan_string_set_t` reference to a Swfit `Set<String>`, assuming the individual string references
```
cltnschlosser, line 253: typo: `Swfit`. Maybe fix on your next PR if you remember.

artemcm: Thanks, @cltnschlosser!

PR #483, opened 2021-02-11
[Incremental] Add functions and tests to BidirectionalMap

davidungar:
These will be needed in an upcoming PR.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/BidirectionalMap.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/BidirectionalMap.swift
@@ -11,9 +11,11 @@
 //===----------------------------------------------------------------------===//
 
 /// Like a two-way dictionary, only works for accessing present members
-public struct BidirectionalMap<T1: Hashable, T2: Hashable>: Equatable {
+public struct BidirectionalMap<T1: Hashable, T2: Hashable>: Equatable, Sequence {
   private var map1: [T1: T2] = [:]
   private var map2: [T2: T1] = [:]
+
+  public init() {}
 
   public subscript(_ key: T1) -> T2 {
     get {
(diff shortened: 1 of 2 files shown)


davidungar:
@swift-ci please test

CodaFi (approved):

Sources/SwiftDriver/IncrementalCompilation/BidirectionalMap.swift
```diff
@@ -38,4 +40,25 @@ public struct BidirectionalMap<T1: Hashable, T2: Hashable>: Equatable {
     }
     set { self[newValue] = key }
   }
+  public func contains(key: T1) -> Bool {
+    map1.keys.contains(key)
+  }
+  public func contains(key: T2) -> Bool {
```
porglezomp, line 46: Is it possible to distinguish between these methods if `T1` happens to be equal to `T2`?

cltnschlosser: Maybe this should be `contains(value: T2)`. Not sure how naming works with BidirectionalMap, if it's still `<Key, Value>` or if it's `<FirstKey, SecondKey>`.

porglezomp: Ah, I tried it out, it's already impossible to add any elements to the map if `T1` = `T2`, because the subscript operators are also ambiguous then.

CodaFi: There's a simple fix here were these ambiguous operations are given one final overload under the cosntraint `T1 == T2`

@davidungar Would you mind adding those overloads?

```
extension BidirectionalMap where T1 == T2 {
  // ...
}
```

davidungar: Great thoughts, both of you! Tx, will do

davidungar: I looked at it, and what's nice is that the compiler complains if the types are the same. I'm inclined to land this, since the same-type functionality is not needed and would add unused operations to the protocol. If it's ever used with same types, it won't compile and the programmer can use the extension trick then.

PR #484, opened 2021-02-11
Use the FrontendOverride when querying the `hostTriple`.

artemcm:
And fix the `FrontendOverride::setUpForTargetInfo` to clear the overridden compiler.

This should get the compiler tests in `Driver/Dependencies` running again.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -530,8 +530,8 @@ public struct Driver {
 
     self.supportedFrontendFlags =
       try Self.computeSupportedCompilerFeatures(of: self.toolchain, hostTriple: self.hostTriple,
-                                                swiftCompilerPrefixArgs:
-                                                  swiftCompilerPrefixArgs,
+                                                parsedOptions: &self.parsedOptions,
+                                                diagnosticsEngine: diagnosticEngine,
                                                 fileSystem: fileSystem, executor: executor,
                                                 env: env)
 
@@ -2287,7 +2287,7 @@ extension Driver {
     }
   }
 
-  private struct FrontendOverride {
+  internal struct FrontendOverride {
     private let overridePath: AbsolutePath?
     let prefixArgs: [String]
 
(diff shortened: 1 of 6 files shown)


artemcm:
@swift-ci please test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2291,23 +2298,23 @@ extension Driver {
...
       prefixArgs = frontendCommandLine.dropFirst().map {String($0)}
     }
 
     var appliesToFetchingTargetInfo: Bool {
-      path?.basename != "Python"
+      return overridePath?.basename != "Python"
     }
     func setUpForTargetInfo(_ toolchain: Toolchain) {
-      if appliesToFetchingTargetInfo {
-        setUpForCompilation(toolchain)
```
artemcm, line 2303: @davidungar , this threw me off a bit because in the case where the frontend gets overriden, this line never gets executed, and if it were, then it would just call `setUpForCompilation` and still override the path. Am I misunderstanding the intent behind it?

davidungar: It's tricky. There's one setup needed for the triple, and other for actual compilation. For fetching the target info, we need to forget the overrides if they are pointing to Python. For the "compilation" we need to obey them no matter what.

cltnschlosser:
Do we need something like this for: https://github.com/apple/swift-driver/pull/482 or does it hit libSwiftScan and not fallback on the frontend?

Sources/SwiftDriver/Driver/Driver.swift
artemcm: Yes, I understand that part, I was just confused because in the existing code (before this PR), the `setUpForTargetInfo` would not actually cause the override to be forgotten. We  got away with it before because we used the toolchain first, before applying the override, to query the target info, and then we set it to "compilation" mode.

I think this change makes it work, though.

davidungar (approved):

artemcm:
> Do we need something like this for: #482 or does it hit libSwiftScan and not fallback on the frontend?

Oh, great catch! We absolutely do. 
I was testing with a non up-to-date branch. I'll make this change. Thank you, @cltnschlosser.

artemcm:
@swift-ci please test

PR #486, opened 2021-02-11
[Incremental] Centralize fingerprint/swiftmodule invariant, catch bad paths earlier

davidungar:
Centralizes the enforcement of the aforementioned invariant. Also makes the fields of a `SourceModuleDependencyGraph.Node` immutable. Catches bad external dependency paths as soon as possible.


davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/KeyAndFingerprintHolder.swift
```diff
@@ -0,0 +1,53 @@
+//===------- KeyAndFingerprintEnforcer.swift ------------------------------===//
```
CodaFi, line 1:
```suggestion
//===------- KeyAndFingerprintHolder.swift -------------------------------===//
```

Tests/SwiftDriverTests/ModuleDependencyGraphTests.swift
```diff
@@ -1201,8 +1211,14 @@ fileprivate struct SourceFileDependencyGraphMocker {
...
-    use.addDefIDependUpon(def.sequenceNumber)
-    allNodes[use.sequenceNumber] = use
+    var newDefsIDependUpon = use.defsIDependUpon
+    newDefsIDependUpon.append(def.sequenceNumber)
+    let newUse = try! Node(key: use.key,
+                           fingerprint: use.fingerprint,
+                           sequenceNumber: use.sequenceNumber,
+                           defsIDependUpon: newDefsIDependUpon,
+                           isProvides: use.isProvides)
+    allNodes[newUse.sequenceNumber] = newUse
```
CodaFi, line 1221: This is... rather expensive.

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
```diff
@@ -1,39 +1,47 @@
...
-  let fingerprint: String?
+  let file: VirtualPath
 
-  /*@_spi(Testing)*/ public init(_ fileName: String, fingerprint: String? = nil) {
-    self.fileName = fileName
-    self.fingerprint = fingerprint
+  /// Cache this
+  let isSwiftModule: Bool
+
+  /*@_spi(Testing)*/ public init(_ fileName: String)
```
CodaFi, line 12: Thanks!

CodaFi (changes requested):
I appreciate trying to push immutability in more places, but this PR will have the unfortunate effect of making node construction incur roughly factorial copies in the number of def-use arcs added during graph construction. Perhaps we could explore a lowering step where the node data is collected and all the arcs are gathered, then we "commit" and lower that data to the actual graph nodes.

Sources/SwiftDriver/IncrementalCompilation/KeyAndFingerprintHolder.swift
davidungar: D'Oh! Thanks.

Tests/SwiftDriverTests/ModuleDependencyGraphTests.swift
davidungar: Yes, I'll think about a better way.

davidungar:
> I appreciate trying to push immutability in more places, but this PR will have the unfortunate effect of making node construction incur roughly factorial copies in the number of def-use arcs added during graph construction. Perhaps we could explore a lowering step where the node data is collected and all the arcs are gathered, then we "commit" and lower that data to the actual graph nodes.

That's a fair point. I see where happens in the test. I think that's the only place (in the driver). I'll look at doing as you suggest.

CodaFi:
(Correction *the number of copies is roughly a triangular sum, not a product. That makes it quadratic)

davidungar:
@swift-ci please test

davidungar:
> (Correction *the number of copies is roughly a triangular sum, not a product. That makes it quadratic)

Yes, that sounds more accurate.

davidungar:
@swift-ci  please test

davidungar:
@CodaFi How does it look now?

davidungar:
@swift-ci please test

davidungar:
rebased onto master

davidungar:
@swift-ci please test

Tests/SwiftDriverTests/ModuleDependencyGraphTests.swift
```diff
@@ -1200,9 +1212,30 @@ fileprivate struct SourceFileDependencyGraphMocker {
   }
 
   private mutating func addArc(def: Node, use: Node) {
-    var use = getNode(use.sequenceNumber)
-    use.addDefIDependUpon(def.sequenceNumber)
-    allNodes[use.sequenceNumber] = use
+    while dependencyAccumulator.count <= use.sequenceNumber {
```
cltnschlosser, line 1215: I don't know the context of these numbers, so maybe it doesn't make sense, but should you reserveCapacity in the array before the loop?

Tests/SwiftDriverTests/ModuleDependencyGraphTests.swift
```diff
@@ -1200,9 +1212,30 @@ fileprivate struct SourceFileDependencyGraphMocker {
...
+      dependencyAccumulator.append(nil)
+    }
+    var dh = dependencyAccumulator[use.sequenceNumber] ?? {
+      let newOne = DependencyHolder()
+      dependencyAccumulator[use.sequenceNumber] = newOne
+      return newOne
+    }()
```
cltnschlosser, line 1222: I wrote this thinking this was a dictionary at first, but it's not, so this doesn't really apply. But maybe could make something like this. I think it increases readability, but completely up to you. Might be weird for an array to have that subscript.
```suggestion
    var dh = dependencyAccumulator[use.sequenceNumber, default: DependencyHolder()]
```

Related, any reason or advantage to having this be an array of optionals? Or would in make sense to append DependencyHolders in the loop instead of nil

davidungar: It is an optimization, since a number have no entries. But its a good question.

Tests/SwiftDriverTests/ModuleDependencyGraphTests.swift
davidungar: Yes, I should. Great point. Thanks for reviewing.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -564,15 +564,15 @@ extension IncrementalCompilationState {
     var externalDependencySources = Set<DependencySource>()
     for extDepAndPrint in moduleDependencyGraph.fingerprintedExternalDependencies {
       let extDep = extDepAndPrint.externalDependency
-      let extModTime = extDep.file.flatMap {try? fileSystem.getFileInfo($0).modTime}
+      let extModTime = (try? fileSystem.getFileInfo(extDep.file).modTime)
```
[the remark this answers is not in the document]
davidungar: Thanks; nice catch!

davidungar:
@swift-ci please test

CodaFi (approved):

CodaFi:
⛵

davidungar:
> ⛵

Thank you!

PR #485, opened 2021-02-11
Revert "Make the driver aware of all the flags the selected toolchain's `swift-frontend` supports."

artemcm:
Reverts apple/swift-driver#482

Making `SwiftScan` internal instead of private seems to be breaking the ability to build SwiftDriver from SwiftPM PRs.

PR #487, opened 2021-02-12
Workaround a compiler error with SPI public type using `implementationOnly` properties.

artemcm:
Keep `SwiftScan` internal becuase using `@spi public` combined with using an `implementationOnly` property type does not work when building with 5.3 (as SwiftPM self-hosted PR testing does).

Doing this instead of https://github.com/apple/swift-driver/pull/485.


artemcm:
@swift-ci please test

cltnschlosser (approved):

PR #488, opened 2021-02-12
Allow libSwiftDriver client passing down a compiler executable directory

nkcsgexi:
When using swift-driver as a library, we cannot assume compiler is adjacent to the
build system executable.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -353,6 +353,7 @@ public struct Driver {
     fileSystem: FileSystem = localFileSystem,
     executor: DriverExecutor,
     integratedDriver: Bool = true,
+    compilerExecutableDir: AbsolutePath? = nil,
     // FIXME: Duplication with externalBuildArtifacts and externalTargetModulePathMap
     // is a temporary backwards-compatibility shim to help transition SwiftPM to the new API
     externalBuildArtifacts: ExternalBuildArtifacts? = nil,
@@ -415,7 +416,8 @@ public struct Driver {
           &self.parsedOptions, diagnosticsEngine: diagnosticEngine,
           compilerMode: self.compilerMode, env: env,
           executor: self.executor, fileSystem: fileSystem,
-          useStaticResourceDir: self.useStaticResourceDir)
+          useStaticResourceDir: self.useStaticResourceDir,
+          compilerExecutableDir: compilerExecutableDir)
 
     // Compute the host machine's triple
     self.hostTriple =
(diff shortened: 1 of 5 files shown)


nkcsgexi:
@swift-ci please test

artemcm:
@nkcsgexi what are your thoughts on using this versus passing in a path to a toolchain "root"? We need many different things from the toolchain beyond a compiler executable so I wonder if it's better to find them relative to a toolchain itself, vs. relative to a compiler executable. Either way, we are beholden to a toolchain's directory structure being fixed.

artemcm (approved):

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -353,6 +353,7 @@ public struct Driver {
     fileSystem: FileSystem = localFileSystem,
     executor: DriverExecutor,
     integratedDriver: Bool = true,
+    compilerExecutableDir: AbsolutePath? = nil,
```
cltnschlosser, line 356: It's not as explicitly typed, but you can already do this by changing the env that you provide the driver.
Using `SWIFT_DRIVER_<TOOLNAME>_EXEC` environment variables. It also happens sooner.

```
if let overrideString = envVar(forExecutable: executable) {
      return try AbsolutePath(validating: overrideString)
    } else if let toolDir = toolDirectory,
              let path = lookupExecutablePath(filename: executable, searchPaths: [toolDir]) {
      // Looking for tools from the tools directory.
      return path
    } else if let path = lookupExecutablePath(filename: executable, searchPaths: [executableDir]) {
      return path
      return path
    } else if !["swift-frontend", "swift"].contains(executable),
              let parentDirectory = try? getToolPath(.swiftCompiler).parentDirectory,
              parentDirectory != executableDir,
              let path = lookupExecutablePath(filename: executable, searchPaths: [parentDirectory]) {
      // If the driver library's client and the frontend are in different directories,
      // try looking for tools next to the frontend.
      return path
    } else if let path = lookupExecutablePath(filename: executable, searchPaths: searchPaths) {
      return path
    } else if executable == "swift-frontend" {
      // Temporary shim: fall back to looking for "swift" before failing.
      return try lookup(executable: "swift")
    } else if fallbackToExecutableDefaultPath {
      return AbsolutePath("/usr/bin/" + executable)
    } else {
      throw ToolchainError.unableToFind(tool: executable)
    }
```

Also using the `-tools-directory` option is of higher precedence as well.

nkcsgexi:
@artemcm all tools are currently found with an anchor of the executable dir where swift-driver is installed, e.g. the scanner library. I think this still makes sense and more versatile in situations that we don't have a fully-defined toolchain (in build directories for example).

Sources/SwiftDriver/Driver/Driver.swift
nkcsgexi: Both `SWIFT_DRIVER_<TOOLNAME>_EXEC` and `-tools-directory` are global settings that the build system will honor no matter which toolchain is currently under use. This parameter `compilerExecutableDir` is mostly for the build system  to create different driver instances for different toolchains.

cltnschlosser: Right you pass the args and environment to the driver as you create it. So you can modify them there. Or just append, the argument to the existing logic. Users of libSwiftDriver are probably already creating argument lists.

It seems weird that the Driver creator needs another way to override this path.

PR #491, opened 2021-02-13
[Incremental] Handle nodes imported from swiftmodules

davidungar:
A small change to keep `printPath` from crashes when nodes are integrated from swiftmodules, when we start doing cross-module dependencies.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
@@ -114,7 +114,6 @@ extension ModuleDependencyGraph.Tracer {
     currentPathIfTracing = currentPath
   }
 
-
   private func printPath(_ path: [Graph.Node]) {
     guard path.first?.dependencySource != path.last?.dependencySource
     else {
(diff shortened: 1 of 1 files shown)


davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
```diff
@@ -126,8 +125,13 @@ extension ModuleDependencyGraph.Tracer {
         path
           .compactMap { node in
             node.dependencySource
-              .flatMap {graph.inputDependencySourceMap[$0] }
-              .map { "\(node.key) in \($0.file.basename)"}
+              .flatMap {
+                // swiftmodules won't be in the map
+                graph.inputDependencySourceMap.contains(key: $0)
+                  ? "\(node.key) in \(graph.inputDependencySourceMap[$0].file.basename)"
+                  : "\(node.key)"
```
cltnschlosser, line 132: `if let` won't be one line but it will prevent need to query the map twice

cltnschlosser (approved):

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
davidungar: Interesting... I always thought that ?: used short-circuit evaluation. I'll have to try that out. Thanks for taking a look.

davidungar: Indeed, according to a hasty experiment, I think that ?: does short-circuit.

davidungar: But, now I see what you mean. Sorry--there's an odd aspect to this map; unlike normal maps, it crashes if subscripted with an absent key. We plan to fix that in a later PR.

davidungar: At least this isn't in code that normally runs.

davidungar: And thanks!

CodaFi (approved):

PR #495, opened 2021-02-15
Ignore interruption signal sent from the build system

nkcsgexi:
The underlying swift compiler isn't ready to be safely canceled yet and
interrupting them may cause red-herring build failures that pollute the build log.

rdar://74292387


nkcsgexi:
@swift-ci please test

artemcm (approved):

PR #498, opened 2021-02-16
[NFC] Workaround compiler issue by renaming some optionals

davidungar:
In some cases, some versions of the compiler cannot cope with `if let foo = foo`. So just rename the variables in the problematic cases.


davidungar:
@swift-ci please test

artemcm (approved):

PR #499, opened 2021-02-16
Emit a remark on interrupt of the driver

artemcm:
This should help distinguish cases where the driver invocation was interrupted by the client (e.g. a build system) vs. exited with a failure on its own.


nkcsgexi (approved):

artemcm:
@swift-ci please test

PR #500, opened 2021-02-17
Add support for symbol graphs and access notes

beccadax:
This commit adds support for the `-access-notes-path`, `-emit-symbol-graph`, and `-emit-symbol-graph-dir` flags to the new driver.

I have also updated Options.swift; I *think* I've handled all of the new options that can be passed into the driver, but I'd like a second pair of eyes to confirm that.

Note that the symbol graph tests still don't pass because of an issue with the frontend's handling of the `-emit-symbol-graph-dir` flag; I've filed rdar://74418885 about that.

No new tests because I couldn't find anywhere that similar passthrough flags were being tested in the driver. If I've missed something, please let me know.

### DIFF
--- a/Sources/makeOptions/makeOptions.cpp
+++ b/Sources/makeOptions/makeOptions.cpp
@@ -59,6 +59,7 @@ enum SwiftFlags {
   ArgumentIsPath = (1 << 12),
   ModuleInterfaceOption = (1 << 13),
   SupplementaryOutput = (1 << 14),
+  SwiftAPIExtractOption = (1 << 15),
 };
 
 static std::set<std::string> swiftKeywords = { "internal", "static" };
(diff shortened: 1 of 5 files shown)


beccadax:
@swift-ci please test

Sources/makeOptions/makeOptions.cpp
```diff
@@ -59,6 +59,7 @@ enum SwiftFlags {
   ArgumentIsPath = (1 << 12),
   ModuleInterfaceOption = (1 << 13),
   SupplementaryOutput = (1 << 14),
+  SwiftAPIExtractOption = (1 << 15),
```
cltnschlosser, line 62: What's this do?

cltnschlosser (approved):
LGTM

cltnschlosser:
> I have also updated Options.swift; I think I've handled all of the new options that can be passed into the driver, but I'd like a second pair of eyes to confirm that.

You used the makeOptions thing to generate the Swift file, right? It looks like you did.

Sources/makeOptions/makeOptions.cpp
beccadax: It was recently added to the Swift compiler in apple/swift#35690; the only reason we care about it is that it's used in Options.td, so we need to support it in the `makeOptions` tool.

QuietMisdreavus:
Paging @bitjammer who implemented `-emit-symbol-graph[-dir]`

QuietMisdreavus (approved):
Symbol graph flags look fine, but i'll also defer to @bitjammer if she has anything else to add.

artemcm (approved):

PR #502, opened 2021-02-18
Optimize performance by replacing a set of nodes with a bit in each node.

davidungar:


davidungar:
@swift-ci please test

davidungar:
rdar://74466708

davidungar:
@swift-ci please test

davidungar:
@CodaFi I am wondering if the driver still needs this transitive machinery at all. For example: if external dependency E is used by A in a fashion that will change B when E changes, will the frontend record that B depends on E? If so, we could remove all the transitive tracing code.

davidungar:
@swift-ci please test

nkcsgexi (approved):
LGTM!

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
```diff
@@ -53,6 +58,14 @@ extension ModuleDependencyGraph {
     }
   }
 }
+
+// MARK: - trace status
+extension ModuleDependencyGraph.Node {
+  var isUntraced: Bool { !isTraced }
+  func beTraced() { isTraced = true }
```
artemcm, line 65: Nit: why not `setTraced`/`setUntraced`? 

Otherwise, LGTM!

artemcm (approved):

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
davidungar: Good idea! Tx for the review. I'll make those changes, retest, and merge.

davidungar:
@swift-ci please test

PR #505, opened 2021-02-19
[Incremental] Fix fingerprint integration bug, add a test, and add debugging affordances.

davidungar:
When integrating a node where the only change is its fingerprint, it's critical to set the fingerprint of the integrated-to node on the module dependency graph to match the node being integrated. That way, if its fingerprint changes back and it gets integrated again, the change will be noticed. 

The above was not happening. This PR adds a test case, fixes the bug, adds a nice `description` to source file dependency nodes (to aid in debugging), and also adds a check to ensure that a graph is serializable.


davidungar:
@swift-ci please test

davidungar:
rdar://74513210

davidungar:
@swift-ci please test

CodaFi (approved):

PR #506, opened 2021-02-19
Interrupt handler: terminate (SIGINT) running compiler processes and allow the driver to exit gracefully.

artemcm:
We need to ensure that we terminate (SIGINT) the running compiler processes without emitting an error that those processes got terminated, and allow the driver to continue execution so that it is able to emit parseable-output which communicates that the terminated compiler jobs were interrupted. 

Resolves rdar://74527749


nkcsgexi (approved):
LGTM!

artemcm:
@swift-ci please test

akyrtzi:
Yay, cancellation that stops the compiler as intended, nice! 🎉

PR #507, opened 2021-02-20
[Incremental] Improve printing for debugging

davidungar:
When printing out swiftdeps and swiftmodules for the debugging remarks and in the dot graphs, just use the basename of the swiftdeps and the parent-directory of the swiftmodules. (Because every swiftmodule has the same basename, the parent carries the info.)

Also, get the source name for a `SourceFileDependencyGraph` from its `DependencySource`, which is statically guaranteed to be present.


davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

CodaFi (approved):

davidungar:
@CodaFi Thanks!

PR #509, opened 2021-02-23
Incremental: ensure we add post-compile jobs if no new compile jobs are found

nkcsgexi:
rdar://74447605

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -30,7 +30,7 @@ public class IncrementalCompilationState {
 
   /// Track required jobs that haven't finished so the build record can record the corresponding
   /// input statuses.
-  private var unfinishedJobs: Set<Job>
+  private var unfinishedCompileJobs: Set<Job>
 
   /// Keyed by primary input. As required compilations are discovered after the first wave, these shrink.
   private var skippedCompileGroups = [TypedVirtualPath: CompileJobGroup]()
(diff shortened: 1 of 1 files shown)


nkcsgexi:
@swift-ci please test

artemcm (approved):
LGTM, @davidungar what do you think?

nkcsgexi:
@swift-ci please test

davidungar (commented):
Thank you for looking at this issue 

I suspect that, with this fix, the post-compile jobs would get to run before the compile-pch job finished.

I'm starting to wonder if a better fix is to remove the early return from `addRuleBeyondMandatoryCompiles` and instead to put an early return in `collectJobsDiscoveredToBeNeededAfterFinishing` , returning empty if a non-compile job?

davidungar:
Hate to say it, but I'm now waffling.
Here's what I'm thinking... in a non-incremental build, maybe there would be nothing to sequence the jobs other than llbuild input-output matching. (Is that right?) So, could all the pre-compile jobs just get run in parallel with the post-compile jobs modulo inputs and outputs? If so, your fix is better: unfinishedJobs -> unfinishedCompileJobs.

nkcsgexi:
Ok, let's merge this for now. @davidungar any advice for adding test for this change?

davidungar (approved):

davidungar:
> Ok, let's merge this for now. @davidungar any advice for adding test for this change?

Good question, @nkcsgexi . I guess we need a test that creates a precompiled header. I would look at `testIncremental` to see what that does. Much of that machinery should be reusable.

davidungar:
One last thought: with this change are we guaranteed that the whole driver process won't terminate before the pch job?

artemcm:
> One last thought: with this change are we guaranteed that the whole driver process won't terminate before the pch job?

I think this job is always going to be a part of the initial set of jobs we submit to llbuild, (not one discovered when other jobs finish), so the `ExecuteAllJobsRule` should ensure it actually gets run before we return from `MultiJobExecutor::execute`.

davidungar:
> > One last thought: with this change are we guaranteed that the whole driver process won't terminate before the pch job?
> 
> I think this job is always going to be a part of the initial set of jobs we submit to llbuild, (not one discovered when other jobs finish), so the `ExecuteAllJobsRule` should ensure it actually gets run before we return from `MultiJobExecutor::execute`.

Thanks! Sounds good to me.

PR #510, opened 2021-02-25
Emit-module separately: fix use with -emit-library and -whole-module-optimization

xymus:
Fix two issue observed with early adopters of `-experimental-emit-module-separately`:
- Pass on the `-parse-as-library` flag to the frontend avoiding considering single file target as a script. rdar://74713625
- Ignore the flag in whole-module builds leaving it to the single job to generate the module files.


xymus:
@swift-ci Please test

xymus:
@swift-ci Please test

artemcm (approved):
LGTM!

PR #512, opened 2021-02-27
Remove a couple of unused test inputs

owenv:
These are leftover from tests using the older swiftdeps format


owenv:
@swift-ci please test

cltnschlosser (approved):

PR #513, opened 2021-03-01
Offer suggestions if it looks like an option was misspelled

owenv:
The implementation here isn't very efficient, but it seems fast enough in practice and the option list grows slowly so I think it'll be fine


owenv:
@swift-ci please test

davidungar (commented):
I love the idea. How about determining the close match algorithmically?

PR #514, opened 2021-03-01
Waiting for all parsable output to be printed before exit

nkcsgexi:


nkcsgexi:
@swift-ci please test

cltnschlosser (approved):

nkcsgexi:
@neonichu does this change make sense to you?

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -64,6 +64,9 @@ public final class MultiJobExecutor {
     /// Queue for executor delegate.
     let delegateQueue: DispatchQueue = DispatchQueue(label: "org.swift.driver.job-executor-delegate")
 
+    /// Create a dispatch group so we can wait on these tasks to be finished.
+    let delegateGroup: DispatchGroup = DispatchGroup()
```
davidungar, line 68: A better name would be the *purpose* of the group, something like "tasksToBeWaitedFor" but better.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -128,9 +131,14 @@ public final class MultiJobExecutor {
       self.processType = processType
     }
 
+    fileprivate func async(execute work: @escaping @convention(block)() -> Void) -> Void {
```
davidungar, line 134: Calling this `async` really confused me because I know we have an `async` keyword coming. What's the purpose of the function here?

davidungar (changes requested):
This is a great idea! But I would really like to see more comments and better names so that someone who has to pick this up and maintain it in the future will know better why the code does what it does.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -64,6 +64,9 @@ public final class MultiJobExecutor {
     /// Queue for executor delegate.
     let delegateQueue: DispatchQueue = DispatchQueue(label: "org.swift.driver.job-executor-delegate")
 
+    /// Create a dispatch group so we can wait on these tasks to be finished.
```
davidungar, line 67: This comment is good, but even better would be another sentence explaining what happens without the group. Something like: "without this group, output could be emitted after the driver seems to exit" but more accurate!

PR #515, opened 2021-03-01
[Integration Tests] Skip currently-failing lit tests

porglezomp:
The lit tests in the integration suite aren't run by default—this marks all of the failing ones as skipped so that it will be safe to make integration testing default for CI.


porglezomp:
@swift-ci please test

porglezomp:
Note this specifically leaves Driver/Dependencies enabled since that's currently passing.

davidungar (approved):

cltnschlosser:
Assuming we don't get all the tests fixed before Xcode 12.5 ships, we should update to https://developer.apple.com/documentation/xctest/3726077-xctexpectfailure once it does :) That way the tests still run.

PR #516, opened 2021-03-02
Add driver support to specify an overriding output path to record in the index data

nathawes:
The frontend supports this via new options -index-unit-output-path and -index-unit-output-path-filelist that mirror -o and -output-filelist. These are intended to allow sharing index data across builds in separate directories (so different -o values) that are otherwise equivalent as far as the index data is concerned (e.g. an ASAN build and a non-ASAN build) by supplying the same -index-unit-output-path for both.

This change updates the driver to add these new options to the frontend invocation 1) when a new "index-unit-output-path" entry is specified for one or more input files in the -output-file-map json or 2) when -index-file is specified (guaranteeing a single primary input) and a new -index-unit-output-path driver option is passed.

Equivalent change for the old driver here: https://github.com/apple/swift/pull/36272

Resolves rdar://problem/74816412

### DIFF
--- a/Sources/SwiftDriver/Jobs/CompileJob.swift
+++ b/Sources/SwiftDriver/Jobs/CompileJob.swift
@@ -26,6 +26,19 @@ extension Driver {
     }
   }
 
+  mutating func computeIndexUnitOutput(for input: TypedVirtualPath, outputType: FileType, topLevel: Bool) -> TypedVirtualPath? {
+    if let path = outputFileMap?.existingOutput(inputFile: input.file, outputType: .indexUnitOutputPath) {
+      return TypedVirtualPath(file: path, type: outputType)
+    }
+    if topLevel {
+      if let baseOutput = parsedOptions.getLastArgument(.indexUnitOutputPath)?.asSingle,
+         let baseOutputPath = try? VirtualPath(path: baseOutput) {
+        return TypedVirtualPath(file: baseOutputPath, type: outputType)
+      }
+    }
+    return nil
+  }
+
   mutating func computePrimaryOutput(for input: TypedVirtualPath, outputType: FileType,
                                         isTopLevel: Bool) -> TypedVirtualPath {
     if let path = outputFileMap?.existingOutput(inputFile: input.file, outputType: outputType) {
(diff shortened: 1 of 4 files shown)


nathawes:
@swift-ci please test

nathawes:
@swift-ci please test

nathawes:
@swift-ci please test

nathawes:
@swift-ci please test

Sources/SwiftDriver/Jobs/CompileJob.swift
```diff
@@ -265,6 +304,20 @@ extension Driver {
       }
     }
 
+    // Add index unit output paths if needed.
+    if !primaryIndexUnitOutputs.isEmpty {
+      if primaryOutputs.count > fileListThreshold {
```
artemcm, line 309: Nit: I know we expect them to be equal, coming out of the producer (`addCompileInputs`), but here it is not immediately obvious why we're using `primaryOutputs` instead of `primaryIndexUnitOutputs` for the count/threshold.

Sources/SwiftDriver/Utilities/FileType.swift
```diff
@@ -114,6 +114,11 @@ public enum FileType: String, Hashable, CaseIterable, Codable {
   /// The extension isn't real.
   case indexData
 
+  /// Output path to record in the indexing data store
+  ///
+  /// This is only needed for use as a key in the output file map.
+  case indexUnitOutputPath
```
artemcm, line 120: Hmm, I'm now realizing that `indexData` is the same pattern; but, everything else here is a file type, whereas `indexData` and `indexUnitOutputPath` are just directories. 
It may be worth refactoring in the future to be a common directory `FileType`, with an enum associated value for the specific kind of directory. But it's alright in this PR.

Sources/SwiftOptions/Options.swift
```diff
@@ -308,6 +308,8 @@ extension Option {
   public static let indexIgnoreSystemModules: Option = Option("-index-ignore-system-modules", .flag, attributes: [.noInteractive], helpText: "Avoid indexing system modules")
   public static let indexStorePath: Option = Option("-index-store-path", .separate, attributes: [.frontend, .argumentIsPath], metaVar: "<path>", helpText: "Store indexing data to <path>")
   public static let indexSystemModules: Option = Option("-index-system-modules", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Emit index data for imported serialized swift system modules")
+  public static let indexUnitOutputPath: Option = Option("-index-unit-output-path", .separate, attributes: [.frontend, .noInteractive, .argumentIsPath], metaVar: "<file>", helpText: "Record <file> as the output file in the produced index data")
```
artemcm, line 311: It looks like these were added manually because they haven't been merged in `swift` yet, right?
This file is automatically generated. Once these options exist in the `swift` repo, please follow instructions here to add them:

artemcm (approved):
The way we generate options in `swift-driver` makes it preferable to first merge the options update in `swift`, and then automatically re-generate here from that and then merge.

But other than that this change looks good!

Sources/SwiftOptions/Options.swift
nathawes: Ah, I didn't realize that. Will do!

nathawes:
@swift-ci test and merge

nathawes:
@swift-ci please test

PR #517, opened 2021-03-02
Block on emitting parseable-output messages in `executeJob`.

artemcm:
`executeJob` itself runs in a concurrent context.
Adding further asynchrony in emitting parseable-output here isn't worth the trouble it can cause because the performance implications of waiting for these messages to be emitted are likely negligible.

Blocking on calls to `jobStarted` and `jobFinished` defines away various possibilities for error states, for example - the executor concluding its workload and returning before all parseable-output messages have been emitted, potentially confusing the build system in the process.

Resolves rdar://74058113

### DIFF
--- a/Sources/SwiftDriverExecution/MultiJobExecutor.swift
+++ b/Sources/SwiftDriverExecution/MultiJobExecutor.swift
@@ -523,7 +523,7 @@ class ExecuteJobRule: LLBuildRule {
       }
 
       // Inform the delegate.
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci please test

nkcsgexi (approved):

davidungar:
Looks OK.

davidungar (commented):
Looks OK.

PR #518, opened 2021-03-02
[Integration Tests] Calculate the swiftRootDir more accurately

porglezomp:
Previously the swiftRootDir would be calculated by taking `swiftBuildDir/../../..`, which is accurate on most developer machines, but it is not accurate for some CI bots that use one fewer layer of directories. We already know the `swiftRootDir` more accurately, we're using it for calculating relative paths, just use that one.


porglezomp:
@swift-ci please test

artemcm (approved):
Thank you!

porglezomp:
This should hopefully fix the failure we saw in https://github.com/apple/swift/pull/36222

PR #522, opened 2021-03-03
Tweak test relying on versioned clang module re-scanning to not need it.

artemcm:
There is an intermittent failure of this test where the order of batch scanning entries affects scanning result.
Temporarily make this test not need this behavior, while we address the issue in the dependency scanner.


artemcm:
@swift-ci please test

shahmishal:
Thanks!

nkcsgexi (approved):

PR #523, opened 2021-03-03
Use 'Incremental Imports' Terminology

CodaFi:
Resync the options flag and drop the experimental cross-module incremental build terminology. The flags are sync'd to ToT Swift at https://github.com/apple/swift/pull/36254.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -60,7 +60,7 @@ public class IncrementalCompilationState {
       ? "Disabling"
       : "Enabling"
     reporter?.report(
-      "\(enablingOrDisabling) incremental cross-module building")
+      "\(enablingOrDisabling) incremental imports")
 
 
     guard let outputFileMap = driver.outputFileMap else {
@@ -120,8 +120,8 @@ extension IncrementalCompilationState {
     /// When this information is combined with the output file map, swiftdeps
     /// files can be located and loaded into the graph.
     ///
-    /// In a cross-module build, the dependency graph is derived from prior
-    /// state that is serialized alongside the build record.
+    /// In a build that is aware of incremental imports, the dependency graph is
+    /// derived from prior state that is serialized alongside the build record.
     let graph: ModuleDependencyGraph
     /// The set of compile jobs we can definitely skip given the state of the
     /// incremental dependency graph and the status of the input files for this
@@ -406,7 +406,7 @@ extension IncrementalCompilationState {
 
 extension IncrementalCompilationState {
   @_spi(Testing) public func writeDependencyGraph() {
-    // If the cross-module build is not enabled, the status quo dictates we
+    // If incremental imports not enabled, the status quo dictates we
     // not emit this file.
     if moduleDependencyGraph.info.areIncrementalImportsDisabled {
       return
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -56,9 +56,9 @@ public class IncrementalCompilationState {
       self.reporter = nil
     }
 
-    let enablingOrDisabling = options.contains(.enableCrossModuleIncrementalBuild)
-      ? "Enabling"
-      : "Disabling"
+    let enablingOrDisabling = options.contains(.disableIncrementalImports)
+      ? "Disabling"
+      : "Enabling"
     reporter?.report(
       "\(enablingOrDisabling) incremental cross-module building")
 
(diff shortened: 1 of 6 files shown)


CodaFi:
@swift-ci test

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -152,12 +152,19 @@ extension Driver {
...
+    // Propagate the disable flag for cross-module incremental builds
+    // if necessary. Note because we're interested in *disabling* this feature,
+    // we consider the disable form to be the positive and enable to be the
+    // negative.
+    if self.parsedOptions.hasFlag(positive: .disableIncrementalImports,
+                                  negative: .enableIncrementalImports,
+                                  default: false) {
+      options.formUnion(.disableIncrementalImports)
+    } else {
       options.formUnion(.readPriorsFromModuleDependencyGraph)
```
CodaFi, line 165: @davidungar If you'd like, I can also flip the polarity on `readPriorsFromModuleDependencyGraph` to be something like `readPriorsFromSourceFileDependencyGraphs`

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -56,11 +56,11 @@ public class IncrementalCompilationState {
       self.reporter = nil
     }
 
-    let enablingOrDisabling = options.contains(.enableCrossModuleIncrementalBuild)
-      ? "Enabling"
-      : "Disabling"
+    let enablingOrDisabling = options.contains(.disableIncrementalImports)
```
davidungar, line 59: What if options contains both?

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -382,18 +382,19 @@ extension IncrementalCompilationState {
...
+    /// the incremental build to ignore dependency information embedded in
+    /// swiftmodule files.
     ///
-    /// FIXME: This option is transitory. We intend to make this the
-    /// default behavior. This option should flip to a "disable" bit after that.
-    public static let enableCrossModuleIncrementalBuild      = Options(rawValue: 1 << 4)
+    /// Using this flag pessimizes the incremental compilation session by forcing
+    /// the driver to only consider changes to the modification time of imported
+    /// swiftmodules, which necessarily requires cascading rebuilds.
+    public static let disableIncrementalImports              = Options(rawValue: 1 << 4)
```
davidungar, line 397: Should there be a similar comment-rewording for the enable flag?

Sources/SwiftDriver/IncrementalCompilation/InitialStateComputer.swift
```diff
@@ -69,7 +69,7 @@ extension IncrementalCompilationState {
       self.readPriorsFromModuleDependencyGraph = maybeBuildRecord != nil &&
         options.contains(.readPriorsFromModuleDependencyGraph)
       self.alwaysRebuildDependents = options.contains(.alwaysRebuildDependents)
-      self.isCrossModuleIncrementalBuildEnabled = options.contains(.enableCrossModuleIncrementalBuild)
+      self.areIncrementalImportsDisabled = options.contains(.disableIncrementalImports)
```
davidungar, line 72: Nit: I tend to favor the bit being true when the feature is enabled.

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -152,12 +152,19 @@ extension Driver {
...
-                                  default: true) {
-      options.formUnion(.enableCrossModuleIncrementalBuild)
+
+    // Propagate the disable flag for cross-module incremental builds
+    // if necessary. Note because we're interested in *disabling* this feature,
+    // we consider the disable form to be the positive and enable to be the
+    // negative.
+    if self.parsedOptions.hasFlag(positive: .disableIncrementalImports,
+                                  negative: .enableIncrementalImports,
+                                  default: false) {
```
davidungar, line 162: Wouldn't it be clearer to write ```
positive: .enableIncrementalImports,
negative: .disableIncrementalImports,
default: true```

davidungar (changes requested):
Thanks for tackling this. I think there is one critical issue:
`let enablingOrDisabling = options.contains(.disableIncrementalImports)`

And in a few places, IMO it would be clearer to use the bit for the positive, even if the positive is the default. (You'll see it in my comments). But that's not an essential change.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
CodaFi: These are the bitflags, not the parsed options.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
CodaFi: I'm not sure what you mean here. The opposite of this flag is `Options(rawValue: ~(1 << 4))`

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Oops! my bad

davidungar (approved):
Would still like to be positive, though.

CodaFi:
Will try to re-invert this bit but use terminology that isn't enable/disable to help.

PR #526, opened 2021-03-03
Temporarily Disable Reading Priors from the Module DepGraph

CodaFi:
Turning this on by default without considering the legacy tests means
those tests that try to e.g. make sure the driver does the right thing
when there are malformed Swiftdeps will not run correctly. Since, with
this flag enabled, we never even read them!


CodaFi:
@swift-ci smoke test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

davidungar (approved):
Looks OK, though I'd rather fix the tests to delete the priors file.

CodaFi:
@swift-ci smoke test

davidungar (commented):
I think we found a better solution offline.

CodaFi:
W

PR #528, opened 2021-03-05
llbuild: model post-compile jobs as a llbuild build rule

nkcsgexi:
Instead of piggy-backing on the last compile job for these post-compile jobs,
we should model them as a proper llbuild rule that requests all other compile jobs
as inputs.

### DIFF
--- a/Sources/SwiftDriverExecution/MultiJobExecutor.swift
+++ b/Sources/SwiftDriverExecution/MultiJobExecutor.swift
@@ -222,9 +222,6 @@ public final class MultiJobExecutor {
         let newJobIndices = Self.addJobs(newJobs, to: &jobs, producing: &producerMap)
         needInputFor(indices: newJobIndices)
       }
-      else {
-        needInputFor(indices: postCompileIndices)
-      }
     }
     fileprivate func needInputFor<Indices: Collection>(indices: Indices)
     where Indices.Element == Int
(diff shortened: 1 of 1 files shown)


nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -222,9 +222,6 @@ public final class MultiJobExecutor {
         let newJobIndices = Self.addJobs(newJobs, to: &jobs, producing: &producerMap)
         needInputFor(indices: newJobIndices)
```
davidungar, line 223: I think that if you leave this in, line 233has to change because the call to `executeAllJobsTaskBuildEngine` in `needInputFor` needs to add the input to your new all-compile-jobs rule.

davidungar (changes requested):
Adding an all-compile-jobs rule is a fine idea, though as far as I know, things are working fine without it. (Must be something I don't know!)
However, required compilation jobs seem to be getting added to the wrong rule. (See my comments.) And gating the completion of the all-compile-jobs rule might be tricky.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
nkcsgexi: We definitely need to change this as well, because it's also jobs added asynchronously in `inputsAvailable`.

davidungar: OK, thanks!

nkcsgexi:
close this in favor of a more complete fix here: https://github.com/apple/swift-driver/pull/535

PR #530, opened 2021-03-05
Do not add precompiled bridging header to a Compile job's display inputs

artemcm:
"Display inputs" are used to convey to the user/client the set of input files that a given compile job will build, e.g. in parseable-output.

The driver is relying on the presence of the bridging PCH in `displayInputs` because the `printActions` code relies on `displayInputs`, when they are present. Unlike parseable-output, action printing is used for compiler debugging purposes only, and given that `printActions` is a best-effort emulation of the legacy driver behavior that doesn't really exist in the new driver, we can just as well get away with using `primaryInputs` and ensure `compile` job actions refer to the corresponding precompiled header, to keep the C++ tests passing (`test/Driver/actions.swift` & `test/Driver/actions-dsym.swift`).

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -1080,8 +1080,13 @@ extension Driver {
       }
       // All input action IDs for this action.
       var inputIds = [UInt]()
+
+      var jobInputs = job.primaryInputs.isEmpty ? job.inputs : job.primaryInputs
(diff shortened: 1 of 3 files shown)


cltnschlosser (approved):
Makes sense, thanks for the documentation updates too. I have a much better understanding of what the differences are now.

nkcsgexi (approved):
Thanks for fixing it!

artemcm:
@swift-ci please test

BenchR267 (approved):
Thanks!

PR #531, opened 2021-03-06
[Incremental Imports; testing NFC] Add new testing framework for incremental imports, etc

davidungar:
The main item here is a new testing framework testing the incremental import feature of the driver.
It uses enums and protocols to get a lot of static checking. A test is typically:

- a `SourceVersion` enum conforming to `SourceVersionProtocol` that defines cases for each version of source code, & defines the filename for each version.
- a `Module` enum conforming to `ModuleProtocol` that defines cases for each module, which modules are libraries vs apps, and which modules import which other modules.
- a `State` enum conforming to `StateProtocol` that defines cases for each state, and the jobs required to put the test in that state. (A job is defined by the module it is building, and the set of source versions being compiled.)
- a struct conforming to `TestProtocol` that defines the start state of the test, and a series of steps to perform. Each step consists of a state to be reached, and expectations for what should be compiled in reaching that state.

With this framework, files can be added, changed, or removed from state to state. Any topology of modules can be specified. Thanks to @CodaFi for suggesting this degree of versatility.

This PR also removes uses of `@testable` imports, as well as incorporating a few tests using the new framework.

### DIFF
--- a/Tests/TestUtilities/DriverExtensions.swift
+++ b/Tests/TestUtilities/DriverExtensions.swift
@@ -34,37 +34,38 @@ extension Driver {
                   executor: executor)
   }
 
-  // For tests that need to set the sdk path:
-  public static func sdkArgumentsForTesting() throws -> [String] {
-    ["-sdk", try cachedSDKPath.get()]
+  /// For tests that need to set the sdk path.
+  /// Only works on hosts with `xcrun`, so return nil if cannot work on current host.
+  public static func sdkArgumentsForTesting() throws -> [String]? {
+    try cachedSDKPath.map {["-sdk", try $0.get()]}
   }
 }
 
-private let cachedSDKPath = Result<String, Error> {
-  if let pathFromEnv = ProcessEnv.vars["SDKROOT"] {
-    return pathFromEnv
-  }
+/// Set to nil if cannot perform on this host
+private let cachedSDKPath: Result<String, Error>? = {
   #if !os(macOS)
-  enum NotMacOS: LocalizedError {
-    case xcrunOnlyOnMacOS
-  }
-  throw NotMacOS.xcrunOnlyOnMacOS
+  return nil
   #endif
-  let process = Process(arguments: ["xcrun", "-sdk", "macosx", "--show-sdk-path"])
-  try process.launch()
-  let result = try process.waitUntilExit()
-  guard result.exitStatus == .terminated(code: EXIT_SUCCESS) else {
-    enum XCRunFailure: LocalizedError {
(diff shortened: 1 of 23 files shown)


davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

davidungar:
(Testing failure with non-verbose mode)

davidungar:
@swift-ci please test

Tests/TestUtilities/DriverExtensions.swift
```diff
@@ -46,7 +45,10 @@ private let cachedSDKPath = Result<String, Error> {
     return pathFromEnv
   }
   #if !os(macOS)
-  throw XCTSkip("xcrun only available on macOS")
+  enum NotMacOS: LocalizedError {
+    case xcrunOnlyOnMacOS
+  }
+  throw NotMacOS.xcrunOnlyOnMacOS
```
CodaFi, line 51: Throwing from this test will cause it to fail on non-Darwin hosts.

davidungar:
@swift-ci please test

Tests/TestUtilities/DriverExtensions.swift
davidungar: Yes, tests that need this fail on those hosts because `xcrun` is not available. Is there a better way? I think when I reorganized this into a separate target, `import XCTest` stopped working without a custom build setting, which I don't know how to do in swiftpm. So it had to just throw an error.
Maybe what I forgot is to ensure that all callers in a test target caught this error and threw `XCSkip` instead?

davidungar: And this is causing the Linux failure in CI. I'm thinking returning an optional with nil in this case is better. Will fix. Thanks.

davidungar:

davidungar:

Tests/TestUtilities/DriverExtensions.swift
CodaFi: You need to throw `XCTSkip` here. Throwing any other exception will just fail.

davidungar:
@swift-ci please test

Tests/TestUtilities/DriverExtensions.swift
davidungar: Thanks. Fixed!

davidungar: @CodaFi The fix is working; it is passing!

davidungar:
Closing in favor of a simpler solution

PR #532, opened 2021-03-08
Remove LinuxMain.swift to enable automatic test discovery on non-Darwin

owenv:
As of https://github.com/apple/swift-package-manager/pull/3053, test discovery is on by default for non-Darwin platforms unless a LinuxMain.swift file is present. If we remove the LinuxMain and XCTTestManifest files, we should be able to adopt it now since swift-driver already more or less requires a nightly toolchain.


owenv:
@swift-ci please test

owenv:
Linux: `Executed 218 tests, with 10 tests skipped and 0 failures (0 unexpected) in 19.174 (19.174) seconds`
macOS: `Executed 218 tests, with 4 tests skipped and 0 failures (0 unexpected) in 49.202 (49.218) seconds`

CodaFi (approved):
Thank you!

PR #535, opened 2021-03-10
llbuild: refactor llbuild rules in MultiJobExecutor

nkcsgexi:
After this change, we now have three rules: (a) ExecuteJobRule, (b) ExecuteAllJobsRule and (c) ExecuteAllCompilationJobsRule. The start callback of (c) requests all mandatory compilations via (a). The provideInput callback of (c) synchronously requests additional incremental build jobs by requesting more (a) as inputs. (b) requests (c) as the only input and runs post-compilation jobs after the only input is ready.

This refactoring allows us to (1) avoid requesting new inputs asynchronously, which may lead to race condition; and (2) avoid keeping track of a LLTaskBuildEngine instance in the context.

### DIFF
--- a/Sources/SwiftDriverExecution/MultiJobExecutor.swift
+++ b/Sources/SwiftDriverExecution/MultiJobExecutor.swift
@@ -347,6 +347,9 @@ struct DriverBuildValue: LLBuildValue {
     return .init(success: success, kind: .jobExecution)
   }
 }
+
+/// A rule represents all jobs to finish compiling a module, including mandatory jobs,
+/// incremental jobs, and post-compilation jobs.
 class ExecuteAllJobsRule: LLBuildRule {
   struct RuleKey: LLBuildKey {
     typealias BuildValue = DriverBuildValue
@@ -368,12 +371,15 @@ class ExecuteAllJobsRule: LLBuildRule {
   }
 
   override func start(_ engine: LLTaskBuildEngine) {
+    // Requests all compilation jobs to be done
     engine.taskNeedsInput(ExecuteAllCompilationJobsRule.RuleKey(), inputID: allCompilationId)
   }
 
   override func provideValue(_ engine: LLTaskBuildEngine, inputID: Int, value: Value) {
     do {
       let subtaskSuccess = try DriverBuildValue(value).success
+      // After all compilation jobs are done, we can schedule post-compilation jobs,
+      // including merge module and linking jobs.
       if inputID == allCompilationId && !context.primaryIndices.isEmpty && subtaskSuccess {
         context.postCompileIndices.forEach {
           engine.taskNeedsInput(ExecuteJobRule.RuleKey(index: $0), inputID: $0)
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -203,36 +193,17 @@ public final class MultiJobExecutor {
...
-    fileprivate func needInputFor<Indices: Collection>(indices: Indices)
-    where Indices.Element == Int
-    {
-      for index in indices {
-        let key = ExecuteJobRule.RuleKey(index: index)
-        executeAllJobsTaskBuildEngine!.taskNeedsInput(key, inputID: index)
+          .collectJobsDiscoveredToBeNeededAfterFinishing(job: jobs[jobIdx]) {
+        return Self.addJobs(newJobs, to: &jobs, producing: &producerMap)
       }
+      return 0..<0
```
davidungar, line 206: I don't understand how this let's the driver distinguish a fixpoint from "no additional jobs needed *at this time*"

davidungar (commented):
I'm probably missing something, but I'm not seeing how a fixpoint is distinguished from: no additional jobs needed at this time.

nkcsgexi:
@davidungar By `fixpoint`, do you mean the additional layer of rule? The additional rule allows us to request post-compilation jobs after compilation jobs are all done.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
nkcsgexi: Not sure I understand the question. An empty range indicates no additional jobs are needed. Those post-compilation jobs are entirely refactored out to another rule.

nkcsgexi:
@swift-ci please test

Sources/SwiftDriverExecution/MultiJobExecutor.swift
nkcsgexi: ah, we don't need to explicitly capture the case of "no additional compile jobs are needed ever" because it's handled implicitly by llbuild for us. After evaluating ExecuteAllCompilationJobsRule, we have both mandatory and incremental jobs finished and ready to schedule post-compilation jobs. This is achieved by introducing the post-compile rule.

davidungar: OK, sounds good. 

If I put myself in the position of a newbie trying to maintain this code, what would really help me is a good comment thoroughly explaining your elegant solution. I would need it to explain what the rule structure is, and why it is what it is, including the complexity of not knowing what to compile before all previous compiles finish, and also the llbuild restrictions being worked around.

Is this a reasonable ask?

nkcsgexi: sure, will add more documentation.

artemcm (approved):
This is a really great refactoring and simplification. LGTM!

nkcsgexi:
@swift-ci please test

Sources/SwiftDriverExecution/MultiJobExecutor.swift
nkcsgexi: @davidungar added more documentation.

nkcsgexi:
@swift-ci please test

PR #537, opened 2021-03-12
[Build Script Helper] Add an option that indicates the driver is being built for use with a local just-built frontend

artemcm:
When used, we mostly just need to be aware that the build is in a single-target mode and to simplify where the build products are placed, so that they are easier to find from the compiler's `build-script`.

This will be used from swift's `build-script` to always build an instance of `swift-driver` using a host toolchain, in order to patch the `swift` build directory with the new driver.


artemcm:
@swift-ci please test

nkcsgexi (approved):

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #538, opened 2021-03-12
[Incremental] Try a cache for external dependency mod times

davidungar:
Playing a hunch to fix a performance issue. 

rdar://75351845


davidungar:
@swift-ci please test

CodaFi:
Do we know if this is the root cause?

davidungar:
> Do we know if this is the root cause?

It's suspicious, given the trace in the radar, but I'm waiting to hear from the person reporting the problem to see if it fixes it. I didn't see enough info the radar to reproduce the issue myself.

davidungar:
> Do we know if this is the root cause?

Fixed the issue, 40 secs -> 2 secs.

CodaFi (approved):
Approved. Please add a regression test that uses a mock file system instance to make sure we don’t see this again.

PR #539, opened 2021-03-12
Use a global DispatchQueue to synchronize parseable outputs and driver error messages

nkcsgexi:
Both of these outputs will be printed to stderr. Without a queue, interleaving them can
lead to malformed parsable output.

rdar://75370288

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -307,28 +307,33 @@ public struct Driver {
   /// A collection of all the flags the selected toolchain's `swift-frontend` supports
   @_spi(Testing) public let supportedFrontendFlags: Set<String>
 
+  /// A global queue for emitting non-interrupted messages into stderr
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

artemcm (approved):
Makes sense to me. Thanks!

nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -307,28 +307,33 @@ public struct Driver {
   /// A collection of all the flags the selected toolchain's `swift-frontend` supports
   @_spi(Testing) public let supportedFrontendFlags: Set<String>
 
+  /// A global queue for emitting non-interrupted messages into stderr
+  public static let stdErrQueue = DispatchQueue(label: "org.swift.driver.emit-to-stderr")
```
cltnschlosser, line 311: Should this be public? Or just internal?

PR #542, opened 2021-03-15
[Incremental] Delay converting external dependency file names to paths as an optimization.

davidungar:
Delay converting external dependency file names to paths as an optimization. For example, on a clean build, the external dependency names might never have to be used as paths; but this PR may not go quite that far. Still more than doubles the speed I measured in a no-priors build.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
@@ -166,7 +166,7 @@ fileprivate extension DependencyKey.Designator {
   static let oneOfEachKind: [DependencyKey.Designator] = [
       .topLevel(name: ""),
       .dynamicLookup(name: ""),
-      .externalDepend(try! ExternalDependency(".")),
+    .externalDepend(try! ExternalDependency(fileName: ".")),
       .sourceFileProvide(name: ""),
       .nominal(context: ""),
       .potentialMember(context: ""),
--- a/Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
@@ -128,7 +128,8 @@ extension ExportableNode {
     key.designator.externalDependency != nil
   }
   fileprivate var isAPINotes: Bool {
-    key.designator.externalDependency?.file.extension == "apinotes"
+    key.designator.externalDependency?.fileName.hasSuffix("apinotes")
+    ?? false
   }
 
   fileprivate var shape: Shape {
--- a/Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
@@ -128,7 +128,7 @@ extension ExportableNode {
     key.designator.externalDependency != nil
   }
   fileprivate var isAPINotes: Bool {
-    key.designator.externalDependency?.fileName.hasSuffix("apinotes")
+    key.designator.externalDependency?.fileName.hasSuffix(".apinotes")
     ?? false
   }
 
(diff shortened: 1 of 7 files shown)


davidungar:
@swift-ci please test

davidungar:
Also, I want to test it's effect before merging.

Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
```diff
@@ -128,7 +128,8 @@ extension ExportableNode {
     key.designator.externalDependency != nil
   }
   fileprivate var isAPINotes: Bool {
-    key.designator.externalDependency?.file.extension == "apinotes"
+    key.designator.externalDependency?.fileName.hasSuffix("apinotes")
```
cltnschlosser, line 131: Should this be `.apinotes` now?

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
```diff
@@ -15,46 +15,56 @@ import TSCBasic
...
 
   /// Cache this
   let isSwiftModule: Bool
 
-  /*@_spi(Testing)*/ public init(_ fileName: String)
-  throws {
-    self.init(try VirtualPath(path: fileName))
+  /*@_spi(Testing)*/ public init(fileName: String) {
+    self.fileName = fileName
+    self.isSwiftModule = fileName.hasSuffix(FileType.swiftModule.rawValue)
```
cltnschlosser, line 27: Check for the `.` here too?

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
```diff
@@ -15,46 +15,56 @@ import TSCBasic
...
   }
 
-  init(_ file: VirtualPath) {
-    self.file = file
-    self.isSwiftModule = file.extension == FileType.swiftModule.rawValue
+  /// Should only be called by debugging functions or functions that are cached
+  private func getPath() -> VirtualPath? {
+    try? VirtualPath(path: fileName)
```
cltnschlosser, line 32: Can this be a lazy property instead? Or would we need a reference type for that?

CodaFi: Agreed. VirtualPath is potentially twice as expensive as AbsolutePath since it calls both validating initializers in sequence. We should create them sparingly.

Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
davidungar: Thanks! Great catch.

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
davidungar: Yes!

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
davidungar: Good idea, since it will only be asked if there is also a fingerprint, I think. I'm wondering if we perhaps ought to canonicalize ExternalDependencies, but there could be concurrency issues there.

davidungar:
Just did some measurements. Very small effect when reading prior`ModuleDependencyGraph`. However when no prior graph was present, I saw a decrease in time from 7.3 seconds to 2.7 secs for the driver. @cltnschlosser , @CodaFi  , would you care to weigh in again?

davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
davidungar: See my comments below; lazy didn't help, but the code looks better.

cltnschlosser (approved):

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
cltnschlosser: Recreating the same path will be cached with the VirtualPath.Handle PR. So if you want to go back to a struct I think that’s fine too.

davidungar:
Thank you, @cltnschlosser

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
davidungar: Thanks for that. At this point, I feel the need to go on to the next optimization, so I'm hitting merge.

PR #545, opened 2021-03-16
Use an Array When Computing Primary Inputs

CodaFi:
Unfortunately, (Typed)VirtualPath's Hashable conformance is wildly expensive to the point that constructing and searching a hashed collection is much more expensive than a linear scan. Preliminary tests show a factor of 4 improvement here by switching to a linear scan for large workloads.

While I'm here, avoid materializing the filtered collection of input files and just use a loop condition. This saves an enormous amount of transient memory during the planning step.

### DIFF
--- a/Sources/SwiftDriver/Jobs/CompileJob.swift
+++ b/Sources/SwiftDriver/Jobs/CompileJob.swift
@@ -133,11 +133,14 @@ extension Driver {
     // If we will be passing primary files via -primary-file, form a set of primary input files so
     // we can check more quickly.
     let usesPrimaryFileInputs: Bool
-    let primaryInputFiles: Set<TypedVirtualPath>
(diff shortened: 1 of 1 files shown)


CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

cltnschlosser:
What makes TypedVirtualPath hash so much slower than it’s equality check?

CodaFi:
From initial profiling, it's dispatching through String's Hashable witnesses, but it's also incurring a retain on the path implementation type. I'm not sure if we're hitting some pathological case in String's implementation of Hashable, but it's enough that I'm going to look into amortizing the cost of hashing these things across the driver.

artemcm (approved):
Looks good to me. 
Happy about these performance improvements!

CodaFi:
⛵

PR #546, opened 2021-03-16
Bump Default Package Version to 10.15

CodaFi:
Xcode 12.3 and 12.4 require macOS 10.15.4.


CodaFi:
@swift-ci test

cltnschlosser (approved):
Should we got to 11? Xcode 12.5 will require Big Sur.

CodaFi:
We should bump again when the bots bump. Right now, ci.swift.org reports we're using `12C33`.

PR #547, opened 2021-03-16
Amortize the Cost of VirtualPath Creation

CodaFi:
`VirtualPath` has a particularly expensive implementation of `Hashable`
that makes using it as the key in hashed collections expensive as well. By
using a `VirtualPath.Handle`, hashing costs are amortized. Additionally,
`VirtualPath.init(path:)` is quite an expensive operation. Interned
`VirtualPath.Handle` instances represent fully-validated paths, so sharing
in the global path table ensures we only pay the cost at most once per
path string.

Benchmarks show a significant decrease in peak heap usage when under heavy load. Additionally, the cost of executing the build is often dominated by the cost of hashing inputs for large projects. Creating and subscripting hash tables of VirtualPath.Handle keys is about 5x faster than `VirtualPath` keys. This leads to a 2-5x reduction in overhead during build planning and execution.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -326,12 +326,11 @@ extension ModuleDependencyGraph {
     if let hasChanged = externalDependencyModTimeCache[externalDependency] {
       return hasChanged
     }
-    guard let depFile = externalDependency.path
-    else {
+    guard let depFile = externalDependency.path else {
       return true
     }
-    let hasChanged = ((try? info.fileSystem.lastModificationTime(for: depFile)) ?? .distantFuture)
-      >= info.buildTime
+    let fileModTime = (try? info.fileSystem.lastModificationTime(for: depFile)) ?? .distantFuture
+    let hasChanged = fileModTime >= info.buildTime
     externalDependencyModTimeCache[externalDependency] = hasChanged
     return hasChanged
   }
@@ -356,7 +355,7 @@ extension OutputFileMap {
     diagnosticEngine: DiagnosticsEngine
   ) -> DependencySource? {
     assert(sourceFile.type == FileType.swift)
-    guard let swiftDepsPath = existingOutput(inputFile: sourceFile.file,
+    guard let swiftDepsPath = existingOutput(inputFile: sourceFile.fileHandle,
                                              outputType: .swiftDeps)
     else {
       // The legacy driver fails silently here.
@@ -365,7 +364,7 @@ extension OutputFileMap {
       )
       return nil
     }
-    assert(swiftDepsPath.extension == FileType.swiftDeps.rawValue)
+    assert(VirtualPath.lookup(swiftDepsPath).extension == FileType.swiftDeps.rawValue)
     let typedSwiftDepsFile = TypedVirtualPath(file: swiftDepsPath, type: .swiftDeps)
     return DependencySource(typedSwiftDepsFile)
   }
@@ -563,7 +562,7 @@ extension ModuleDependencyGraph {
           let hasFingerprint = Int(record.fields[6]) != 0
           let fingerprint = hasFingerprint ? fingerprintStr : nil
           guard let dependencySource = try swiftDepsStr
-                  .map({ try VirtualPath(path: $0) })
+                  .map({ try VirtualPath.intern(path: $0) })
                   .map(DependencySource.init)
           else {
             throw ReadError.unknownDependencySourceExtension
@@ -599,15 +598,17 @@ extension ModuleDependencyGraph {
           }
           let inputPathString = identifiers[Int(record.fields[0])]
           let dependencySourcePathString = identifiers[Int(record.fields[1])]
-          let inputPath = try VirtualPath(path: inputPathString)
-          let dependencySourcePath = try VirtualPath(path: dependencySourcePathString)
+          let inputHandle = try VirtualPath.intern(path: inputPathString)
+          let inputPath = VirtualPath.lookup(inputHandle)
+          let dependencySourceHandle = try VirtualPath.intern(path: dependencySourcePathString)
+          let dependencySourcePath = VirtualPath.lookup(dependencySourceHandle)
           guard inputPath.extension == FileType.swift.rawValue,
                 dependencySourcePath.extension == FileType.swiftDeps.rawValue,
-                let dependencySource = DependencySource(dependencySourcePath)
+                let dependencySource = DependencySource(dependencySourceHandle)
           else {
             throw ReadError.malformedMapRecord
           }
-          let input = TypedVirtualPath(file: inputPath, type: .swift)
+          let input = TypedVirtualPath(file: inputHandle, type: .swift)
           inputDependencySourceMap.append((input, dependencySource))
         case .externalDepNode:
           guard record.fields.count == 2,
--- a/Sources/SwiftDriver/Utilities/VirtualPath.swift
+++ b/Sources/SwiftDriver/Utilities/VirtualPath.swift
@@ -202,7 +202,14 @@ public enum VirtualPath: Hashable {
   }
 
   public static func == (lhs: VirtualPath, rhs: VirtualPath) -> Bool {
-    return lhs.description == rhs.description
+    switch (lhs, rhs) {
+    case (.standardOutput, .standardOutput), (.standardInput, .standardInput):
+      return true
+    case (.standardOutput, _), (.standardInput, _), (_, .standardOutput), (_, .standardInput):
+      return false
+    default:
+      return lhs.description == rhs.description
+    }
   }
 }
 
@@ -212,7 +219,15 @@ extension VirtualPath.Handle: Codable {
   }
 
   public init(from decoder: Decoder) throws {
-    self = try .constant(VirtualPath(from: decoder))
+    let vp = try VirtualPath(from: decoder)
+    switch vp {
+    case .standardOutput:
+      self = .standardOutput
+    case .standardInput:
+      self = .standardInput
+    default:
+      self = vp.intern()
+    }
   }
 }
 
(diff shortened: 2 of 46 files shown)


CodaFi:
@swift-ci test

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -201,6 +198,160 @@ public enum VirtualPath: Hashable {
...
+          let nextSlot = self.table.count
+          self.uniquer[key] = .init(.handle(nextSlot))
+          let path: VirtualPath
+          if let absolute = try? AbsolutePath(validating: key) {
+            path = .absolute(absolute)
+          } else {
+            let relative = try RelativePath(validating: key)
+            path = .relative(relative)
+          }
+          self.table.append(path)
```
cltnschlosser, line 301: I think you need the following to make sure that normalized paths are still equal. Otherwise `/` and `/.` won't be equal, whereas `VirtualPath("\")` is equal to `VirtualPath("\.")`.
```swift
if key != path.pathString {
  self.uniquer[path.pathString] = .init(.handle(nextSlot))
}
```

Actually I think it might need to be a bit more complex:
```swift
        guard let idx = self.uniquer[key] else {
          let nextSlot = self.table.count
          let nextHandle = VirtualPath.Handle(.handle(nextSlot))
          let path: VirtualPath
          if let absolute = try? AbsolutePath(validating: key) {
            path = .absolute(absolute)
          } else {
            let relative = try RelativePath(validating: key)
            path = .relative(relative)
          }
          if path.pathString != key {
            // Normalized path in cache, use existing handle
            if let index = self.uniquer[path.pathString] {
              let originalHandle = VirtualPath.Handle(.handle(index))
              // Cache new key for this handle
              self.uniquer[key] = originalHandle
              return originalHandle
            } else {
              // Cache both input and normalized paths
              self.uniquer[path.pathString] = nextHandle
              self.uniquer[key] = nextHandle
              self.table.append(path)
              return .init(nextHandle)
            }
          } else {
            self.uniquer[key] = nextHandle
            self.table.append(path)
            return .init(nextHandle)
          }
        }
        return idx
```

Can probably clean that up a bit, but should get the idea across.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -593,10 +593,11 @@ extension ModuleDependencyGraph {
...
           let dependencySourcePathString = identifiers[Int(record.fields[1])]
-          let inputPath = try VirtualPath(path: inputPathString)
-          let dependencySourcePath = try VirtualPath(path: dependencySourcePathString)
-          guard inputPath.extension == FileType.swift.rawValue,
-                dependencySourcePath.extension == FileType.swiftDeps.rawValue,
+          let inputPath = try VirtualPath.intern(path: inputPathString)
+          let dependencySourcePath = try VirtualPath.intern(path: dependencySourcePathString)
+          let ext = VirtualPath.lookup(dependencySourcePath).extension
+          guard ext == FileType.swift.rawValue,
+                ext == FileType.swiftDeps.rawValue,
```
cltnschlosser, line 600: Logic changed here, this guard will never be true as currently written.

cltnschlosser (commented):
From the most part looks good. Pretty curious about what is going wrong with hashing the strings, but this should always perform better!

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -201,6 +198,160 @@ public enum VirtualPath: Hashable {
...
+          return .init(.handle(nextSlot))
+        }
+        return idx
```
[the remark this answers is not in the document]
CodaFi: We're amortizing the creation of paths too, so you're correct in that the initializer overall has gotten slower, but from my numbers it looks like an arbitrary path string has a 1 in 7 chance of being re-created so we win in the long run.

Sources/SwiftDriver/Utilities/VirtualPath.swift
CodaFi: This is a very good point, thanks. I think all we need to do is ensure that the canonicalized path and the raw user-provided key have entries pointing to the same handle. That'll ensure that entering e.g. a user-defined path into the cache but looking up a constructed virtual path and vice-versa will always get you a shared entry. So just

```
          self.uniquer[key] = .init(.handle(nextSlot))
          self.uniquer[path.description] = .init(.handle(nextSlot))
```

CodaFi: Good point! I think this is all we need tho

```
    fileprivate func intern(_ key: String) throws -> VirtualPath.Handle {
      return try self.queue.sync(flags: .barrier) {
        guard let idx = self.uniquer[key] else {
          let nextSlot = self.table.count
          let path: VirtualPath
          if let absolute = try? AbsolutePath(validating: key) {
            path = .absolute(absolute)
          } else {
            let relative = try RelativePath(validating: key)
            path = .relative(relative)
          }
          if let existing = self.uniquer[path.description] {
            // If there's an entry for the canonical path for this key, we just
            // need to vend its handle.
            self.uniquer[key] = existing
            return existing
          } else {
            // Otherwise we need to add an entry for the key and its canonical
            // path.
            self.uniquer[path.description] = .init(.handle(nextSlot))
            self.uniquer[key] = .init(.handle(nextSlot))
            self.table.append(path)
            return .init(.handle(nextSlot))
          }
        }
        return idx
      }
    }
```

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -38,12 +40,7 @@ public enum VirtualPath: Hashable {
```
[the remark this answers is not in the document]
CodaFi: That's a great point! But I don't see where they would be entering into the hashed collections here. I also don't see Array's hashable witness showing up in the instruments traces...

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
constants potentially not being interned seems like a footgun. I'll see about removing it entirely. For now, every place that isn't throwing VirtualPaths into dictionaries and sets should be good.

CodaFi:
Perhaps I could ban the hashing of non-interned constants entirely. It would make this approach more future-proof while still allowing clients to opt out of adding entries to the interning table they know are transient.

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -203,6 +200,190 @@ public enum VirtualPath: Hashable {
...
+            // If there's an entry for the canonical path for this key, we just
+            // need to vend its handle.
+            self.uniquer[key] = existing
+            return existing
+          } else {
+            // Otherwise we need to add an entry for the key and its canonical
+            // path.
+            let nextSlot = self.table.count
+            self.uniquer[path.description] = .init(.handle(nextSlot))
+            self.uniquer[key] = .init(.handle(nextSlot))
```
cltnschlosser, line 327: Probably makes no difference, but only need this like if key and path.description are different. If that equality check is significantly less expensive than hashing the string, it might be good to add. Otherwise this is probably best.

cltnschlosser (approved):
I agree the constant handle does open up the potential for accidentally hurting performance.

Should we remove the Hashable conformance of VirtualPath to prevent people from using it that way in the future?

artemcm (approved):
This is pretty cool! The changes look good. 

> Perhaps I could ban the hashing of non-interned constants entirely. It would make this approach more future-proof while still allowing clients to opt out of adding entries to the interning table they know are transient.

I like this idea quite a bit, it'll help folks new to the code-base.

davidungar (commented):
I hate to say this, but I just did a bake-off and, unless I did something wrong, I didn't see any benefit from this PR vs the hash-caching, even without serializing the hashes. In fact, it was a tiny bit slower: 1.10 user time for this one vs 1.08 for the hash-caching. Please hold off on merging till we have more info.
EDIT: Really need to double-check the bake-off results mentioned above. Will do so later today.

cltnschlosser:
https://github.com/apple/swift-driver/pull/557#issuecomment-803645944

@davidungar I think you posted this comment on the wrong PR, but I'll continue here. 

I haven't benchmarked it myself, but my comment was going off of this:
> Benchmarks show a significant decrease in peak heap usage when under heavy load. Additionally, the cost of executing the build is often dominated by the cost of hashing inputs for large projects. Creating and subscripting hash tables of VirtualPath.Handle keys is about 5x faster than VirtualPath keys. This leads to a 2-5x reduction in overhead during build planning and execution.

It makes sense though that improvements in hashing VirtualPaths would affect all builds, we use VirtualPath keyed dictionaries in a quite a few places.

Regarding benchmarks, do you guys have a set of open source projects that you are using?

davidungar:
> [#557 (comment)](https://github.com/apple/swift-driver/pull/557#issuecomment-803645944)
> 
> @davidungar I think you posted this comment on the wrong PR, but I'll continue here.
> 
> I haven't benchmarked it myself, but my comment was going off of this:
> 
> > Benchmarks show a significant decrease in peak heap usage when under heavy load. Additionally, the cost of executing the build is often dominated by the cost of hashing inputs for large projects. Creating and subscripting hash tables of VirtualPath.Handle keys is about 5x faster than VirtualPath keys. This leads to a 2-5x reduction in overhead during build planning and execution.
> 
> It makes sense though that improvements in hashing VirtualPaths would affect all builds, we use VirtualPath keyed dictionaries in a quite a few places.
> 
> Regarding benchmarks, do you guys have a set of open source projects that you are using?

Thanks, @cltnschlosser . Yes, I need to move my benchmarking to an open-source project. As I see it, the benefit of the interning is, as you say, that it helps every project that uses virtual paths, *in which the same paths are used more than once.*
The costs are:
1. By introducing another concept, it adds to the complexity of the code. Every time you write something involving a virtual path, you have to think about whether to use a handle, is it already a handle, where to convert it, etc., and
2. It adds the run-time cost of synchronization. This cost can potentially be non-trivial.

In my opinion, we need to understand the net benefit, not just the reduction in overhead--after all the relative amount of overhead could be large or small--but rather the effect upon the experience of the user. It's not easy, because we have to deal with a plethora of use cases: a project with one module and 1,000 source files, a project with 1,000 stacked modules, each with 2 source files, or whatever.

Perhaps I weigh cost # 1 more heavily than others. But I would love to be convinced by some numbers.

davidungar:
> I agree the constant handle does open up the potential for accidentally hurting performance.
> 
> Should we remove the Hashable conformance of VirtualPath to prevent people from using it that way in the future?

I like this idea, and also removing constant handle.

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
⛵

PR #548, opened 2021-03-17
Add some directives to enable OpenBSD support.

3405691582:
These are not compile blockers, but it makes sense to set these anyway.


CodaFi (approved):

CodaFi:
@swift-ci test

PR #550, opened 2021-03-17
[Incremental Imports; performance] Reserve use capacity when deserializing module dep graph

davidungar:
Reserve space in the `usesByDef` dictionary. May reserve too much.


davidungar:
Can't prove that it pays off; closing.

PR #551, opened 2021-03-18
[Incremental, Optimization] Only verifyUseIsOK in a debug build.

davidungar:
What should have always been an assertion is now an assertion.


davidungar:
@swift-ci please test

cltnschlosser (approved):
Seems weird that these assertions are fatalErrors that are conditionally evaluated. Instead of being assertions. But this change by itself seems fine.

davidungar:
@swift-ci please test

davidungar:
Thank you, @cltnschlosser

davidungar:
> Seems weird that these assertions are fatalErrors that are conditionally evaluated. Instead of being assertions. But this change by itself seems fine.

Good point! I'll fix in another PR.

PR #554, opened 2021-03-19
[Perf] Drop Some Places We Re-Materialize Input Files

CodaFi:
Creating in-memory collections that are derived from the set of input files is potentially expensive for large compilations.


CodaFi:
@swift-ci test

cltnschlosser (approved):

PR #556, opened 2021-03-20
Maintain previously changed public API

BenchR267:
Public API was changed in edd0ee6 but needs to be maintained for compatibility.
ProcessResult will not be used in function body.

rdar://75650542

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -191,21 +191,19 @@ extension Diagnostic.Message {
   }
 }
 
-
 // MARK: - Scheduling the 2nd wave
 extension IncrementalCompilationState {
   /// Remember a job (group) that is before a compile or a compile itself.
   /// `job` just finished. Update state, and return the skipped compile job (groups) that are now known to be needed.
   /// If no more compiles are needed, return nil.
   /// Careful: job may not be primary.
-
   public func collectJobsDiscoveredToBeNeededAfterFinishing(
-    job finishedJob: Job) throws -> [Job] {
+    job finishedJob: Job) throws -> [Job]? {
     // Find and deal with inputs that now need to be compiled
     let invalidatedInputs = collectInputsInvalidatedByRunning(finishedJob)
     assert(Set(invalidatedInputs).isDisjoint(with: finishedJob.primaryInputs),
            "Primaries should not overlap secondaries.")
-
+    
     if let reporter = self.reporter {
       for input in invalidatedInputs {
         reporter.report(
@@ -213,6 +211,13 @@ extension IncrementalCompilationState {
       }
     }
     return try getJobs(for: invalidatedInputs)
+  }
+
+  /// Needed for API compatibility, `result` will be ignored
+  public func collectJobsDiscoveredToBeNeededAfterFinishing(
+    job finishedJob: Job, result: ProcessResult
+  ) throws -> [Job]? {
+    try collectJobsDiscoveredToBeNeededAfterFinishing(job: finishedJob)
   }
 
   /// After `job` finished find out which inputs must compiled that were not known to need compilation before


BenchR267:
@swift-ci test

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -191,16 +191,32 @@ extension Diagnostic.Message {
...
   /// Remember a job (group) that is before a compile or a compile itself.
   /// `job` just finished. Update state, and return the skipped compile job (groups) that are now known to be needed.
   /// If no more compiles are needed, return nil.
   /// Careful: job may not be primary.
+  public func collectJobsDiscoveredToBeNeededAfterFinishing(
+    job finishedJob: Job) throws -> [Job]? {
+    return try collectJobsDiscoveredToBeNeededAfterFinishing(job: finishedJob, result: ProcessResult.dummy)
+  }
 
+  /// Deprecated API, `result` will not be considered.
```
davidungar, line 216: Why is this one the deprecated one?

davidungar (commented):
It's not clear to me what the need for compatibility is. Can we not make a corresponding change in the client? I'm missing something here, sorry.

artemcm:
> It's not clear to me what the need for compatibility is. Can we not make a corresponding change in the client?

The compatibility break here is from the fact that some clients may require building against a current SwiftDriver sometimes, but also against SwiftDriver that contains the [previously-removed](https://github.com/apple/swift-driver/commit/edd0ee6fd0395bbf040e3f1c2d7ecee90e5ee6e4#diff-612f32c953b438ae8f4362d003441a78dc367e2dff8d2bd76d6b671f1c76ca7bL209-L212) API at other times:
```
job finishedJob: Job, result: ProcessResult) throws -> [Job]? {
```
One nit I have is for which one contains the implementation. It'd make it easier to reason about this code without context if our "new" API (without `result`) is where the logic lives, while the one we maintain strictly for compatibility (deprecated) should forward to it.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
BenchR267: This is the version of the API that was changed in edd0ee6. This change breaks clients linking against different versions of SwiftDriver (by linking against it as a library in the toolchain, not as a Swift package). Changing public API in a product that gets distributed like that needs to be additive, so my change brings back the old API, 'marks' it as deprecated and make the new call into it.
Note that I also changed the return type back to an optional.

BenchR267:
> One nit I have is for which one contains the implementation. It'd make it easier to reason about this code without context if our "new" API (without `result`) is where the logic lives, while the one we maintain strictly for compatibility (deprecated) should forward to it.

Oh yeah, absolutely. Thanks! That makes it much clearer.

BenchR267:
@swift-ci please test

artemcm (approved):

davidungar:
> > It's not clear to me what the need for compatibility is. Can we not make a corresponding change in the client?
> 
> The compatibility break here is from the fact that some clients may require building against a current SwiftDriver sometimes, but also against SwiftDriver that contains the [previously-removed](https://github.com/apple/swift-driver/commit/edd0ee6fd0395bbf040e3f1c2d7ecee90e5ee6e4#diff-612f32c953b438ae8f4362d003441a78dc367e2dff8d2bd76d6b671f1c76ca7bL209-L212) API at other times:
> 
> ```
> job finishedJob: Job, result: ProcessResult) throws -> [Job]? {
> ```
> 
> One nit I have is for which one contains the implementation. It'd make it easier to reason about this code without context if our "new" API (without `result`) is where the logic lives, while the one we maintain strictly for compatibility (deprecated) should forward to it.

Thanks, @artemcm for the explanation. And I also very much agree with your "nit".

davidungar (approved):

PR #557, opened 2021-03-21
[Incremental] ModuleDependencyNode hash caching + bad prior recovery

davidungar:
1. Cache the hash of a `ModuleDependencyGraph.Node` in the node.
2. Serialize and deserialize the hash to and from the priors.
3. When a read of the priors fails, stay in incremental mode so as to write a corrected priors.

So far, I've measured on one module with 800 files, this change has cut driver time (when priors are present) by about one third. (Without priors, about 15%). Brings the with-prior time for the new driver down to only 15% more than the old.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -544,27 +544,23 @@ extension ModuleDependencyGraph {
           self.compilerVersionString = compilerVersionString
         case .moduleDepGraphNode:
           let kindCode = record.fields[0]
-          guard record.fields.count == 9,
+          guard record.fields.count == 7,
                 let declAspect = DependencyKey.DeclAspect(record.fields[1]),
-                record.fields[4] < identifiers.count,
-                record.fields[5] < identifiers.count,
+                record.fields[2] < identifiers.count,
+                record.fields[3] < identifiers.count,
                 case .blob(let fingerprintBlob) = record.payload,
                 let fingerprintStr = String(data: fingerprintBlob, encoding: .utf8)
           else {
             throw ReadError.malformedModuleDepGraphNodeRecord
           }
-          let hashHi = record.fields[2]
-          let hashLo = record.fields[3]
-          let hashValue = Int(bitPattern: UInt((hashHi << 32) | hashLo))
-          
-            let context = identifiers[Int(record.fields[4])]
-          let identifier = identifiers[Int(record.fields[5])]
+          let context = identifiers[Int(record.fields[2])]
+          let identifier = identifiers[Int(record.fields[3])]
           let designator = try DependencyKey.Designator(
             kindCode: kindCode, context: context, name: identifier, fileSystem: fileSystem)
           let key = DependencyKey(aspect: declAspect, designator: designator)
-          let hasSwiftDeps = Int(record.fields[6]) != 0
-          let swiftDepsStr = hasSwiftDeps ? identifiers[Int(record.fields[7])] : nil
-          let hasFingerprint = Int(record.fields[8]) != 0
+          let hasSwiftDeps = Int(record.fields[4]) != 0
+          let swiftDepsStr = hasSwiftDeps ? identifiers[Int(record.fields[5])] : nil
+          let hasFingerprint = Int(record.fields[6]) != 0
           let fingerprint = hasFingerprint ? fingerprintStr : nil
           guard let dependencySource = try swiftDepsStr
                   .map({ try VirtualPath(path: $0) })
@@ -574,8 +570,7 @@ extension ModuleDependencyGraph {
           }
           self.finalize(node: Node(key: key,
                                    fingerprint: fingerprint,
-                                   dependencySource: dependencySource,
-                                   hashValue: hashValue))
+                                   dependencySource: dependencySource))
         case .dependsOnNode:
           let kindCode = record.fields[0]
           guard record.fields.count == 4,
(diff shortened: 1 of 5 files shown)


davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
```diff
@@ -84,9 +89,11 @@ extension ModuleDependencyGraph.Node: Equatable, Hashable {
...
     lhs.dependencySource == rhs.dependencySource
   }
-  public func hash(into hasher: inout Hasher) {
-    hasher.combine(keyAndFingerprint.key)
-    hasher.combine(dependencySource)
+  static private func computeHash(_ key: DependencyKey, _ source: DependencySource?) -> Int {
+    var h = Hasher()
+    h.combine(key)
+    h.combine(source)
+    return h.finalize()
```
cltnschlosser, line 96: Does it matter that hashes won’t be stable across builds?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -824,11 +829,15 @@ extension ModuleDependencyGraph {
         .fixed(bitWidth: 3),
         // dependency decl aspect discriminator
         .fixed(bitWidth: 1),
-        // dependency context
+        // hash hi value
+        .fixed(bitWidth: 32),
+        // hash lo value
+        .fixed(bitWidth: 32),
```
cltnschlosser, line 835: Can this be one 64 bit value? Or no? Or is there some advantage to this way?

CodaFi: It's a limitation of bitstream, which only allocates 5 bits of metadata to hold the length of fixed fields.

CodaFi:
I measure a similar speedup with the interning patch. Should I land that instead?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
CodaFi: It might. @davidungar Suppose a prior graph includes a node that should collide with a newly-created node with identical key material.  The hasher in this case will have a different seed and we may miscompile by scheduling a broken build.

CodaFi (changes requested):
We cannot serialize an unstable hash.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
davidungar: Great point!

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: As far as I can tell, the serializer requires it to be broken up.

davidungar:
> We cannot serialize an unstable hash. https://github.com/apple/swift/blob/main/stdlib/public/core/Hasher.swift#L270

Thank you--great point! What we could do, is compute our own hashes, down to the bits.

davidungar:
Thank you both, @cltnschlosser and @CodaFi . It's really great to get this feedback so quickly. Here's what I'm thinking:
1. I'll do a separate PR with just the read-error recover fix.
2. I'll measure how much benefit there is in this sort of fix, without the serialization of the hash.
3. Let's compare the performance with # 2 with the interning patch. I see plusses and minuses comparing the two.
   -- I can do the experiment, or you can, or better yet, we can each experiment on different projects.

What I saw this this PR and the follow-on was that the new driver, with priors, was within 5% of the old driver. That feels to me like where we need to be. So, we'll want to adopt either something like this, or the interning patch.

cltnschlosser:
I think no matter what, we want the interning, because it benefits non incremental builds as well.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
davidungar: I have removed the serialization of hashes from this PR.
A quick bake-off looks like this is slightly faster than with the interning patch added to this. And 10% faster than the interning patch alone. But I want to recheck later today. And invite you all to recheck, too!

davidungar:
@swift-ci please test

davidungar:
> I think no matter what, we want the interning, because it benefits non incremental builds as well.

It adds a fair bit of complexity. So, I would want to measure the benefit before we take it.

davidungar:
In https://github.com/apple/swift-driver/pull/557, @cltnschlosser asserts that we ought to take this PR because it will help non-incremental builds, too. Can we get some measurements before we decide?

davidungar:
@CodaFi Does this version avoid the serialization issue?

davidungar:
I will be busy for a while this afternoon, but will return to the bake-off later today. I'm not sure I trust my results so far.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
CodaFi: OK, but is that 10% across the board or a 10% faster hash? Because it comes at the cost of implementing two deprecated protocol requirements and I don't think it's worth it unless the win is really that big.

CodaFi:
It avoids the serialization issue, but as-written it's going to add two unsilenceable warnings to the build. If this optimization is worthwhile to take on top of the virtual path amortization, please implement `hashValue` accessors under a different name feed those pre-computed values to `hash(into:)` and document the whole shebang.

davidungar:
> It avoids the serialization issue, but as-written it's going to add two unsilenceable warnings to the build. If this optimization is worthwhile to take on top of the virtual path amortization, please implement `hashValue` accessors under a different name feed those pre-computed values to `hash(into:)` and document the whole shebang.

Yes, I'll look at that tomorrow. I saw the warnings, but was another 10% of the total time IIRC! I can share more tomorrow.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Node.swift
davidungar: Yes, 10% across the board. It really is that big, at least in the case I measured.

davidungar: I will explore doing it in a way that doesn't get the warning, as you mentioned elsewhere.

davidungar:
Closed in favor of https://github.com/apple/swift-driver/pull/561 and https://github.com/apple/swift-driver/pull/547

PR #558, opened 2021-03-21
[Incremental] Key hash caching

davidungar:
Builds on https://github.com/apple/swift-driver/pull/557 to also cache `DependencyKey` hashes. Saves another 10% according to one measured case.

The changes beyond the other PR are all in `DependencyKey`

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
@@ -300,6 +300,7 @@ public struct DependencyKey: Hashable, CustomStringConvertible {
 
   /*@_spi(Testing)*/ public let aspect: DeclAspect
   /*@_spi(Testing)*/ public let designator: Designator
+  public let hashValue: Int
 
 
   /*@_spi(Testing)*/ public init(
@@ -308,6 +309,10 @@ public struct DependencyKey: Hashable, CustomStringConvertible {
   {
     self.aspect = aspect
     self.designator = designator
+    var h = Hasher()
+    h.combine(aspect)
+    h.combine(designator)
+    self.hashValue = h.finalize()
   }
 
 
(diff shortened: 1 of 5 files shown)


davidungar:
@swift-ci please test

CodaFi (changes requested):
`hashValue` is not a form of hash interning - you will always incur at least one `hash(into:)` call because none of Swift's keyed collections use `hashValue` alone to assign buckets. This implementation will also add two unsilenceable warnings to the build. If you want to do this, please implement these hash value accessors with a different name and write a `hash(into:)` that feeds that value to it. I guarantee it will be just as fast, if not faster.

CodaFi:
Could you also give me some more context for the 10% speedup? Are you seeing a 10% improvement in the cost of hashing, or 10% across the board? I think the most valuable metric here is the cost of insertion and lookup into hashed collections.

davidungar:
> Could you also give me some more context for the 10% speedup? Are you seeing a 10% improvement in the cost of hashing, or 10% across the board? I think the most valuable metric here is the cost of insertion and lookup into hashed collections.

Assuming I did everything correctly, I tried an invocation of the compiler when the priors were present. I used three versions of the new driver, plus the old driver. Version 1 of the new driver was main, version 2: main + your intern version (I had to rebase), version 3: the other version of the new driver was main + your intern version + the hash optimizations (just as in this PR). Times reported by the time command.

old driver: 0.95 (user), 1.05 (total) 
new driver (main): 1.6 (user), 1.77 (total)
new driver (main + intern): 1.10 (user), 1.24 (total)
new driver (main + intern + hash opts): 0.81 (user), 1.07 (total)

It is certainly possible I did something wrong; would love you to try the experiment for yourself. Contact me in Slack for details.

davidungar:
Closing in favor of https://github.com/apple/swift-driver/pull/560

PR #559, opened 2021-03-21
[Incremental] Recover better from malformed prior ModuleDependencyGraph

davidungar:
If the prior graph gets corrupted, or we change the format, the driver needs to recover from reading a malformed one by writing a good one. By remaining in incremental mode, this change will do that.


davidungar:
@swift-ci please test

CodaFi (approved):
We really oughta write a regression test for this...

davidungar:
> We really oughta write a regression test for this...

Yes.

davidungar:
Thanks, @CodaFi !

PR #562, opened 2021-03-22
Remove VirtualPath's Hashable Conformance

CodaFi:
This conformance was quite expensive. Remove it to force future authors to use VirtualPath.Handle.

This patch is performance-neutral.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -485,7 +485,7 @@ public struct Driver {
     if shouldUseInputFileList {
       let swiftInputs = inputFiles.filter(\.type.isPartOfSwiftCompilation)
       let path = RelativePath(createTemporaryFileName(prefix: "sources"))
-      self.allSourcesFileList = .fileList(path, .list(swiftInputs.map(\.file)))
+      self.allSourcesFileList = .fileList(path, .list(swiftInputs.map(\.fileHandle)))
     } else {
       self.allSourcesFileList = nil
     }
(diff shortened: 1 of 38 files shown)


CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

davidungar (commented):
Looks fine, but I would really like to see some measurement before we land it. Just before-and-after total CPU times, ideally in a pessimal case.

CodaFi:
Stress test total user time varies from about 11.37 in the baseline to 11.56 in this patch. The slight increase in user time can be explained by stats getting slower by the time I ran the variant benchmarks (2.19s baseline vs 2.35s multiplied by each worker thread) which is something we have no control over. Wall time is essentially unchanged.

davidungar:
> Stress test total user time varies from about 11.37 in the baseline to 11.56 in this patch. The slight increase in user time can be explained by stats getting slower by the time I ran the variant benchmarks (2.19s baseline vs 2.35s multiplied by each worker thread) which is something we have no control over. Wall time is essentially unchanged at 1:35 for both the baseline and incremental build of 1000 files.

Hmm... that's not quite the right direction, and I can't tell you sure you are that the increase is just noise. Let me look at the PR again.

davidungar (commented):
I haven't looked over every change, but I am a bit concerned about this change.
The largest concern is that using a handle is likely slower than not unless the hash is computed more than some number of times.  For example, if we had 1,000 files, each of which were only hashed once, it could be significantly slower to use handles for each. What do you think? Am I wrong? Did you already consider this and thus only selectively use the handles? What if I wanted an old-fashioned `TypedVirtualFile` without the `intern` overhead?

CodaFi:
The only places this patch changed were the ones relying on the Hashable conformance. If you want to traffic in a (contiguous) collection of VirtualPaths, that is still an option. The lookups required costs O(n) memops at construction time, which is a cost I have yet to see appear in any traces thus far.

Consider that the cost of interning is amortized across the Driver for *all* uses of a particular handle. Local analysis is not sufficient to see the win this provides - interning is a hot path, sure, but the cost of not re-validating and not re-hashing VirtualPaths in general crushes that.

CodaFi:
This has since become quite stale. If somebody wants to pick this up later, here it is.

PR #563, opened 2021-03-23
Perform Input Modification Checks At the End of the Build

CodaFi:
Also deduplicate them while we're at it. We don't need to ask each job for its input set, we just need to check the same set of files we recorded mod times for in the first place. This saves a quadratic amount of disk I/O and syscalls in the incremental and non-batch builds, which can be a huge wall time savings for large compiles.

In the incremental stress test, 15 seconds are just gone from total wall time. Total user time spent in stating files went from 7 seconds to 20 milliseconds!

### DIFF
--- a/Sources/SwiftDriver/Utilities/VirtualPath.swift
+++ b/Sources/SwiftDriver/Utilities/VirtualPath.swift
@@ -648,7 +648,7 @@ extension TSCBasic.FileSystem {
     try resolvingVirtualPath(path, apply: exists)
   }
 
-  func lastModificationTime(for file: VirtualPath) throws -> Date {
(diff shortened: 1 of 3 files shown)


CodaFi:
@swift-ci test

owenv (approved):
LGTM, thanks!

owenv:
testInputModifiedDuringMultiJobBuild is failing because this is now only reporting one error for a two-job compile, which should be the expected behavior IMO

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
⛵

davidungar (commented):
LGTM

PR #565, opened 2021-03-23
Use a Confinement Queue to Protect IncrementalCompilationState

CodaFi:
This class is serendipitously used only from the main queue for the
Driver's usual use cases. But we have no reason to expect clients of
libSwiftDriver to manipulate this class in the same manner. Provide a serial
confinement queue and protect the mutable members of
IncrementalCompilationState with it.

In the future, we ought to drop the queue and use actors.

rdar://75744005

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -13,7 +13,25 @@ import TSCBasic
 import Foundation
 import SwiftOptions
 
-public class IncrementalCompilationState {
+/// An instance of `IncrementalCompilationState` encapsulates the data necessary
+/// to make incremental build scheduling decisions.
+///
+/// The primary form of interaction with the incremental compilation state is
+/// using it as an oracle to discover the jobs to execute as the incremental
+/// build progresses. After a job completes, call
+/// `IncrementalCompilationState.collectJobsDiscoveredToBeNeededAfterFinishing(job:)`
+/// to both update the incremental state and recieve an array of jobs that
+/// need to be executed in response.
+///
+/// Jobs become "unstuck" as their inputs become available, or may be discovered
+/// by this class as fresh dependency information is integrated.
+///
+/// Threading Considerations
+/// ========================
+///
+/// The public API surface of this class is thread safe, but not re-entrant.
+/// FIXME: This should be an actor.
+public final class IncrementalCompilationState {
 
   /// The oracle for deciding what depends on what. Applies to this whole module.
   @_spi(Testing) public let moduleDependencyGraph: ModuleDependencyGraph
@@ -25,14 +43,23 @@ public class IncrementalCompilationState {
   /// Already batched, and in order of input files.
   public let mandatoryJobsInOrder: [Job]
 
+  /// Jobs to run *after* the last compile, for instance, link-editing.
+  public let jobsAfterCompiles: [Job]
+
+  /// A  high-priority confinement queue that must be used to protect the incremental compilation state.
+  private let confinementQueue: DispatchQueue = DispatchQueue(label: "com.apple.swift-driver.incremental-compilation-state", qos: .userInteractive)
+
   /// Sadly, has to be `var` for formBatchedJobs
+  ///
+  /// After initialization, mutating accesses to the driver must be protected by
+  /// the confinement queue.
   private var driver: Driver
 
   /// Keyed by primary input. As required compilations are discovered after the first wave, these shrink.
+  ///
+  /// This state is modified during the incremental build. All accesses must
+  /// be protected by the confinement queue.
   private var skippedCompileGroups = [TypedVirtualPath: CompileJobGroup]()
-
-  /// Jobs to run *after* the last compile, for instance, link-editing.
-  public let jobsAfterCompiles: [Job]
 
 // MARK: - Creating IncrementalCompilationState if possible
   /// Return nil if not compiling incrementally
@@ -124,10 +151,6 @@ extension IncrementalCompilationState {
     /// for the first wave to execute.
     let mandatoryJobsInOrder: [Job]
   }
-
-
-
-
 }
 
 extension Driver {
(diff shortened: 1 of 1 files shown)


CodaFi:
@swift-ci test

cltnschlosser (approved):

akyrtzi:
It would be good to consider at some point using a `Lock` wrapper over `pthread_mutex_t`, since in general it can be more efficient than locking via dispatch queue.

BenchR267 (approved):

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -13,7 +13,25 @@ import TSCBasic
 import Foundation
 import SwiftOptions
 
-public class IncrementalCompilationState {
+/// An instance of `IncrementalCompilationState` encapsulates the data necessary
+/// to make incremental builds scheduling decisions.
+///
+/// The primary form of interaction with the incremental compilation state is
```
davidungar, line 19: Nice comment! May I offer a suggestion? "The foo is a bar" can usually be strengthened with active voice--who is doing what to whom? Also, here I would be tempted to mention  the first wave, too. Something like:

"When performing an incremental compilation, the driver must determine which source files much be compiled and which may be skipped. This determination is complicated by the fact that one compilation may reveal the need to perform another compilation if, for instance, the user has changed a declaration in one file that another (unchanged) file, depends upon. These decisions are made here."

Then follow up with the "how":

"Additionally the IncrementalCompilationState must account for the batching of compilations into jobs."

And a chronology:

"It starts with the planning process, which creates the jobs for a non-incremental compilation and passes them of to the InitialCompilationState constructor, which builds the ModuleDependencyGraph and initializes two key fields: mandatoryJobsInOrder, and jobsAfterCompiles. The driver will then schedule the mandatory jobs, which are always needed.  This phase is also called the first wave.

Next, as each job finishes, collectJobsDiscoveredToBeNeededAfterFinishing will discover additional jobs that must be run. collectJobsDiscoveredToBeNeededAfterFinishing is also called after each of these jobs. This phase is called the "second wave". When there is no longer any possibility of requiring additional jobs, the driver will run jobsAfterCompiles."

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -222,17 +245,21 @@ extension IncrementalCompilationState {
...
-    return job.primaryInputs.reduce(into: Set()) { invalidatedInputs, primaryInput in
-      invalidatedInputs.formUnion(collectInputsInvalidated(byCompiling: primaryInput))
+    return self.confinementQueue.sync {
+      guard job.kind == .compile else {
+        return Set<TypedVirtualPath>()
+      }
+      return job.primaryInputs.reduce(into: Set()) { invalidatedInputs, primaryInput in
+        invalidatedInputs.formUnion(collectInputsInvalidated(byCompiling: primaryInput))
+      }
+      .subtracting(job.primaryInputs) // have already compiled these
```
davidungar, line 255: I think that, instead of protected part of `collectInputsInvalidatedByRunning` and then separately protecting the body of `getJobs`, we need to just protect the whole body of `collectInputsInvalidatedByRunning`. I'm worried about another thread getting in and making changes in between the two synchronizations. In that event, the call to `getJobs` might see a state inconsistent with the rest of collectInputs...

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -242,34 +269,41 @@ extension IncrementalCompilationState {
   }
 
   /// Find the jobs that now must be run that were not originally known to be needed.
-  private func getJobs(for invalidatedInputs: Set<TypedVirtualPath>
+  private func getJobs(
```
davidungar, line 272: See above comment. I would protect this in the caller.

davidungar (changes requested):
Thank you for undertaking this fix. I would like to see two changes:
1. You wrote a nice comment, but I think it could be better. How about something more like what I suggested?
2. More importantly, unless I'm missing something, the synchronization in `getJobs` does not seem to be sufficient. Again, see my inline comments. Do you agree?

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
CodaFi: I suppose I don't understand, because the critical section extends to the entire body of this function already...

CodaFi: Did you mean `collectJobsDiscoveredToBeNeededAfterFinishing`?

davidungar: Yes, that's right. Sorry I wrote the wrong thing. Yes the whole of `collectJobsDiscoveredToBeNeededAfterFinishing` needs to be protected.

CodaFi:
@swift-ci test

davidungar (approved):
LGTM, thanks!

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Yes, a document and a pointer is good, but that proposal does not answer my objection:
What bothers me is the general practice if comments that merely summarize the "what" without much, if any, of the "why". The comment, as is, feels like a "what" summary, with very little why. 

Why do we have comments in the code at all? To help future maintainers. The most vital help is that which cannot be understood from reading the code, especially non-local information. I don't mean to single this PR out; many comments I read in the whole llvm/swift-compiler community fail this test.

Sure, what I wrote may feel like a bit much, but as it stands--no such document as you propose exists as of this writing--your comment could be more helpful by adding a little more of the nonobvious, nonlocal information.

However, I won't block the PR for this. I will, however, *unresolve* the conversation.

CodaFi:
@swift-ci test

CodaFi:
⛵

PR #566, opened 2021-03-23
[Test] Harden Explicit Module Build test

artemcm:
Make sure we always find the stdlib on Darwin, and use custom "fake" interfaces for stdlib and OnoneSupport in the dependency scanning test.


artemcm:
@swift-ci please test

nkcsgexi (approved):
Makes sense to me.

PR #567, opened 2021-03-24
Add an option that causes individual `swift-frontend` invocations to emit parseable output, instead of the driver.

artemcm:
Resolves rdar://75787629


artemcm:
@swift-ci please test

nkcsgexi (approved):
LGTM!

PR #569, opened 2021-03-24
[Incremental, Optimization] Defer use hashing when reading a ModuleDependencyGraph

davidungar:
When deserializing a `ModuleDependencyGraph`, don't record uses in a dictionary twice. Just save in an array while reading. On a project with 1,000 files, when rebuilding after just adding a newline to one file, swift-driver-new time in Instruments was reduced from 5.3 secs to 4.7.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -479,7 +479,7 @@ extension ModuleDependencyGraph {
       // The empty string is hardcoded as identifiers[0]
       private var identifiers: [String] = [""]
       private var currentDefKey: DependencyKey? = nil
-      private var nodeUses: [DependencyKey: [Int]] = [:]
+      private var nodeUses: [(DependencyKey, Int)] = []
       private var inputDependencySourceMap: [(TypedVirtualPath, DependencySource)] = []
       public private(set) var allNodes: [Node] = []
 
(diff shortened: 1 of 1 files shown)


davidungar:
@swift-ci please test

cltnschlosser (approved):

davidungar:
Thank you, @cltnschlosser !

davidungar:
@CodaFi , would you like to review before this lands?

CodaFi (approved):
I suspect the win you're seeing is because we're no longer hashing DependencyKeys. There's no double registration here AFAICT.

davidungar:
> I suspect the win you're seeing is because we're no longer hashing DependencyKeys. There's no double registration here AFAICT.
Not sure why you thought of double registration. But there was double dictionary access. Thanks for reviewing!

CodaFi:
There is no double access. The mutation here goes through the modify coroutine. We don't lookup once to read and once to append https://github.com/apple/swift/blob/main/stdlib/public/core/Dictionary.swift#L893

davidungar:
> There is no double access either. The mutation here goes through the modify coroutine. We don't lookup once to read and once to append https://github.com/apple/swift/blob/main/stdlib/public/core/Dictionary.swift#L893

I understand that. The other access is here: ```
         let isNewUse = self.graph.nodeFinder
            .record(def: dependencyKey, use: self.allNodes[useID])
```
inside `record`. There was one dictionary access while reading, and another while finalizing.

PR #572, opened 2021-03-26
[Parsable Output] Break down batch compile jobs into constituent messages

artemcm:
We have, historically, treated batch mode as an implementation detail of the compiler. As a consequence, we have been lying to clients of parsable-output by emitting multiple `began`, `finished`, `signalled` messages for each batch job - one message per primary input. The time has come to teach swift-driver to tell the same old lies. 

Emulating existing C++ driver behavior:
    - Cargo-cult the concept of a `quasiPID`
    (https://github.com/apple/swift/blob/main/include/swift/Basic/ParseableOutput.h#L36)
    - Break up one compile job with multiple primaries into multiple `[Began|Finished|Signalled]Message` messages
    - Collect `((job, primaryInput) -> quasiPID)` information to generate correct `Finished` messages
    - Fixup individual command lines of each `Began` message (w.r.t. `-o` and `-primary-input` options)
    - Make `InputOutputMap` a part of a `.compile` `Job`'s state so that we can later query which of its many outputs correspond to a specific primary

Resolves rdar://75636235

### DIFF
--- a/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
+++ b/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
@@ -24,6 +24,15 @@ import Glibc
 
 /// Delegate for printing execution information on the command-line.
 final class ToolExecutionDelegate: JobExecutionDelegate {
+  /// Quasi-PIDs are _negative_ PID-like unique keys used to
+  /// masquerade batch job constituents as (quasi)processes, when writing
+  /// parseable output to consumers that don't understand the idea of a batch
+  /// job. They are negative in order to avoid possibly colliding with real
+  /// PIDs (which are always positive). We start at -1000 here as a crude but
+  /// harmless hedge against colliding with an errno value that might slip
+  /// into the stream of real PIDs (say, due to a TaskQueue bug).
+  static let QUASI_PID_START = -1000
+
   public enum Mode {
     case verbose
     case parsableOutput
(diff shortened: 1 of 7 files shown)


Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
```diff
@@ -151,6 +159,187 @@ final class ToolExecutionDelegate: JobExecutionDelegate {
...
+  func constructSingleSignalledMessage(error: String, output: String?, signal: Int32,
+                                       pid: Int, realPid: Int)
+  -> SignalledMessage {
+    return SignalledMessage(pid: pid, realPid: realPid, output: output,
+                            errorMessage: error, signal: Int(signal))
+  }
+
+  /// Best-effort attempt to "fix-up" the individual swift-frontend invocation command line, to pretend
+  /// it is an individual single-primary compile job, rather than a batch mode compile with multiple primaries
+  static func filterPrimaryArguments(in arguments: [String],
```
nkcsgexi, line 306: Is this really necessary? I think it's a good idea to keep the command line as truthful as possible so we could understand/debug the compiler.

artemcm: That's a fair point. The C++ driver first generates individual jobs, and then forms messages using those. Which results in these "fake" jobs actually having usable command-lines that are not batch-mode. But here, since the new driver generates batched jobs directly, I thought it may help to try and also generate commands that look like individual compiles.

But, as you mention, being able to copy these from the build log and debug using them seems much more important so, on second thought, I think we should just show the command line as the identical batch mode invocation for all the individual batch entry. Since that is more truthful to what the compiler is doing.

artemcm:
@swift-ci please test

nkcsgexi (approved):

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #573, opened 2021-03-29
Add `free()` workaround that causes error when built with some compiler versions.

artemcm:
Otherwise some building compilers will see errors:
```
error: value of optional type 'UnsafeMutablePointer<CChar>?' (aka 'Optional<UnsafeMutablePointer<Int8>>') must be unwrapped to a value of type 'UnsafeMutablePointer<CChar>' (aka 'UnsafeMutablePointer<Int8>')
  for ptr in cstrings { free(ptr) }
```


artemcm:
@swift-ci please test

nkcsgexi (approved):
Thanks!

PR #576, opened 2021-03-31
Set dsym-job's output to be the primary output file + `.dSYM`

artemcm:
Instead of taking the input file and replacing its extension with `.dSYM`.
This change causes the behavior to match what the C++ driver does and what LLDB expects.

Resolves rdar://76073797


artemcm:
@swift-ci please test

nkcsgexi (approved):
Thank you for fixing this!

PR #577, opened 2021-04-01
[Merge Modules] Omit merge-modules job when only one input is given.

artemcm:
And said input matches the expected output path.

Otherwise both the `.compile` and the `.mergeModule` jobs have an identical output:
`.temporary(<module_name>.swiftmodule)` which will trigger a fatal error in the MultiJobExecutor because it may cause LLBuild cycles downstream.

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -392,6 +392,16 @@ extension Driver {
           !(moduleInputs.isEmpty && moduleInputsFromJobOutputs.isEmpty),
           compilerMode.usesPrimaryFileInputs
     else { return nil }
+
+    // If there is only 1 module input to this job and its path matches that which we intend
+    // to produce with this merge-modules job, skip it
+    let mergeOutput = moduleOutputInfo.output!.outputPath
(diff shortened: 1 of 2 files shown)


nkcsgexi:
Are we certain that we will use the correct output file name `Foo.swiftmodule` instead of `filename~particial.swiftmodule` in the frontend command?

artemcm:
> Are we certain that we will use the correct output file name `Foo.swiftmodule` instead of `filename~particial.swiftmodule` in the frontend command?

We ensure with this check:
```
soleInput.fileHandle == mergeOutput
```
That this will only omit merge-modules if the final module output matches that from the compile job.
In a plain compile with one source file and with `-emit-module`, we will have the situation you describe where we have one module `Foo~partial.swiftmodule` in a temporary directory somewhere and the output of merge-modules is `Foo.swiftmodule` where the user wants it. In that case, the job will not be skipped.

This is motivated by the `-g` case where the final module output is in a temporary directory and matches *exactly* the module output of the single-source-file.

nkcsgexi:
ah, that makes sense. Thank you for the explanation! @artemcm

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #578, opened 2021-04-01
Revert "[Merge Modules] Omit merge-modules job when only one input is given."

artemcm:
Reverts apple/swift-driver#577

Because we are relying on the presence of merge-modules job for emitting interface files in various contexts (only -emit-module jobs are capable of that).


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #579, opened 2021-04-02
Implement swift_build_sdk_interfaces.py using libSwiftDriver

nkcsgexi:
This PR is an implementation of swift_build_sdk_interfaces.py using
libSwiftDriver. The tool is used to traverse an entire SDK to find
Swift interface files and to generate prebuilt module cache for each one
of them. Comparing to the python implementation, this implementation generates
the prebuilt modules in the dependency order so that:

(1) It's more performant than the python counterpart. Several preliminary
experiments showed an average of 40% less CPU time spent on generating modules in an SDK.

(2) It fails fast. If an interface is broken in the SDK, this new implementation
won't schedule jobs for building other modules depending on this interface. It should help us avoid
issues like rdar://76201633

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,6 +21,9 @@ let package = Package(
     .executable(
       name: "swift-help",
       targets: ["swift-help"]),
+    .executable(
+      name: "swift-build-sdk-interfaces",
+      targets: ["swift-build-sdk-interfaces"]),
     .library(
       name: "SwiftDriver",
       targets: ["SwiftDriver"]),
@@ -92,6 +95,11 @@ let package = Package(
     .target(
       name: "swift-help",
       dependencies: ["SwiftOptions", "ArgumentParser", "SwiftToolsSupport-auto"]),
+
+    /// The help executable.
+    .target(
+      name: "swift-build-sdk-interfaces",
+      dependencies: ["SwiftDriver", "SwiftDriverExecution"]),
 
     /// The `makeOptions` utility (for importing option definitions).
     .target(
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -496,7 +496,7 @@ extension Driver {
     return try explicitDependencyBuildPlanner!.generateExplicitModuleDependenciesBuildJobs()
   }
 
-  private mutating func gatherModuleDependencies()
+  mutating func gatherModuleDependencies()
   throws -> InterModuleDependencyGraph {
     var dependencyGraph = try performDependencyScan()
 
@@ -550,6 +550,7 @@ extension Driver {
   }
 
 }
+
 
 /// MARK: Planning
 extension Driver {
--- a/Sources/swift-build-sdk-interfaces/main.swift
+++ b/Sources/swift-build-sdk-interfaces/main.swift
@@ -1,0 +1,142 @@
+import SwiftDriverExecution
+import SwiftDriver
+import TSCLibc
+import TSCBasic
+import TSCUtility
+
+let diagnosticsEngine = DiagnosticsEngine(handlers: [Driver.stderrDiagnosticsHandler])
+
+guard let sdkPathRaw = ProcessEnv.vars["SDKROOT"] else {
+  diagnosticsEngine.emit(.error("need to set SDKROOT"))
+  exit(1)
+}
+
+var rawOutputDir = ""
+if let oid = CommandLine.arguments.firstIndex(of: "-o") {
+  let dirId = oid.advanced(by: 1)
+  if dirId < CommandLine.arguments.count {
+    rawOutputDir = CommandLine.arguments[dirId]
+  }
+}
+if rawOutputDir.isEmpty {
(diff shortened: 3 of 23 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -496,6 +496,92 @@ extension Driver {
...
+      for dep in dependencies {
+        if let path = getOutputPath(dep.moduleName) {
+          results.append(path)
+        }
+      }
+      return results
+    }
+
+    func addCommonFlags(commandLine: inout [Job.ArgTemplate]) {
+      commandLine.appendFlag(.buildModuleFromParseableInterface)
```
harlanhaskins, line 545: tiny nitpick: the current version of this is `compile-module-from-interface`

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -496,6 +496,84 @@ extension Driver {
...
+            commandLine.appendFlag(.sdk)
+            commandLine.append(.path(VirtualPath.lookup(sdkPath)))
+            commandLine.appendFlag(.prebuiltModuleCachePath)
+            commandLine.appendPath(prebuiltModuleDir)
+            commandLine.appendFlag(.moduleName)
+            commandLine.appendFlag(moduleName)
+            commandLine.appendFlag(.o)
+            commandLine.appendPath(outputPath.file)
+            commandLine.appendPath(inputPath.file)
+            jobs.append(Job(
```
artemcm, line 560: Could we re-use the `generateExplicitModuleDependenciesBuildJobs` here?
It should be sufficient to instantiate a `ExplicitDependencyBuildPlanner` with the `dependencyGraph` here and invoke this method. 

Or do we not want the _explicit_ aspect of the jobs that planner generates yet?

nkcsgexi: That's a great point and was my original thoughts too! However, I found the command in explicit module builds are unnecessarily complicated for prebuilt module generation. Also, there are some differences, e.g. prebuilt module generation doesn't like `-track-system-dependencies` for performance reasons. So I've decided to keep them separate and simple.

Sources/SwiftDriver/Jobs/Planning.swift
nkcsgexi: Fixed!

artemcm (approved):

Package.swift
```diff
@@ -21,6 +21,9 @@ let package = Package(
     .executable(
       name: "swift-help",
       targets: ["swift-help"]),
+    .executable(
+      name: "swift-build-sdk-interfaces",
```
artemcm, line 25: I think some changes in `build-script-helper.py` are required to make sure this gets installed when built with `build-script`.

artemcm (approved):

Package.swift
nkcsgexi: Yeah, will make those build-script/cmake changes later.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -79,12 +79,6 @@ extension Driver {
       commandLine.appendFlag(Triple(variant, normalizing: true).triple)
     }
 
-    // Enable address top-byte ignored in the ARM64 backend.
```
artemcm, line 82: Is this no longer needed?

Sources/swift-build-sdk-interfaces/main.swift
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+do {
+  let sdkPath = try VirtualPath(path: sdkPathRaw).absolutePath!
+  let swiftcPathRaw = ProcessEnv.vars["SWIFT_EXEC"]
+  var swiftcPath: AbsolutePath
+  if let swiftcPathRaw = swiftcPathRaw {
+    swiftcPath = try VirtualPath(path: swiftcPathRaw).absolutePath!
+  } else {
+    swiftcPath = sdkPath.parentDirectory.parentDirectory.parentDirectory
```
artemcm, line 45: Do you think it could be more reliable to use:
`xcrun --toolchain default --find swiftc`?

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
nkcsgexi: ah, I accidentally removed this. Will add back.

Sources/swift-build-sdk-interfaces/main.swift
nkcsgexi: We should be always using the default toolchain in the same Xcode bundle with the target SDK to generate prebuilt modules. The only exception is for our local testing purposes, and we can specify `SWIFT_EXEC` in that case.

nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

PR #580, opened 2021-04-03
[Incremental] Skip post-compile jobs only if no compiles ran, and all the outputs are up-to-date.

davidungar:
If the merge-modules step fails or is cancelled, all the compile steps may be skipped, yet the post-compile jobs still must run.
Improve the logic that was skipping the post-compile jobs whenever there are no compile jobs.
When the compile jobs would otherwise be skipped, check their outputs against the time the previous build finished.
In order to perform the check, enhance the build record to store the end time as well as the start time of the prior build.

rdar://76127929

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -46,6 +46,14 @@ public final class IncrementalCompilationState {
   /// Jobs to run *after* the last compile, for instance, link-editing.
   public let jobsAfterCompiles: [Job]
 
+  /// Can compare input mod times against this.
+  private let buildStartTime: Date
+
+  /// In order to decide while postCompile jobs to run, need to compare their outputs against the build time
+  private let buildEndTime: Date
+
+  private let fileSystem: FileSystem
+
   /// A  high-priority confinement queue that must be used to protect the incremental compilation state.
   private let confinementQueue: DispatchQueue = DispatchQueue(label: "com.apple.swift-driver.incremental-compilation-state", qos: .userInteractive)
 
@@ -124,6 +132,9 @@ public final class IncrementalCompilationState {
     self.mandatoryJobsInOrder = initial.mandatoryJobsInOrder
     self.jobsAfterCompiles = jobsInPhases.afterCompiles
     self.moduleDependencyGraph = initial.graph
+    self.buildStartTime = initial.buildStartTime
+    self.buildEndTime = initial.buildEndTime
+    self.fileSystem = driver.fileSystem
     self.driver = driver
   }
 }
@@ -150,6 +161,10 @@ extension IncrementalCompilationState {
     /// All of the pre-compile or compilation job (groups) known to be required
     /// for the first wave to execute.
     let mandatoryJobsInOrder: [Job]
+    /// The last time this compilation was started. Used to compare against e.g. input file mod dates.
+    let buildStartTime: Date
+    /// The last time this compilation finished. Used to compare against output file mod dates
+    let buildEndTime: Date
   }
 }
 
@@ -288,6 +303,14 @@ extension IncrementalCompilationState {
       }
     }
     return try driver.formBatchedJobs(unbatched, showJobLifecycle: driver.showJobLifecycle)
+  }
+}
+// MARK: - Scheduling post-compile jobs
+extension IncrementalCompilationState {
+  public func canSkipPostCompile(job: Job) -> Bool {
+    job.outputs.allSatisfy {output in
+      let fileModTime = (try? fileSystem.lastModificationTime(for: output.file)) ?? .distantFuture
+      return fileModTime <= buildEndTime}
   }
 }
 
--- a/Sources/SwiftDriver/IncrementalCompilation/BuildRecord.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/BuildRecord.swift
@@ -19,24 +19,32 @@ public struct BuildRecord {
   public let swiftVersion: String
   /// When testing, the argsHash may be missing from the build record
   public let argsHash: String?
-  public let buildTime: Date
+  /// Next compile, will compare an input mod time against the start time of the previous build
+  public let buildStartTime: Date
+  /// Next compile, will compare an output mod time against the end time of the previous build
+  public let buildEndTime: Date
   /// The date is the modification time of the main input file the last time the driver ran
   public let inputInfos: [VirtualPath: InputInfo]
 
   public init(argsHash: String?,
               swiftVersion: String,
-              buildTime: Date,
+              buildStartTime: Date,
+              buildEndTime: Date,
               inputInfos: [VirtualPath: InputInfo]) {
     self.argsHash = argsHash
     self.swiftVersion = swiftVersion
-    self.buildTime = buildTime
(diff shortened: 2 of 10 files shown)


davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -46,6 +46,14 @@ public final class IncrementalCompilationState {
   /// Jobs to run *after* the last compile, for instance, link-editing.
   public let jobsAfterCompiles: [Job]
 
+  /// Can compare input mod times against this.
+  private let buildStartTime: Date
+
+  /// In order to decide while postCompile jobs to run, need to compile their outputs against the build time
```
cltnschlosser, line 52:
```suggestion
  /// In order to decide while postCompile jobs to run, need to compare their outputs against the build time
```

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -290,6 +305,14 @@ extension IncrementalCompilationState {
     return try driver.formBatchedJobs(unbatched, showJobLifecycle: driver.showJobLifecycle)
   }
 }
+// MARK: - Scheduling post-compile jobs
+extension IncrementalCompilationState {
+  public func canSkipPostCompile(job: Job) -> Bool {
+    job.outputs.allSatisfy {output in
+      let fileModTime = (try? fileSystem.lastModificationTime(for: output.file)) ?? .distantFuture
+      return fileModTime < buildEndTime}
```
cltnschlosser, line 313: `<=`?

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -290,6 +305,14 @@ extension IncrementalCompilationState {
     return try driver.formBatchedJobs(unbatched, showJobLifecycle: driver.showJobLifecycle)
   }
 }
+// MARK: - Scheduling post-compile jobs
+extension IncrementalCompilationState {
+  public func canSkipPostCompile(job: Job) -> Bool {
+    job.outputs.allSatisfy {output in
+      let fileModTime = (try? fileSystem.lastModificationTime(for: output.file)) ?? .distantFuture
```
cltnschlosser, line 312: Is this the function that just does stat?

cltnschlosser (approved):

davidungar:
FYI, I plan to add a regression test before landing this PR.

davidungar:
@cltnschlosser Thank you for such a quick review! I'll look at your concerns today.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Good catch! A thinko on my part.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Hmmm.... Good question. `<` counts on there being a little extra time for the driver to finish up... Sure, `<=` might be better. Thanks for asking the question.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Not sure what you are asking. If you mean to ask why `canSkipPostCompile` is a separate function, it is because I have found it helpful to use "intention-revealing names", factoring functions where needed: http://wiki.c2.com/?IntentionRevealingNames .

OTOH, if you are asking about `fileSystem.lastModificationTime`, the answer is "yes".

Let me know if I misunderstood. TIA!

davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
cltnschlosser: Yeah the lastModificationTime, I was just double checking that this was the new function that calls stat directly. It’s much faster than the other one that did something like getFileInfo().modTime

davidungar: Thanks for clarifying.

davidungar:
@swift-ci please test

CodaFi:
Err, can we change the format of the build record? If the user needs to fall back/upgrade to or from the old and new drivers, will this break anything?

davidungar:
> Err, can we change the format of the build record? If the user needs to fall back/upgrade to or from the old and new drivers, will this break anything?

Thanks for taking a look at this PR. I'm not sure I understand. The new build end time has a new section name, and the code should accept either the old name for the old time or the new name for the old time. At least, that's the code I thought I wrote.
If the user upgrades and there are files to be compiled, it should just work as before, writing out the more complete build record.
If the user upgrades and there are no files to be compiled, the post-compile jobs might run one time, even though they are not needed.
If the user downgrades, the read of the build record start date will fail, and everything will be recompiled on the first post-downgrade compile.
The tradeoff here is between using a clearer name for the start date in the build record, and avoiding the recompilation-when-downgrading penalty. I had chosen the former, but did you mean to argue for the latter? If so, what is the rationale? I'm open to it.

CodaFi:
The question is more base than that: This new format appears to replace a key from the legacy driver’s build record format outright. So, if I use the new driver and need to downgrade then the old driver will be missing information. Is the legacy driver tolerant of that use case? Similarly if I upgrade, the new driver will not have these new keys to read.

davidungar:
> The question is more base than that: This new format appears to replace a key from the legacy driver’s build record format outright. So, if I use the new driver and need to downgrade then the old driver will be missing information. Is the legacy driver tolerant of that use case? Similarly if I upgrade, the new driver will not have these new keys to read.
> 
> If we are unsure of the answer to this question, I would prefer we make only additive changes to the build record format.

Thanks for explaining. This sort of issue is exactly why I like to reviews! I just performed a few experiments: Tried an upgrade from a main swift-driver to this one, a downgrade from this one to the main, and also a downgrade from this one to the legacy driver. I tried these transitions in one of two scenarios, with "build_time" changed to "build_start_time" and with "build_time" left alone.

Surprisingly, it turned out that there was no benefit to the latter scenario. Both the legacy driver and the main (status quo) swift driver reject the build record when there is a new field (in this case "build_end_time") added to it. I conclude that the better option is to go with the former scenario. We might as well rename the field since the added field will cause rejection anyway.

To summarize the results:

Upgrading from the main swift driver to this PR:
- Works fine

Downgrading from this PR to a main swift driver: 
- "Incremental compilation has been disabled due to malformed build record file Unexpected key 'build_start_time'"
- Subseqenent compilations work fine

Downgrading from this PR or from main swift driver to legacy driver:
- "Incremental compilation has been disabled, because different arguments were passed to the compiler."

Thanks for asking the question. I don't see any better solution than this PR as it stands. Do you?

CodaFi (approved):

davidungar:
@swift-ci please test

davidungar:
Also, cleaned up the commit history

PR #582, opened 2021-04-06
Use unique temporary file names

artemcm:
This PR augments `VirtualPath` with an ability to create a temporary file that is unique a given driver execution context: `createUniqueTemporaryFile`. i.e. if a temporary file with the same filename as one being requested has been created before, using this API, then the newly-created file will have a suffix appended to the base name to distinguish it.

This brings swift-driver closer to the legacy driver's behavior, which generates unique temporary files by-default, using LLVM's `llvm::sys::fs::createTemporaryFile`.

Doing so is necessary to resolve several kinds of problems, for example: 
Compiling a module `Foo` consisting of two source files `Foo.swift` and `Bar.swift` will result in a temporary partial module `Foo.swiftmodule` being produced from building `Foo.swift` and a full `Foo.swiftmodule` produced from the merge-modules job. If the full module does not need to be exposed to the user as the output of compilation, it will be a temporary file, in which case it will conflict with the partial `Foo.swiftmodule` causing the downstream executor to rightly complain about potential cycles in the build.

I entertained the idea of instead resolving `.temporary` paths to unique files at argument resolution time, but that proved troublesome because in that case a specific unique file has to be tracked via a unique `VirtualPath.handle`, where different such handles may refer to the same `VirtualPath`, which is an opaque distinction to many path clients in the driver, and would be really hard to enforce throughout the code-base. Instead, this approach relies on the file's creator to make sure it is unique.

Having enum case access control would be super useful here in order to prohibit clients from instantiating `.temporary` paths explicitly, instead of via the new API. This is left for future work once Swift has such a feature.

Resolves rdar://76123509

### DIFF
--- a/Sources/SwiftDriver/Jobs/BackendJob.swift
+++ b/Sources/SwiftDriver/Jobs/BackendJob.swift
@@ -15,6 +15,7 @@ import Foundation
 extension Driver {
   /// Form a backend job.
   mutating func backendJob(input: TypedVirtualPath,
+                           baseInput: TypedVirtualPath?,
                            addJobOutputs: ([TypedVirtualPath]) -> Void)
   throws -> Job {
     var commandLine: [Job.ArgTemplate] = swiftCompilerPrefixArgs.map { Job.ArgTemplate.flag($0) }
@@ -55,7 +56,9 @@ extension Driver {
 
     // Add the output file argument if necessary.
     if let compilerOutputType = compilerOutputType {
-      let output = computePrimaryOutput(for: input,
+      // If there is no baseInput (singleCompileMode), primary output computation
+      // is not input-specific, therefore it does not matter which input is passed.
+      let output = computePrimaryOutput(for: baseInput ?? input,
                                         outputType: compilerOutputType,
                                         isTopLevel: isTopLevelOutput(type: compilerOutputType))
       commandLine.appendFlag(.o)
--- a/Sources/SwiftDriver/Utilities/VirtualPath.swift
+++ b/Sources/SwiftDriver/Utilities/VirtualPath.swift
@@ -20,6 +20,8 @@ import Darwin
 public enum VirtualPath: Hashable {
   private static var pathCache = PathCache()
 
+  private static var temporaryFileStore = TemporaryFileStore()
+
   /// A relative path that has not been resolved based on the current working
   /// directory.
   case relative(RelativePath)
@@ -33,12 +35,14 @@ public enum VirtualPath: Hashable {
   /// Standard output
   case standardOutput
 
+  /// We would like to direct clients to use the temporary file creation utilities `createUniqueTemporaryFile`, etc.
+  /// To ensure temporary files are unique.
+  /// TODO: If/When Swift gains enum access control, we can prohibit direct instantiation of temporary file cases,
+  /// e.g. `private(init) case temporary(RelativePath)`.
   /// A temporary file with the given name.
   case temporary(RelativePath)
-
   /// A temporary file with the given name and contents.
   case temporaryWithKnownContents(RelativePath, Data)
-
   /// A temporary file that holds a list of paths.
   case fileList(RelativePath, FileList)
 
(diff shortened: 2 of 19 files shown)


artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -403,6 +407,68 @@ extension VirtualPath {
...
+/// This extension provides a set of utilities to create unique (within driver context) relative paths to temporary files.
+/// Clients are still allowed to instantiate `.temporary` `VirtualPath` values directly because of our inability to specify
+/// enum case access control, but are discouraged from doing so.
+extension VirtualPath {
+  internal static func createUniqueTemporaryFile(_ path: RelativePath) -> VirtualPath {
+    let uniquedRelativePath = getUniqueTemporaryPath(for: path)
+    return .temporary(uniquedRelativePath)
+  }
+
+  internal static func createUniqueTemporaryFileWithKnownContents(_ path: RelativePath, _ data: Data)
```
cltnschlosser, line 422: Should these be public if this is what we want people to use?

Sources/SwiftDriver/Jobs/BackendJob.swift
```diff
@@ -15,6 +15,7 @@ import Foundation
 extension Driver {
   /// Form a backend job.
   mutating func backendJob(input: TypedVirtualPath,
+                           baseInput: TypedVirtualPath?,
```
artemcm, line 18: This was a separate bug that uniquing temporary files exposed. 
Instead of using the base primary input file's name as a base (with a new extension), we were relying on a previous job's intermediate output file's name. This correction aligns this behavior with the legacy driver.

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -403,6 +407,68 @@ extension VirtualPath {
...
+  -> VirtualPath {
+    let uniquedRelativePath = getUniqueTemporaryPath(for: path)
+    return .temporaryWithKnownContents(uniquedRelativePath, data)
+  }
+
+  private static func getUniqueTemporaryPath(for path: RelativePath) -> RelativePath {
+    let uniquedBaseName = Self.temporaryFileStore.getUniqueFilename(for: path.basenameWithoutExt)
+    // Avoid introducing the the leading dot
+    let dirName = path.dirname == "." ? "" : path.dirname
+    let fileExtension = path.extension == nil ? "" : "." + path.extension!
```
cltnschlosser, line 432: This can be written without the force unwrap using map. `path.extension.map { “.\($0)”} ?? “”`
The quotes are wrong because I’m reviewing from my phone.

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -403,6 +407,68 @@ extension VirtualPath {
...
+          newCount = 1
+        }
+        uniqueFileCountDict[temporaryPathStr] = newCount
+        return "\(temporaryPathStr)-\(newCount)"
+      }
+    }
+
+    // Used for testing purposes only
+    fileprivate func reset() {
+      uniqueFileCountDict.removeAll()
```
cltnschlosser, line 462: Should be on the queue

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -403,6 +407,68 @@ extension VirtualPath {
...
+
+  /// A cache of created temporary files
+  private final class TemporaryFileStore {
+    private var uniqueFileCountDict: [String: Int]
+    private var queue: DispatchQueue
+
+    init() {
+      self.uniqueFileCountDict = [String: Int]()
+      self.queue = DispatchQueue(label: "com.apple.swift.driver.temp-file-store",
+                                 qos: .userInteractive, attributes: .concurrent)
```
cltnschlosser, line 444: If we always use barrier does it make more sense to use serial queue?

cltnschlosser (commented):
This is awesome, we’ve needed this for a while. There is somewhere else where I added a hack that uses a static int to differentiate temporary files. We can remove that and use this instead.

Sources/SwiftDriver/Utilities/VirtualPath.swift
artemcm: Fixed.

Sources/SwiftDriver/Utilities/VirtualPath.swift
artemcm: It would. Changed to serial queue, thanks.

Sources/SwiftDriver/Utilities/VirtualPath.swift
artemcm: Neat, thanks!

Sources/SwiftDriver/Utilities/VirtualPath.swift
artemcm: Yup, done.

artemcm:
Thanks for taking a look, @cltnschlosser!

cltnschlosser:
https://github.com/apple/swift-driver/blob/2f2a3f85d7467667dc43c25b589254c6f5672de7/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift#L482 Was the other function I was talking about that this can replace.

artemcm:
> createTemporaryFileName

Ah, cool, very nice to unify these things. Thanks!

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -383,6 +384,7 @@ final class SwiftDriverTests: XCTestCase {
 
     try assertNoDriverDiagnostics(args: "swiftc", "foo.swift", "-debug-prefix-map", "foo=bar=baz", "-debug-prefix-map", "qux=") { driver in
         let jobs = try driver.planBuild()
+      print(jobs[0].commandLine)
```
cltnschlosser, line 387: Can delete this now

cltnschlosser (approved):
All looks good to me, I'd bet this fixes some integration tests. I'll have to rerun them and compare after this gets merged. :)

artemcm:
@swift-ci please test

artemcm:
When cross-tested from `swift` with Together with apple/swift-package-manager#3391, the tests pass in https://github.com/apple/swift/pull/36799, and apple/swift-package-manager#3391 is now merged, so I think this is good to go.

PR #583, opened 2021-04-07
[Incremental] Add remarks explaining post-compile builds

davidungar:
https://github.com/apple/swift-driver/pull/580 added some logic to better schedule post-compile jobs. Explain what is happening for that when `-driver-show-incremental` is enabled.


davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -308,9 +308,20 @@ extension IncrementalCompilationState {
 // MARK: - Scheduling post-compile jobs
 extension IncrementalCompilationState {
   public func canSkipPostCompile(job: Job) -> Bool {
-    job.outputs.allSatisfy {output in
+    let neededOutputs = job.outputs.filter {output in
       let fileModTime = (try? fileSystem.lastModificationTime(for: output.file)) ?? .distantFuture
-      return fileModTime <= buildEndTime}
+      return fileModTime > buildEndTime
+    }
```
cltnschlosser, line 314: These changes make sense given the current logic, but I'm not following the current logic. Maybe I'll explain what I'm thinking and you can explain what I'm not understanding.

First, it seems like we would want to run a job if it's inputs have changed or we don't have the output.

I see we are looking at the outputs only and saying we need them if they don't exist or if they were changed since the last build. What would be changing output files other than their producing jobs?

Shouldn't we be looking at inputs?

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -401,16 +401,25 @@ class ExecuteAllJobsRule: LLBuildRule {
...
+      engine.taskNeedsInput(ExecuteJobRule.RuleKey(index: postCompileIndex),
+                            inputID: postCompileIndex)
+    }
+    let didAnyCompileJobsRun = !context.primaryIndices.isEmpty
+    /// If any compile jobs ran, skip the expensive mod-time checks
+    let scheduleEveryPostCompileJob = didAnyCompileJobsRun
+    if scheduleEveryPostCompileJob {
+      context.postCompileIndices.forEach(schedule)
+      context.incrementalCompilationState?.reporter?.report(
+        "Scheduling all post-compile jobs because something was compiled")
```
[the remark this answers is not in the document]
davidungar: Nice catch, good idea! Thank you.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -401,16 +401,25 @@ class ExecuteAllJobsRule: LLBuildRule {
...
+    }
     for postCompileIndex in context.postCompileIndices {
       let job = context.jobs[postCompileIndex]
-      /// If any compile jobs ran, skip the expensive mod-time checks
-      if context.primaryIndices.isEmpty,
-         let incrementalCompilationState = context.incrementalCompilationState,
-         incrementalCompilationState.canSkipPostCompile(job: job) {
+      if context.incrementalCompilationState?.canSkipPostCompile(job: job) ?? false {
         continue
       }
```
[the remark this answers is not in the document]
davidungar: Interesting...I've taking another attempt at it, take a look as soon as I push.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Sigh... I'll explain the thinking & maybe then you'll want to suggest a comment to explain it :)
The old driver didn't look at the inputs of post-compile steps, as far as I know--I haven't gone back and rechecked, but this code was cargo-culted when written. I didn't want to introduce any more change that was needed to fix the problem, which happened if the build were interrupted during the merge-module phase.
Then the build would end, the source files would seem up to date, but there would be no swiftmodule produced.
(If I introduced code to check the inputs, the driver would be depending on the inputs being correct for each post-compile job.)
OK, I need to step away from the keyboard! ;)
Thanks for your thoughtful work. I'll be back tomorrow.

cltnschlosser: Yeah I still don't really understand the current logic, but I've worked my way into realizing why it works out, so I think we can merge this and maybe come back to it.

This works out because we are very aggressive in running these jobs currently. If we ran any compile job we run all of the post compile jobs. This is effectively our inputs have changed check.

The output time check is probably only functioning as a does this output exist check.

But that should be enough to run the jobs at least as often as they need to be (most likely more often).

cltnschlosser (approved):

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: > Yeah I still don't really understand the current logic, but I've worked my way into realizing why it works out, so I think we can merge this and maybe come back to it.
> 
> This works out because we are very aggressive in running these jobs currently. If we ran any compile job we run all of the post compile jobs. This is effectively our inputs have changed check.
> 
> The output time check is probably only functioning as a does this output exist check.
> 
> But that should be enough to run the jobs at least as often as they need to be (most likely more often).

@cltnschlosser Thanks a lot for your thought and patience! After sleeping on it, I think that an inputs vs outputs check may be the best solution. Take a look at this version and let me know what you think.

@CodaFi , would you like to weigh in?

davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -307,21 +307,62 @@ extension IncrementalCompilationState {
...
+
+    guard let (oldestOutput, oldestOutputModTime) =
+            findOldestOutputForSkipping(postCompileJob: postCompileJob)
+    else {
+      report(skipping: false, "No outputs")
+      return false
+    }
+    guard .distantPast < oldestOutputModTime
+    else {
+      report(skipping: false, "Missing output", oldestOutput)
```
cltnschlosser, line 332: This only reports the first missing output, which is probably fine.

cltnschlosser (approved):
Change makes sense to me, I can follow the logic better now.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Yes, by design.

davidungar:
> Change makes sense to me, I can follow the logic better now.

Thank you.

CodaFi (approved):

PR #584, opened 2021-04-07
Only output serialized diagnostics for verify jobs if explicitly requested in the output file map

artemcm:
Otherwise, today, we are always emitting these serialized diagnostic files incorrectly.
We use either:
   - `ModuleName` + `.dia` for both plain and private module interfaces (which is wrong because we have multiple producers of an identical file).
   - `InterfaceFileBasename` + `.dia` for both plain and private module interfaces (which is wrong for the plain interface, for example, because it may conflict with a compile job emitting an identically-named `.dia` file).

Looking at the legacy driver, we don't actually emit these most of the time, even when `-serialize-diagnostics` is specified on the invocation, because that refers to the overall build, not the verification. Which leads me to believe there aren't any consumers for these specific serialized diagnostics.

Resolves rdar://76122268


nkcsgexi (approved):
Makes sense to me. Otherwise no clients consume these files.

PR #586, opened 2021-04-08
[Incremental] Only warn if cannot write prior module dependencies, and print more error info

davidungar:
A mitigation to rdar://76359678.
If the write fails, import incrementality is lost, but it is not a fatal error.
@CodaFi , what do you think?

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -988,10 +988,7 @@ extension Driver {
     if !childJobs.isEmpty {
       do {
         defer {
-          self.incrementalCompilationState?.writeDependencyGraph()
-          buildRecordInfo?.writeBuildRecord(
-            jobs,
-            incrementalCompilationState?.skippedCompilationInputs)
+          writeIncrementalBuildInformation(jobs)
         }
         try performTheBuild(allJobs: childJobs,
                             jobExecutionDelegate: toolExecutionDelegate,
@@ -1053,6 +1050,20 @@ extension Driver {
       numParallelJobs: numParallelJobs ?? 1,
       forceResponseFiles: forceResponseFiles,
       recordedInputModificationDates: recordedInputModificationDates)
+  }
+
+  private func writeIncrementalBuildInformation(_ jobs: [Job]) {
+    if let incrementalCompilationState = self.incrementalCompilationState {
+      let hadError = incrementalCompilationState.writeDependencyGraph()
+      /// Ensure that a bogus dependency graph is not used next time
+      guard !hadError else {
+        buildRecordInfo?.removeBuildRecord()
+        return
+      }
+    }
+    buildRecordInfo?.writeBuildRecord(
+      jobs,
+      incrementalCompilationState?.skippedCompilationInputs)
   }
 
   private func printBindings(_ job: Job) {
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -1057,11 +1057,15 @@ extension Driver {
     // A mitigation to rdar://76359678.
     // If the write fails, import incrementality is lost, but it is not a fatal error.
     if let incrementalCompilationState = self.incrementalCompilationState {
-      let hadError = incrementalCompilationState.writeDependencyGraph()
-      /// Ensure that a bogus dependency graph is not used next time.
-      guard !hadError else {
-        buildRecordInfo?.removeBuildRecord()
-        return
+      do {
+        try incrementalCompilationState.writeDependencyGraph()
+      }
+      catch {
+        diagnosticEngine.emit(
+          .warning("next compile won't be incremental; could not write dependency graph: \(error.localizedDescription)"))
+          /// Ensure that a bogus dependency graph is not used next time.
+          buildRecordInfo?.removeBuildRecord()
+          return
       }
     }
     buildRecordInfo?.writeBuildRecord(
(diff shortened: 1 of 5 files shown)


CodaFi:
This works iff the build record is updated to reflect that we failed. Otherwise we will startup again with bogus priors.

davidungar:
> This works iff the build record is updated to reflect that we failed. Otherwise we will startup again with bogus priors.

Yes, I was imagining the case where there was a path error or permission error and there were no priors written. 

If this were a first-time compile--there had been no priors read--then this failure could be non-fatal.

davidungar:
> This works iff the build record is updated to reflect that we failed. Otherwise we will startup again with bogus priors.

Would like to get your thoughts on this version. The idea is to move the build record if the graph write fails. TIA

davidungar:
@swift-ci please test

davidungar:
> Looks fine for a temporary workaround. Please add a reference to a radar and a note somewhere about what this is working around.
There's a reference to a radar here, in the PR, so I'm guessing you mean a reference to the radar in the code, as well as a note. Will do.

davidungar:
@swift-ci please test

CodaFi:
We got CI blue again. Could you scale this patch back to just printing the error info?

davidungar:
> We got CI blue again. Could you scale this patch back to just printing the error info?

Not sure I follow. Are you proposing to keep the behavior of having the whole compile fail if writing the priors fails? If so, could you explain the superiority of failing the compile, as opposed to letting it succeed but warning about the loss of incrementality?

davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

davidungar:

CodaFi:
Okay, now I really don't think we need this. Mishal was able to fix this with a sandbox configuration change, and I was able to improve the error message in #594

davidungar:
> Okay, now I really don't think we need this. Mishal was able to fix this with a sandbox configuration change, and I was able to improve the error message in #594

Right. Do we need this change right now? Nope. 
Congrats on your work to track down this error!

But, would our code base be better with this change or without it?

Here is the argument against it, that I can see: It adds some code and complexity to the driver.

Here are the arguments for it: 

1. It brings the priors error handling into consistency with the build-record-writing error handling. In other words, when the build-record write fails, the driver does not fail the whole compile, it just emits a warning. (See the `writeBuildRecord` function.)

2. If, in the future, something causes the write to fail, wouldn't it be better to allow the compilation to succeed? The compilation might take a very long time, as we've heard from some users. Do you really want to have them waste that time on an unnecessary failure?

If you still want to veto this PR, you'd have to: a) be in favor of the inconsistency between build-record errors and priors errors, *and* b) want to potentially fail a long compile when it is not strictly necessary.

If you could choose between two kinds of bug reports, a warning disabling incrementality, vs. a completely failed compile, wouldn't you choose the former?

Thinking about it in these broader terms, is there any chance you might agree to land this PR?

davidungar:

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -496,23 +496,26 @@ extension IncrementalCompilationState {
 // MARK: - Serialization
 
 extension IncrementalCompilationState {
-  @_spi(Testing) public func writeDependencyGraph() {
+  @_spi(Testing) public func writeDependencyGraph() -> Bool {
```
CodaFi, line 499: Let’s throw something.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -663,19 +663,22 @@ extension ModuleDependencyGraph {
   ///   - fileSystem: The file system for this location.
   ///   - compilerVersion: A string containing version information for the
   ///                      driver used to create this file.
+  /// - Returns: true if had error
   @_spi(Testing) public func write(
     to path: VirtualPath,
     on fileSystem: FileSystem,
     compilerVersion: String
-  ) {
+  ) -> Bool {
```
CodaFi, line 671: Same here

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1055,6 +1052,23 @@ extension Driver {
       recordedInputModificationDates: recordedInputModificationDates)
   }
 
+  private func writeIncrementalBuildInformation(_ jobs: [Job]) {
+    // In case the write fails, don't crash the build.
+    // A mitigation to rdar://76359678.
```
CodaFi, line 1057: Now it’s a correctness fix, no? We need to ensure that when the write fails, subsequent runs of the driver do not trust the information on disk.

CodaFi (approved):
Just nits

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Okey-dokey.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: Sure!

davidungar:
@swift-ci please test

davidungar:
> Just nits

Thanks!

PR #590, opened 2021-04-09
cmake: set up build-script support for swift-build-sdk-interfaces

nkcsgexi:

PR #593, opened 2021-04-11
Revert "cmake: set up build-script support for swift-build-sdk-interfaces"

compnerd:
Reverts apple/swift-driver#590

This seems to have regressed the Windows port.


compnerd:
@swift-ci please test

nkcsgexi (changes requested):
Could you fix this instead of reverting since you have the setup?

compnerd:
Well, its unclear if this code even makes sense on other platforms - I think it makes some sense to revert for the time being.  It will be a pretty invasive change to get this fixed.

PR #597, opened 2021-04-13
PrebuiltModuleGen: Allow arm64 interfaces to depend on arm64e interfaces

nkcsgexi:
Prebuit modules of arm64e interfaces can be loaded to help building arm64 interfaces.


nkcsgexi:
@swift-ci please test

artemcm (approved):
Looks reasonable to me.

PR #598, opened 2021-04-13
Fix `testTargetTriple` to not always expect macOS to be x86_64

artemcm:


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #599, opened 2021-04-13
Temporarily disable tests that fail on Apple Silicon CI.

artemcm:
The following tests currently fail on Apple Silicon:
```
testDarwinBasic
testSaveTemps
testPrebuiltModuleGenerationJobs
testLitDriverDependenciesTests
```
e.g. https://ci.swift.org/job/oss-swift-incremental-RA-macos-apple-silicon/433/consoleText

For the most part, the test code just does not account for the fact that macOS can run on anything other than x86. 
These tests are going to be disabled on Apple Silicon for the time being to unblock CI, while we go and fix the test code.


artemcm:
@swift-ci please test

hborla (approved):

PR #601, opened 2021-04-14
PrebuiltModuleGen: run dangling jobs only when Foundation is built successfully for macabi

nkcsgexi:


nkcsgexi:
@swift-ci please test

artemcm (approved):

PR #602, opened 2021-04-15
Disable `testInputForwarding` on Apple Silicon while figuring out why it is failing to unblock CI.

artemcm:


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #603, opened 2021-04-15
Generalize disabled tests to also work on Apple Silicon 

artemcm:
https://github.com/apple/swift-driver/pull/599 and https://github.com/apple/swift-driver/pull/602 disabled the following tests:
- `testDarwinBasic`
- `testSaveTemps`
- `testInputForwarding`
- `testPrebuiltModuleGenerationJobs`
This PR generalizes those tests to be agnostic to the macOS architecture and re-enables them on Apple Silicon. 

The only remaining Apple Silicon-disabled test after this is:
- `testLitDriverDependenciesTests `

### DIFF
--- a/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
+++ b/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
@@ -754,12 +754,6 @@ final class ExplicitModuleBuildTests: XCTestCase {
 // We only care about prebuilt modules in macOS.
 #if os(macOS)
   func testPrebuiltModuleGenerationJobs() throws {
-    #if arch(arm64)
-      // Disabled on Apple Silicon
-      // rdar://76609781
-      throw XCTSkip()
-    #endif
-
     func getInputModules(_ job: Job) -> [String] {
       return job.inputs.map { input in
         return input.file.absolutePath!.parentDirectory.basenameWithoutExt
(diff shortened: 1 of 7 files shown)


Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
```diff
@@ -830,22 +822,30 @@ final class ExplicitModuleBuildTests: XCTestCase {
       XCTAssertTrue(danglingJobs.allSatisfy { job in
         job.moduleName == "MissingKit"
       })
-      XCTAssertTrue(jobs.count == 13)
+      XCTAssertTrue(jobs.count == 18)
```
artemcm, line 825: @nkcsgexi , I may not have the full context for some of the things being tested here so please take a look to ensure it's still testing the right thing.

nkcsgexi: Thank you for fixing this! The test updates look correct to me.

nkcsgexi (approved):

Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
```diff
@@ -821,7 +814,6 @@ final class ExplicitModuleBuildTests: XCTestCase {
       var driver = try Driver(args: ["swiftc", main.pathString,
```
nkcsgexi, line 814: Do we really need to always use the host triple? Can we give a specific `-target` here to make the test arch-agnostic?

artemcm: I'm having a hard time reproducing the original CI failures for this test so I'm investigating why that may be.

artemcm: I finally reproduced this locally, and the failure is because these are some of the few tests actually need to link to succeed (as opposed to just invoking the driver and inspecting the generated jobs). And when being tested with a just-built compiler, we do not have a fully-formed fat toolchain, we just have the host slice of everything. 

So I think in this case it's simpler to just compile for the host to make sure these tests are still exercised on whatever platform we are testing on.

artemcm:
@swift-ci please test

PR #604, opened 2021-04-16
PrebuiltModuleGen: consider failures of Foundation and anything below critical

nkcsgexi:
Foundation failures can lead to noticeable slowness in compilation, so we should
consider them critical.

### DIFF
--- a/Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
+++ b/Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
@@ -12,29 +12,28 @@
 import TSCBasic
 import SwiftOptions
 
-func isIosMac(_ path: TypedVirtualPath) -> Bool {
+func isIosMac(_ path: VirtualPath) -> Bool {
   // Infer macabi interfaces by the file name.
   // FIXME: more robust way to do this.
-  return path.file.basenameWithoutExt.contains("macabi")
+  return path.basenameWithoutExt.contains("macabi")
 }
 
 public class PrebuitModuleGenerationDelegate: JobExecutionDelegate {
   var failingModules = Set<String>()
-  var succeededJobs: [Job] = []
   var commandMap: [Int: String] = [:]
   let diagnosticsEngine: DiagnosticsEngine
   let verbose: Bool
-  public init(_ diagnosticsEngine: DiagnosticsEngine, _ verbose: Bool) {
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
```diff
@@ -54,21 +53,25 @@ public class PrebuitModuleGenerationDelegate: JobExecutionDelegate {
     }
   }
 
+  static func getCriticalOutput(_ job: Job) -> VirtualPath? {
+    return job.moduleName == "Foundation" ? job.outputs[0].file : nil
```
artemcm, line 57: Before we also considered `Swift` and `_Concurrency` critical, do we no longer want to exit early on those?

Sources/swift-build-sdk-interfaces/main.swift
```diff
@@ -107,13 +107,13 @@ do {
                             executor: executor,
                             compilerExecutableDir: swiftcPath.parentDirectory)
     let (jobs, danglingJobs) = try driver.generatePrebuitModuleGenerationJobs(with: inputMap, into: outputDir, exhaustive: !coreMode)
-    let delegate = PrebuitModuleGenerationDelegate(diagnosticsEngine, verbose)
+    let delegate = PrebuitModuleGenerationDelegate(jobs, diagnosticsEngine, verbose)
     do {
       try executor.execute(workload: DriverExecutorWorkload.init(jobs, nil, continueBuildingAfterErrors: true),
                            delegate: delegate, numParallelJobs: 128)
     } catch {
       // Only fail the process if stdlib failed
```
artemcm, line 115: This comment seems stale now, see comment above.

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
nkcsgexi: This new condition implies the previous conditions as well. If `Swift` or `_Concurrency` failed, `Foundation` will surely fail too.

Sources/swift-build-sdk-interfaces/main.swift
nkcsgexi: Will update!

artemcm (approved):

nkcsgexi:
@swift-ci please test

PR #606, opened 2021-04-19
Attempt to diagnose arguments to `-sdk` that point to unsupported SDKs

artemcm:
Resolves rdar://70215624

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2005,13 +2005,42 @@ extension Driver {
 
       if !fileSystem.exists(path) {
         diagnosticsEngine.emit(.warning_no_such_sdk(sdkPath))
+      } else if isSDKTooOld(sdkPath: path, fileSystem: fileSystem,
+                            diagnosticsEngine: diagnosticsEngine) {
+        diagnosticsEngine.emit(.error_sdk_too_old(sdkPath))
+        return nil
       }
-      // .. else check if SDK is too old (we need target triple to diagnose that).
 
       return .absolute(path)
     }
 
     return nil
+  }
+}
+
+// SDK checking: attempt to diagnose if the SDK we are pointed at is too old.
+extension Driver {
+  static func isSDKTooOld(sdkPath: AbsolutePath, fileSystem: FileSystem,
+                          diagnosticsEngine: DiagnosticsEngine) -> Bool {
+    let sdkInfo = DarwinToolchain.readSDKInfo(fileSystem, VirtualPath.absolute(sdkPath).intern())
+    guard let sdkInfo = sdkInfo else {
+      diagnosticsEngine.emit(.warning_no_sdksettings_json(sdkPath.pathString))
+      return false
+    }
+    guard let sdkVersion = Version(potentiallyIncompleteVersionString: sdkInfo.versionString) else {
+      diagnosticsEngine.emit(.warning_fail_parse_sdk_ver(sdkInfo.versionString, sdkPath.pathString))
+      return false
+    }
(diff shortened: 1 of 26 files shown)


artemcm:
@swift-ci please test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2005,8 +2005,11 @@ extension Driver {
 
       if !fileSystem.exists(path) {
         diagnosticsEngine.emit(.warning_no_such_sdk(sdkPath))
+      } else if targetTriple != nil,
+                isSDKTooOld(sdkPath: path, target: targetTriple!) {
```
cltnschlosser, line 2009:
```suggestion
      } else if let targetTriple = targetTriple,
                isSDKTooOld(sdkPath: path, target: targetTriple) {
```
Don't have to force unwrap here if you use `if let`

nkcsgexi:
Can we use the `DarwinSDKInfo` for checks like this? We may use an SDK path that's without the version number in it (a symlink to the actual SDK dir).

artemcm:
> Can we use the `DarwinSDKInfo` for checks like this? We may use an SDK path that's without the version number in it (a symlink to the actual SDK dir).

Oh, I hadn't realized we already had this implemented. Great, I'll look into that.

Sources/SwiftDriver/Driver/Driver.swift
artemcm: This code looks a bit different now, but thank you for taking a look.

artemcm:
> Can we use the `DarwinSDKInfo` for checks like this? We may use an SDK path that's without the version number in it (a symlink to the actual SDK dir).

@nkcsgexi, I changed the approach to use the `DarwinSDKInfo`.

artemcm:
@swift-ci please test

TestInputs/SDKChecks/watchOS3.0.sdk/SDKSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+	"Version": "3.0",
+	"VersionMap": {
+		"macOS_iOSMac": {},
+		"iOSMac_macOS": {}
+	},
```
nkcsgexi, line 6: Thanks! I think we shouldn't need `VersionMap` in non-macosx SDKs.

artemcm: Good catch, thanks. 👍🏼

artemcm:
@swift-ci please test

nkcsgexi (approved):
LGTM!

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2015,6 +2018,32 @@ extension Driver {
...
+    guard let sdkVersion = Version(potentiallyIncompleteVersionString: sdkInfo.versionString) else {
+      diagnosticsEngine.emit(.warning_fail_parse_sdk_ver(sdkInfo.versionString, sdkPath.pathString))
+      return false
+    }
+    if sdkInfo.canonicalName.hasPrefix("macos") {
+      return sdkVersion < Version(10, 15, 0)
+    } else if sdkInfo.canonicalName.hasPrefix("iphoneos") ||
+                sdkInfo.canonicalName.hasPrefix("appletvos") {
+      return sdkVersion < Version(13, 0, 0)
+    } else if sdkInfo.canonicalName.hasPrefix("watchos") {
```
nkcsgexi, line 2039: Sorry about missing this. Do we need to handle simulator SDKs here?

nkcsgexi: Thanks! Using a more concise prefix will just work.

PR #609, opened 2021-04-20
Update `isSDKTooOld` to be compatible with older compilers.

artemcm:
5.3 and ealier compilers may hit:
```
swift-driver/Sources/SwiftDriver/Driver/Driver.swift:2044:15: error: definition conflicts with previous value
    guard let sdkInfo = sdkInfo else {
```
e.g.
https://ci.swift.org/job/swift-PR-toolchain-osx/949/console


artemcm:
@swift-ci please test

bnbarham (approved):
Could also just inline it, but LGTM

PR #615, opened 2021-04-22
Verify module interfaces generated from emit-module jobs

xymus:
Verify interfaces generated by single compile jobs and emit-module jobs. The verify job was previously only added after merge-module jobs which is more prone to generate broken swift interfaces.

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -253,6 +253,8 @@ extension Driver {
     return linkerInputs
   }
 
+  /// When in single compile, add one compile job and possiblity multiple backend jobs.
+  /// Return the compile job if one was created.
   private mutating func addSingleCompileJobs(
     addJob: (Job) -> Void,
     addJobOutputs: ([TypedVirtualPath]) -> Void,
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -104,7 +104,7 @@ extension Driver {
 
     try addPrecompileModuleDependenciesJobs(addJob: addJobBeforeCompiles)
     try addPrecompileBridgingHeaderJob(addJob: addJobBeforeCompiles)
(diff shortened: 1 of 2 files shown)


xymus:
@swift-ci test

xymus:
@swift-ci test

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -253,9 +257,9 @@ extension Driver {
     addJob: (Job) -> Void,
     addJobOutputs: ([TypedVirtualPath]) -> Void,
     emitModuleTrace: Bool
-  ) throws {
+  ) throws -> Job? {
```
artemcm, line 260: Just because the name states the plural `Jobs`, and we may be adding multiple (with backend jobs), it's worth a comment to explain exactly which job this function is going to return.

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -221,9 +223,11 @@ extension Driver {
       }
     }
 
-    try addSingleCompileJobs(addJob: addJobBeforeCompiles,
+    if let compileJob = try addSingleCompileJobs(addJob: addJobBeforeCompiles,
                              addJobOutputs: addJobOutputs,
-                             emitModuleTrace: loadedModuleTracePath != nil)
+                             emitModuleTrace: loadedModuleTracePath != nil) {
+      try addVerifyJobs(mergeJob: compileJob, addJob: addJobAfterCompiles)
```
artemcm, line 229: Nit: it may be time to re-name this argument name from `mergeJob` to something that better reflects that this can be any interface-producing job.

artemcm (approved):
A couple of nits, but overall looks good to me!

beccadax (approved):

Sources/SwiftDriver/Jobs/Planning.swift
xymus: Done!

xymus:
@swift-ci please test

artemcm:
@swift-ci please test Linux platform

PR #618, opened 2021-04-28
Re-enable `testLitDriverDependenciesTests` when run on Apple Silicon

artemcm:
This test was previously failing in `swift` CI.
I cannot trigger the failure locally so re-enabling here to test with cross-repository PR testing.


artemcm:
The failure was due to having an x86 binary for `python3` on the CI bots, which is now resolved.
Running cross-repository test in https://github.com/apple/swift/pull/36936 to confirm, and then this is good to merge.

artemcm:
@swift-ci please test

nkcsgexi (approved):
Thank you!

PR #619, opened 2021-04-28
Add workaround for building executables with older CMake versions.

artemcm:
Some time between CMake 3.16 and 3.17 the default behaviour seems to have changed as follows:
For a project which will build an executable, e.g.:
```
add_executable(exe_target
    main.swift)
target_link_libraries(exe_target PUBLIC
    libA
    libB)
```
In 3.17, the invocation command to create the `exe_target` executable includes `-Xlinker -rpath -Xlinkier <libA_path/lib>` and so forth for the linked libraries and their dependencies.
In 3.16, the linker is not passed the rpath flags, so the executable ends up without the `LC_RPATH` required for it to run.
This affects executable targets only, libraries do still have rpaths propagated correctly.

The correct solution (which will hopefully happen soon) is to bump the compiler's minimum CMake version to 3.18.
This is under discussion now here: https://forums.swift.org/t/bump-cmake-version-to-3-18

### DIFF
--- a/Sources/swift-driver/CMakeLists.txt
+++ b/Sources/swift-driver/CMakeLists.txt
@@ -12,4 +12,18 @@ target_link_libraries(swift-driver PUBLIC
   SwiftDriver
   SwiftDriverExecution)
 
+# This is a fairly egregious workaround for the fact that in versions < 3.17,
+# executables do not get `-rpath` linker arguments for their linked library
+# dependencies (direct and transitive)
+if(CMAKE_VERSION VERSION_LESS 3.17)
+  get_target_property(TSC_UTIL_LIB TSCUtility LOCATION)
+  get_filename_component(TSC_LIB_DIR ${TSC_UTIL_LIB} DIRECTORY)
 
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci please test

nkcsgexi (approved):

artemcm:
It's looking like this PR:
https://github.com/apple/swift/pull/36094
Is moving ahead so it'll be better to have that merged to avoid introducing this kind of workaround. 

We should also bump the corresponding required CMake on this repository though, I'll put up a PR.

artemcm:
Closing in favour of:
https://github.com/apple/swift-driver/pull/624

artemcm:
I'm going to re-open because the CMake version bump logistics will require more orchestration and I would like to merge the respective `swift` PR. I will track removing this code once we update the required CMake in a way that makes it redundant.

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
Just re-basing to get cross-repo testing to pick up the right changes.

@swift-ci please test

PR #624, opened 2021-04-30
Bump required CMake version to 3.17

artemcm:
There are several key livability fixes in 3.17 for building Swift code.
This should prevent scenarios like this:
https://github.com/apple/swift/pull/37114

Where CMake `<3.17` missed setting up rpaths on the `swift-driver` executable.

PR #625, opened 2021-05-02
Revert "Bump required CMake version to 3.17"

xymus:
Reverts apple/swift-driver#624

This is breaking many Linux bots as they only have CMake 3.16.5. Reverting until the bots are updated.

rdar://77440032


xymus:
@swift-ci Please test

artemcm:
Thanks, @xymus.

PR #629, opened 2021-05-05
Verify emitted module interfaces by default

xymus:
Verifying swiftinterface files after they are emitted will detect framework breaking issues much earlier. It can be turned off in offending projects with `-no-verify-emitted-module-interface`.

rdar://77534863


xymus:
@swift-ci Please test

artemcm (approved):

xymus:
We may want to still wait a bit before landing this but let's make sure it's ready.

xymus:
Known offending projects have been notified, let's merge this! 🤞

PR #633, opened 2021-05-06
[Incremental Imports] Emit more meaningful diagnostics when there's an input map failure.

davidungar:
Refine the warning emitted in this case to be something the user might actually understand.
Also emit a remark when `-driver-show-incremental` is present to help diagnose the issue.


davidungar:
@swift-ci please test

artemcm (approved):
Nice cleanup of the diagnostic, I like it!

davidungar:
> Nice cleanup of the diagnostic, I like it!

Thanks!

PR #635, opened 2021-05-07
Validate that `-static-executable` is not supported on Darwin.

artemcm:
Otherwise using it on Darwin does not error out early and results in really confusing errors downstream in compilation. 

Resolves rdar://77631444


artemcm:
@swift-ci please test

nkcsgexi (approved):
LGTM!

PR #639, opened 2021-05-07
PrebuiltModuleGen: allow retrying after hitting bad file descriptor error

nkcsgexi:


nkcsgexi:
@swift-ci please test

artemcm (approved):
We still don't have a good idea on why these file descriptor operations fail, do we?

nkcsgexi:
Not yet. We only hit them on certain build environments only.

PR #641, opened 2021-05-07
Revert "Add workaround for building executables with older CMake versions."

compnerd:
Reverts apple/swift-driver#619

This breaks the Windows build.  I'd like to revert this and re-enable it with a fix for Windows.

### DIFF
--- a/Sources/swift-driver/CMakeLists.txt
+++ b/Sources/swift-driver/CMakeLists.txt
@@ -12,18 +12,4 @@ target_link_libraries(swift-driver PUBLIC
   SwiftDriver
   SwiftDriverExecution)
 
-# This is a fairly egregious workaround for the fact that in versions < 3.17,
-# executables do not get `-rpath` linker arguments for their linked library
-# dependencies (direct and transitive)
-if(CMAKE_VERSION VERSION_LESS 3.17)
-  get_target_property(TSC_UTIL_LIB TSCUtility LOCATION)
-  get_filename_component(TSC_LIB_DIR ${TSC_UTIL_LIB} DIRECTORY)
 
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

compnerd:
RPATH is a bad thing to use - please never use it directly, use CMake to control rpaths

artemcm:
Ah. This means that https://github.com/apple/swift/pull/37122 needs to be reverted as well because it depends on it.

@compnerd, #619 is working around the fact that rpath is automatically propagated in CMake >= 3.17, but not 3.16. What would an alternative workaround look like?

compnerd:
Is this for testing purposes or for distribution purposes?

artemcm:
This is for the driver that gets built in the build directory of `swift`, and gets configured to be the default compiler driver in https://github.com/apple/swift/pull/37122. 

For the driver that gets built to be distributed, we have rpaths configured in https://github.com/apple/swift-driver/blob/main/Utilities/build-script-helper.py.

artemcm:
To clarify, when just building `swift`, and using the build directory compiler, this is used.
 
When we run `--install-swift-driver`, rpaths are configured separately in the `build-script-helper.py`.

compnerd:
The question still stands: is it for _testing_ or do you intend these paths to be used on a random host (this is a problem because rpaths are an attack vector)

artemcm:
This is for use by compiler developers only, in the compiler build directory. It is not meant to be distributed or to be used on a random host.

compnerd:
Ah, okay, then you should be using the `BUILD_RPATH` property on the targets.

compnerd:
See #642 for a possible fix

PR #642, opened 2021-05-07
use `BUILD_RPATH` to make the build more portable

compnerd:
The `-rpath` linker options are not portable and persist into
installation providing an attack vector.  This uses CMake to control the
RPATH for testing and ensures that the paths are erased at install time.
This is more secure and portable.


compnerd:
@swift-ci please test

artemcm (approved):
Thank you!

PR #644, opened 2021-05-07
Pass down `-Xlinker` flags to the underlying clang linker call using `-Xlinker`

artemcm:
Passing them in without a `-Xlinker` prefix argument is not valid when invoking the linker using the clang executable as `clang -fuse-ld=gold`. We still need to use the prefix to make sure clang forwards them to the underlying linker invocation.


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #645, opened 2021-05-08
[rdar://77619051] implement sanitize-recover validation

aralisza:
port validation for `-sanitize-recover` over from the c swift driver

rdar://77619051

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2232,7 +2232,6 @@ extension Driver {
       if sanitizer != .address {
         diagnosticEngine.emit(
           .error_unsupported_argument(argument: arg, option: .sanitizeRecoverEQ))
-//          .error_unsupported_argument(argument: arg, arg: Option.sanitizeRecoverEQ))
         continue
       }
       
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -567,6 +567,8 @@ public struct Driver {
       diagnosticEngine: diagnosticEngine,
       toolchain: toolchain,
       targetInfo: frontendTargetInfo)
+    
+    Self.validateSanitizerRecoverArgValues(&parsedOptions, diagnosticEngine: diagnosticsEngine, enabledSanitizers: enabledSanitizers)
 
     Self.validateSanitizerCoverageArgs(&parsedOptions,
                                        anySanitizersEnabled: !enabledSanitizers.isEmpty,
@@ -1795,6 +1797,10 @@ extension Diagnostic.Message {
   static var verify_debug_info_requires_debug_option: Diagnostic.Message {
     .warning("ignoring '-verify-debug-info'; no debug info is being generated")
   }
+  
+  static func warning_option_requires_sanitizer(currentOption: Option, currentOptionValue: String, sanitizerRequired: Sanitizer) -> Diagnostic.Message {
+      .warning("option '\(currentOption.spelling)\(currentOptionValue)' has no effect when '\(sanitizerRequired)' sanitizer is disabled. Use \(Option.sanitizeEQ.spelling)\(sanitizerRequired) to enable the sanitizer")
+  }
 }
 
 // Module computation.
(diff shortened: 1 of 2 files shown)


aralisza:
@swift-ci please test

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1584,6 +1584,70 @@ final class SwiftDriverTests: XCTestCase {
...
   }
+  
+  func testSanitizerRecoverArgs() throws {
+    let commonArgs = ["swiftc", "foo.swift", "bar.swift",]
+    do {
+      // address sanitizer + address sanitizer recover
+      var driver = try Driver(args: commonArgs + ["-sanitize=address", "-sanitize-recover=address"])
+      let plannedJobs = try driver.planBuild()
+
+      XCTAssertEqual(plannedJobs.count, 3)
```
cltnschlosser, line 1595: This is failing on linux: `SwiftDriverTests.testSanitizerRecoverArgs : XCTAssertEqual failed: ("4") is not equal to ("3")`

My guess is you need `let plannedJobs = try driver.planBuild().removingAutolinkExtractJobs()` instead

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2196,6 +2202,46 @@ extension Driver {
...
+        diagnosticEngine.emit(
+          .error_invalid_arg_value(arg: .sanitizeRecoverEQ, value: arg))
+        continue
+      }
+      
+      // only -sanitize-recover=address is supported
+      if sanitizer != .address {
+        diagnosticEngine.emit(
+          .error_unsupported_argument(argument: arg, option: .sanitizeRecoverEQ))
+//          .error_unsupported_argument(argument: arg, arg: Option.sanitizeRecoverEQ))
```
cltnschlosser, line 2235:
```suggestion
```
Can remove this line

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1584,6 +1584,70 @@ final class SwiftDriverTests: XCTestCase {
...
+
+      let compileJob = plannedJobs[0]
+      let compileCmd = compileJob.commandLine
+      XCTAssertTrue(compileCmd.contains(.flag("-sanitize=address")))
+      XCTAssertTrue(compileCmd.contains(.flag("-sanitize-recover=address")))
+    }
+    do {
+      // invalid sanitize recover arg
+      try assertDriverDiagnostics(args: commonArgs + ["-sanitize-recover=foo"]) {
+        $1.expect(.error("invalid value 'foo' in '-sanitize-recover='"))
```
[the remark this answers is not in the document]
cltnschlosser: I think this makes sense too. Only concern is that the there is a test in `Driver/sanitize_recover.swift` that checks for `Missing: unsupported argument 'foo' to option '-sanitize-recover='`, but we can update old driver, or disable this test when running with swift-driver. Unfortunately the disable is file wide, so might be easier to update old driver in this case, if we want to change this and keep the tests passing for both.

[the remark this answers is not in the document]
artemcm: I like this change as well. And I do not think we need to update the legacy driver and its tests. The tests added in this PR give us enough confidence that we are validating the right things here without needing to rely on lit tests in the `swift` repo, it should be okay to just have the lit test disabled.

artemcm (approved):

aralisza:
@swift-ci please test

aralisza:
@swift-ci please test

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1584,6 +1584,73 @@ final class SwiftDriverTests: XCTestCase {
...
+      // thread sanitizer + address sanitizer recover
+      try assertDriverDiagnostics(args: commonArgs + ["-sanitize=thread", "-sanitize-recover=address"]) {
+        $1.expect(.warning("option '-sanitize-recover=address' has no effect when 'address' sanitizer is disabled. Use -sanitize=address to enable the sanitizer"))
+      }
+    }
+    // "-sanitize=undefined" is not available on x86_64-unknown-linux-gnu
+    #if os(macOS)
+    do {
+      // multiple sanitizers separately
+      try assertDriverDiagnostics(args: commonArgs + ["-sanitize=undefined", "-sanitize=address", "-sanitize-recover=address"]) {
```
delcypher, line 1642: I'm kind of surprised `undefined` is supported **at all**. UBSan doesn't make in Swift because AFAIK we don't support emitting UB checks.

This is more of a comment. This doesn't block landing this patch.

delcypher (approved):

PR #646, opened 2021-05-09
Provide swift-frontend path when running integration tests

cltnschlosser:
Fixes the issues I was running into: https://forums.swift.org/t/lots-of-errors-running-swift-driver-integration-tests/47986

Still 90 failing tests, which is more than the 50 from March 27: https://github.com/apple/swift-driver/pull/365/commits/dadb0a3707f979dedd72d984bbd6b21dea33829d

A lot of Dependencies tests failing with similar looking errors, but first step is actually being able to run the tests :)
This is an example error snippet if anyone is curious:
```
            8: Failed to read dependency file
check:12'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            9: Traceback (most recent call last):
check:12'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           10:  File "SOURCE_DIR/test/Driver/Dependencies/Inputs/update-dependencies.py", line 53, in <module>
check:12'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           11:  shutil.copyfile(primaryFile, depsFile)
check:12'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           12:  File "/usr/local/Cellar/python@3.9/3.9.1_7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/shutil.py", line 264, in copyfile
check:12'0     
```

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2427,9 +2427,9 @@ extension Driver {
 
     var appliesToFetchingTargetInfo: Bool {
       guard let path = overridePath else { return true }
-      return path.basename != "Python" &&
-        // starts(with:) to handle both python3 and point versions (Ex: python3.9)
-        !path.basename.starts(with: "python3")
+      // lowercased() to handle Python
+      // starts(with:) to handle both python3 and point versions (Ex: python3.9). Also future versions (Ex: Python4).
+      return !path.basename.lowercased().starts(with: "python")
     }
     func setUpForTargetInfo(_ toolchain: Toolchain) {
       if !appliesToFetchingTargetInfo {
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2426,8 +2426,10 @@ extension Driver {
     }
 
     var appliesToFetchingTargetInfo: Bool {
-      return overridePath?.basename != "Python" &&
-             overridePath?.basename != "python3"
+      guard let path = overridePath else { return true }
+      return path.basename != "Python" &&
+        // starts(with:) to handle both python3 and point versions (Ex: python3.9)
+        !path.basename.starts(with: "python3")
     }
     func setUpForTargetInfo(_ toolchain: Toolchain) {
       if !appliesToFetchingTargetInfo {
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2426,8 +2426,10 @@ extension Driver {
     }
 
     var appliesToFetchingTargetInfo: Bool {
-      return overridePath?.basename != "Python" &&
-             overridePath?.basename != "python3"
+      guard let path = overridePath else { return true }
+      return path.basename != "Python" &&
```
davidungar, line 2430: Hmm... What about "Python4?" What you have is fine, but would
`path.lowercased().hasPrefix("python")` be better? Don't know, what do you think?

davidungar (approved):
Aside from a thought about "python4" (!) it looks fine to me. Would be interested in what you think about that.

Sources/SwiftDriver/Driver/Driver.swift
cltnschlosser: I don’t think it can hurt. I can’t imagine we ever have a legitimate need for a python prefixed frontend, where we would be worried about breaking it.

cltnschlosser:
@swift-ci please test

cltnschlosser:
Running a cross repo test run before merging, just to double check that I don't break anything.

Also figured out what was causing all those `Driver/Dependencies` failures. It was that I was running the tests in parallel. I'm down to 46 Failures total now, so that's good :)

Any ideas what might be causing the hiccups when running `Driver/Dependencies` tests in parallel with the other tests? Never had any issues in the past. Maybe some new shared dependencies file location? Is this expected? or a regression that needs to be investigated?

cltnschlosser:
> Any ideas what might be causing the hiccups when running `Driver/Dependencies` tests in parallel with the other tests? Never had any issues in the past. Maybe some new shared dependencies file location? Is this expected? or a regression that needs to be investigated?

Just realized the cause of the failures is rather obvious. These will both run `test/Driver/Dependencies`. If the same tests are running at the same time they will do things like delete the temporary directory that another instance of the same test was using.
```
  func testLitDriverTests() throws {
    guard ProcessEnv.vars.keys.contains("SWIFT_DRIVER_ENABLE_FAILING_INTEGRATION_TESTS") else {
      throw XCTSkip("Not all Driver tests supported")
    }
    try runLitTests(suite: "test", "Driver")
  }

  func testLitDriverDependenciesTests() throws {
    try runLitTests(suite: "test", "Driver", "Dependencies")
  }
  ```

PR #650, opened 2021-05-11
[5.5] Pass down `-Xlinker` flags to the underlying clang linker call using `-Xlinker` prefix

artemcm:
Passing them in as-is is not valid when invoking the linker using the clang executable as `clang -fuse-ld=gold`.

This is a 5.5 cherry-pick of a change merged on `main` in: https://github.com/apple/swift-driver/pull/644


artemcm:
@swift-ci please test

airspeedswift (approved):

PR #652, opened 2021-05-12
Verify emitted module interfaces by default (Take 2)

xymus:
Verifying swiftinterface files after they are emitted will detect framework breaking issues much earlier. It can be turned off in offending projects with `-no-verify-emitted-module-interface`.

Only swiftinterfaces emitted from non-merge-module builds at verified by default. Merge-module builds are more likely to generate broken swiftinterfaces even when a whole-module or emit-module-separately build generates a valid one.

rdar://77534863

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -405,7 +405,7 @@ extension Driver {
        parsedOptions.hasArgument(.enableLibraryEvolution),
        parsedOptions.hasFlag(positive: .verifyEmittedModuleInterface,
                              negative: .noVerifyEmittedModuleInterface,
-                             default: false)
+                             default: true)
     else { return }
 
     func addVerifyJob(forPrivate: Bool) throws {
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -405,7 +405,14 @@ extension Driver {
        parsedOptions.hasArgument(.enableLibraryEvolution),
        parsedOptions.hasFlag(positive: .verifyEmittedModuleInterface,
                              negative: .noVerifyEmittedModuleInterface,
-                             default: true)
+                             default: true),
+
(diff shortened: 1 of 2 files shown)


xymus:
@swift-ci Please test

xymus:
@swift-ci Please test Apple Silicon

xymus:
@swift-ci test Apple Silicon

artemcm:
> @swift-ci test Apple Silicon

Sorry, I misunderstood. This does not work on this repo, it needs to be invoked as a cross-repository test with this PR from the `swift` repo.

xymus:
@swift-ci Please test

xymus:
The broken swiftinterface in the Swift repo has been fixed and the tests are looking good (except maybe for one that I'm testing again). https://github.com/apple/swift/pull/38454

@artemcm Can you give this another review?

nkcsgexi:
Related to the discussion in a Swift-side PR: https://github.com/apple/swift/pull/38465

Can we sanitize `-disable-availability-checking` when verifying module interfaces?

xymus:
@nkcsgexi What do you mean by "sanitize"? I don't think `-disable-availability-checking` is printed in the swiftinterface at this time.

nkcsgexi:
ah, sorry for missing explanation. `verifyModuleInterfaceJob` calls `addCommonFrontendOptions` which adds all `-Xfronend` options. If the framework authors specified `-Xfrontend -disable-availability-checking` in the build setting, interface verification job will succeed even if there are availability violations in the interface, resulting in the escape of broken interfaces. By sanitizing, I meant after calling `addCommonFrontendOptions`, we specifically remove `-disable-availability-checking` to avoid situations like this.

xymus:
Ah ok, good point. We could use that phase as an early sanity check for all kind of library issues.

xymus:
I think we can merge this now. The remaining failure on Apple Silicon is unrelated to this change.

We should add extra verifications once this lands properly and we have a reliable verification in place.

artemcm (approved):

PR #653, opened 2021-05-12
implmement validation for -sanitize-address-use-odr-indicator

aralisza:
port over validation from c swift driver instead of relying on frontend

rdar://77619161

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -582,6 +582,8 @@ public struct Driver {
       toolchain: toolchain,
       targetInfo: frontendTargetInfo)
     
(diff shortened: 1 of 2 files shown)


aralisza:
@swift-ci please test

cltnschlosser (approved):

delcypher (approved):
LGTM

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2218,6 +2220,16 @@ extension Driver {
     }
   }
   
+  private static func validateSanitizerAddressUseOdrIndicatorFlag(
+    _ parsedOptions: inout ParsedOptions,
+    diagnosticEngine: DiagnosticsEngine,
+    addressSanitizerEnabled: Bool
+  ) {
+    if (parsedOptions.hasArgument(.sanitizeAddressUseOdrIndicator) && !addressSanitizerEnabled) {
```
devincoughlin, line 2228: Minor nit that can be cleaned up in a later commit: the outer parentheses in the if condition here are not idiomatic; they can be removed.

PR #654, opened 2021-05-12
Fix quadratic COW copies in DictionaryOfDictionaries

harlanhaskins:
DictioanryOfDictionaries implemented its subscript in terms of updateValue
and removeValue, both of which keep track of and return the old value, and then
threw away the old value.

Instead, flip it so removeValue and updateValue are implemented in terms of the
subscript, and make the subscript modify the dictionary in-place.

This ended up being a 700x speedup on an Intel Mac and a 180x speedup on an M1.

Fixes rdar://77882768


CodaFi:
@swift-ci test

davidungar:
> This ended up being a 700x speedup on an Intel Mac and a 180x speedup on an M1 for a particular stress test.

Great numbers! How where they measured? Overall driver time?? Just removal of an entry? If the latter, how much overall driver speedup?

PR #655, opened 2021-05-12
Fix quadratic COW copies in DictionaryOfDictionaries

CodaFi:
Cherry pick #654 

----------------

DictioanryOfDictionaries implemented its subscript in terms of updateValue
and removeValue, both of which keep track of and return the old value, and then
threw away the old value.

Instead, flip it so removeValue and updateValue are implemented in terms of the
subscript, and make the subscript modify the dictionary in-place.

This ended up being a 700x speedup on an Intel Mac and a 180x speedup on an M1.

Fixes rdar://77882768


CodaFi:
@swift-ci test

artemcm (approved):

airspeedswift (approved):

airspeedswift (approved):

PR #657, opened 2021-05-12
Improve TwoLevelMap (nee DictionaryOfDictionaries)

CodaFi:
Replace DictionaryOfDictionaries with TwoLevelMap - a name that more accurately captures the interface rather than the implementation. Clean up all of its interfaces and document them all. Finally, clean up the two-level subscript operation so it no longer incurs a triple access in order to remove a key-value pair.


CodaFi:
@swift-ci test

CodaFi:
Oh boy, we cannot be mutating data structures while we're iterating over them.

```
  private func verifyNodeMap() {
    var nodes = [Set<Graph.Node>(), Set<Graph.Node>()]
    nodeMap.verify {
      _, v, submapIndex in
      if let prev = nodes[submapIndex].update(with: v) {
        fatalError("\(v) is also in nodeMap at \(prev), submap: \(submapIndex)")
      }
      v.verify()
    }
  }
```

CodaFi:
@swift-ci test

artemcm (approved):
Thank you.
And the docs are super helpful.

CodaFi:
⛵

Sources/SwiftDriver/IncrementalCompilation/DictionaryOfDictionaries.swift
```diff
@@ -93,20 +93,7 @@ extension DictionaryOfDictionaries {
...
-        // Remove the inner key from the inner dictionary.
-        outerDict[key.0]?[key.1] = nil
-
-        // If that was the last entry in the inner dictionary,
-        // remove the entire inner entry.
-        if outerDict[key.0]?.isEmpty == true {
-            outerDict[key.0] = nil
-        }
-      }
-    }
```
davidungar, line 109: At a glance, it looks as if the performance issue occurred when writing a nil value into the structure. If so, when was that happening?
Also, the new change appears to never reclaim an inner dictionary. Is that correct?

PR #660, opened 2021-05-13
Fix startIndex Calculation

CodaFi:
Using `startIndex` of the outerDict is wrong. We go to the trouble of
searching for the first non-empty inner dictionary for a reason. We need
to use *that* index as the start instead.


CodaFi:
@swift-ci test

artemcm (approved):

PR #661, opened 2021-05-13
Do Not Serialize Ephemeral Input Mappings in Priors

CodaFi:
Build systems reserve the right to alter the mapping from swift files to
swiftdeps files at will. We cannot rely on output file maps having the
same entries from run-to-run, so we cannot rely on combining
deserialized inputs from priors and the output file map to agree.

In rare circumstances, XCBuild would alter the GUID suffix it used for
a swiftdeps file in the OFM, and the priors file would enter the old
swiftdeps path into the dependency graph. This would cause a crash when
that bogus file was looked up during integration.

The easy solution is to just ignore the mapping nodes here. The only
source of truth for this mapping is the output file map, and serializing
this into the priors is just duplicative and wrong.

This change drops swiftdeps in priors from consideration during
(de)serialization. A subsequent commit will remove them from the priors
format and bump the format.

The proper answer to rdar://77804636


CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

PR #663, opened 2021-05-14
[5.5] Fixup Bidirectional Map

CodaFi:
Cherry pick #662 

------------

The setters here were both implemented in such a way that the map would
lose the invariant that it was truly bidirectional. In particular, one
must ensure that previous mappings from keys to values were struck from
the bidirectional map before emplacing a new mapping, else the mirror
image of the mapping would retain extra entries.

In addition, document the misuse of this data structure in the driver.
The mapping from swiftdeps to swift files is a not a bijection. There
are three sources of data that can all give conflicting mappings from
swiftdeps files to swift files in particular:

1) The output file map
2) The priors
3) Swiftdeps files after the frontends have run

This means we can have at most three conflicting maps from some
swiftdeps files to some swift file. Currently, the driver overwrites
this mapping and the last writer wins. This is the root cause of
a family of crashers where the output file map's contents disagreed from
run to run.

When a project workspace is set up with multiple files with the same
base name but differing extensions, XCBuild will disambiguate any
products by appending a GUID derived from the structure of the project.
That means, if foo.swift and foo.c live in the same target's inputs,
then the output file map will have an entry

foo-<GUID_1>.swiftdeps

This GUID is derived from the structure of the project itself, and
changes with that structure. So, for example, one can delete and re-add
a reference to a swift file and the output file map entry for that
swiftdeps file will regenerate

foo-<GUID_2>.swiftdeps

This means that the priors' swiftdeps files now disagree in content with
the output file map, and the output file map's entries currently win.
But, only partially. Because the bidirectional map was broken, it was
able to successfully answer queries about the extra injective mapping
between swift files and swiftdeps files. It was, however, not equipped
to handle the other direction, and so it used to crash. We have since
changed this to redo the incremental build to flush this state.

There is a remaining lurking bug here that involves an extra step.
the output file map.

All this to say, we should take the earliest opportunity to remove

rdar://77804636

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/BidirectionalMap.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/BidirectionalMap.swift
@@ -11,50 +11,63 @@
 //===----------------------------------------------------------------------===//
 
 /// Like a two-way dictionary
+///
+// FIXME: The current use of this abstraction in the driver is
+// fundamentally broken. This data structure should be retired ASAP.
+// See the extended FIXME in its use in
+// `ModuleDependencyGraph.inputDependencySourceMap`
 public struct BidirectionalMap<T1: Hashable, T2: Hashable>: Equatable, Sequence {
   private var map1: [T1: T2] = [:]
   private var map2: [T2: T1] = [:]
 
   public init() {}
 
+  /// Accesses the value associated with the given key for reading and writing.
+  ///
+  /// - Parameter key: The key to find in the dictionary.
+  /// - Returns: The value associated with key if key is in the bidirectional
+  /// map; otherwise, `nil`.
   public subscript(_ key: T1) -> T2? {
     get {
-      guard let value = map1[key]
-      else {
-        return nil
-      }
-      return value
+      return self.map1[key]
     }
     set {
-      if let someNewValue = newValue {
-        map1[key] = someNewValue
-        map2[someNewValue] = key
-        return
+      // First, strike any existing mappings.
+      if let oldTarget = self.map1.removeValue(forKey: key) {
+        self.map2.removeValue(forKey: oldTarget)
       }
-      if let oldValue = map1.removeValue(forKey: key) {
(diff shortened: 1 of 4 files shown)


CodaFi:
@swift-ci test

Sources/SwiftDriver/IncrementalCompilation/BidirectionalMap.swift
```diff
@@ -11,50 +11,63 @@
...
-        return
+      // First, strike any existing mappings.
+      if let oldTarget = self.map1.removeValue(forKey: key) {
+        self.map2.removeValue(forKey: oldTarget)
       }
-      if let oldValue = map1.removeValue(forKey: key) {
-        map2.removeValue(forKey: oldValue)
+      // Then construct the forward mapping (or removal).
+      self.map1[key] = newValue
+      if let newValue = newValue {
```
artemcm, line 41: I think this will not build with 5.3 or older compilers. We've recently been [catching](https://github.com/apple/swift-driver/pull/609) these in other CI bots after being tested here (swift-driver CI uses a recent compiler).

Sources/SwiftDriver/IncrementalCompilation/BidirectionalMap.swift
```diff
@@ -11,50 +11,63 @@
...
-        return
+      // First, strike any existing mappings.
+      if let oldSource = self.map2.removeValue(forKey: key) {
+        self.map1.removeValue(forKey: oldSource)
       }
-      if let oldValue = map2.removeValue(forKey: key) {
-        map1.removeValue(forKey: oldValue)
+      // Then construct the reverse mapping (or removal).
+      self.map2[key] = newValue
+      if let newValue = newValue {
```
artemcm, line 64: Same here

artemcm (approved):
The fix LGTM, other than the above concern with not breaking the build when using an older compiler.

airspeedswift (approved):

PR #664, opened 2021-05-14
[NFC, Incremental] Refactoring to facilitate replacing the inputDependencySourceMap with the right thing

davidungar:
A refactoring of `inputDependencySourceMap` prior to really fixing the below:

See https://github.com/apple/swift-driver/pull/662

> There is a remaining lurking bug here that involves an extra step.
the output file map.

> All this to say, we should take the earliest opportunity to remove

Start on that.

rdar://77998890

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -54,19 +54,19 @@ import SwiftOptions
     self.creationPhase = phase
   }
 
-  @_spi(Testing) public func getRequiredSource(for input: TypedVirtualPath,
-                                               function: String = #function,
-                                               file: String = #file,
-                                               line: Int = #line) -> DependencySource {
-    guard let source = inputDependencySourceMap.getSourceIfKnown(for: input)
+  @_spi(Testing) public func source(requiredFor input: TypedVirtualPath,
+                                    function: String = #function,
+                                    file: String = #file,
+                                    line: Int = #line) -> DependencySource {
+    guard let source = inputDependencySourceMap.sourceIfKnown(for: input)
     else {
       fatalError("\(input.file.basename) not found in inputDependencySourceMap, \(file):\(line) in \(function)")
     }
     return source
   }
 
-  @_spi(Testing) public func getNeededInput(for source: DependencySource) -> TypedVirtualPath? {
-    guard let input = inputDependencySourceMap.getInputIfKnown(for: source)
+  @_spi(Testing) public func input(neededFor source: DependencySource) -> TypedVirtualPath? {
+    guard let input = inputDependencySourceMap.input(ifKnownFor: source)
     else {
       info.diagnosticEngine.emit(warning: "Failed to find source file for '\(source.file.basename)', recovering with a full rebuild. Next build will be incremental.")
       return nil
@@ -141,7 +141,7 @@ extension ModuleDependencyGraph {
       return TransitivelyInvalidatedInputSet()
     }
     return collectInputsRequiringCompilationAfterProcessing(
-      dependencySource: getRequiredSource(for: input))
+      dependencySource: source(requiredFor: input))
   }
 }
 
@@ -163,14 +163,14 @@ extension ModuleDependencyGraph {
   /// speculatively scheduled in the first wave.
   func collectInputsInvalidatedBy(input: TypedVirtualPath
   ) -> TransitivelyInvalidatedInputArray {
-    let changedSource = getRequiredSource(for: input)
+    let changedSource = source(requiredFor: input)
     let allDependencySourcesToRecompile =
       collectSwiftDepsUsing(dependencySource: changedSource)
 
     return allDependencySourcesToRecompile.compactMap {
       depedencySource in
       guard depedencySource != changedSource else {return nil}
-      let dependentInput = inputDependencySourceMap.getInputIfKnown(for: depedencySource)
+      let dependentInput = inputDependencySourceMap.input(ifKnownFor: depedencySource)
       info.reporter?.report(
         "Found dependent of \(input.file.basename):", dependentInput)
       return dependentInput
@@ -190,7 +190,7 @@ extension ModuleDependencyGraph {
   /// Does the graph contain any dependency nodes for a given source-code file?
   func containsNodes(forSourceFile file: TypedVirtualPath) -> Bool {
     precondition(file.type == .swift)
-    guard let source = inputDependencySourceMap.getSourceIfKnown(for: file) else {
+    guard let source = inputDependencySourceMap.sourceIfKnown(for: file) else {
       return false
     }
     return containsNodes(forDependencySource: source)
@@ -203,7 +203,7 @@ extension ModuleDependencyGraph {
   
   /// Returns: false on error
   func populateInputDependencySourceMap(
-    `for` purpose: InputDependencySourceMap.AdditionPurpose
+    for purpose: InputDependencySourceMap.AdditionPurpose
   ) -> Bool {
     let ofm = info.outputFileMap
     let diags = info.diagnosticEngine
@@ -250,7 +250,7 @@ extension ModuleDependencyGraph {
   func collectInputsRequiringCompilation(byCompiling input: TypedVirtualPath
   ) -> TransitivelyInvalidatedInputSet? {
     precondition(input.type == .swift)
-    let dependencySource = getRequiredSource(for: input)
+    let dependencySource = source(requiredFor: input)
     return collectInputsRequiringCompilationAfterProcessing(
       dependencySource: dependencySource)
   }
@@ -345,7 +345,7 @@ extension ModuleDependencyGraph {
   ) -> TransitivelyInvalidatedInputSet? {
     var invalidatedInputs = TransitivelyInvalidatedInputSet()
     for invalidatedSwiftDeps in collectSwiftDepsUsingInvalidated(nodes: directlyInvalidatedNodes) {
-      guard let invalidatedInput = getNeededInput(for: invalidatedSwiftDeps)
+      guard let invalidatedInput = input(neededFor: invalidatedSwiftDeps)
       else {
         return nil
       }
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -209,10 +209,9 @@ extension ModuleDependencyGraph {
     let diags = info.diagnosticEngine
     var allFound = true
     for input in info.inputFiles {
-      if let source = ofm.getDependencySource(for: input, diagnosticEngine: diags) {
+      if let source = ofm.dependencySource(for: input, diagnosticEngine: diags) {
         inputDependencySourceMap.addEntry(input, source, for: purpose)
-      }
-      else {
+      } else {
         // Don't break in order to report all failures.
         allFound = false
       }
@@ -221,7 +220,7 @@ extension ModuleDependencyGraph {
   }
 }
 extension OutputFileMap {
-  fileprivate func getDependencySource(
+  fileprivate func dependencySource(
     for sourceFile: TypedVirtualPath,
     diagnosticEngine: DiagnosticsEngine
   ) -> DependencySource? {
@@ -345,8 +344,7 @@ extension ModuleDependencyGraph {
   ) -> TransitivelyInvalidatedInputSet? {
     var invalidatedInputs = TransitivelyInvalidatedInputSet()
     for invalidatedSwiftDeps in collectSwiftDepsUsingInvalidated(nodes: directlyInvalidatedNodes) {
-      guard let invalidatedInput = input(neededFor: invalidatedSwiftDeps)
-      else {
+      guard let invalidatedInput = input(neededFor: invalidatedSwiftDeps) else {
         return nil
       }
       invalidatedInputs.insert(invalidatedInput)
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -24,19 +24,7 @@ import SwiftOptions
   @_spi(Testing) public var nodeFinder = NodeFinder()
   
   /// Maps input files (e.g. .swift) to and from the DependencySource object.
-  ///
-  // FIXME: The map between swiftdeps and swift files is absolutely *not*
-  // a bijection. In particular, more than one swiftdeps file can be encountered
-  // in the course of deserializing priors *and* reading the output file map
-  // *and* re-reading swiftdeps files after frontends complete
-  // that correspond to the same swift file. These cause two problems:
-  // - overwrites in this data structure that lose data and
-  // - cache misses in `getInput(for:)` that cause the incremental build to
-  // turn over when e.g. entries in the output file map change. This should be
-  // replaced by a multi-map from swift files to dependency sources,
-  // and a regular map from dependency sources to swift files -
-  // since that direction really is one-to-one.
-  @_spi(Testing) public private(set) var inputDependencySourceMap = BidirectionalMap<TypedVirtualPath, DependencySource>()
+  @_spi(Testing) public private(set) var inputDependencySourceMap = InputDependencySourceMap()
 
   // The set of paths to external dependencies known to be in the graph
   public internal(set) var fingerprintedExternalDependencies = Set<FingerprintedExternalDependency>()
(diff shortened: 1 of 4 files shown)


davidungar:
@swift-ci please test macOS Platform

davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,64 @@
...
+// MARK: - Populating
+extension InputDependencySourceMap {
+  public enum AdditionPurpose {
+    case mocking,
+         buildingFromSwiftDeps,
+         readingPriors,
+         inputsAddedSincePriors }
+  @_spi(Testing) public mutating func addEntry(_ input: TypedVirtualPath,
+                                               _ dependencySource: DependencySource,
+                                               `for` _ : AdditionPurpose) {
```
CodaFi, line 60: Nit: You don't need to quote keywords when they're external labels. If Xcode is highlighting this incorrectly, please file a radar.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,64 @@
...
+    _ eachFn: (TypedVirtualPath, DependencySource) -> Void
+  ) {
+    biMap.forEach(eachFn)
+  }
+}
+
+// MARK: - Populating
+extension InputDependencySourceMap {
+  public enum AdditionPurpose {
+    case mocking,
```
CodaFi, line 54: Please document these. Their names alone don't reveal much at the point of usage.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -219,17 +204,47 @@ extension ModuleDependencyGraph {
     return nodeFinder.findNodes(for: source).map {!$0.isEmpty}
       ?? false
   }
-
-  /// Return true on success
-  func populateInputDependencySourceMap() -> Bool {
+  
+  /// Returns: false on error
+  func populateInputDependencySourceMap(
+    `for` purpose: InputDependencySourceMap.AdditionPurpose
```
CodaFi, line 210: Nit: Same here. Don't need the quotes around `for`.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,64 @@
...
+extension InputDependencySourceMap {
+  @_spi(Testing) public func getSourceIfKnown(for input: TypedVirtualPath) -> DependencySource? {
+    biMap[input]
+  }
+
+  @_spi(Testing) public func getInputIfKnown(for source: DependencySource) -> TypedVirtualPath? {
+    biMap[source]
+  }
+
+  @_spi(Testing) public func enumerateToSerializePriors(
```
CodaFi, line 44: This name conflates the API's purpose with its usage. I think the swift-ier name is something closer to just `enumerate(_:)`.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,64 @@
...
+  // and a regular map from dependency sources to swift files -
+  // since that direction really is one-to-one.
+  
+  public typealias BiMap = BidirectionalMap<TypedVirtualPath, DependencySource>
+  @_spi(Testing) public var biMap = BiMap()
+}
+
+// MARK: - Accessing
+extension InputDependencySourceMap {
+  @_spi(Testing) public func getSourceIfKnown(for input: TypedVirtualPath) -> DependencySource? {
```
CodaFi, line 36: Just `source(for:)` would do here.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,64 @@
...
+    biMap[input]
+  }
+
+  @_spi(Testing) public func getInputIfKnown(for source: DependencySource) -> TypedVirtualPath? {
```
CodaFi, line 40: just `input(for:)` as well

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: Thanks, will do.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: Okey-doke!

CodaFi (changes requested):
Some documentation of this new interface would be welcome. API notes are in-line.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: The conflation is quite deliberate here. The enclosing type is meant to operate at the level of *intention*. This level distinction paves the way for the greater structural change to come.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: Yes, it would. And I'll remove the "get" prefix.

 The idea here was to draw a very explicit distinction with `ModuleDependencyGraph.getRequiredSource`. In an earlier draft, both were named `getSource` and it was not easy to search for one as opposed to the other, because the only difference in usage was the result type. So, I decided to rename at least one, for the sake of clarity. Then, for the sake of symmetry (and clarity), I renamed them both.
So, I would rather keep the "IfKnown" part.

davidungar: Hmm... after eliminating the "get"s, the "IfKnown" etc., moves to the keyword. You'll see after I push. Does that look better?

davidungar:
> Some documentation of this new interface would be welcome. API notes are in-line.

Yes, I'll add.

davidungar:
Thanks for the comments; they helped! Have addressed them, and now will fix the conflicts.

davidungar:
@swift-ci please test

davidungar:
Thanks, @CodaFi , for productive comments. I've made the changes requested, modulo keeping some intentionality in the function names. This rev may not be there yet, so please let me know what you see.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
CodaFi: I understand that, but Swift does not use the name of the API in the same way that Objective-C or Smalltalk might encourage the use of intention-revealing selectors.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,93 @@
...
+  /// - Returns: the corresponding `DependencySource`, or nil if none known.
+  @_spi(Testing) public func sourceIfKnown(for input: TypedVirtualPath) -> DependencySource? {
+    biMap[input]
+  }
+
+  /// Find where the source file is for a given swiftdeps file.
+  ///
+  /// - Parameter source: A `DependencySource` (containing a swiftdeps file)
+  /// - Returns: the corresponding input source file, or nil if none known.
+  @_spi(Testing) public func input(ifKnownFor source: DependencySource) -> TypedVirtualPath? {
```
CodaFi, line 51: `ifKnownFor` has the same issue as below. Just `for` suffices.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
```diff
@@ -121,7 +121,8 @@ extension ModuleDependencyGraph.Tracer {
         path.compactMap { node in
           node.dependencySource.map {
             source in
-            graph.inputDependencySourceMap[source].map { input in
+            graph.inputDependencySourceMap.input(ifKnownFor: source).map {
```
CodaFi, line 124: Any reason this is here? Seems superfluous.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /// Add a mapping to & from input source file to swiftDeps file path
+  ///
+  /// - Parameter input: the source file path
+  /// - Parameter dependencySource: the dependency source (holding the swiftdeps path)
+  /// - Parameter why: the purpose for this addition. Will be used for future restructuring.
+  @_spi(Testing) public mutating func addEntry(_ input: TypedVirtualPath,
+                                               _ dependencySource: DependencySource,
+                                               for why: AdditionPurpose) {
```
CodaFi, line 89: It looks like the `for` parameter is dead. Is that intentional?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -214,17 +200,47 @@ extension ModuleDependencyGraph {
...
+      else {
+        // Don't break in order to report all failures.
+        allFound = false
+      }
+    }
+    return allFound
+  }
+}
+extension OutputFileMap {
+  fileprivate func getDependencySource(
```
CodaFi, line 224: Nit from before: `get` prefixes usually imply indirect return values in Cocoa and Swift. Just `dependencySource(for:diagnosticEngine:)` suffices.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -214,17 +200,47 @@ extension ModuleDependencyGraph {
...
-    return info.inputFiles.reduce(true) { okSoFar, input in
-      ofm.getDependencySource(for: input, diagnosticEngine: de)
-        .map {source in addMapEntry(input, source); return okSoFar } ?? false
+    let diags = info.diagnosticEngine
+    var allFound = true
+    for input in info.inputFiles {
+      if let source = ofm.getDependencySource(for: input, diagnosticEngine: diags) {
+        inputDependencySourceMap.addEntry(input, source, for: purpose)
+      }
+      else {
```
CodaFi, line 215: Nit: OTBS here please

```suggestion
      if let source = ofm.getDependencySource(for: input, diagnosticEngine: diags) {
         inputDependencySourceMap.addEntry(input, source, for: purpose)
       } else {
```

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -66,22 +54,20 @@ import SwiftOptions
...
-    assert(input.type == .swift && dependencySource.typedFile.type == .swiftDeps)
-    inputDependencySourceMap[input] = dependencySource
-  }
-
-  @_spi(Testing) public func getSource(for input: TypedVirtualPath,
-                                       function: String = #function,
-                                       file: String = #file,
-                                       line: Int = #line) -> DependencySource {
-    guard let source = inputDependencySourceMap[input] else {
+  @_spi(Testing) public func source(requiredFor input: TypedVirtualPath,
```
CodaFi, line 57: The `requirement` part of this API is communicated by the type signature being non-optional. You could reinforce this in documentation, but I think `for` suffices.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -329,7 +345,8 @@ extension ModuleDependencyGraph {
   ) -> TransitivelyInvalidatedInputSet? {
     var invalidatedInputs = TransitivelyInvalidatedInputSet()
     for invalidatedSwiftDeps in collectSwiftDepsUsingInvalidated(nodes: directlyInvalidatedNodes) {
-      guard let invalidatedInput = getInput(for: invalidatedSwiftDeps) else {
+      guard let invalidatedInput = input(neededFor: invalidatedSwiftDeps)
+      else {
```
CodaFi, line 349: Put this back if you could. This fits onto 80 columns just fine.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: Thanks--done!

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: I did a look-up on "OTBS" and it came back with "off track horse betting"! :)

davidungar: Looks like the request is to put the else on the same line as the closing brace--no prob!

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: What's the rule you follow in general for the "else" in a guard? 80-columns?

davidungar: Will do.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: By putting the requirement in the function name, it becomes more salient at the point of use, *without* clicking to see the documentation, or the type of the assigned-to variable.
Consider: `let foo = source(requiredFor: someInput)` vs `let foo = source(for: someInput)`.
The first construction makes the required nature of the mapping more salient.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: Yes. As the documentation says, it will be used in the future. For now, it makes clear at the point-of-use what the purpose is.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
davidungar: The "ifKnownFor" is there because it makes two aspects more salient:
1. That the function might not return a value. (In this use, it reinforces that the `map` is to handle an optional, rather than a collection.), and
2. That it depends on if the mapping is "known". I'd be happy to find a better word than "known". "Has been seen" feels too wordy.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: Please see my comment for the next issue. I have a rule: Correlate salience with importance in order to ease the comprehension for new maintainer.

davidungar:
@CodaFi Thank you for your review. I have responded to your comments and made changes where I agree. You have a wealth of experience with Swift style, and that helps my code fit in. To clarify, I think we agree that this PR makes no functional change, and that (for me) is the most critical part of your review, because I don't want to break the driver. There are some remaining disagreements on function naming, but I think they follow from my practice of choosing salience at the point-of-use. If Xcode were better at showing inferred types, or Doxygen documentation, I'd be in agreement with you. 

Would you be able to approve this PR at this point?

davidungar:
@swift-ci please test

CodaFi (approved):
I remain unconvinced that this is a meaningful change, and moreover that the new API surface here is an improvement over the status quo. But this PR has also been blocked on my review and I don’t see a reason for that much, so

davidungar:
> I remain unconvinced that this is a meaningful change, and moreover that the new API surface here is an improvement over the status quo. But this PR has also been blocked on my review and I don’t see a reason for that much, so

I understand. Thank you so very much for compromising here.

PR #665, opened 2021-05-14
[Build Script Helper] Restrict cross-compilation to producing fat macOS build products

artemcm:
For now, allow `x86_64` and `arm64` fat builds, mimicking how SwiftPM builds:
https://github.com/apple/swift-package-manager/blob/main/Utilities/bootstrap#L808
 
Other cross-compilation flows will require more testing.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -617,8 +617,14 @@ def main():
   else:
     args.sysroot = None
 
-  if args.cross_compile_hosts and not all('apple-macos' in target for target in args.cross_compile_hosts):
-    error('Cross-compilation is currently only supported for the Darwin platform.')
+  swift_exec = os.path.join(os.path.join(args.toolchain, 'bin'), 'swiftc')
+  build_target = get_build_target(swift_exec, args)
+  if (build_target == 'x86_64-apple-macosx' and 'macosx-arm64' in args.cross_compile_hosts):
+      args.cross_compile_hosts = [build_target + macos_deployment_target, 'arm64-apple-macosx%s' % macos_deployment_target]
+  elif (build_target == 'arm64-apple-macosx' and 'macosx-x86_64' in args.cross_compile_hosts):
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci please test

nkcsgexi (approved):
Thank you fixing this! Do we need to have this logic in every build-script-helper.py?

artemcm:
> Thank you fixing this! Do we need to have this logic in every build-script-helper.py?

I think we might, at least in some of them. 
Just looking at the latest toolchain snapshot from https://swift.org/download/, `swift-driver` is not the only executable that is only built for the host arch, instead of universal.
`sk-stress-test` and `sourcekit-lsp`, `libSwiftSyntax.dylib`, `lib_InternalSwiftSyntaxParser.dylib`, `libArgumentParser.dylib` and more are not built universal.

nkcsgexi:
> > Thank you fixing this! Do we need to have this logic in every build-script-helper.py?
> 
> I think we might, at least in some of them.
> Just looking at the latest toolchain snapshot from https://swift.org/download/, `swift-driver` is not the only executable that is only built for the host arch, instead of universal.
> `sk-stress-test` and `sourcekit-lsp`, `libSwiftSyntax.dylib`, `lib_InternalSwiftSyntaxParser.dylib`, `libArgumentParser.dylib` and more are not built universal.

cc: @akyrtzi

akyrtzi:
`lib_InternalSwiftSyntaxParser.dylib` is coming from the swift compiler build, is it the only binary from the compiler build that is not universal?

artemcm:
> `lib_InternalSwiftSyntaxParser.dylib` is coming from the swift compiler build, is it the only binary from the compiler build that is not universal?

At a quick glance through the `lib` directory of a snapshot toolchain, yes.
I think we will need to include it under a change similar to:
https://github.com/apple/swift/pull/36960

I did not include it right then because I was looking at a copy of `lib_InternalSwiftSyntaxParser.dylib` built separately so it was built universal. But in the package build, it appears that it does not.

akyrtzi:
Ah good catch! I'll look into adding it.

PR #668, opened 2021-05-20
Intro a flag to disable emit-module-separately

xymus:
We want to make emit-module-separately the default incremental build mode soon. Let's first add a flag to disable it to be safe. Once emit-module-separately is the default, passing this flag will force the driver to revert to using the merge-module incremental build mode.


xymus:
@swift-ci Please test

artemcm (approved):

PR #673, opened 2021-05-21
[5.5] New Driver: do not generate a dSYM bundle for static archives

gottesmm:
This is the new driver version of apple/swift#37539.

rdar://78334766

----

Explanation: When emitting a static library, the driver attempts to run dsymutil on it. dsymutil is not supposed to be run on archives so it fails. As a result, one cannot compile static libraries with debug info at all. This patch just fixes the behavior by checking when adding the dsymutil job if we are emitting a static library and in such a case, do not add the dsymutil job.
Scope: Small. Just ensures that if we are compiling a static library we do not add a dsymutil job in the driver.
Risk: Low.
Testing: Adding a compiler test that validates the behavior.
Issue: rdar://78271443
Reviewer: @artemcm

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -469,8 +469,18 @@ extension Driver {
 
     let linkJ = try linkJob(inputs: linkerInputs)
     addJob(linkJ)
-    guard targetTriple.isDarwin, debugInfo.level != nil
-    else {return }
+    guard targetTriple.isDarwin
+    else { return }
+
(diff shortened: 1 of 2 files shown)


gottesmm:
@swift-ci test

artemcm (approved):
LGTM for the 5.5 branch.

nkcsgexi (approved):

gottesmm:
@swift-ci nominate

airspeedswift (approved):

gottesmm:
@airspeedswift thanks!

PR #674, opened 2021-05-23
Delay most tool lookup until after build planning

owenv:
This makes testing easier because fewer tools need to be present locally to plan builds for various targets, and in theory it simplifies distributed builds a little. 

I'd be interested in hearing thoughts on the approach used here. The main change here is that `Job` now has a `Tool` instead of the `AbsolutePath` to a tool, and the `DriverExecutor` entry points now accept the locations of the tools needed to run their workload. I thought about moving the tool lookup operations into `ArgsResolver`, but since they require a `Toolchain` I thought it made sense to keep them in `Driver` for now.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -935,7 +935,9 @@ extension Driver {
     // If we're only supposed to print the jobs, do so now.
     if parsedOptions.contains(.driverPrintJobs) {
       for job in jobs {
-        print(try executor.description(of: job, forceResponseFiles: forceResponseFiles))
+        print(try executor.description(of: job,
+                                       toolPath: try toolchain.getToolPath(job.tool),
+                                       forceResponseFiles: forceResponseFiles))
       }
       return
     }
@@ -952,7 +954,7 @@ extension Driver {
 
     if parsedOptions.contains(.driverPrintBindings) {
       for job in jobs {
-        printBindings(job)
+        try printBindings(job)
       }
       return
     }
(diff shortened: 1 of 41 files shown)


owenv:
@swift-ci please test

owenv:
Leaving this as a draft for now because I'll need to update SwiftPM before merging

artemcm:
I like the idea of treating tools similarly to how we treat arguments, and separating resolving their FS paths from build planning. 

Bit it is a bit of a shame though to have to make `libSwiftDriver` clients have to implement yet another path resolution step in their executors. I imagine the plan here is for SwiftPM to call out to the Driver's `Toolchain.getToolPath` to resolve a tool for a given job before passing it in to `resolveArgumentList`? [`Toolchain`](https://github.com/apple/swift-driver/blob/main/Sources/SwiftDriver/Driver/Driver.swift#L119) is not public API so we'd need to provide a wrapper of some kind (`ToolsResolver`?). 

Perhaps we can get most of the benefit if we make tool resolution an explicit step that happens at the end of build planning (in a way the tests or other clients can disable)? With something like `case resolved(AbsolutePath)` in `Tool` we could even keep the current API by forwarding `Job.tool` to the resolved case so that the executors don't need to be aware of the change. 

@BenchR267 what do you think about this?

cltnschlosser:
I haven’t looked at this yet. I like the idea though. One question came to mind: Does the presence of a tool ever change which jobs are created?

owenv:
> I imagine the plan here is for SwiftPM to call out to the Driver's Toolchain.getToolPath to resolve a tool for a given job before passing it in to resolveArgumentList? 

Hmm, I hadn't thought this all the way through yet, I'll have to think about what the interface should be for library clients. A new cas on Tool might work, I'll look into it. 

> Does the presence of a tool ever change which jobs are created?

Some parts of build planning check for specific resources in the file system, but as far as I know nothing depends on the presence of a tool right now

PR #675, opened 2021-05-24
[Incremental] Harness immutability to make `inputDependencySourceMap` robust

davidungar:
Make the `inputDependencyMap` immutable, solely reflecting the `OutputFileMap`.
Do not include it in the priors.

In a previous PR, @CodaFi uncovered and explained issues with the `inputDependencySourceMap:
```
   // FIXME: The map between swiftdeps and swift files is absolutely *not*`
   // a bijection. In particular, more than one swiftdeps file can be encountered
   // in the course of deserializing priors *and* reading the output file map
   // *and* re-reading swiftdeps files after frontends complete
   // that correspond to the same swift file. These cause two problems:
   // - overwrites in this data structure that lose data and
   // - cache misses in `getInput(for:)` that cause the incremental build to
   // turn over when e.g. entries in the output file map change. This should be
   // replaced by a multi-map from swift files to dependency sources,
   // and a regular map from dependency sources to swift files -
   // since that direction really is one-to-one.
```

This PR solves the problem in a simpler way. The key insight is that the output file map contains the information linking a source file to the swiftdeps file for any build. Thus, the output file map should just determine the entries in the `inputDependencySourceMap`, just as it did before the introduction of incremental imports. Then the map in question does become a true bijection. It can also be made immutable, which greatly eases reasoning about the code.

It will also make it easy to cull removed file nodes from priors in a future PR.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
@@ -256,7 +256,6 @@ extension IncrementalCompilationState.IncrementalDependencyAndInputSetup {
     else {
       return nil
     }
-    
     var inputsInvalidatedByChangedExternals = TransitivelyInvalidatedInputSet()
     for input in sourceFiles.currentInOrder {
        guard let invalidatedInputs =
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
@@ -227,9 +227,6 @@ extension IncrementalCompilationState.IncrementalDependencyAndInputSetup {
     else {
       return buildInitialGraphFromSwiftDepsAndCollectInputsInvalidatedByChangedExternals()
     }
-    guard graph.populateInputDependencySourceMap(for: .inputsAddedSincePriors) else {
-      return nil
-    }
     graph.dotFileWriter?.write(graph)
 
     // Any externals not already in graph must be additions which should trigger
@@ -255,13 +252,11 @@ extension IncrementalCompilationState.IncrementalDependencyAndInputSetup {
   private func buildInitialGraphFromSwiftDepsAndCollectInputsInvalidatedByChangedExternals()
   -> (ModuleDependencyGraph, TransitivelyInvalidatedInputSet)?
   {
-    let graph = ModuleDependencyGraph(self, .buildingWithoutAPrior)
-    assert(outputFileMap.onlySourceFilesHaveSwiftDeps())
-    
-    guard graph.populateInputDependencySourceMap(for: .buildingFromSwiftDeps) else {
+    guard let graph = ModuleDependencyGraph(self, .buildingWithoutAPrior)
+    else {
       return nil
     }
-
+    
     var inputsInvalidatedByChangedExternals = TransitivelyInvalidatedInputSet()
     for input in sourceFiles.currentInOrder {
        guard let invalidatedInputs =
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -24,7 +24,7 @@ import SwiftOptions
   @_spi(Testing) public var nodeFinder = NodeFinder()
   
   /// Maps input files (e.g. .swift) to and from the DependencySource object.
-  @_spi(Testing) public private(set) var inputDependencySourceMap = InputDependencySourceMap()
+  @_spi(Testing) public let inputDependencySourceMap: InputDependencySourceMap
 
   // The set of paths to external dependencies known to be in the graph
   public internal(set) var fingerprintedExternalDependencies = Set<FingerprintedExternalDependency>()
@@ -43,7 +43,7 @@ import SwiftOptions
   /// Minimize the number of file system modification-time queries.
   private var externalDependencyModTimeCache = [ExternalDependency: Bool]()
 
-  public init(_ info: IncrementalCompilationState.IncrementalDependencyAndInputSetup,
+  public init?(_ info: IncrementalCompilationState.IncrementalDependencyAndInputSetup,
               _ phase: Phase
   ) {
     self.info = info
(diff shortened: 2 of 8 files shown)


davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
```diff
@@ -121,7 +121,8 @@ extension ModuleDependencyGraph.Tracer {
         path.compactMap { node in
           node.dependencySource.map {
             source in
-            graph.inputDependencySourceMap[source].map { input in
+            graph.inputDependencySourceMap.input(ifKnownFor: source).map {
+              input in
```
CodaFi, line 125: Please remove this change

Tests/SwiftDriverTests/Helpers/MockingIncrementalCompilation.swift
```diff
@@ -139,3 +144,38 @@ extension Collection where Element: StringProtocol {
...
+  }
+
+  func mockUpAGraph() -> ModuleDependencyGraph {
+    ModuleDependencyGraph(info, .buildingWithoutAPrior)!
+  }
+}
+
+
+extension OutputFileMap {
+  fileprivate static func mock(maxIndex: Int) -> Self {
```
CodaFi, line 172: Nit: Whitespace here is quite weird

davidungar:
@swift-ci please test

Tests/SwiftDriverTests/Helpers/MockingIncrementalCompilation.swift
davidungar: Is the white space there still weird? I hope I've fixed it.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
davidungar: Is this request still relevant now that the map has been turned into its own, domain-specific type?

davidungar:
@CodaFi Would you mind taking a look at this? Because you uncovered and explored the issues before with this map, you'll be a great reviewer to assess the correctness of this change. Thanks!

davidungar:
@swift-ci please test

davidungar:
@CodaFi Any word on when you might have time for this? Thanks!

Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
```diff
@@ -255,13 +252,11 @@ extension IncrementalCompilationState.IncrementalDependencyAndInputSetup {
...
-    let graph = ModuleDependencyGraph(self, .buildingWithoutAPrior)
-    assert(outputFileMap.onlySourceFilesHaveSwiftDeps())
-    
-    guard graph.populateInputDependencySourceMap(for: .buildingFromSwiftDeps) else {
+    guard let graph = ModuleDependencyGraph(self, .buildingWithoutAPrior)
+    else {
       return nil
     }
-
+
```
artemcm, line 259: Nit: trailing whitespace.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -671,27 +634,6 @@ extension ModuleDependencyGraph {
             throw ReadError.malformedDependsOnRecord
           }
           self.nodeUses.append( (key, Int(record.fields[0])) )
-        case .mapNode:
```
artemcm, line 674: This tells me that we are no longer reading in swiftdeps files from serialized priors. I'm lacking context here to understand the implications of this, where were these deserialized files used? 

They are one of the examples that made the `SourceMap` not-truly-bijective, is it the case that they never led to correct behavior? Or if they resulted in another behavior, where were we relying on it that we now choose not to?

artemcm (approved):
On the surface, this change seems like and improvement to me, and the problems Robert outlined shouldn't be possible with this data-structure being immutable. It seems reasonable to have the outputFileMap be the canonical source of dependencies for each input and it simplifies the logic quite a bit. 

I'd still really like @CodaFi to weigh in because I'm less familiar with the scenarios that previously led to conflict and whether we were relying on serialized priors containing dependency sources in a way I'm not familiar with. And whether that could have implications on incremental build performance.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: I think I understand the confusion here.

The intention of this change is to no longer serialize and deserialize the `inputDependencySourceMap`. Serialized `mapNode`s were only used for that purpose. We still read in `ModuleDependencyGraph`s from the priors.

Yes, serializing the map and then deserialized it created confusion because the deserialized version might be inconsistent with the info from the `OutputFileMap`. This PR is premised on the believe that it is the `OutputFileMap` that should be the source of truth.

Does this explanation satisfy? Have I missed something?

artemcm: Yes that makes sense, thank you. And it also makes sense for the output-file-map to be the canonical source of truth. Probably so much so that I am wondering why we serialized the `inputDependencySourceMap` in the first place, hence the confusion.

davidungar:
> On the surface, this change seems like and improvement to me, and the problems Robert outlined shouldn't be possible with this data-structure being immutable. It seems reasonable to have the outputFileMap be the canonical source of dependencies for each input and it simplifies the logic quite a bit.
> 
> I'd still really like @CodaFi to weigh in because I'm less familiar with the scenarios that previously led to conflict and whether we were relying on serialized priors containing dependency sources in a way I'm not familiar with. And whether that could have implications on incremental build performance.

Thank you for your review. Yes, it would still be ideal for @CodaFi to weigh in. As I haven't received any indication of how long it will be till he can review this PR, I plan to fix the white-space issue and merge.

davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: Thank you for validating the notion of output-file-map as canonical. Wrt to serializing the `inputDependencySourceMap` I have to plead mea culpa. It seemed like a good idea at the time because the `inputDependencySourceMap` was being dynamically build as `swiftdeps` files were processed. When @CodaFi added the prior deserialization code, IIRC, there was a moment when we had deserialized priors with an empty map. So I (wrongly) fixed it by serializing the `inputDependencySourceMap`. My mistake, in conjuction with the build system's proclivity to change file names when a C++ file and Swift file had the same basename-without-extension, led to hilarity ensuing. @CodaFi got to the bottom of things there, and identified the problems with that design.

Later, I realized that the output-file-map should just be the truth, and that, consequently, the `inputDependencySourceMap` *could* be both immutable and a bijection. I hope I'm right this time!

davidungar:
rdar://77998890

PR #685, opened 2021-05-28
[NFC, Incremental] Refactoring to facilitate replacing the inputDependencySourceMap with the right thing

davidungar:
Redo of reverted https://github.com/apple/swift-driver/pull/664 ; reverted because I forgot to add a new file to CMakeLists.txt.

This PR paves the way for bug fixes by adding an intentions layer to the usage of a bidirectional map to go from input files (.swift) to and from dependency files (.swiftdeps).

1. Wrap the `BidirectionalMap` into its own type, and put that into `InputDependencySourceMap.swift`
2. Make the interface very explicit as to intentions:
   - Add an `enum` to explain why entries are being added (`AdditionPurpose`) and pass that through calls that add entries
   - `addMapEntry` in the `ModuleDependencyGraph` becomes `addEntry` on the new type
   - The new type does lookups via `sourceIfKnown(for:)` and `inputIfKnown(for:)` to contrast the possibility of failure with `sourceRequired(for:)`, which is the new name for `getSource(for:)` in the graph method.
3. Fixes a misnomer in a local variable in `collectInputsInvalidatedBy(input:)`
4. The `reduce` in `populateInputDependencySourceMap` has been rewritten for clarity.

In preparation for: rdar://77998890

### DIFF
--- a/Sources/SwiftDriver/CMakeLists.txt
+++ b/Sources/SwiftDriver/CMakeLists.txt
@@ -33,6 +33,7 @@ add_library(SwiftDriver
   Execution/ParsableOutput.swift
   Execution/ProcessProtocol.swift
 
+  "IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift"
   "IncrementalCompilation/ModuleDependencyGraphParts/Integrator.swift"
   "IncrementalCompilation/ModuleDependencyGraphParts/Node.swift"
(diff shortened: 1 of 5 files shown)


davidungar:
@swift-ci please test

akyrtzi (approved):

davidungar:
Thanks, @akyrtzi !

PR #686, opened 2021-05-28
[NFC] Refactor some incremental compilation tests, prior to adding more tests

davidungar:
The behavior-neutral changes that will be wanted when more tests are added:

1. Add an `inputPath` method and use it where needed.
2. Add a `write` method to factor out future input writing.
3. Eliminate many cases where the tests would just crash. Raise exceptions instead. Add 'try's.
4. Rewrite `doABuild` and return a non-optional.


davidungar:
@swift-ci please test

artemcm (approved):
Nice cleanup, thanks!

davidungar:
My pleasure and thank you, @artemcm !

davidungar:
rdar://77998890

PR #687, opened 2021-05-28
[Incremental] Add input addition test and fix nondeterminism bug

davidungar:
This PR adds a test to test input additions and fixes a bug that can manifest in nondeterministic outputs from `-driver-show-incremental`.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
@@ -193,9 +193,7 @@ extension IncrementalCompilationState.FirstWaveComputer {
   }
 
   private func sortByCommandLineOrder(_ inputs: Set<TypedVirtualPath>) -> [TypedVirtualPath] {
-    let indices = Dictionary(uniqueKeysWithValues: inputs.enumerated()
-                              .map {offset, element in (element, offset)})
-    return inputs.sorted {indices[$0]! < indices[$1]!}
+    inputFiles.filter (inputs.contains)
   }
 
   /// Encapsulates information about an input the driver has determined has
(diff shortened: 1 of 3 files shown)


davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

artemcm (approved):

CodaFi (changes requested):
(whoops, read the wrong variable there)

CodaFi (approved):

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
```diff
@@ -193,9 +193,7 @@ extension IncrementalCompilationState.FirstWaveComputer {
   }
 
   private func sortByCommandLineOrder(_ inputs: Set<TypedVirtualPath>) -> [TypedVirtualPath] {
```
CodaFi, line 195: This function can be removed and replaced by a `where` clause on the appropriate loops. This will avoid re-materializing the inputs into a separate collection.

davidungar: @CodaFi Thank you very much for a quick review and your approval despite this comment.
I am reluctant to perform the optimization you suggest because:
1. Factoring the functionality out allows it to possess a descriptive name, and
2. It is called only when `-driver-show-incremental`  is used. Thus, it would not affect performance in normal use-cases.

Would there be any way to preserve the factoring and gain the optimization, too?
For example, suppose the loop with `where` clause were in the function and it took a closure parameter?
Would that be more efficient? How would the cost of the closure compare with the cost of materializing the resultant collection?

I'd be happy to do a future PR if both goals could be met. And thanks again, for your help here.

davidungar:
Thank you, @artemcm and @CodaFi !

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
CodaFi: > Would there be any way to preserve the factoring and gain the optimization, too?

Somebody has to have a loop with a where clause somewhere. This method could be transformed to take a non-escaping closure and turned into an iterator. After optimization, it should be equivalent to the loop+where clause form since its body should be quite profitable from the inliner's point of view.

CodaFi: You can also return a `LazyFilterSequence` here.

davidungar:
rdar://77998890

PR #697, opened 2021-06-04
Use `stat` Instead of `lstat`

CodaFi:
Using `lstat` here will do the wrong thing when the input path is a symlink since it's going to give us mod time info on the symlink instead of the file it references. Since mod time updates are intentionally not propagated across links, this has the potential to cause us to miscompile if a file's contents change on disk but the symlink passed as input hasn't actually changed. It also has the potential to defeat the incremental build when build systems that generate symlinks on every build interact with the driver.

On Darwin (and, technically, other UNIX-likes) the fix is simple: use stat instead of lstat which will resolve symlinks for us. On Windows the answer is... complicated. Instead, on the fallback path, resolve symlinks before retrieving the modification time of the file.

rdar://78828871

### DIFF
--- a/Sources/SwiftDriver/Utilities/VirtualPath.swift
+++ b/Sources/SwiftDriver/Utilities/VirtualPath.swift
@@ -721,18 +721,33 @@ extension TSCBasic.FileSystem {
     try resolvingVirtualPath(path, apply: exists)
   }
 
+  /// Retrieves the last modification time of the file referenced at the given path.
+  ///
+  /// If the given file path references a symbolic link, the modification time for the *linked file*
+  /// - not the symlink itself - is returned.
+  ///
+  /// - Parameter file: The path to a file.
+  /// - Throws: `SystemError` if the underlying `stat` operation fails.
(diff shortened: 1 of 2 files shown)


CodaFi:
This isn't quite finished yet, just wanted to solicit opinions about the approach, particularly on Windows.

CC @compnerd

compnerd:
Going through Foundation means that it _should_ do the right thing, and if it doesn't we should fix it in Foundation.  That said, generally, you don't want to use `stat` and there is no real `lstat` on Windows.  However, the `stat` behaviour on Windows isn't the Unix behaviour, and we need to manually chase the symlink :-(.  Implementing the default behaviour of chasing the symlinks is possible, though slightly expensive (recursive open + ioctl) if we really need.  I think that this approach is the simplest to implement while being correct and generally portable.

CodaFi:
Alright, good to go

@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

artemcm (approved):

PR #705, opened 2021-06-08
[Explicit Module Builds] Use context hash in Swift dependency module file names

artemcm:
This change ensures that explicit swift module dependencies are built into binary modules with filenames that encode their context hash (as reported by the dependency scanner). e.g. `A.swiftinterface` --> `A-<CONTEXT_HASH>.swiftmodule`.

This is required because in some situations builds of the same module for different contexts (e.g. target architectures) otherwise overlap and cause chaos.

Resolves rdar://78820404

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -11,7 +11,7 @@
         }
       },
       {
-        "package": "llbuild",
+        "package": "swift-llbuild",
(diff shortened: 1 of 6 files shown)


artemcm:
@swift-ci please test

PR #707, opened 2021-06-10
Disable output file map diagnostic for non-incremental mode

keith:
This re-implements this logic https://github.com/apple/swift/blob/3116eed2e465a2688f070b3b87adc106cbbaf09f/lib/Driver/Driver.cpp#L414-L417 to not warn when output file maps don't have a main entry when using WMO

### DIFF
--- a/Tests/TestUtilities/OutputFileMapCreator.swift
+++ b/Tests/TestUtilities/OutputFileMapCreator.swift
@@ -18,23 +18,28 @@ public struct OutputFileMapCreator {
   private let module: String
   private let inputPaths: [AbsolutePath]
   private let derivedData: AbsolutePath
+  // The main entry isn't required for some WMO builds
+  private let excludeMainEntry: Bool
 
-  private init(module: String, inputPaths: [AbsolutePath], derivedData: AbsolutePath) {
+  private init(module: String, inputPaths: [AbsolutePath], derivedData: AbsolutePath, excludeMainEntry: Bool) {
     self.module = module
     self.inputPaths = inputPaths
     self.derivedData = derivedData
+    self.excludeMainEntry = excludeMainEntry
   }
 
   public static func write(module: String,
                            inputPaths: [AbsolutePath],
                            derivedData: AbsolutePath,
-                           to dst: AbsolutePath) {
-    let creator = Self(module: module, inputPaths: inputPaths, derivedData: derivedData)
+                           to dst: AbsolutePath,
+                           excludeMainEntry: Bool = false) {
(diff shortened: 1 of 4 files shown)


brentleyjones (approved):

davidungar (changes requested):
Thank you for doing this. It looks good, but before merging I'd like to see an XCTest added to check this. It might be easiest to add a test in `IncrementalCompilationTests.swift`, in which the test runs the driver both with- and without the `-incremental` flag and checks to see whether the diagnostic is emitted when appropriate.

Would you like to add such a test?

brentleyjones:
FYI, this is still an issue in Xcode 13.0 beta 4.

keith:
Updated with tests!

artemcm:
@swift-ci please test

artemcm (approved):

artemcm:
@davidungar could you please take another look? It looks good to go to me.

keith:
@davidungar mind taking another look here?

keith:
@davidungar mind taking another look here?

Tests/TestUtilities/OutputFileMapCreator.swift
```diff
@@ -18,23 +18,27 @@ public struct OutputFileMapCreator {
   private let module: String
   private let inputPaths: [AbsolutePath]
   private let derivedData: AbsolutePath
+  private let excludeMainEntry: Bool
```
davidungar, line 21: Would it be possible to add a comment somewhere explaining when you would want to exclude the main entry?

davidungar (approved):
Sorry this PR slipped under my radar. Thanks to @CodaFi for pinging me. In future, please feel free to ping me directly. 
It would be great if you could add a comment somewhere (if it's not already there) explaining why one would want to exclude the main entry, or under what circumstances. Otherwise, LGTM and thank you for your patience!

keith:
Done, thanks for following up. To be clear where do you mean when you say "ping you directly"?

artemcm (approved):

artemcm:
@swift-ci please test

brentleyjones:
@artemcm Is this able to get CP'ed into 5.5?

artemcm:
> @artemcm Is this able to get CP'ed into 5.5?

Yeah, I think this change is pretty safe.
Please create a cherry-pick PR against `release/5.5` and use something like https://github.com/apple/swift/pull/39140 as a template for the PR description.

keith:
Thanks! Submitted https://github.com/apple/swift-driver/pull/828 for the CP

PR #709, opened 2021-06-10
PrebuiltModuleGen: teach the tool to dump a .dot file for module dependency visualization

nkcsgexi:


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
```diff
@@ -283,14 +283,68 @@ extension Driver {
     )
   }
 
+
+  func dumpDotGraph(_ graph: InterModuleDependencyGraph, _ path: AbsolutePath, _ includingPCM: Bool) throws {
```
artemcm, line 287: Should we make this an `InterModuleDependencyGraph` extension so that other uses can potentially do the same?

artemcm (approved):
Very nice! 
We have something similar in SwiftPM:
https://github.com/apple/swift-package-manager/blob/main/Sources/LLBuildManifest/DOTManifestSerializer.swift

But this will be super useful also. 
My only comment is that this looks like it'd be handy for general debugging with explicit module builds so perhaps we can make the dot-emission routine more accessible?

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
nkcsgexi: sure, will refactor that

nkcsgexi:
@swift-ci please test

PR #710, opened 2021-06-11
[5.5, NFC, Incremental] Create a domain-specific type for the inputDependencySourceMap

davidungar:
### Cherry-pick of  https://github.com/apple/swift-driver/pull/685 for the `release/5.5` branch.

>This PR paves the way for bug fixes by adding an intentions layer to the usage of a bidirectional map to go from input files (.swift) to and from dependency files (.swiftdeps).

> Wrap the BidirectionalMap into its own type, and put that into InputDependencySourceMap.swift
Make the interface very explicit as to intentions:
> - Add an enum to explain why entries are being added (AdditionPurpose) and pass that through calls that add entries
> - addMapEntry in the ModuleDependencyGraph becomes addEntry on the new type
> - The new type does lookups via sourceIfKnown(for:) and inputIfKnown(for:) to contrast the possibility of failure with sourceRequired(for:), which is the new name for getSource(for:) in the graph method.
> - Fixes a misnomer in a local variable in collectInputsInvalidatedBy(input:)
> - The reduce in populateInputDependencySourceMap has been rewritten for clarity.
> - In preparation for: rdar://77998890
rdar://79211265


davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

artemcm (approved):
LGTM for the 5.5 branch.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,70 @@
...
+  // - cache misses in `getInput(for:)` that cause the incremental build to
+  // turn over when e.g. entries in the output file map change. This should be
+  // replaced by a multi-map from swift files to dependency sources,
+  // and a regular map from dependency sources to swift files -
+  // since that direction really is one-to-one.
+  
+  public typealias BiMap = BidirectionalMap<TypedVirtualPath, DependencySource>
+  @_spi(Testing) public var biMap = BiMap()
+
+  private let simulateGetInputFailure: Bool
```
CodaFi, line 33: Why is this a format part of the API? We know how to create this failure mode in tests for the current implementation, and the new one should structurally avoid the issue altogether.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,70 @@
...
+    _ eachFn: (TypedVirtualPath, DependencySource) -> Void
+  ) {
+    biMap.forEach(eachFn)
+  }
+}
+
+// MARK: - Populating
+extension InputDependencySourceMap {
+  public enum AdditionPurpose {
+    case mocking,
```
CodaFi, line 60: Please document these

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,70 @@
...
+    simulateGetInputFailure ? nil : biMap[source]
+  }
+
+  @_spi(Testing) public func enumerateToSerializePriors(
+    _ eachFn: (TypedVirtualPath, DependencySource) -> Void
+  ) {
+    biMap.forEach(eachFn)
+  }
+}
+
```
CodaFi, line 56: Would very much prefer we not conflate the point of use with the names of these APIs.

davidungar:
> LGTM for the 5.5 branch.

Thank you!

davidungar:
> It is not clear to me what invariants are captured by this data structure and why this refactoring needs to go into 5.5.

I understand your preference, and the custom type does add one more term to the vocabulary here. Here's why I think that this type eases maintenance:

Because it:

Provides a focus for an overall comment,
Moves the invariants into a dedicated initializer,
Replaces subscript invocations with domain-specific, intention-revealing names such as sourceIfKnown(for:)
Provides a separate place for an extension to OutputFileMap that is specific to this data structure
I believe that it enhances the maintainability of the code by making it easier for someone to understand what's going on.
The invariants are spelled out in the comment: ```
/// Maps input files (e.g. .swift) to and from the DependencySource object.
///
// This map caches the same information as in the OutputFileMap, but it
// optimizes the reverse lookup, and includes path interning via `DependencySource`.
// Once created, it does not change.

Most of all, it lays the foundation for https://github.com/apple/swift-driver/pull/712, by isolating the locus of the change.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: `simulateGetInputFailure` is gone on the main branch. It remains here in the interest of minimal changes to this branch. If you insist, I'm happy with removing it. Whatever the folks in charge of this branch prefer.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: Good point! Will do.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: Could you please elaborate? I want to be sure I understand the point here. What would you suggest instead?
If you feel that` enumerateToSerializePriors` is too specific, then we probably disagree about the value of intention-revealing names.

If that is the source of disagreement, here is a nice quote I found that explains what I am trying to accomplish:

>If a developer must consider the implementation of a component in order to use it, the value of encapsulation is lost. If someone other than the original developer must infer the purpose of an object or operation based on its implementation, that new developer may infer a purpose that the operation or class fulfills only by chance. If that was not the intent, the code may work for the moment, but the conceptual basis of the design will have been corrupted, and the two developers will be working at cross-purposes.
>
>Therefore,
>
>Name classes and operations to describe their effect and purpose, without reference to the means by which they do what they promise. This relieves the client developer of the need to understand the internals.
- From http://ddd.fed.wiki.org/view/intention-revealing-interfaces

davidungar:
@swift-ci please test

davidungar:
@CodaFi Thanks a lot for taking a look at this. I was hoping you would verify that this refactoring introduces no functional change. From what I can tell your comments relate to: the maintenance value of the refactoring, and the choice of names.  (I have added documentation per your request.) Is this assessment of your review correct? If so, I am heartened by your assessment. Thank you again!

davidungar:
@swift-ci please test

davidungar:

CodaFi:
>  I was hoping you would verify that this refactoring introduces no functional change. 

This is sidestepping the primary issues with this patch(set). I have expressed concerns from the start about merging a refactoring patchset to the release branch without additional qualification or sufficient time to seed to users both internally and externally so we can detect issues before release. We should be shooting for *narrow* fixes to issues, as you yourself articulated upthread, and there is a much more narrow course of action here: Only using the output file map as a source of truth. I do not understand why the rest of the refactoring needs to come along for the ride.

davidungar:
> > I was hoping you would verify that this refactoring introduces no functional change.
> 
> This is sidestepping the primary issues with this patch(set). I have expressed concerns from the start about merging a large patchset including refactoring to the release branch without additional qualification or sufficient time to seed to users both internally and externally so we can detect issues before release. We should be shooting for _narrow_ fixes to issues, as you yourself articulated upthread, and there is a much more narrow course of action here: Only using the output file map as a source of truth. I do not understand why the rest of the refactoring needs to come along for the ride.

Do you mean that, although you cannot identify a functional change from the refactoring, in the interest of reducing risk for the beta, you advocate the functional change of making the `OutputFileMap` the source of truth *without* also doing the refactoring? If so, then I could agree. Although IMO, the refactoring adds value, and we have both looked at it, the greater value is getting the immutability and single-source-of-truth into the beta, ASAP.

PR #711, opened 2021-06-11
Verify usage of `-o` arguments in multi-threading mode.

artemcm:
Causing the behavior to be analagous to the Legacy driver's here:
https://github.com/apple/swift/blob/main/lib/Driver/Driver.cpp#L2415
Otherwise, this isn't a supported flow and leads to driver crashes downstream.

Also ensure that we do not perform the check that ensures the same output is not produced more than once for standard-out outputs. This allows things like `-emit-ir` to actually function with `-wmo -num-threads` and be output correctly.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -24,6 +24,7 @@ public struct Driver {
     case relativeFrontendPath(String)
     case subcommandPassedToDriver
     case integratedReplRemoved
+    case cannotSpecify_OForMultipleOutputs
     case conflictingOptions(Option, Option)
     case unableToLoadOutputFileMap(String)
     case unableToDecodeFrontendTargetInfo(String?, [String], String)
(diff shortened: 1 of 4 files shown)


artemcm:
@swift-ci please test

cltnschlosser (approved):

artemcm:
Thanks @cltnschlosser !

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -24,6 +24,7 @@ public struct Driver {
     case relativeFrontendPath(String)
     case subcommandPassedToDriver
     case integratedReplRemoved
+    case cannotSpecify_OForMultipleOutputs
```
davidungar, line 27: It might help users to list the outputs in here, and print them with the error message.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -191,7 +191,8 @@ public final class MultiJobExecutor {
                                     to producerMap: inout [VirtualPath.Handle: Int]
     ) {
       for output in job.outputs {
-        if let otherJobIndex = producerMap.updateValue(index, forKey: output.fileHandle) {
+        if output.file != .standardOutput,
```
davidungar, line 194: A comment explaining why the error is skipped for .standardOutput could help here.

davidungar: Also, the `updateValue` gets skipped. Maybe producerMap should exclude stdouts? Does it already?? Is there a comment on the var??

PR #712, opened 2021-06-11
[5.5, Incremental] Harness immutability to make inputDependencySourceMap robust

davidungar:
Partial cherry-pick of https://github.com/apple/swift-driver/pull/675

Follow-on to https://github.com/apple/swift-driver/pull/710. Suggest not reviewing this until that PR is merged and this one is rebased.
Does all of the below, *except* it does not change the format or contents of the saved prior ModuleDependencyGraph, in order to minimize risk.

> Make the inputDependencyMap immutable, solely reflecting the OutputFileMap.
Do not include it in the priors.

> In a previous PR, @CodaFi uncovered and explained issues with the `inputDependencySourceMap:

```
   // FIXME: The map between swiftdeps and swift files is absolutely *not*
   // a bijection. In particular, more than one swiftdeps file can be encountered
   // in the course of deserializing priors *and* reading the output file map
   // *and* re-reading swiftdeps files after frontends complete
   // that correspond to the same swift file. These cause two problems:
   // - overwrites in this data structure that lose data and
   // - cache misses in `getInput(for:)` that cause the incremental build to
   // turn over when e.g. entries in the output file map change. This should be
   // replaced by a multi-map from swift files to dependency sources,
   // and a regular map from dependency sources to swift files -
   // since that direction really is one-to-one.
```
> This PR solves the problem in a simpler way. The key insight is that the output file map contains the information linking a source file to the swiftdeps file for any build. Thus, the output file map should just determine the entries in the inputDependencySourceMap, just as it did before the introduction of incremental imports. Then the map in question does become a true bijection. It can also be made immutable, which greatly eases reasoning about the code.

> It will also make it easy to cull removed file nodes from priors in a future PR.


davidungar:
rdar://79211265

davidungar:
@swift-ci please test

CodaFi (changes requested):
The change in the serialization format requires a bump in the format version number. Let's just do that and drop the key instead of leaving behind an needless branch in the deserialization code.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,96 @@
...
+  @_spi(Testing) public let biMap: BiMap
+
+  private let simulateGetInputFailure: Bool
+
+  /// Based on entries in the `OutputFileMap`, create the bidirectional map to map each source file
+  /// path to- and from- the corresponding swiftdeps file path.
+  ///
+  /// - Returns: the map, or nil if error
+  init?(_ info: IncrementalCompilationState.IncrementalDependencyAndInputSetup) {
+    self.simulateGetInputFailure = info.simulateGetInputFailure
```
CodaFi, line 32: Why do we need this still? We know how to reproduce the issue in the old code, and the new code should not be subject to the same problem.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,96 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+import Foundation
+import TSCBasic
+
+@_spi(Testing) public struct InputDependencySourceMap: Equatable {
```
CodaFi, line 14: I would still prefer we not build custom data structures to solve this problem. It's unclear what invariants this structure is encapsulating and whether those are worth the maintenance burden.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @_spi(Testing) public func inputIfKnown(for source: DependencySource) -> TypedVirtualPath? {
+    simulateGetInputFailure ? nil : biMap[source]
+  }
+
+  @_spi(Testing) public func enumerateToSerializePriors(
+    _ eachFn: (TypedVirtualPath, DependencySource) -> Void
+  ) {
+    biMap.forEach(eachFn)
```
CodaFi, line 78: Would very much prefer we not conflate the point of use with the names of these APIs.

CodaFi:
Does this refactoring need to go into Beta 3? Do we have enough time to qualify the new implementation that will follow on after this PR before then?

davidungar:
@CodaFi Welcome aboard! Didn't know you'd have time, or else would have invited you up front. Happy to address your review.

Sources/SwiftDriver/IncrementalCompilation/BidirectionalMap.swift
```diff
@@ -74,6 +70,20 @@ public struct BidirectionalMap<T1: Hashable, T2: Hashable>: Equatable, Sequence
...
+    _ = oldValue.map {map2.removeValue(forKey: $0)}
+    map2[newValue] = key
+    return oldValue
+  }
+  public mutating func updateValue(_ newValue: T1, forKey key: T2) -> T1? {
+    let oldValue = map2.updateValue(newValue, forKey: key)
+   _ = oldValue.map {map1.removeValue(forKey: $0)}
+    map1[newValue] = key
+    return oldValue
+  }
```
CodaFi, line 85: This is duplicating the implementation of the subscripts above. Please define one in terms of the other.

davidungar:
> The change in the serialization format requires a bump in the format version number. Let's just do that and drop the key instead of leaving behind an needless branch in the deserialization code.

That suggestion is perfect for the "main" branch! And that is what I did, modulo the format version number. (I had thought about what would happen without a format number change, but will go back and do a PR on main.)

For for this PR, because we want to minimize risk, I decided to leave the format alone. For example, were there to be a bug that resulted in a new priors not being written when the version number changed, a user could be stuck in never-incremental-import land.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: `simulateGetInputFailure` is gone on the main branch. It remains here in the interest of minimal changes to this branch. If you insist, I'm happy with removing it. Whatever the folks in charge of this branch prefer.

CodaFi:
> For example, were there to be a bug that resulted in a new priors not being written when the version number changed

We have no evidence of such a bug though. The same qualification that this code should be put through will detect such a case anyhow. Do you have additional regression tests that you could commit to verify this doesn't happen?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: I understand your preference, and the custom type does add one more term to the vocabulary here. Here's why I think that this type *eases* maintenance: 

Because it:
1. Provides a focus for an overall comment,
2. Moves the invariants into a dedicated initializer,
3. Replaces subscript invocations with domain-specific, intention-revealing names such as `sourceIfKnown(for:)`
4. Provides a separate place for an extension to `OutputFileMap` that is specific to this data structure
I believe that it *enhances* the maintainability of the code by making it easier for someone to understand what's going on.

The invariants are spelled out in the comment: ```
  /// Maps input files (e.g. .swift) to and from the DependencySource object.
  ///
  // This map caches the same information as in the `OutputFileMap`, but it
  // optimizes the reverse lookup, and includes path interning via `DependencySource`.
  // Once created, it does not change.
```

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
davidungar: Could you please elaborate? I want to be sure I understand the point here. What would you suggest instead?
If you feel that `enumerateToSerializePriors` is too specific, then we probably disagree about the value of intention-revealing names.

If that is the source of disagreement, here is a nice quote I found that explains what I am trying to accomplish:
>If a developer must consider the implementation of a component in order to use it, the value of encapsulation is lost. If someone other than the original developer must infer the purpose of an object or operation based on its implementation, that new developer may infer a purpose that the operation or class fulfills only by chance. If that was not the intent, the code may work for the moment, but the conceptual basis of the design will have been corrupted, and the two developers will be working at cross-purposes.
>
>Therefore,
>
>Name classes and operations to describe their effect and purpose, without reference to the means by which they do what they promise. This relieves the client developer of the need to understand the internals. 
- From http://ddd.fed.wiki.org/view/intention-revealing-interfaces

davidungar:
> Does this refactoring need to go into Beta 3? Do we have enough time to qualify the new implementation that will follow on after this PR before then?

Great question! The refactoring lets me cherry-pick the immutability in a fashion that reduces risk by following what we've already got in main. It is my considered professional judgement that it would be best to put this into Beta 3, but it is not my ultimate decision. I am heartened by the fact that you have looked this over and have not (AFAICT) found any bugs or risks in it.

davidungar:
> > For example, were there to be a bug that resulted in a new priors not being written when the version number changed
> 
> We have no evidence of such a bug though. The same qualification that this code should be put through will detect such a case anyhow. Do you have additional regression tests that you could commit to verify this doesn't happen?

It is true that we have no evidence, that is why I used the subjunctive. I plan to create a PR for main, to both change the version number, as I should have before, and to add a test to ensure correct behavior when the number changes.

davidungar:
I see you have also commented in the same changes in in https://github.com/apple/swift-driver/pull/710.
I'll copy my responses over there. That's probably the right place to debate the refactorings.

CodaFi:
> I am heartened by the fact that you have looked this over and have not (AFAICT) found any bugs or risks in it.

That is manifestly not what I said, either now or in the past. The refactorings present in this patchset don't make sense to include in 5.5 unless you intend to actually make a behavior changing patch based on them, and it is that one that I am afraid of getting merged into the release branch.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
CodaFi: 1. Provides a focus for an overall comment,
2. 
Which are not included here

2. Moves the invariants into a dedicated initializer


> The invariants are spelled out in the comment:

These are not invariants, it's the purpose of the data structure... And that purpose could be fulfilled by having the output file map expose a read-only reverse view of itself. We could (and should) create it on-demand.

CodaFi: 1. Provides a focus for an overall comment,

Every decl does...

2. Moves the invariants into a dedicated initializer

You are sequestering invariants, but it's unclear *what they actually are*.

3. Replaces subscript invocations with domain-specific, intention-revealing names such as sourceIfKnown(for:)

I have articulated in the past that Swift does not usually use intention-revealing selectors because signatures and labels can communicate far more information at the point of use so you have an expanded vocabulary at your fingertips that you are not using.

4. Provides a separate place for an extension to OutputFileMap that is specific to this data structure
I believe that it enhances the maintainability of the code by making it easier for someone to understand what's going on.

Maintaining an entire other data structure for that purpose seems like a huge hammer to throw at that particular nail. 


> The invariants are spelled out in the comment:

Are you verifying them anywhere?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InputDependencySourceMap.swift
CodaFi: https://swift.org/documentation/api-design-guidelines/#naming

davidungar:
> > I am heartened by the fact that you have looked this over and have not (AFAICT) found any bugs or risks in it.
> 
> That is manifestly not what I said, either now or in the past. The refactorings present in this patchset don't make sense to include in 5.5 unless you intend to actually make a behavior changing patch based on them, and it is that one that I am afraid of getting merged into the release branch.

I am confused. This patchset *does* include a behavior-changing patch, making the `inputSourceDependencyMap` immutable.

Which of your review comments expressed that you found a bug, or a risk of a bug?

Sources/SwiftDriver/IncrementalCompilation/BidirectionalMap.swift
davidungar: Yes, good point! I will.

PR #718, opened 2021-06-18
[Incremental] When reading priors filter out nodes from removed inputs.

davidungar:
If a source file was part of the build but is no longer, it is possible to read a prior `ModuleDependencyGraph` node that corresponds to a definition in the removed file. Filter those out on deserialization in order to remove a potential source of bugs and bloat.


davidungar:
@swift-ci please test

davidungar:
Measured the additional time impact with 1,000 source files and a null build. The additional user time was 80 user ms out of a total user time of 4.24 user secs.

davidungar:

PR #719, opened 2021-06-18
Clean up testRemoval

davidungar:
Now that all cases of the removal test work, there is no need for separate tests.
Also, add some more information.


davidungar:
@swift-ci please test

artemcm (approved):

PR #720, opened 2021-06-19
Get Android to parity with upstream driver and enable remaining tests

buttaface:
This is the only remaining Android difference with the C++ driver. The new test was tested natively on Android with my Swift 5.4.1 toolchain package in [the Termux app](https://termux.com/), all tests passed.

### DIFF
--- a/Sources/SwiftDriver/Toolchains/GenericUnixToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/GenericUnixToolchain.swift
@@ -105,6 +105,7 @@ import TSCBasic
     targetTriple: Triple,
     isShared: Bool
   ) throws -> String {
-    return "libclang_rt.\(sanitizer.libraryName)-\(targetTriple.archName).a"
+    let environment = (targetTriple.environment == .android) ? "-android" : ""
+    return "libclang_rt.\(sanitizer.libraryName)-\(targetTriple.archName)\(environment).a"
   }
 }
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -1689,13 +1689,13 @@ final class SwiftDriverTests: XCTestCase {
   }
 
   func testSanitizerArgs() throws {
-  // FIXME: This doesn't work on Linux.
-  #if os(macOS)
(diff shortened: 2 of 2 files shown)


Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1808,6 +1808,31 @@ final class SwiftDriverTests: XCTestCase {
     }
     #endif
```
buttaface, line 1809: These linux tests above are disabled on linux, mistake or deliberate decision?

cltnschlosser: It looks like they are enabled, only on Linux?

buttaface: See the enclosing `#if os(macOS)` conditional compile at the top of this test.

Sources/SwiftDriver/Toolchains/GenericUnixToolchain.swift
```diff
@@ -105,6 +105,7 @@ import TSCBasic
     targetTriple: Triple,
     isShared: Bool
   ) throws -> String {
-    return "libclang_rt.\(sanitizer.libraryName)-\(targetTriple.archName).a"
+    let environment = (targetTriple.environment == .android) ? "-android" : ""
+    return "libclang_rt.\(sanitizer.libraryName)-\(targetTriple.archName)\(environment).a"
```
buttaface, line 109: This is a port of [my earlier patch for the C++ driver](https://github.com/apple/swift/commit/14cc62001670032c2ba4fc3acde6af37910f9edd#diff-55fd9483c4660c88c6c3e942cdc38a2cae2647839c581e322c7bf292775e5c43R46).

Tests/SwiftDriverTests/SwiftDriverTests.swift
cltnschlosser: Oh I see. Hmm :/

cltnschlosser (approved):

artemcm (approved):

buttaface:
I think this just needs to run through the CI and it can be merged?

DougGregor:
@swift-ci please test

PR #721, opened 2021-06-20
Add an integration with swift-api-digester

owenv:
The goal here is to add support to the driver for generating an API/ABI baseline or comparing to one immediately after emitting a module. This means higher-level build systems like SwiftPM can share the implementation and each won't need to have their own support for generating swift-api-digester command lines to add breaking-change detection.

A few notes on the implementation:
- Right now, all the flags related to this functionality have separate API and ABI variants. The makes it possible to generate or compare both types of baseline in a single driver invocation, but I'm not entirely sure it's worth the trouble
- The baselines are treated as 'auxilary' outputs, similar to swiftsourceinfo files, so if an output path isn't specified they'll look for a Project/ directory if one is present. If this doesn't make sense, we could make them regular supplementary outputs.
- Right now, these features only work if `-emit-module` is used to emit a top-level module. I implemented it this way because if the module is considered 'auxilary', the swiftmodule has a temporary filename and swift-api-digester won't be able to load it. This limitation could probably be lifted in the future.
- I'm using .api.json and .abi.json as the baseline file types, but only because I couldn't think of anything better.
- Currently this depends on `https://github.com/apple/swift/pull/37991`, and uses the new -disable-fail-on-error option so that when a breaking change is detected we don't mark the build as failed. I think this is the usually the right behavior, but we could make it configurable.

The new flags are:
- `-emit-(api|abi)-baseline`, `-emit-(api|abi)-baseline-path`: emit a baseline JSON file for the module being emitted
- `-compare-to-(api|abi)-baseline-path`: compare the API/ABI of the module being emitted to the baseline at <path>
- `-serialize-(api|abi)-breaking-changes-path`: serialize the errors emitted by swift-api-digester instead of printing them to stdout
- `-(api|abi)-breakage-allowlist-path`: specify a list of breaking changes which should be ignored

This is just my first pass at the feature, so any and all feedback is welcome. Thanks to @nkcsgexi for the idea!

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -34,6 +34,8 @@ public struct Driver {
     case missingProfilingData(String)
     case conditionalCompilationFlagHasRedundantPrefix(String)
     case conditionalCompilationFlagIsNotValidIdentifier(String)
+    case baselineGenerationRequiresTopLevelModule(String)
+    case optionRequiresAnother(String, String)
     // Explicit Module Build Failures
     case malformedModuleDependency(String, String)
     case missingPCMArguments(String)
@@ -41,7 +43,6 @@ public struct Driver {
     case missingContextHashOnSwiftDependency(String)
     case dependencyScanningFailure(Int, String)
     case missingExternalDependency(String)
-    case baselineGenerationRequiresTopLevelModule(String)
 
     public var description: String {
       switch self {
@@ -103,6 +104,8 @@ public struct Driver {
         return "Missing External dependency info for module: \(moduleName)"
       case .baselineGenerationRequiresTopLevelModule(let arg):
         return "generating a baseline with '\(arg)' is only supported with '-emit-module' or '-emit-module-path'"
+      case .optionRequiresAnother(let first, let second):
+        return "'\(first)' cannot be specified if '\(second)' is not present"
       }
     }
   }
@@ -291,11 +294,11 @@ public struct Driver {
   /// Path to the Swift module source information file.
   let moduleSourceInfoPath: VirtualPath.Handle?
 
-  /// Path to the module API baseline file.
-  let apiBaselinePath: VirtualPath.Handle?
+  /// Path to the module's digester baseline file.
+  let digesterBaselinePath: VirtualPath.Handle?
 
-  /// Path to the module ABI baseline file.
-  let abiBaselinePath: VirtualPath.Handle?
+  /// The mode the API digester should run in.
+  let digesterMode: DigesterMode
 
   /// Force the driver to emit the module first and then run compile jobs. This could be used to unblock
   /// dependencies in parallel builds.
@@ -544,6 +547,16 @@ public struct Driver {
     self.numThreads = Self.determineNumThreads(&parsedOptions, compilerMode: compilerMode, diagnosticsEngine: diagnosticEngine)
     self.numParallelJobs = Self.determineNumParallelJobs(&parsedOptions, diagnosticsEngine: diagnosticEngine, env: env)
 
+    var mode = DigesterMode.api
+    if let modeArg = parsedOptions.getLastArgument(.digesterMode)?.asSingle {
+      if let digesterMode = DigesterMode(rawValue: modeArg) {
+        mode = digesterMode
+      } else {
+        diagnosticsEngine.emit(Error.invalidArgumentValue(Option.digesterMode.spelling, modeArg))
+      }
+    }
+    self.digesterMode = mode
+
     Self.validateWarningControlArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
     Self.validateProfilingArgs(&parsedOptions,
                                fileSystem: fileSystem,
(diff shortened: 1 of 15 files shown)


Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -462,6 +463,24 @@ extension Driver {
     }
   }
 
+  private mutating func addAPIDigesterJobs(addJob: (Job) -> Void) throws {
```
nkcsgexi, line 466: ABI checking is only relevant when `-enable-library-evolution` and `-emit-module-interface` are specified. Could you check these conditions before proceeding and diagnose as well?

Sources/SwiftDriver/Jobs/APIDigesterJobs.swift
```diff
@@ -0,0 +1,152 @@
...
+
+    commandLine.appendFlag(.o)
+    commandLine.appendPath(VirtualPath.lookup(outputPath))
+
+    return Job(
+      moduleName: moduleOutputInfo.name,
+      kind: mode.baselineGenerationJobKind,
+      tool: .absolute(try toolchain.getToolPath(.swiftAPIDigester)),
+      commandLine: commandLine,
+      inputs: [.init(file: modulePath, type: .swiftModule)],
```
nkcsgexi, line 61: For `-abi`, we should also require the interface file as the input. Please also add digester flag `-use-interface-for-module CurrentMainModule` when generating ABI baselines.

nkcsgexi (commented):
This is amazing! @owenv Some comments are inlined.

Sources/SwiftOptions/ExtraOptions.swift
```diff
@@ -18,6 +18,18 @@ extension Option {
   public static let noEmitModuleSeparately: Option = Option("-no-emit-module-separately", .flag, attributes: [.helpHidden], helpText: "Force using merge-module as the incremental build mode")
   public static let useFrontendParseableOutput: Option = Option("-use-frontend-parseable-output", .flag, attributes: [.helpHidden], helpText: "Emit parseable-output from swift-frontend jobs instead of from the driver")
 
+  // API digester operations
+  public static let emitAPIBaseline: Option = Option("-emit-api-baseline", .flag, attributes: [.noInteractive, .supplementaryOutput], helpText: "Emit an API baseline file for the module")
```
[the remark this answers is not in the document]
owenv: Ok, that sounds good to me! Instead of `-abi`, right now I'm using `-digester-mode abi` or `-digester-mode api` (with the default being `api` if it's unspecified), which I think makes it a little clearer what the impact of the setting is.

Sources/SwiftDriver/Jobs/Planning.swift
owenv: 👍 makes sense, done!

Sources/SwiftDriver/Jobs/APIDigesterJobs.swift
owenv: 👍 makes sense, done!

owenv:
@nkcsgexi thanks for the feedback! I hadn't spent as much time thinking about the ABI-checking use case, so this is really helpful.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2650,6 +2717,31 @@ extension Driver {
...
+    mode: DigesterMode,
+    compilerOutputType: FileType?,
+    compilerMode: CompilerMode,
+    outputFileMap: OutputFileMap?,
+    moduleName: String,
+    projectDirectory: VirtualPath.Handle?
+  ) throws -> VirtualPath.Handle? {
+    // Only emit a baseline if at least of the arguments was provided.
+    guard parsedOptions.hasArgument(.emitDigesterBaseline, .emitDigesterBaselinePath) else { return nil }
+    return try computeModuleAuxiliaryOutputPath(&parsedOptions,
```
nkcsgexi, line 2732: Thanks for fixing the previous round of comments! Here will give us a location next to `.swiftmodule` file. Can we change it so that we emit next to `.swiftsourceinfo` file instead? `.swiftsourceinfo` is in the `Project` directory if you use XCBuild, which is a sub-directory reserved for local artifacts.

owenv: I think this is already using the same logic as `.swiftsourceinfo`, they both call `computeModuleAuxiliaryOutputPath` and should use the `Project` folder if it's present. The only difference is that source info is opt-out and baseline generation is opt-in. I have a couple basic tests for this in `testBaselineOutputPath` which roughly correspond to the ones in `testSourceInfoFileEmitOption`. It's possible I made a mistake somewhere though.

owenv: Oh I see, that makes sense! I added the new rule and a couple tests involving output file maps

nkcsgexi (approved):
This looks great! @owenv

nkcsgexi:
@swift-ci please test

owenv:
Thanks! Once I merge https://github.com/apple/swift/pull/37991 and it makes it into a nightly toolchain the tests here should pass

nkcsgexi:
Sounds good!

owenv:
@swift-ci please test

owenv:
The tests which run jobs are now checking for flag availability so they won't fail when run with an older toolchain. I verified in https://github.com/apple/swift/pull/38251 that they run & pass with new toolchains, so this is good to go.

PR #724, opened 2021-06-22
Refactor InputDependencySourceMap to Remove a Dependency on BidirectionalMap

CodaFi:
This was the last user of this custom data structure. The forward mapping is maintained in the output file map itself. Maintaining the bidirectional map cost us an extra copy of that information, which is O(inputs)-many duplicated entries. Instead, construct the mapping directly from the output file map itself, and keep an immutable copy around for the forward mapping. To better drive home that point, rename `InputDependencySource` to `OutputFileMap.DependencyView` as it is now an immutable view of the swift-to-swiftdeps mapping for dependencies.


CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

davidungar (changes requested):
Thank you very much for requesting my review on this matter. Given that some of these issues represent long-standing disagreements we have had since you reviewed https://github.com/apple/swift-driver/pull/664, I am especially honored that you sought my review.  This PR does raise some good points, but it seems that we remain in disagreement concerning the best ways to structure the code and the concepts it represents.

** Details: **

1. The issue of whether an extra copy is maintained is orthogonal to this reorganization. The same approach of creating a `DependencySource` on demand could be adopted in the status quo. 

2. Although the status quo currently does copy, it then saves the cost of manufacturing a new `DependencySource` object each time the same source is requested. If many "forward" lookups are done, the copying approach might be cheaper. After a brief inspection of the code it does appear that there could be multiple lookups of the same input. For example, if a change to a source file causes the same input to be invalidated multiple times, then it is possible that the copying could be cheaper. Thus, saving the extra copy could actually *save* time in some cases. 

3. It is true that `InputDependencySourceMap` is a bit of a mouthful, and is more *what* than *why.* But, `DependencyView` fails to communicate the intent, which is to find swiftdeps files from swift files and vice-versa.  `DependencyView` is not a view on dependencies, i.e. arcs from declarations to uses of them. Whether this PR ends up merging or not, we ought to find a better name than either for this type. How about `SourceAndDependencyFileFinder`?

4. This PR nests the new type inside of the `OutputFileMap` because the information in the new type is derived from some of the information in the `OutputFIleMap`. However, there are several drawbacks to this approach:
 - The `OutputFileMap` type becomes less clear because now the struct moves further away from just representing the deserialized JSON of the output file map file.
 - Eliminating the dedicated file `InputDependencySourceMap.swift`, located in the folder `ModuleDependencyGraph parts` in the status quo, removes a place to put code that supports the mapping, and weakens the link to the `ModuleDependencyGraph`. Whatever the type is named, be it `DependencyView` or `InputDependencySourceMap` or `SourceAndDependencyFileFinder`, it exists solely in service of the `ModuleDependencyGraph`, whereas the `OutputFileMap` has other uses, including forming compile jobs, etc.
 
5. A necessary invariant, which is checked by `BidirectionalMap` in the status quo, is that if a mapping is created from a -> b, then b must map back to a. At the `InputDependencySourceMap` level, this means that if a.swift maps to x.swiftdeps, then x.swiftdeps will map back to a.swift. This PR loses that test.

6. This PR renames `sourceIfKnown(for:)` with `source(for:)`, and you have previous expressed a preference for the latter name. In this case, since there is also a `sourceRequired(for:)` which does *almost* the same thing, `sourceIfKnown(for:)` makes the distinction clearer to an unfamiliar reader.  If `sourceRequired(for:)` did not exist, I would agree with your renaming, but it does.

**Summary**

This PR raises meritorious issues:
- `InputDependencySourceMap` is a poor name, however `DependencyView` seems even more obscure. I would suggest `SourceAndDependencyFileFinder`, but lets discuss it.
- There is a cost to copy each `VirtualPath.Handle` for each `swiftdeps` file. However, this may be offset by the cost to recreate a `DependencySource` object for each such lookup.
By contrast, this PR makes changes that will significantly weaken its comprehensibility:
- The `ModuleDependencyGraph` relies on a mapping from .swift to and from .swiftdeps files, yet this PR makes that reliance less clear by moving the new type into the `OutputFileMap`. The move also dilutes the focus of the `OutputFileMap` type.
- By dropping the `BidirectionalMap` this PR sacrifices the reification of the required bidirectionality, as well as losing a test.  If we do want to avoid the copying cost, I would suggest replacing `BidirectionalMap` with `BidirectionalView` in order to maintain the clarity of the bidirectionality and also to provide a locus in which to test it.
Finally, this PR makes a change that will somewhat weaken its comprehensibilty:
- By dropping the `IfKnown` from  `sourceIfKnown(for:)` and `inputIfKnown(for)`, a potential for confusion and misuse is increased wrt `sourceRequired(for:)`.

**Changes to this PR that would make it acceptable to me**
1. Keep the new type in its own file, in the `ModuleDependencyGraph` parts folder.
2. Rename `DependencySource` to something more descriptive, such as `SourceAndDependencyFileFinder`.
3. Maintain the reification in the status quo of the bidirectional requirement and keep a test for it, perhaps with a `BidirectionalView` replacing the current `BidirectionalMap`. The copying cost could still be avoided.

**Changes I would like but will not insist upon**
1. Retaining the "IfKnown" in the appropriate functions

Thank you again for your thoughtful consideration. I regret that you and I favor different tradeoffs between brevity and clarity.

CodaFi:
> The issue of whether an extra copy is maintained is orthogonal to this reorganization. The same approach of creating a DependencySource on demand could be adopted in the status quo.

But it is extremely relevant to performance tuning. I was able to wipe 50 MB of peak heap usage off of SwiftUI's build with this branch relative to ToT. In general, we must remember that anything that copies or iterates over the input source files has the potential to be thousands, if not tens of thousands, of files worth of work.

> Although the status quo currently does copy, it then saves the cost of manufacturing a new DependencySource object each time the same source is requested...

Copying a dependency source costs copying a `TypedVirtualPath` which costs copying a `VirtualPath.Handle` and a `FileType` which are both trivial types. This type should be POD.


> It is true that InputDependencySourceMap...

It does not communicate the intent in quite the same way, it is in fact an immutable view of this data structure, in the same vein as our [String.UTF{8,16,32}View](https://developer.apple.com/documentation/swift/string/utf16view). I feel the name is quite appropriate.

> A necessary invariant, which is checked by BidirectionalMap in the status quo, is that if a mapping is created from a -> b, then b must map back to a. 

Does that invariant actually *hold* yet? Because if it doesn't, what the data structure does is centralize a *bug* where we encounter data loss when duplicated pairs of inputs are written into the map from external sources. Moreover, that invariant is enforced on construction by `makeDependencyView()` so.. I think we're good here.

CodaFi:
> yet this PR makes that reliance less clear by moving the new type into the OutputFileMap.

The type is related to the output file map, should it not appear near it? I can break it into a file if you wish.

> The move also dilutes the focus of the OutputFileMap type.

How so? I have added one operation that uses its data to project yet more data. The link between the two is quite sharp to me.

>. By dropping the BidirectionalMap this PR sacrifices the reification of the required bidirectionality, as well as losing a test.

For an entire data structure, there were a family of two tests, neither of which were sufficient to catch the issues I encountered with it in f8f3c67. If we are to keep this structure, we would need exponentially more tests covering the entire structure before I would feel comfortable keeping it in the codebase.

> By dropping the IfKnown from sourceIfKnown(for:) and inputIfKnown(for), a potential for confusion and misuse is increased wrt sourceRequired(for:).

https://swift.org/documentation/api-design-guidelines/#naming

davidungar:
The points about invariants are good--I'll be adding some assertions to enforce them.

davidungar:
Also, I'll be factoring the implementation of `BidirectionalMap` better, per a suggestion of yours in another PR.

davidungar:
OK, after working on the refactoring of `BidirectionalMap`, I do see that it provides more functionality than is used at present. For instance, it supports removal. I'm thinking that the happy medium could be to remove the unused functionality, while still reifying the bidirectionality. I'll work on such a PR, for your consideration.

CodaFi:
I don't really want a happy medium. I want us to arrive at a much better place, and right now I'm not convinced that that future involves any of the custom data structures we've built up here. The maintenance cost and the bugs we've seen so far have been far too high to justify their continued existence. We should leave this to the experts and focus on improving the performance and stability of the driver in other ways.

davidungar:
> I don't really want a happy medium. I want us to arrive at a much better place, and right now I'm not convinced that that future involves any of the custom data structures we've built up here. The maintenance cost and the bugs we've seen so far have been far too high to justify their continued existence. We should leave this to the experts and focus on improving the performance and stability of the driver in other ways.

I understand, and still hope that we can find something that we both like. We both make some good points.

PR #725, opened 2021-06-22
Disable explicit module builds test

CodaFi:
It's failing in CI for some as-yet-undetermined reason.

rdar://79594631

PR #727, opened 2021-06-22
Revert "Disable explicit module builds test #725" and fix the test.

artemcm:
Recent changes that introduced `-clang-target` have introduced a conflict with how explicit module builds work. We need to address this, but in the meantime, re-enable the test with `-disable-clang-target`.

Resolves rdar://79594631


artemcm:
@swift-ci please test

nkcsgexi (approved):
Thanks!

PR #731, opened 2021-06-25
Do not emit a dependency-scanner fallback warning when in integrated-driver mode

artemcm:
This warning is mostly useful for debugging the driver, so let's hide it when `libSwiftDriver` is used, instead of a `swift-driver `executable.
There are also valid scenarios where the scanning action can succeed just-as-well using the fallback path where this warning is extraneous and may confuse users because it is not actionable.


nkcsgexi (approved):

DougGregor:
@swift-ci please test

PR #733, opened 2021-06-28
Do not set `DYLD_LIBRARY_PATH` to point the installed toolchain/sdk on macOS.

artemcm:
This has not been a good idea since Swift started shipping in the OS. 
The OS runtime is the one we would like Swift scripts to use. Using the installed toolchain can potentially cause sync issues.

Continue setting this variable in other UNIX toolchains, because there is no runtime shipped in the OS there. 

Resolves rdar://78951185

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -11,7 +11,7 @@
         }
       },
       {
-        "package": "swift-llbuild",
+        "package": "llbuild",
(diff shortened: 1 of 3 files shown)


artemcm:
@swift-ci please test

nkcsgexi (approved):
Seems reasonable to me.

mikeash (approved):

PR #734, opened 2021-06-29
Improve Deserialization Performance

CodaFi:
Consume the new non-owning bitstream deserialization in apple/swift-tools-support-core#236 to provide a 3-5x boost to deserialization performance for source and module dependency graph files. Also, undo a cached hash optimization that was only really necessary because of the quadratic COW issue resolved in #654. Otherwise, the hash was being eagerly computed during deserialization which forced an extra half-second of wall time.


artemcm (approved):
All of these changes look great to me.

davidungar (approved):
It would be nice to add a comment somewhere about the strategy for handling fingerprint strings, which in this PR is designed for performance. Either here, or in a subsequent PR.
But that said, LGTM!

CodaFi:
apple/swift-tools-support-core#236

@swift-ci test

CodaFi:
⛵

PR #737, opened 2021-06-29
[SIL] Deprecating the '-gsil' frontend flag

mshockwave:
The `-gsil` is deprecated in favor of `-sil-based-debuginfo`.

This PR is a synergy with https://github.com/apple/swift/pull/38167.

### DIFF
--- a/Sources/SwiftOptions/Options.swift
+++ b/Sources/SwiftOptions/Options.swift
@@ -336,7 +336,7 @@ extension Option {
   public static let glineTablesOnly: Option = Option("-gline-tables-only", .flag, attributes: [.frontend], helpText: "Emit minimal debug info for backtraces only", group: .g)
   public static let gnone: Option = Option("-gnone", .flag, attributes: [.frontend], helpText: "Don't emit debug info", group: .g)
   public static let groupInfoPath: Option = Option("-group-info-path", .separate, attributes: [.helpHidden, .frontend, .noDriver], helpText: "The path to collect the group information of the compiled module")
-  public static let debugOnSil: Option = Option("-gsil", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Write the SIL into a file and generate debug-info to debug on SIL  level.")
+  public static let legacyGsil: Option = Option("-gsil", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Deprecated, use '-sil-based-debuginfo' instead")
   public static let g: Option = Option("-g", .flag, attributes: [.frontend], helpText: "Emit debug info. This is the preferred setting for debugging with LLDB.", group: .g)
   public static let helpHidden: Option = Option("-help-hidden", .flag, attributes: [.helpHidden, .frontend], helpText: "Display available options, including hidden options")
   public static let helpHidden_: Option = Option("--help-hidden", .flag, alias: Option.helpHidden, attributes: [.helpHidden, .frontend], helpText: "Display available options, including hidden options")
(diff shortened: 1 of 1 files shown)


mshockwave:
@swift-ci test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -240,6 +240,11 @@ extension Driver {
       if shouldSupportAnonymousContextMangledNames {
         commandLine.appendFlag(.enableAnonymousContextMangledNames)
       }
+
+      if compilerOutputType == .sil {
+        commandLine.appendFlag("-Xllvm")
+        commandLine.appendFlag("-sil-print-debuginfo")
```
nkcsgexi, line 246: You should be able to use `.Xllvm` and `.debugOnSil` here.

Sources/SwiftOptions/Options.swift
```diff
@@ -336,7 +336,7 @@ extension Option {
   public static let glineTablesOnly: Option = Option("-gline-tables-only", .flag, attributes: [.frontend], helpText: "Emit minimal debug info for backtraces only", group: .g)
   public static let gnone: Option = Option("-gnone", .flag, attributes: [.frontend], helpText: "Don't emit debug info", group: .g)
   public static let groupInfoPath: Option = Option("-group-info-path", .separate, attributes: [.helpHidden, .frontend, .noDriver], helpText: "The path to collect the group information of the compiled module")
-  public static let debugOnSil: Option = Option("-gsil", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Write the SIL into a file and generate debug-info to debug on SIL  level.")
+  public static let debugOnSil: Option = Option("-sil-based-debuginfo", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Write the SIL into a file and generate debug-info to debug on SIL  level.")
```
nkcsgexi, line 339: We usually don't rename an argument unless it's really appealing, so that driver is always in compatible with the compiler.

nkcsgexi (changes requested):

Sources/SwiftOptions/Options.swift
adrian-prantl: In this case, I think we do want to rename the option. The problem is that `-gsil` invades in the `-g` option namespace and thus can't be combined with the other `-g{0,line-tables-only,dwarf-types,{}}` options. The option is used only by a handful of Swift compiler developers.

adrian-prantl (approved):
LGTM form my side, but please wait for @nkcsgexi to also approve the PR.

adrian-prantl:
We may have an exhaustive list of the existing `-gsil` userbase in the list of Reviewers now 😀

Sources/SwiftOptions/Options.swift
nkcsgexi: ah, OK. Thanks for the explanation! Another question: is `Options.swift` updated using the steps mentioned [here](https://github.com/apple/swift-driver#rebuilding-optionsswift)?

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
mshockwave: I'm changing "-Xllvm" to `.Xllvm` but "-sil-print-debuginfo" doesn't have any corresponding enum (`.debugOnSil` is mapped to "-sil-based-debuginfo").

atrick (approved):
Sweet!

Sources/SwiftOptions/Options.swift
mshockwave: > ah, OK. Thanks for the explanation! Another question: is Options.swift updated using the steps mentioned here?
Currently not, I'm going to update it using that method (and update the PR here is there is any difference). Thanks for the pointer

adrian-prantl:
@mshockwave For the odd chance that there are other users of `-gsil`: Since we don't use `-gsil` for anything else — should we define it and print a "this option is deprecated" error or alias it to the new option?

mshockwave:
> @mshockwave For the odd chance that there are other users of `-gsil`: Since we don't use `-gsil` for anything else — should we define it and print a "this option is deprecated" error or alias it to the new option?

Yes I can do that

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
atrick: I'm curious, why this is in the driver and not the frontend? Don't we want `swift-frontend -g -emit-sil` to do the right thing? I generally expect the driver and frontend flags to have the same meaning.

adrian-prantl: I think I asked a similar question in https://github.com/apple/swift-driver/pull/737 — I think it would make sense to update the frontend flags in a similar manner (and a separate patch).

atrick: That's great if we add -g `.sil` handling for the frontend in a separate patch. I'm just not sure why we need to add a `-Xllvm` flag to the frontend invocation if the frontend already does the right thing.

mshockwave:
> > @mshockwave For the odd chance that there are other users of `-gsil`: Since we don't use `-gsil` for anything else — should we define it and print a "this option is deprecated" error or alias it to the new option?
> 
> Yes I can do that

Please refer to https://github.com/apple/swift/pull/38167#issuecomment-871006641

@swift-ci test

nkcsgexi (approved):
LGTM!

nkcsgexi:
The patch looks great to me! For future reference, we don't have to update the option list if the driver code doesn't use any new options in the implementation. The option list will be eventually updated by later changes.

PR #740, opened 2021-07-03
[Incremental] When reading priors, don't use nodes for removed inputs.

davidungar:
If priors contain nodes from input files that have been removed, do not add those nodes to the `ModuleDependencyGraph` when deserializing priors.

I have measured driver time on a 1000+ source-file project, and the extra time for this check was about 1 ms according to instruments, out of 2.8 secs spent reading the priors. Total driver time was 26 secs.


davidungar:
@swift-ci please test

artemcm (approved):
Nice cleanup.

artemcm:
@swift-ci please test

davidungar:
Thank you, @artemcm

PR #741, opened 2021-07-03
[Incremental] Compare modification time of priors to build times in build record

davidungar:
Ensure that saved priors actually comes from the prior build. Guards against reading stale priors.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -587,6 +587,7 @@ extension ModuleDependencyGraph {
     case bogusNameOrContext
     case unknownKind
     case unknownDependencySourceExtension
+    case timeTravellingPriors(Date, ClosedRange<Date>)
   }
 
   /// Attempts to read a serialized dependency graph from the given path.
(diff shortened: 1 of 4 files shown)


davidungar:
@swift-ci please test

CodaFi:
I'm a bit unclear on why stale priors are an issue. Seems like the build record's mod time check covers us here.

davidungar:
> I'm a bit unclear on why stale priors are an issue. Seems like the build record's mod time check covers us here.

In the normal course of events, you are right, a stale priors file should not exist, as far as I can tell (though I'm not exactly sure what you mean about the build record's mod time check). However, in the case of unforeseen circumstances, stale priors might be possible. For the cost of one stat on the priors file, we get to cover that case, and prevent potential future problems. Ideally, as you've proposed before, the priors would be merged into the build record. Until that happens, I'd like to add this additional check.

davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -775,6 +775,36 @@ extension ModuleDependencyGraph {
...
+    info: IncrementalCompilationState.IncrementalDependencyAndInputSetup
+  ) throws -> ByteString? {
+    guard try info.fileSystem.exists(path) else {
+      return nil
+    }
+    try ensurePriorsCreatedDuringPriorBuild(at: path, info: info)
+    return try info.fileSystem.readFileContents(path)
+  }
+
+  fileprivate static func ensurePriorsCreatedDuringPriorBuild(
```
artemcm, line 796: Could you please add a comment here that explains the rationale and maybe gives an example of what could have gone wrong for this check to save us?

artemcm (approved):
This change looks good to me. 
Even if such inconsistencies should be rare and hard to hit, this check is inexpensive so it's nice to have.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: Thank you for looking at this. Will do, great suggestion!

davidungar:

davidungar:
@swift-ci please test

PR #742, opened 2021-07-04
[Draft] Clean build: Only integrate module dependencies once

cltnschlosser:
Not sure if this change is safe, or if it needs to go further to impact incremental builds as well.

Currently this is shaving close to 2.5 minutes off of a clean build for a project with >150 modules and ~350 swift files in the main target.

Issue before was this was reading and integrating the module dependency information (M) for each input swift file (N), so (N*M), now it's only happening M times.

It needs more testing, but there is probably an incremental build situation where dependency information is being read and integrated N * M times where M is modules that have changed since last build.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -376,9 +376,16 @@ extension ModuleDependencyGraph {
   private func invalidationReason(
     for fed: ExternalIntegrand
   ) -> ExternalDependency.InvalidationReason? {
-    let isNewToTheGraph = !fed.isKnown && fingerprintedExternalDependencies.insert(fed.externalDependency).inserted
-    if self.phase.shouldNewExternalDependenciesTriggerInvalidation && isNewToTheGraph {
-      return .added
+    guard fed.isKnown else {
+      if fingerprintedExternalDependencies.insert(fed.externalDependency).inserted {
+        return .added
+      } else {
+        // If a file is unknown, but previously added, then this is probably a clean build, which wouldn't have a build record.
+        // This will result in dependencies being processed for each input (because it will be considered always changed).
+        // We can skip, because a previous input has already integrated this dependency,
+        // which is why it's in `fingerprintedExternalDependencies`.
+        return nil
+      }
     }
 
     if self.hasFileChanged(fed.externalDependency.externalDependency) {
--- a/Package.resolved
+++ b/Package.resolved
@@ -11,7 +11,7 @@
         }
       },
       {
-        "package": "llbuild",
+        "package": "swift-llbuild",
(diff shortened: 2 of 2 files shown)


CodaFi (approved):
Makes sense to me. Thanks!

CodaFi:
@swift-ci test

davidungar:
Please don't merge this yet. I'm in the midst of analysis and I suspect it may be introducing a bug when incremental imports is disabled. Hope to know for sure very soon!

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -376,9 +376,16 @@ extension ModuleDependencyGraph {
...
-      return .added
+    guard fed.isKnown else {
+      if fingerprintedExternalDependencies.insert(fed.externalDependency).inserted {
+        return .added
+      } else {
+        // If a file is unknown, but previously added, then this is probably a clean build, which wouldn't have a build record.
+        // This will result in dependencies being processed for each input (because it will be considered always changed).
+        // We can skip, because a previous input has already integrated this dependency,
+        // which is why it's in `fingerprintedExternalDependencies`.
+        return nil
```
davidungar, line 387: I think there may be a problem here: What if the external dependency has changed? It would need to be reintegrated. If the external dependency was in a priors, `ifKnown` will be true, so you don't get here. *But*, what if an external dependency in the priors has changed, it is integrated, and *it* includes a changed external dependency that is already in `fingerprintedExternalDependencies`? `isKnown` will not be true and I think you get into trouble here.

I believe there is a problem: `isKnown` really means "is guaranteed to be in  `fingerprintedExternalDependencies` because it was found directly in the priors. But `!isKnown` only means it was found in a `swiftdeps` or `swiftmodule` file.

I have an idea for what I hope will be a much better fix. Let me see if it looks good. I should have the PR posted in a day.

davidungar (commented):
See my comments. If I'm right, this fix is unsound, but it's tricky!

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
cltnschlosser: Yeah, I’m thinking maybe just a separate set that we can use to guarantee that dependencies are only read once per driver invocation.

Alternatively we could store mod time with the dependencies we know about, so we could compare that, instead of comparing to build start time.

Does the fingerprint change when the file contents change? Could that be used? Or is that something else?

davidungar: Please see https://github.com/apple/swift-driver/pull/750 , which is my proposal to fix this issue. I would love to get your thoughts on it.

cltnschlosser:
Closing in favor of #750

PR #743, opened 2021-07-05
[Build Script Helper] Add support for non-Darwin cross-compilation hosts, starting with Android

buttaface:
Add a flag, `--cross-compile-config`, to pass in a SPM cross-compilation file and cross-compile with that instead of CMake. Also, fix non-Darwin installs to use the `--prefix` list passed in, rather than simply installing into the build toolchain.

I tested this with the latest trunk source snapshot from June 12. The `build-script` side of this pull was just added to apple/swift#36917. This pull is the equivalent of apple/swift-package-manager#3403. I plan to use this pull to build and install this swift-driver as part of [my upcoming Swift 5.5 toolchain for Android](https://github.com/buttaface/termux-packages/tree/swift55/packages/swift).

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -11,6 +11,7 @@ import shutil
 import subprocess
 import sys
 import errno
+import re
 
 if platform.system() == 'Darwin':
     shared_lib_ext = '.dylib'
@@ -96,7 +97,11 @@ def get_swiftpm_options(args):
       os.path.join(args.toolchain, 'lib', 'swift', 'Block'),
     ]
 
-    if 'ANDROID_DATA' in os.environ:
+    if args.cross_compile_hosts:
+      swiftpm_args += ['--destination', args.cross_compile_config]
+
+    if 'ANDROID_DATA' in os.environ or (args.cross_compile_hosts and re.match(
+        'android-', args.cross_compile_hosts[0])):
       swiftpm_args += [
         '-Xlinker', '-rpath', '-Xlinker', '$ORIGIN/../lib/swift/android',
         # SwiftPM will otherwise try to compile against GNU strerror_r on
@@ -180,8 +185,13 @@ def handle_invocation(args):
   if args.sysroot:
     env['SDKROOT'] = args.sysroot
 
+  env['SWIFT_EXEC'] = '%sc' % (swift_exec)
+
   if args.action == 'build':
-    build_using_cmake(args, toolchain_bin, args.build_path, targets)
+    if args.cross_compile_hosts and not re.match('-macosx', args.cross_compile_hosts[0]):
+      swiftpm('build', swift_exec, swiftpm_args, env)
+    else:
+      build_using_cmake(args, toolchain_bin, args.build_path, targets)
 
   elif args.action == 'clean':
     print('Cleaning ' + args.build_path)
@@ -191,7 +201,6 @@ def handle_invocation(args):
         tool_path = os.path.join(toolchain_bin, tool)
         if os.path.exists(tool_path):
             env['SWIFT_DRIVER_' + tool.upper().replace('-','_') + '_EXEC'] = '%s' % (tool_path)
-    env['SWIFT_EXEC'] = '%sc' % (swift_exec)
     test_args = swiftpm_args
     test_args += ['-Xswiftc', '-enable-testing']
     if should_test_parallel():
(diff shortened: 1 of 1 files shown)


Utilities/build-script-helper.py
```diff
@@ -626,6 +640,8 @@ def add_common_args(parser):
       args.cross_compile_hosts = [build_target + macos_deployment_target, 'arm64-apple-macosx%s' % macos_deployment_target]
   elif (build_target == 'arm64-apple-macosx' and 'macosx-x86_64' in args.cross_compile_hosts):
       args.cross_compile_hosts = [build_target + macos_deployment_target, 'x86_64-apple-macosx%s' % macos_deployment_target]
+  elif args.cross_compile_hosts and re.match('android-', args.cross_compile_hosts[0]):
```
buttaface, line 643: Gates this new cross-compilation path for Android only, since it has only been tested for that non-Darwin platform.

Utilities/build-script-helper.py
```diff
@@ -180,8 +185,13 @@ def handle_invocation(args):
   if args.sysroot:
     env['SDKROOT'] = args.sysroot
 
+  env['SWIFT_EXEC'] = '%sc' % (swift_exec)
```
buttaface, line 188: The only change in this pull that will affect native compilation too, by applying this environment variable when invoking `swift-build` when installing on non-Darwin platforms. I had to add this when cross-compiling for Android with Swift 5.5 or trunk on a linux host with a Swift 5.4 package installed, as SPM would then try to use that system Swift 5.4 and error because of the Swift stdlib version mismatch.

This fixes that bug in the native build too.

Utilities/build-script-helper.py
```diff
@@ -180,8 +185,13 @@ def handle_invocation(args):
...
+
   if args.action == 'build':
-    build_using_cmake(args, toolchain_bin, args.build_path, targets)
+    if args.cross_compile_hosts:
+      swiftpm('build', swift_exec, swiftpm_args, env)
```
buttaface, line 192: Since SPM is used for the install step below anyway, I see no reason to cross-compile with CMake first, which would require passing in a separate list of Swift flags to cross-compile with CMake.

Utilities/build-script-helper.py
```diff
@@ -208,16 +217,17 @@ def handle_invocation(args):
...
     assert False, 'unknown action \'{}\''.format(args.action)
 
 # Installation flow for non-darwin platforms, only copies over swift-driver and swift-help
 # TODO: Unify CMake-based installation flow used on Darwin with this
-def non_darwin_install(swiftpm_bin_path, toolchain, verbose):
-  toolchain_bin = os.path.join(toolchain, 'bin')
-  for exe in executables_to_install:
-    install_binary(exe, swiftpm_bin_path, toolchain_bin, verbose)
+def non_darwin_install(args, swiftpm_bin_path):
+  for prefix in args.install_prefixes:
```
buttaface, line 227: This should keep working as it always has when a `--prefix` list isn't passed in, since [this list is default-initialized to the toolchain path anyway](https://github.com/apple/swift-driver/blob/148e9d8818363ddfcb0c2b4db41bbebd04819634/Utilities/build-script-helper.py#L646) and `build-script` never passed `--prefix` in.

Utilities/build-script-helper.py
buttaface: Also, it only checks the first element of the hosts with `args.cross_compile_hosts[0]`, as apple/swift#36917 only passes in one at a time. I have proposed changing that flag to a single `--cross_compile_host` instead.

buttaface: I just noticed that you're changing the format of the strings in `cross_compile_hosts` for Darwin in the lines above, whereas I'm keeping the passed-in format for non-Darwin. Since the triple format you use on Darwin is only needed for the CMake build and Darwin `install()`, both of which are unused with non-Darwin hosts in this pull, that works fine for now with these non-Darwin hosts.

Utilities/build-script-helper.py
artemcm: I'd really like to unify these one day, but for now, I agree, this is simpler.

artemcm (approved):
Thank you, @buttaface.

buttaface:
Just needs to be run through the CI and merged?

artemcm:
@swift-ci please test

artemcm:
@buttaface does this need to have a cross-repo test from Swift repo side to make sure that this doesn't affect the `build-script` flow?

buttaface:
No, if you're happy with how this works, we can get this in now, then the build-script side can be tested separately for apple/swift#36917. Btw, I asked you a question about that pull over there.

PR #746, opened 2021-07-06
[NFC] Migrate Off of Deprecated XCTest Failure API

CodaFi:
Now that our baseline deployment target is 10.15, we can unconditionally use the nice API here.


CodaFi:
@swift-ci test

compnerd:
`XCTSourceCodeContext` is not available in swift-corelibs-xctest yet.  (CC: @tomerd who has additional context about the state of the divergence)

tomerd:
cc @andrewnesbitt

CodaFi:
@tomerd Do we have a radar tracking this?

tomerd:
@CodaFi that would be a question for @andrewnesbitt

PR #748, opened 2021-07-07
Reapply swift-api-digester integration

owenv:
This reapplies the patch that caused failure on the package bots, then updates build-script-helper to include swift-api-digester in the list of tools whose lookup is override by the --toolchain flag

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -15,7 +15,7 @@
         "repositoryURL": "https://github.com/apple/swift-llbuild.git",
         "state": {
           "branch": "main",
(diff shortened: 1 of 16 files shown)


owenv:
I'm not 100% sure this is the problem, so I'm going to try and build a toolchain with this patch and see if that succeeds

@swift-ci please test

owenv:
@swift-ci test

owenv:
It looks like the fix in this PR worked - I was able to build a toolchain in the linked PR and verified the tests ran and passed there

artemcm (approved):

PR #752, opened 2021-07-13
Add `.silent` mode to the ToolExecutionDelegate to be used in integrated driver mode

artemcm:
When the integrated driver is being used, the individual job output should be tracked by the library client, instead of the driver forwarding it into stderr.

Resolves rdar://80520371

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -11,7 +11,7 @@
         }
       },
       {
-        "package": "llbuild",
+        "package": "swift-llbuild",
         "repositoryURL": "https://github.com/apple/swift-llbuild.git",
         "state": {
           "branch": "main",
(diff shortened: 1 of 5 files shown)


artemcm:
@swift-ci please test

Package.resolved
```diff
@@ -11,7 +11,7 @@
         }
       },
       {
-        "package": "llbuild",
+        "package": "swift-llbuild",
```
nkcsgexi, line 14: Accidental change?

nkcsgexi (approved):

Package.resolved
artemcm: I don't think so, from what I can tell the llbuild package updated this field so it will resolve to this from now on, so we might as well keep it up-to-date here?

artemcm:
@swift-ci please test

Package.resolved
nkcsgexi: ah, I see. Thanks for the explanation!

owenv (commented):
LGTM 👍

PR #758, opened 2021-07-15
[Incremental] Relax currency test to fix ci errors

davidungar:
Some lit tests are failing on CI, but they won't fail locally. Based on the output, I suspect the priors currency test is failing and it shouldn't. So let's try relaxing the test: using the build-record mod time instead of the build-start time, and don't do the build end time test, since the build record had better be from the previous run anyway.


davidungar:
@swift-ci please test

davidungar:
Addresses rdar://80587167

CodaFi (approved):

PR #761, opened 2021-07-15
[Incremental] Relax the priors currency test by what ought to be enough to get through CI

davidungar:
The CI mod times seem to be very coarse, so relax the priors currency test a bit more.


davidungar:
@swift-ci please test

artemcm (approved):

PR #762, opened 2021-07-15
Specify -clang-target only when explicit module is on

nkcsgexi:
We should turn clang target on for implicit module after we are confident that the feature works nicely in explicit module builds.


nkcsgexi:
@swift-ci please test

artemcm (approved):

PR #749, opened 2021-07-07
[Gardening] Delete deprecated API for tracking external target information.

artemcm:
This was in-place as a temporary shim while SwiftPM adapted the `InterModuleDependencyOracle` API and a lighter-weight `ExternalTargetModulePathMap`.

PR #763, opened 2021-07-15
[Dependency Scanning] Restore passing external target map as inputs to the scanning action.

artemcm:
I got overzealous when deleting code in https://github.com/apple/swift-driver/pull/749 and removed one piece of functionality that we still rely on.
Although the external target map is no longer necessary when scanning dependencies using libSwiftScan, we still use it for resolution of `canImport` statements during the scan.
We need to figure out a better way to deal with `canImport` statements in the scanner, but in the meantime this will restore functionality.


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #765, opened 2021-07-16
Revert "[Dependency Scanning] Restore passing external target map as inputs to the scanning action."

artemcm:
Reverts apple/swift-driver#763

Possibly causing SwiftPM failure on Linux, when combined with a recent compiler change.
https://ci.swift.org/job/oss-swift-pr-test/18142/


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #768, opened 2021-07-20
[Incremental, DNM] Try reading swiftdeps in parallel

davidungar:
Experimental PR to try reading swiftdeps files in parallel.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -215,7 +215,6 @@ extension IncrementalCompilationState {
   /// Careful: job may not be primary.
   public func collectJobsDiscoveredToBeNeededAfterFinishing(
     job finishedJob: Job) throws -> [Job]? {
-    return try self.confinementQueue.sync {
       // Find and deal with inputs that now need to be compiled
       let invalidatedInputs = collectInputsInvalidatedByRunning(finishedJob)
       assert(Set(invalidatedInputs).isDisjoint(with: finishedJob.primaryInputs),
@@ -229,7 +228,6 @@ extension IncrementalCompilationState {
       }
       return try getJobs(for: invalidatedInputs)
     }
-  }
 
   /// Needed for API compatibility, `result` will be ignored
   public func collectJobsDiscoveredToBeNeededAfterFinishing(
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -98,13 +98,15 @@ extension ModuleDependencyGraph {
   /// Integrates `input` as needed and returns any inputs that were invalidated by external dependencies
   /// When creating a graph from swiftdeps files, this operation is performed for each input.
   func collectInputsRequiringCompilationFromExternalsFoundByCompiling(
-    input: TypedVirtualPath
+    input: TypedVirtualPath,
+    confinementQueue: DispatchQueue? = nil
```
cltnschlosser, line 102: So `nil` is valid when already on the queue?

In this case it might be better to drop the default parameter, forcing the user to specify nil means they have to think about if the queue is needed or not. It's easy to accidentally not provide it when it is needed.

Also should document what nil means here.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -266,21 +263,22 @@ extension IncrementalCompilationState {
...
-        assert(primaryInputs[0] == input)
-        self.reporter?.report("Scheduling invalidated", input)
-        return group.allJobs()
-      }
-      else {
-        self.reporter?.report("Tried to schedule invalidated input again", input)
-        return []
+    try confinementQueue.sync {
+      let unbatched = invalidatedInputs.flatMap { input -> [Job] in
+        if let group = skippedCompileGroups.removeValue(forKey: input) {
```
cltnschlosser, line 268: Is it `skippedCompileGroups` here that needs to be accessed on the queue?

IDK if there is convention for denoting fields that need to be accessed on the queue (leading _ or unsafe prefix, etc).

cltnschlosser (commented):
An actor would be nice here :)

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: Thanks for taking a look! I failed to be clear; this is just a quick-and-dirty experiment for you to try to see if the clean build gets improved. But, if this helps, your comments will be very valuable as it becomes something more real.
I agree with your thought about the parameter. Or maybe there's a better way--parallelizing swiftdeps when running with -disable-incremental-imports.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Good point!

davidungar: And yes, you are correct about `skippedCompiledGroups`.

PR #769, opened 2021-07-21
Fix swiftpm warnings about CMakeLists.txt files

benlangmuir:
Exclude the CMakeLists.txt files explicitly so that swiftpm does not issue warnings about them.

I still see a warning about `Sources/SwiftDriver/SwiftDriver.docc`, but I wasn't sure if it's correct to exclude it or if that might interfere with using it from Xcode or not.


benlangmuir:
@swift-ci please test

artemcm (approved):
Thank you!

PR #770, opened 2021-07-22
Propagate `-warn-concurrency` to frontend invocations.

artemcm:
Matching behavior of this flag in the legacy driver.

Resolves rdar://80925867


artemcm:
@swift-ci please test

benlangmuir (approved):
Thanks for the quick turnaround!

PR #776, opened 2021-07-29
SR-14796: Don't lose '-sdk /' when dropping the trailing slash

buttaface:
This is a regression from [the C++ Driver](https://github.com/apple/swift/blob/35735e7d0c57b57953623eb069d3812cd95d25e6/lib/Driver/Driver.cpp#L1776), ~will add a test later today and this should be~ ready to go.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2100,7 +2100,7 @@ extension Driver {
     }
 
     // Delete trailing /.
-    sdkPath = sdkPath.map { $0.last == "/" ? String($0.dropLast()) : $0 }
+    sdkPath = sdkPath.map { $0.count > 1 && $0.last == "/" ? String($0.dropLast()) : $0 }
 
     // Validate the SDK if we found one.
     if let sdkPath = sdkPath {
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2100,7 +2100,7 @@ extension Driver {
     }
 
     // Delete trailing /.
-    sdkPath = sdkPath.map { $0.last == "/" ? String($0.dropLast()) : $0 }
+    sdkPath = sdkPath.map { $0.length > 1 && $0.last == "/" ? String($0.dropLast()) : $0 }
```
cltnschlosser, line 2103: What is `length`? Should be `count` for String.

buttaface: `length` appears to work the same, but I will switch it to the more Swifty `count`.

buttaface:
Added a test, this is ready for review and CI.

artemcm:
@swift-ci please test

drexin (approved):
Thank you!

buttaface:
@drexin, should I submit for the 5.5 branch too, so this doesn't regress when that is released, or is it not worth it? I'm not sure.

drexin:
@buttaface Yeah, I think that would be good.

PR #777, opened 2021-07-30
Jobs: do not inject the swiftautolink file into archives

compnerd:
It has been observed that a static library may sometimes contain the
autolink extracted rules.  This causes autolink-extract to not be able
to process the archive as a dependency, causing a build failure.

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -290,7 +290,8 @@ extension GenericUnixToolchain {
       commandLine.appendFlag("crs")
       commandLine.appendPath(outputFile)
 
-      commandLine.append(contentsOf: inputs.map { .path($0.file) })
+      commandLine.append(contentsOf: inputs.filter { $0.type == .object }
+                                           .map { .path($0.file) })
       return try getToolPath(.staticLinker(lto))
     }
 


compnerd:
CC: @CodaFi @tomerd @buttaface

compnerd:
@swift-ci please test

buttaface (approved):
Thanks for fixing this.

buttaface:
Would be good to get this in. I'm putting together a pull to download the latest 5.5 snapshots and test them on my daily Android CI, and [I currently have to disable using this new Swift driver](https://github.com/buttaface/swift-android-sdk/pull/10/files#diff-61981150078c2463a1cfa23ce8e7b4a91ca3ecebeec9667958c2a482e51497bfR90) because of this bug.

buttaface:
@DougGregor, would be good to get this in, or using Swift static libraries on linux will regress with the 5.5 release. I'd like to get #784 in too.

buttaface:
@artemcm, can we get this in too? This currently breaks building the Foundation static libraries for Android, when it tries to link `plutil` against them and fails because of the extra autolink file.

artemcm (approved):
This one is good to go.
cc @airspeedswift

buttaface:
Ping @airspeedswift, would be good to get this in.

buttaface:
@artemcm, any reason why this pull and #784 have not gone in yet? It's been a month and a half since you approved them.

PR #778, opened 2021-07-30
Jobs: permit bitcode items when building with LTO

compnerd:
When building an archive, permit the LLVM bitcode content into the
archive as the LTO path will add those.  This should ensure that we can
continue to experiment with the LTO optimizations.


compnerd:
CC: @tomerd @CodaFi @buttaface

compnerd:
@buttaface could you please verify that this resolves the issue(s) you have been seeing?

compnerd:
@swift-ci please test

buttaface:
I do not mess with LTO, I only raised the issue of bitcode because I noticed that the old C++ Driver allowed it. You all know better with the LTO stuff, I was just raising the possibility of including it in #773.

PR #779, opened 2021-07-30
[Explicit Module Builds] Add API to specify richer external target module details

artemcm:
Specifically, whether an external target moduel is a framework. This is required to pass this information down to the compiler in order to generate correct auto-linking directives.

Part of rdar://81177797

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -50,7 +50,7 @@ let package = Package(
       name: "SwiftDriver",
       dependencies: ["SwiftOptions", "SwiftToolsSupport-auto",
                      "CSwiftScan", "Yams"],
-      exclude: ["CMakeLists.txt"]),
+      exclude: ["CMakeLists.txt", SwiftDriver.docc]),
 
     /// The execution library.
     .target(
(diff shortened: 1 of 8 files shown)


artemcm:
@swift-ci please test

nkcsgexi (approved):
Makes sense. Thank you for fixing this!

Package.swift
```diff
@@ -50,7 +50,7 @@ let package = Package(
       name: "SwiftDriver",
       dependencies: ["SwiftOptions", "SwiftToolsSupport-auto",
                      "CSwiftScan", "Yams"],
-      exclude: ["CMakeLists.txt"]),
+      exclude: ["CMakeLists.txt", SwiftDriver.docc]),
```
jakepetroules, line 53: This broke the build because "SwiftDriver.docc" isn't a quoted string. Are we not building swift-driver as a package in swift-ci anywhere? /CC @shahmishal

jakepetroules: Fixing in #782

PR #780, opened 2021-07-30
[Explicit Module Builds] Pass in `-fno-implicit-module-maps` for pre-building Clang explicit dependencies

artemcm:
The dependency scanner specifies which exact modulemaps must be used, we do not need Clang to also search for modulemaps.


nkcsgexi (approved):

artemcm:
@swift-ci please test

PR #784, opened 2021-08-03
SR-14796: Don't lose '-sdk /' when dropping the trailing slash

buttaface:
Cherry-pick of #776

- Explanation: This was [a regression compared to the old C++ Driver](https://github.com/apple/swift/blob/35735e7d0c57b57953623eb069d3812cd95d25e6/lib/Driver/Driver.cpp#L1776), where this new Driver can no longer deal with a passed-in SDK path of `/`.
- Scope: Only affects the single-character SDK path of `/`, allowing it again.
- Risk: Almost none
- Testing: Passes the new test, as indicated in the upstream pull.
- Reviewer: @drexin

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2072,7 +2072,7 @@ extension Driver {
     }
 
     // Delete trailing /.
-    sdkPath = sdkPath.map { $0.last == "/" ? String($0.dropLast()) : $0 }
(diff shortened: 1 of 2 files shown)


drexin:
@swift-ci test

buttaface:
Ping, would be good to get this and #777 in, so I can drop [workarounds like this](https://github.com/buttaface/termux-packages/blob/swift55/packages/swift/swift-corelibs-foundation-Sources-Foundation-CMakeLists.txt.patch#L10) when building the Swift 5.5 toolchain.

artemcm (approved):
LGTM! 
cc @airspeedswift

drexin:
@swift-ci please nominate

PR #781, opened 2021-07-30
[5.5][Explicit Module Builds] Cherry-pick explicit module build changes

artemcm:
#696: [Explicit Module Builds] Add libSwiftScan API for scanner cache serialization/deserialization
#705: [Explicit Module Builds] Use context hash in Swift dependency module file names
#739: [Explicit Module Builds] Unblock package bots by making tests more resilient to environment changes
#749: [Gardening] Delete deprecated API for tracking external target information
#763: [Dependency Scanning] Restore passing external target map as inputs to the scanning action
#779: [Explicit Module Builds] Add API to specify richer external target module details
#780: [Explicit Module Builds] Pass in -fno-implicit-module-maps for pre-building Clang explicit dependencies
#788: [Explicit Module Builds] Include source modulemap path into PCM filename hash.

PR #787, opened 2021-08-04
Temporarily disable `testDependencyScanning`

artemcm:
To unblock updating CI Xcode which is currently failing in:
https://ci.swift.org/job/swift-PR-macOS-new-xcode/117/console

The proper test fix is included in https://github.com/apple/swift-driver/pull/781, which will re-enable this test.


artemcm:
@swift-ci please test

PR #788, opened 2021-08-06
[Explicit Module Builds] Include source modulemap path into PCM filename hash.

artemcm:
Not doing so can lead to dangerous collisions if a build system uses a cetral cache across different targets that have a common dependency that is found in different places in the filesystem. For example, when building targets `A` and `B`, both of which depend on Clang module `C`, but have different search paths, finding `C` in two different locations, we should produce two distinct PCMs. With today's hashing scheme, the second will overwrite the first.


artemcm:
@swift-ci please test

nkcsgexi (approved):

artemcm:
@swift-ci please test

PR #789, opened 2021-08-10
[NFC] Box up consumed options.

davidungar:
Treat a method on the `Driver` whose only state change is the consumption of options as not changing the `Driver`. This change should facilitate an incremental build optimization to come.

### DIFF
--- a/Sources/SwiftOptions/ParsedOptions.swift
+++ b/Sources/SwiftOptions/ParsedOptions.swift
@@ -11,8 +11,8 @@
 //===----------------------------------------------------------------------===//
 
 /// Describes a single parsed option with its argument (if any).
-public struct ParsedOption {
-  public enum Argument {
+public struct ParsedOption: Hashable {
+  public enum Argument: Hashable {
     case none
     case single(String)
     case multiple([String])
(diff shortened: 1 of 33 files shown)


davidungar:
@swift-ci please test

Sources/SwiftOptions/ParsedOptions.swift
```diff
@@ -80,6 +80,27 @@ extension ParsedOption: CustomStringConvertible {
   }
 }
 
+/// In order to someday warn if options are not used, track those. Use a `class` rather than a `struct`
+/// so that consuming an option does not count as mutating the `Driver`.
+/// Otherwise, any ``Driver`` method, such as ``Driver/compileJob(primaryInputs:outputType:addJobOutputs:emitModuleTrace:)``
+/// would have to be `mutating`, but for most purposes, we don't care about consumed options.
+private class UnconsumedOptionsBox {
```
DougGregor, line 87: What's our concurrency story here? Should we pre-allocate all of the storage in the box, so we can have efficient setting of the "consumed" bits without a lock?

davidungar: Yes, that's the question! I assume that this driver must run on platforms that don't support the new concurrency features--apologies for stating the obvious! I haven't worked much on concurrency, so would like to ask you: Is there something we could use (maybe an unfair lock), that would be efficient enough? Or does the concurrency issue make this a non-starter? I don't understand your thought about pre-allocating all the storage in the box. But if you explain it enough so I can do it, I'll be glad to!

PR #730, opened 2021-06-24
Make emit-module-separately the default incremental build mode 

xymus:
Making emit-module-separately the default build mode will improve the reliability of incremental builds. This mode builds the module files in a distinct job which only parses the module API and doesn't use the merge-module phase.

In support, this PR fixes a few issues and update a lot of tests expecting the precise behavior of merge-module.

rdar://77225764

PR #790, opened 2021-08-10
Cleanup to prepare for emit-module-separately by default

xymus:
This PR extracts fixes and test updates from #730 as these are safer and less likely to be reverted. A few tests are hardened, some explicitly use merge-modules and others are preemptively moving to emit-module-separately.


xymus:
@swift-ci Please test

artemcm (approved):

PR #791, opened 2021-08-11
Put default interface verifier behind an env var

nkcsgexi:
Without setting this variable, we won't enable module interface verifier unless
explicitly asked to.

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -439,6 +439,12 @@ extension Driver {
                               default: false)
     else { return }
 
+    // FIXME: remove this when we are confident to enable interface verification
+    // by default.
+    if env["ENABLE_DEFAULT_INTERFACE_VERIFIER"] == nil &&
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -439,6 +439,12 @@ extension Driver {
                               default: false)
     else { return }
 
+    // FIXME: remove this when we are confident to enable interface verification
+    // by default.
+    if env["ENABLE_DEFAULT_INTERFACE_VERIFIER"] == nil &&
```
xymus, line 444: This is mostly a note for the future when we'll turn this var around. We could move this check to the `default` argument on line 432:
```
       parsedOptions.hasFlag(positive: .verifyEmittedModuleInterface,
                             negative: .noVerifyEmittedModuleInterface,
                             default: true),
```

xymus (approved):
Thanks!

PR #792, opened 2021-08-12
PrebuiltModuleGen: generate per-interface log files

nkcsgexi:
This is the behavior of swift_build_sdk_interfaces.py and we should do
the same in the new implementation.


nkcsgexi:
@swift-ci please test

artemcm (approved):
Nice!

PR #797, opened 2021-08-16
Fix `guard let sdkPath = sdkPath` pattern for compatiblity with older compilers

artemcm:
Introduced in https://github.com/apple/swift-driver/pull/796, this passed CI because swift-driver CI uses a recent compiler.
This broke SwiftPM CI because it uses a 5.3 compiler which contains a bug which makes such patterns fail to build.


artemcm:
@swift-ci please test

nkcsgexi (approved):
Thank you!

PR #800, opened 2021-08-17
[WIP, DNM, Incremental] Intern experiment for performance

davidungar:


davidungar:

davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

davidungar:
rdar://75386359

davidungar:
Relative savings on some test:
29% testCleanBuildSwiftDepsPerformance
25% testSavingPriorsPerformance
 9% testReadingPriorsPerformance
However, these won't reflect overall savings.

davidungar:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,76 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+public struct InternedString: Hashable, Comparable, CustomStringConvertible {
+  let index: Int
+  let table: InternedStringTable
+  
+  fileprivate init(_ s: String, host: ModuleDependencyGraph) {
```
artemcm, line 19: The use of the module dependency graph here is confusing. 
Why should a generic interned string have any connection to an incremental-build-related data structure? 

I think it should be the String cache entity itself that we rely on here, not where that cache is meant to live (`ModuleDependencyGraph`).

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/DependencySource.swift
```diff
@@ -55,12 +67,15 @@ extension DependencySource {
   /// Throws if a read error
   /// Returns nil if no dependency info there.
   public func read(
-    info: IncrementalCompilationState.IncrementalDependencyAndInputSetup
+    info: IncrementalCompilationState.IncrementalDependencyAndInputSetup,
+    host: ModuleDependencyGraph
```
artemcm, line 71: We now often have the use of `host` throughout that I can't wrap my head around by locally looking at the code. From the rest of this PR I gather that we are passing around the dependency graph because that's where the interned string cache lives, but the name `host` doesn't tell me that. Perhaps we can pass around a reference to the actual string cache to where it is needed?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,76 @@
...
+  public static func <(lhs: Self, rhs: Self) -> Bool {
+    assert(lhs.table === rhs.table)
+    return lhs.string < rhs.string
+  }
+    
+  public var description: String { string }
+}
+
+/// Hardcode empty as 0
+public class InternedStringTable {
```
artemcm, line 51: This feels very close to our existing `PathCache` implementation (https://github.com/apple/swift-driver/blob/main/Sources/SwiftDriver/Utilities/VirtualPath.swift#L325).

- Is this functionality we can common-out or maybe model this approach after the already-existing one? 
- Why do we not require synchronization in this case? If this table lives in the module-dependency-graph, can't we be accessing that concurrently after compilation jobs finish? 
- With filepaths, we have one global cache with synchronized accesses, whereas here we are relying on the specific instance of the string cache/table to be specified whenever we are dealing with interned strings. I wonder why we can't rely on a global cache here as well, much like `VirtualPath`? That way we won't have to tie this to the dependency graph and plumb the graph to all the use-sites.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,76 @@
...
+  fileprivate init(_ s: String, host: ModuleDependencyGraph) {
+    let t = host.internedStringTable
+    self.init(index: s.isEmpty ? 0 : t.index(s),
+              table: t)
+  }
+  
+  public var isEmpty: Bool { index == 0 }
+  
+  // PRIVATE?
+  init(index: Int, table: InternedStringTable) {
```
artemcm, line 28: Yeah, I don't think we should allow this to be used publicly.

davidungar:
Driver CPU times measured by Instruments, on a 1000+ source-file project, 3 runs each

clean build baseline: 20.65, 20.67, 20.08
clean build this PR: 18.07, 18.72, 19.01

incremental build baseline: 3.63, 3.46, 3.63
incremental build this PR: 2.65, 2.77, 2.85

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -954,50 +1040,17 @@ extension ModuleDependencyGraph {
...
-        // Major version
-        .fixed(bitWidth: 16),
-        // Minor version
-        .fixed(bitWidth: 16),
-        // Frontend version
-        .blob,
-      ])
-      self.abbreviate(.moduleDepGraphNode, [
-        .literal(RecordID.moduleDepGraphNode.rawValue),
+      let dependencyKeyOperands: [Bitstream.Abbreviation.Operand] = [
```
CodaFi, line 1053: Is this factorization relevant to the patch?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -660,21 +726,71 @@ extension ModuleDependencyGraph {
...
+          return Int(u)
+        }
+        func internedString(field i: Int) throws -> InternedString {
+          try InternedString(index: stringIndex(field: i))
+        }
+        func nonemptyInternedString(field i: Int) throws -> InternedString? {
+          let s = try internedString(field: i)
+          return s.isEmpty ? nil : s
+        }
+        func dependencyKey(fields: (kindCode: Int,
```
CodaFi, line 777: Nit: The use of a tuple makes the call sites opaque. Perhaps this should be unpacked back out into separate operands in the signature.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,92 @@
...
+public func isInIncreasingOrder(
+  _ lhs: InternedString, _ rhs: InternedString,
+  in holder: InternedStringTableHolder
+) -> Bool {
+  lhs.lookup(in: holder) < rhs.lookup(in: holder)
+}
+
+extension InternedString {
+  public static func < (lhs: InternedString, rhs: InternedString) -> Bool {
+    lhs.index < rhs.index
```
CodaFi, line 61: This doesn't seem correct. And if it is, it's a very surprising implementation of Comparable for this type. I could expect that we would force callers to go through `isInIncreasingOrder` et al.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,92 @@
...
+    self.index = index
+  }
+  
+  public static var empty: Self {
+    let r = Self(index: 0)
+    assert(r.isEmpty)
+    return r
+  }
+
+  public func lookup(`in` holder: InternedStringTableHolder) -> String {
```
CodaFi, line 40: Nit: `in` does not need to be escaped here.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/// Hardcode empty as 0
+public class InternedStringTable {
+  var strings = [""]
+  fileprivate var indices = ["": 0]
+  
+  public init() {}
+  
+  fileprivate func index(_ s: String) -> Int {
```
CodaFi, line 72: Nit: This is a mutating operation with a name that does not convey as much. I would expect this to be named `intern` as well.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
CodaFi: > Why do we not require synchronization in this case? 

In theory, all usages of this class should be as thread safe as the calling context. Integration is currently a serial process in all of the mutating paths, but that's a somewhat ad-hoc thing at the moment in that it's only really being enforced manually by `IncrementalCompilationState`. In practice, there will be interners on both the main queue and on the private queue maintained by `IncrementalCompilationState` which could lead to data races on the cache.

What this suggests to me is that we need to make it structurally impossible for anything but serialization and the module dependency graph to mutate these things, which is currently not the case with this API. Where that is not possible, we should be asserting that accesses are occurring on the expected queue (`dispatchPrecondition` can help).

CodaFi:
Some general comments:

- Thread safety is a pretty big concern here. Could you try running your profiling with a serial access queue attached to see what the overhead of synchronization is?
- `isInIncreasingOrder` doesn't appear to be used anywhere. I suspect we can drop it from this patch and introduce it later when you add it to the Tracer.
- The current representation gives callers direct access to the `Int` in the handle. That should be opaque, and it should not be possible for a caller to construct an interned string handle themselves without going through the table.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,92 @@
...
+//
+// Copyright (c) 2020 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
CodaFi, line 13: Nit: I don't believe we need this import.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,92 @@
...
+    lhs.index < rhs.index
+  }
+}
+
+/// Hardcode empty as 0
+public class InternedStringTable {
+  var strings = [""]
+  fileprivate var indices = ["": 0]
+  
+  public init() {}
```
CodaFi, line 70: Please also benchmark the high water mark for this table and see if `reserveCapacity` on both tables helps here.

davidungar: I assume you mean to save the table size in the priors, and reserve it when reading the priors. Great idea! I'll do that.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
davidungar: Thanks. Gone!

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
davidungar: Sure, no prob.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
davidungar: Thanks. Done!

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: Easy enough to fix!

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
davidungar: The old code duplicated this information, and yes, I could have done the patch by keeping the duplication. Since the patch involved changing the content of the serialized key, the factorization contributed to the quality of the patch by centralizing the invariant around what was needed in the file.

Someday, I'd love to further centralize so that the structure, the code to read, and the code to write each type of node were factored together.

In the meantime, the round-trip test you wrote provides a way to catch many of the potential bugs.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
davidungar: Yes, a holdover. It's gone.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
davidungar: Yes, the refactor I did before set up `IncrementalCompilationState` as the serialization point for the graph. Maybe the table should live there instead of the graph, and be protected there?
I worry that `dispatchPrecondition` on every access could damage performance. How does its overhead compare to a dictionary-keyed-by-string lookup?

davidungar:
@swift-ci please test

davidungar:
@swift-ci please test

davidungar:
@artemcm @CodaFi I have added thread-safety in the form of `dispatchPrecondition` as suggested. Seems to be very little cost. How does this look now? Thanks!

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationSynchronizer.swift
```diff
@@ -0,0 +1,75 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Dispatch
+
+/// Any instance that can find the confinmentQueue for the incremental compilation state.
+/// By confirming, a type gains the use of the shorthand methods in the extension.
+public protocol IncrementalCompilationSynchronizer {
+  var incrementalCompilationQueue: DispatchQueue {get}
```
CodaFi, line 18: Why not just have the table own a confinement queue, or be accessible only from behind a class (say the module dependency graph) that protects it with a confinement queue? By introducing this extra axis of customizability, the it's harder to reason about the concurrency guarantees for the table since the queue it is accessed from could be (but, in this patch, is not after manual inspection) distinct from the one it is initialized with.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationSynchronizer.swift
```diff
@@ -0,0 +1,75 @@
...
+    try sync(flags: .barrier, execute: fn)
+  }
+  /// Block any concurrent mutation so that argument may access (but not mutate) the protected state.
+  @_spi(Testing) public func blockingConcurrentMutation<T>( _ fn: () throws -> T ) rethrows -> T {
+    try sync(execute: fn)
+  }
+}
+
+/// A test fixture for tests, etc., that require synchronization and  an ``InternedStringTable``
+public struct MockIncrementalCompilationSynchronizer: IncrementalCompilationSynchronizer {
```
CodaFi, line 59: Please keep testing code out of the framework.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,101 @@
...
+
+public protocol InternedStringTableHolder {
+  var internedStringTable: InternedStringTable {get}
+}
+
+public struct InternedString: CustomStringConvertible, Equatable, Hashable {
+  
+  let index: Int
+  
+  fileprivate init(_ s: String, _ table: InternedStringTable) {
```
CodaFi, line 23: Something about this initializer doesn't sit right with me. A modification is ultimately being performed on the table, which vends the handle to the interned string. It seems like the table should be the one to directly construct an `InternedString` and this initializer should just be the usual memberwise init.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
```diff
@@ -0,0 +1,101 @@
...
+    self.init(index: s.isEmpty ? 0 : table.intern(s))
+  }
+  
+  public var isEmpty: Bool { index == 0 }
+  
+  // PRIVATE?
+  init(index: Int) {
```
CodaFi, line 30: Yes, this must be private, or the handle needs to become an opaque type a la VirtualPath.Handle. Callers should never be able to construct these things without going through the table.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -69,38 +68,38 @@ public final class IncrementalCompilationState {
...
-      confinementQueue)
+      &driver)
     self.mandatoryJobsInOrder = firstWave.mandatoryJobsInOrder
     self.jobsAfterCompiles = jobsInPhases.afterCompiles
-    self.info = initialState.graph.info
   }
   
-  /// Block any threads from mutating `ProtectedState`
-  public func blockingConcurrentMutation<R>(
+  /// Allow concurrent access to while preventing mutation of ``IncrementalCompilationState/protectedState``
```
CodaFi, line 80: Nit: sentence fragment

CodaFi: Also the in-source link here isn't valid. DocC doesn't support (at least, ergonomically) linking to instance variables like this. A plain `protectedState` will do.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -34,10 +34,9 @@ import SwiftOptions
 /// The public API surface of this class is thread safe, but not re-entrant.
 /// FIXME: This should be an actor.
 public final class IncrementalCompilationState {
-  
-  /// A  high-priority confinement queue that must be used to protect the incremental compilation state.
-  private let confinementQueue: DispatchQueue
     
+  /// State needed for incremental compilation that can change during a run and must be protected from
+  /// concurrent mutation and access. Concurrent accesses are OK.
   private var protectedState: ProtectedState
```
CodaFi, line 40: The protected state should not have to include `info`

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationProtectedState.swift
```diff
@@ -37,43 +37,37 @@ extension IncrementalCompilationState {
...
-    dispatchPrecondition(condition: .onQueueAsBarrier(confinmentQueue))
-  }
-  fileprivate func checkAccess() {
-    dispatchPrecondition(condition: .onQueue(confinmentQueue))
-  }
 }
 
 // MARK: - 2nd wave
 extension IncrementalCompilationState.ProtectedState {
   mutating func collectBatchedJobsDiscoveredToBeNeededAfterFinishing(
```
CodaFi, line 67: Please mark this fileprivate so it cannot be accessed without going through the confinement queue.

CodaFi:
> Seems to be very little cost.

Do you have numbers here? The confinement queues shouldn't add that much overhead, but it should still appear in traces.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
davidungar: Thanks, done!

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationSynchronizer.swift
davidungar: Sorry, the comment was a bit old. This struct is also used to create the legend for emitting dot files. (See `oneOfEachKind` in `DependencyGraphDotFileWriter`. It's nice to have some filename in the dummy box in the legend. However, if you prefer, the filename could be empty, and this code could be in the testing area.

davidungar:
> > Seems to be very little cost.
> 
> Do you have numbers here? The confinement queues shouldn't add that much overhead, but it should still appear in traces.

Fair question: Just to be sure remeasured, using Instruments and filtering the tree on symbols matching "dispatchPrecondition", inverting the call tree.
Results:
- `testCleanBuildPerformance`: 1ms out of 2.37 s
- clean build of a project with more 1,000 source files: 9 ms out of 16.3 s for the driver
- quasi-null build (just one small change) of the same project: 6 ms out of 2.5 s for the driver.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/InternedStrings.swift
davidungar: When reading a serialized `ModuleDependencyGraph`, this PR saves work by using the indices of identifiers in the serialized graph as the indices into the `InternedStringTable`. So, there must be a bridge from the method in the ModuleDependencyGraph.swift code into this method, because the reader must take an index and turn it into an `InternedString`.

Can you suggest something better? Something like a C++ `friend` would be great.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationSynchronizer.swift
davidungar: Yes, an earlier version of this PR required a module dependency graph to be passed in to every intern / un-intern operation. IIRC, @artemcm correctly thought that code was a bit unclear. I can do away with the protocol if that would help. Again, for the sake of realizing the performance gains, I would love to find an architecture that would be above threshold for you. What can you suggest? I can try going back to the string-table hidden in the module dependency graph, but then operations such as reading a source file dependency graph for testing, or reading the priors will have to start from an empty module dependency graph. Not the worst thing in the world, but also a source of awkwardness, passing in a whole graph just to gain the ability to intern strings.

davidungar:
@swift-ci please test

davidungar:
@artem & @CodaFi : I think this is as far as I can go on this without more input from you. Would you care to take another look? Thanks.

davidungar:
After an offline discussion, we agreed it would be robust enough to make the `init(index:)` constructor private to the deserialization code.

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
```diff
@@ -333,16 +354,60 @@ public struct DependencyKey: CustomStringConvertible {
...
+  case let (.member(lc, ln), .member(rc, rn)):
+    return lc == rc ? f(ln) < f(rn) : f(lc) < f(rc)
+    
+  default: break
+  }
+  
+  /// Preserves the ordering that obtained before interned strings were introduced.
+  func kindOrdering(_ d: DependencyKey.Designator) -> Int {
+    switch d {
+    case .topLevel: return 1
```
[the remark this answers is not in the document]
davidungar: The issue is needing the string table as a parameter to the less-than function.

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
```diff
@@ -333,16 +354,60 @@ public struct DependencyKey: CustomStringConvertible {
...
+  case let (.externalDepend(ld), .externalDepend(rd)):
+    return ld < rd
+    
+  case let (.member(lc, ln), .member(rc, rn)):
+    return lc == rc ? f(ln) < f(rn) : f(lc) < f(rc)
+    
+  default: break
+  }
+  
+  /// Preserves the ordering that obtained before interned strings were introduced.
```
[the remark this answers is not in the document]
davidungar: Not sure how to address the concern here.

Sources/SwiftDriver/IncrementalCompilation/DependencyKey.swift
```diff
@@ -10,27 +10,48 @@
...
 import TSCBasic
+import Dispatch
 
 /// A filename from another module
 /*@_spi(Testing)*/ final public class ExternalDependency: Hashable, Comparable, CustomStringConvertible {
 
 
   /// Delay computing the path as an optimization.
-  let fileName: String
+  let fileName: InternedString
```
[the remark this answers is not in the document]
davidungar: If I interpret Xcode's callers facility correctly, the `fileName` is used in lots of places. Storing a `String` here would add some amount of overhead, as well as needing the string table at the use point.

davidungar:
Also helps rdar://79192750

PR #814, opened 2021-08-30
[Test Change] Use `Driver` to generate scanner invocation in dependency scanning test

artemcm:
This is more robust than just relying on a hand-written command, and unblocks this test passing on Apple Silicon.


artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #817, opened 2021-08-31
[Incremental] Make dot file writer a struct

davidungar:
Since the dot file writer contains a sequence number, it gets mutated whenever a dot file is written.
Make this a struct to pass that info upwards in preparation for making thread-safety for the incremental state clearer.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/DependencyGraphDotFileWriter.swift
@@ -12,7 +12,7 @@
 import TSCBasic
 
 // MARK: - Asking to write dot files / interface
-public class DependencyGraphDotFileWriter {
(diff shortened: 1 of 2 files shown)


davidungar:
@swift-ci please test

davidungar:

artemcm (approved):
Generally, why do we need to update the `versionNumber` here?
By having this be a value type, are we not risking that concurrent uses of this would write to the same location?

davidungar:
> Generally, why do we need to update the `versionNumber` here?
> By having this be a value type, are we not risking that concurrent uses of this would write to the same location?

Thank you for looking at this.
We update the version number so that, after successive imports, one can look the evolution of the dot file graph by opening different versions. I'll add something to that effect.

davidungar:
@swift-ci please test

artemcm:
@swift-ci please test

PR #821, opened 2021-09-02
[SwiftDriver] Make sure the proper file type is used for the optimization record file in a supplementary output file map

akyrtzi:
Previously it was always using `yaml` type even if the format requested was bitstream.


akyrtzi:
@swift-ci Please test

artemcm (approved):
LGTM!

nkcsgexi (approved):

PR #825, opened 2021-09-07
[Explicit Module Builds] Use Clang dependencies' `contextHash` as partof its filename-encoded hash

artemcm:
It is a more-robust way of ensuring different instances of the same module will never end up overlapping.

Ever since we removed the `-target` argument from `PCMArgs`, previous method of using `PCMArgs` and `.modulemap` path could not, for example, distinguish building the same Clang module, from the same `.modulemap`, with different `-target` triples.


artemcm:
@swift-ci please test

nkcsgexi (approved):
LGTM

PR #829, opened 2021-09-09
Improve testFileMapMissingMainEntryWMO

keith:
I realized this test would pass even if the WMO arg was removed, which
would negate the positive case. Now if that happens we fall back to
-incremental and get the expected warning.


CodaFi (approved):

CodaFi:
@swift-ci test

keith:
Thanks!

PR #832, opened 2021-09-10
Verify emitted modules interfaces of public frameworks by default

xymus:
Verify emitted module interfaces of public frameworks by default. These framework's interfaces are critical and must be valid.

We can verify the module interfaces of all frameworks later on.

rdar://82987741


xymus:
@swift-ci Please test

nkcsgexi (approved):
Makes sense to me!

artemcm (approved):

PR #833, opened 2021-09-10
[5.5][Explicit Module Builds] Use Clang dependencies' `contextHash` as part of its filename-encoded hash

artemcm:
Cherry-pick of https://github.com/apple/swift-driver/pull/825
----------------------------------------------------------------

- **Explanation**: It is a more-robust way of ensuring different instances of the same module will never end up overlapping.
Ever since we removed the -target argument from PCMArgs, previous method of using PCMArgs and .modulemap path could not, for example, distinguish building the same Clang module, from the same .modulemap, with different -target triples.
- **Risk**: Low
- **Reviewed by**: @nkcsgexi 
- **Testing**: Automated tests modified to reflect new behavior


nkcsgexi (approved):

artemcm:
@swift-ci please test

airspeedswift (approved):

artemcm:
@swift-ci please test

PR #835, opened 2021-09-13
Pass-down -prebuilt-module-cache-dir when targeting at macCatalyst

nkcsgexi:
All prebuilt modules are emitted into SDK-versioned subdirs. When targeting at macCatalyst,
the value from -target-sdk-version doesn't match the actual SDK version because the SDK under
used is a macosx SDK. This patch computes the prebuilt module cache path and passes it down directly
to the compiler invocation.

rdar://82464316

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -336,6 +336,18 @@ public final class DarwinToolchain: Toolchain {
       commandLine.append(.flag("-target-variant-sdk-version"))
       commandLine.append(.flag(sdkInfo.sdkVersion(for: targetVariantTriple).description))
     }
+    // We should be able to pass down prebuilt module dir for all other SDKs.
+    // For macCatalyst, doing so is specifically necessary because -target-sdk-version
+    // doesn't always match the macosx sdk version so the compiler may fail to find
+    // the prebuilt module in the versioned sub-dir.
+    if frontendTargetInfo.target.triple.isMacCatalyst {
+      commandLine.appendFlag(.prebuiltModuleCachePath)
+      commandLine.appendPath(try getToolPath(.swiftCompiler).parentDirectory/*bin*/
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -336,6 +336,18 @@ public final class DarwinToolchain: Toolchain {
...
       commandLine.append(.flag(sdkInfo.sdkVersion(for: targetVariantTriple).description))
     }
+    // We should be able to pass down prebuilt module dir for all other SDKs.
+    // For macCatalyst, doing so is specifically necessary because -target-sdk-version
+    // doesn't always match the macosx sdk version so the compiler may fail to find
+    // the prebuilt module in the versioned sub-dir.
+    if frontendTargetInfo.target.triple.isMacCatalyst {
+      commandLine.appendFlag(.prebuiltModuleCachePath)
+      commandLine.appendPath(try getToolPath(.swiftCompiler).parentDirectory/*bin*/
+        .parentDirectory/*usr*/
```
artemcm, line 346: You can make this a bit shorter by using `sdkStdlib(sdk: AbsolutePath)` from the `DarwinToolchain`. 

It's a bit of a shame we don't have a nicer way to compute these things other than relative to the compiler location.

artemcm (approved):

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
nkcsgexi: hmm, `sdkStdlib` gives us a path into the SDK I presume? However, the prebuilt modules are inside the toolchain.

artemcm: Hmm, perhaps that function is a bit mis-named. 
It just takes a path and appends:
```
appending(RelativePath("usr/lib/swift"))
```

nkcsgexi:
@swift-ci please test

PR #837, opened 2021-09-14
Add Literate Swift support to the driver.

al45tair:
swift-driver needs to know about Markdown, reStructuredText and LaTeX files to make this work.


al45tair:
This goes with apple/swift#39305

artemcm (approved):

PR #838, opened 2021-09-14
Pass the canonical SDK name to the frontend for swiftmodule checking

xymus:
Pass the canonical SDK name to the frontend for it to control which swiftmodule file can be loaded according to the SDK in use. Swiftmodule files should only be used in the context of the SDK used to create them. This will catch issues with roots and other similar configuration problems.

rdar://83104265

Related compiler-side PR: https://github.com/apple/swift/pull/37768


xymus:
@swift-ci Please test

nkcsgexi:
Could you please (1) update `Options.swift` using the ToT compiler source, and (2) check if `-target-sdk-name` is supported in the driver before feeding it to the compiler?

xymus:
@nkcsgexi How can we check if a flag is supported by the compiler?

xymus:
Nevermind, there's a `supportedFrontendFlags`.

nkcsgexi:
Right, it's especially necessary when we are using one single driver instance to orchestrate multiple compilers.

xymus:
Adding commits for the suggested changes.

@swift-ci Please test

nkcsgexi (approved):

kastiglione:
+1 nice addition

PR #843, opened 2021-09-21
[Explicit Module Builds] Use new form of `-fmodule-file` to work around Clang issue with explicit builds and emit-PCH

artemcm:
This form causes Swift's underlying Clang invocation to take different code-paths when loading pre-built PCM dependencies, which is not affected on whether the compiler is in emit-pre-built-header mode.

Workaround for rdar://83309762


artemcm:
@swift-ci please test

nkcsgexi:
Glad this simple change just works!

nkcsgexi (approved):

PR #848, opened 2021-09-22
[Dependency Scanning] Adapt to changes in dependenncy scanner to capture initial PCM arguments

artemcm:
The graph coming out of the scanner will already contain an initial set of captured PCM arguments as defined by the initial scanner invocation. This will allow us to skip the batch re-scan at argument sets that are redundant because they were already captured via the initial scan.


artemcm:
@swift-ci please test

artemcm:
Corresponding compiler PR that adds the new `libSwiftScan` entry-points:
https://github.com/apple/swift/pull/39397

artemcm:
@swift-ci please test

nkcsgexi (approved):

PR #854, opened 2021-09-28
Support response files with relative paths.

allevato:
This makes swift-driver's behavior consistent with the old driver, which supports relative paths (both on the command line and in other response files). The LLVM response file parsing behavior used by the legacy driver is such that

-   Response file equivalence to prevent recursion is based on the file's device/inode pair, not its path.

-   Relative paths are resolved based on the current working directory of the process, even when the relative path is in another response file. (LLVM's parser has an option to make relative paths resolve based on the containing response file's path, but this is not enabled by the legacy driver.)

-   After resolving, the path is tested to ensure that it resolves to an actual file; this ensures that special linker paths like `@loader_path` fall back to regular arguments instead of throwing an error and exiting by attempting to read them.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -869,6 +869,18 @@ extension Driver {
 
 // MARK: - Response files.
 extension Driver {
+  /// Tracks visited response files by unique file ID to prevent recursion,
+  /// even if they are referenced with different path strings.
+  private struct VisitedResponseFile: Hashable, Equatable {
+    var device: UInt64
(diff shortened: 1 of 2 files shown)


cltnschlosser (approved):
Looks good

allevato:
Thanks! It looks like I don't have write access to apple/swift-driver (just apple/swift and a couple others), so I don't think I can run tests. Can you please run those for me?

artemcm (approved):

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

PR #855, opened 2021-09-29
PrebuiltModuleGen: check if an interface is for macabi by reading interface contents

nkcsgexi:
Previously, we were using the file name to infer whether an interface is for macabi.
However, this seems to be inaccurate because some macabi interfaces don't have "-macabi" in
the file names. This patch teaches the driver to open interface files and check
module targets specifically.


nkcsgexi:
@swift-ci please test

nkcsgexi:
@xymus @artemcm @beccadax This may explain why we saw excessively more macabi interface failures in the prebuilt module job.

nkcsgexi:
@swift-ci please test

xymus (approved):

PR #856, opened 2021-09-29
[5.5] PrebuiltModuleGen: check if an interface is for macabi by reading interface contents

nkcsgexi:
Explanation: Previously, we were using the file name to infer whether an interface is for macabi.
However, this seems to be inaccurate because some macabi interfaces don't have "-macabi" in
the file names. This patch teaches the driver to open interface files and check
module target specifically.

Scope: Swift prebuilt module generation

Risk: Very low

Testing: regression test added

Reviewed by: @xymus 

Cherry-pick of: https://github.com/apple/swift-driver/pull/855


nkcsgexi:
@swift-ci please test

airspeedswift (approved):

PR #857, opened 2021-09-29
Building object files using -c with -lto should respect the output file path (-o) if present

kubamracek:
Currently, when combining -c + -lto + -o, the driver will produce a ".bc" file on disk, even if the -o argument said ".o". This PR fixes that and matches the behavior of the C++ driver in the swift repo.

### DIFF
--- a/Sources/SwiftDriver/Jobs/CompileJob.swift
+++ b/Sources/SwiftDriver/Jobs/CompileJob.swift
@@ -85,7 +85,7 @@ extension Driver {
       } else {
         // When -lto is set, .bc will be used for linking. Otherwise, .bc is
         // top-level output (-emit-bc)
-        return lto == nil
+        return lto == nil || linkerOutputType == nil
       }
     case .swiftModule:
       return compilerMode.isSingleCompilation && moduleOutputInfo.output?.isTopLevel ?? false
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -313,6 +313,28 @@ final class SwiftDriverTests: XCTestCase {
     XCTAssertEqual(driver6.compilerOutputType, .llvmBitcode)
   }
 
+  func testLtoOutputPath() throws {
+    do {
+      var driver = try Driver(args: ["swiftc", "foo.swift", "-lto=llvm-full", "-c", "-target", "x86_64-apple-macosx10.9"])
+      XCTAssertEqual(driver.compilerOutputType, .llvmBitcode)
+      XCTAssertEqual(driver.linkerOutputType, nil)
+      let jobs = try driver.planBuild()
+      XCTAssertEqual(jobs.count, 1)
+      XCTAssertEqual(jobs[0].outputs.count, 1)
+      XCTAssertEqual(jobs[0].outputs[0].file.basename, "foo.bc")
+    }
+
+    do {
+      var driver = try Driver(args: ["swiftc", "foo.swift", "-lto=llvm-full", "-c", "-target", "x86_64-apple-macosx10.9", "-o", "foo.o"])
+      XCTAssertEqual(driver.compilerOutputType, .llvmBitcode)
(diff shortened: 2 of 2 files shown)


kubamracek:
@swift-ci please test

artemcm (approved):
LGTM, thanks @kubamracek.

compnerd (approved):

kubamracek:
@swift-ci please test

kubamracek:
@artemcm @compnerd could I ask for another review? I had to fix up some existing tests, and I am not super sure the change is actually correct now :)

Sources/SwiftDriver/Jobs/CompileJob.swift
```diff
@@ -79,6 +79,10 @@ extension Driver {
     case .object:
       return (linkerOutputType == nil)
     case .llvmBitcode:
+      if linkerOutputType == nil {
+        return true
+      }
```
cltnschlosser, line 84: Do we want this override for all cases or only for lto? There is some lto checks below:
```swift
// When -lto is set, .bc will be used for linking. Otherwise, .bc is
// top-level output (-emit-bc)
return lto == nil
```
Maybe?
```swift
lto == nil || linkerOutputType == nil
```

Looking at the tests it seems like bc is a top level output in cases where maybe it shouldn't be.
Ex: `"-embed-bitcode", "-c"` `.bc` should be intermediary output that is passed to the "backend" job to become `.o`. We don't need to output it next to the final outputs.

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -313,6 +313,27 @@ final class SwiftDriverTests: XCTestCase {
...
+
+    var driver2 = try Driver(args: ["swiftc", "foo.swift", "-lto=llvm-full",
+                                    "-c", "-target", "x86_64-apple-macosx10.9",
+                                    "-o", "foo.o"])
+    XCTAssertEqual(driver2.compilerOutputType, .llvmBitcode)
+    XCTAssertEqual(driver2.linkerOutputType, nil)
+    let jobs2 = try driver2.planBuild()
+    XCTAssertEqual(jobs2.count, 1)
+    XCTAssertEqual(jobs2[0].outputs.count, 1)
+    XCTAssertEqual(jobs2[0].outputs[0].file.basename, "foo.o")
```
cltnschlosser, line 334: nit: you can wrap these in `do` blocks to namespace the variables and not need to number everything:
```suggestion
    do {
      var driver = try Driver(args: ["swiftc", "foo.swift", "-lto=llvm-full",
                                      "-c", "-target", "x86_64-apple-macosx10.9"])
      XCTAssertEqual(driver.compilerOutputType, .llvmBitcode)
      XCTAssertEqual(driver.linkerOutputType, nil)
      let jobs = try driver.planBuild()
      XCTAssertEqual(jobs.count, 1)
      XCTAssertEqual(jobs[0].outputs.count, 1)
      XCTAssertEqual(jobs[0].outputs[0].file.basename, "foo.bc")
    }

    do {
      var driver = try Driver(args: ["swiftc", "foo.swift", "-lto=llvm-full",
                                      "-c", "-target", "x86_64-apple-macosx10.9",
                                      "-o", "foo.o"])
      XCTAssertEqual(driver.compilerOutputType, .llvmBitcode)
      XCTAssertEqual(driver.linkerOutputType, nil)
      let jobs = try driver.planBuild()
      XCTAssertEqual(jobs.count, 1)
      XCTAssertEqual(jobs[0].outputs.count, 1)
      XCTAssertEqual(jobs[0].outputs[0].file.basename, "foo.o")
    }
```

Sources/SwiftDriver/Jobs/CompileJob.swift
kubamracek: Aha, I think that's the part I was missing. This makes sense, let me adjust the PR.

kubamracek:
@swift-ci please test

cltnschlosser (approved):

kubamracek:
@swift-ci please test

PR #858, opened 2021-10-03
Jobs: permit bitcode items when building with LTO

compnerd:
When building an archive, permit the LLVM bitcode content into the
archive as the LTO path will add those.  This should ensure that we can
continue to experiment with the LTO optimizations.


compnerd:
@swift-ci please test

cltnschlosser (approved):

artemcm (approved):

PR #860, opened 2021-10-05
Tests: add support for `cachedSdkPath` on Windows

compnerd:
Like Darwin, Windows uses `-sdk` to locate items required for building
for a target.  This adds support in the test harness to support this
flag.


Tests/TestUtilities/DriverExtensions.swift
```diff
@@ -45,9 +45,20 @@ extension Driver {
...
-  return nil
+  #if os(Windows)
+  return Result {
+    if let sdk = ProcessEnv.vars["SDKROOT"] {
+      return sdk.replacingOccurrences(of: "\\", with: "\\\\")
+    } else if let sdk = ProcessEnv.vars["DEVELOPER_DIR"] {
+      return "\(sdk.replacingOccurrences(of: "\\", with: "\\\\"))\\\\Platforms\\\\Windows.platform\\\\Developer\\\\SDKs\\\\Windows.sdk"
+    } else {
+      enum SwiftDriverTestsError: Error { case missingSDK }
+      throw SwiftDriverTestsError.missingSDK
```
compnerd, line 56: I'm not sure whether to return `nil` or throw an error here.  The problem is that `nil` is correct for when running from a swift build tree.  However, when running with a release toolchain, that is invalid, and the `SDKROOT` environment variable should be set.  One could argue that because the release toolchain would always be installed and the installer adds the environment variable, we should return `nil`.  I'd like to solicit advice from the swift-driver maintainers here.

artemcm: > the release toolchain would always be installed and the installer adds the environment variable, we should return nil

This logic makes sense to me. 
If we return `nil` and find ourselves in an environment where `SDKROOT` is not set, then we will see a ton of test failures which hopefully make it sufficiently obvious that we're missing an SDK.

compnerd: Excellent, I'll adjust this this then.  Sadly, that is a fair amounts away - I do have the driver building on Windows and somewhat working, but the test suite is a a pretty Christmas-y look right now ~50% red, ~50% green.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #862, opened 2021-10-07
gitattributes: add attributes for test inputs

compnerd:
Ensure that the swift interface files are always checked out with unix
line-endings irrespective of the configuration.


compnerd:
@swift-ci please test

PR #865, opened 2021-10-08
TestUtilities: add path utilities

compnerd:
The path handling in the tests needs to account for the fact that path
representation is currently hardcoded and the representation differs on
different platforms.  These are intended to help convert the
representation.


Tests/TestUtilities/PathExtensions.swift
```diff
@@ -0,0 +1,30 @@
+//===-------- DriverExtensions.swift - Driver Testing Extensions ----------===//
```
artemcm, line 1: Stale file name.

PR #866, opened 2021-10-08
SwiftDriverTests: disable some tests

compnerd:
These tests explicitly hardcode the macOS triple into the test, making
them inherently non-portable.  Just disable the tests.


compnerd:
Yeah, there is something definitely odd going on here.  I can hold off on merging this if you want to take a look @cltnschlosser.  I am building with a local patch to implement Windows support and just trying to get through the test suite.  But I believe that the version that I have is better than the version that has been up as a PR for a while (it matches the C++ driver better and supports a bit more functionality).

cltnschlosser:
No feel free, just an observation.

compnerd:
That observation has been bugging me too.  I don't understand why this passes on Linux.  Am I mistaken in that this should be limited to macOS?

PR #867, opened 2021-10-08
[5.5] Pass along the `-emit-symbol-graph` flags with `-emit-module-separately`

artemcm:
Originally merged on `main` in https://github.com/apple/swift-driver/pull/730
------------------------------------------
Resolves rdar://83905653


artemcm:
@swift-ci please test

PR #868, opened 2021-10-11
Stop passing `-no_objc_category_merging` to the linker on Darwin.

artemcm:
ObjC category merging is a really useful optimization, so we should stop doing that. This flag was added in the legacy driver as a workaround in 2014:
```
commit 9a1cf1af1691edc5da237b88be0b0fac548a5a09
Author: Greg Parker <gparker@apple.com>
Date:   Wed Feb 19 09:22:38 2014 +0000
```
And whatever issue it was working around should now be resolved.

Resolves rdar://84099601

### DIFF
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -418,8 +418,6 @@ extension DarwinToolchain {
       sdkPath: targetInfo.sdkPath?.path
     )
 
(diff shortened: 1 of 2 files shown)


Package.resolved
```diff
@@ -15,7 +15,7 @@
       "location" : "https://github.com/apple/swift-llbuild.git",
       "state" : {
         "branch" : "main",
-        "revision" : "08c0a758a697bad5d5849ad5d0b1eedb507b0596"
+        "revision" : "c7fe760c4d8bd2909a05e1fef9638f6374f86b05"
```
mikeash, line 18: Is this change intended?

PR #878, opened 2021-10-15
SwiftDriver: initial (incomplete) windows port

compnerd:
This is sufficient to build on Windows, and the resulting driver is able
to somewhat function.  The tests are still not usable due to assumptions
about path which do not hold and the insufficiency of the path
representation in tools-support-core.  This is likely best approached by
replacing the use of `Path` from t-s-c with the representation in
system.  However, this allows some attempt to validate the swift-driver
which is increasingly becoming required for s-p-m, and thus unblocks
further work.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -787,7 +787,7 @@ extension Driver {
     let execName = try VirtualPath(path: args[0]).basenameWithoutExt
 
     // If we are not run as 'swift' or 'swiftc' or there are no program arguments, always invoke as normal.
-    guard execName == "swift" || execName == "swiftc", args.count > 1 else {
+    guard [executableName("swift"), executableName("swiftc")].contains(execName), args.count > 1 else {
       return (.normal(isRepl: false), args)
     }
 
@@ -797,13 +797,13 @@ extension Driver {
 
     // Check for flags associated with frontend tools.
     if firstArg == "-frontend" {
-      updatedArgs.replaceSubrange(0...1, with: ["swift-frontend"])
-      return (.subcommand("swift-frontend"), updatedArgs)
+      updatedArgs.replaceSubrange(0...1, with: [executableName("swift-frontend")])
+      return (.subcommand(executableName("swift-frontend")), updatedArgs)
     }
 
     if firstArg == "-modulewrap" {
-      updatedArgs[0] = "swift-frontend"
-      return (.subcommand("swift-frontend"), updatedArgs)
+      updatedArgs[0] = executableName("swift-frontend")
+      return (.subcommand(executableName("swift-frontend")), updatedArgs)
     }
 
     // Only 'swift' supports subcommands.
@@ -824,7 +824,7 @@ extension Driver {
         return (.normal(isRepl: true), updatedArgs)
     }
 
-    let subcommand = "swift-\(firstArg)"
+    let subcommand = executableName("swift-\(firstArg)")
 
     updatedArgs.replaceSubrange(0...1, with: [subcommand])
 
@@ -2222,10 +2222,12 @@ extension Driver {
 
       if !fileSystem.exists(path) {
         diagnosticsEngine.emit(.warning_no_such_sdk(sdkPath))
-      } else if isSDKTooOld(sdkPath: path, fileSystem: fileSystem,
-                            diagnosticsEngine: diagnosticsEngine) {
-        diagnosticsEngine.emit(.error_sdk_too_old(sdkPath))
-        return nil
+      } else if !(targetTriple?.isWindows ?? (defaultToolchainType == WindowsToolchain.self)) {
+        if isSDKTooOld(sdkPath: path, fileSystem: fileSystem,
+                       diagnosticsEngine: diagnosticsEngine) {
+          diagnosticsEngine.emit(.error_sdk_too_old(sdkPath))
+          return nil
+        }
       }
 
       return .absolute(path)
@@ -2510,7 +2512,7 @@ extension Triple {
     case .wasi:
       return WebAssemblyToolchain.self
     case .win32:
-      fatalError("Windows target not supported yet")
+      return WindowsToolchain.self
     default:
       diagnosticsEngine.emit(.error_unknown_target(triple))
       throw Diagnostics.fatalError
@@ -2523,7 +2525,7 @@ extension Driver {
   #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
   static let defaultToolchainType: Toolchain.Type = DarwinToolchain.self
   #elseif os(Windows)
-  static let defaultToolchainType: Toolchain.Type = { fatalError("Windows target not supported yet") }()
+  static let defaultToolchainType: Toolchain.Type = WindowsToolchain.self
   #else
   static let defaultToolchainType: Toolchain.Type = GenericUnixToolchain.self
   #endif
--- a/Sources/SwiftDriver/Driver/WindowsExtensions.swift
+++ b/Sources/SwiftDriver/Driver/WindowsExtensions.swift
@@ -1,0 +1,24 @@
+//===------------- WindowsExtensions.swift - Windows Extensions -----------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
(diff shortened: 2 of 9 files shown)


compnerd:
@artemcm - the big problem with this change is that there are a few places where I cannot reliably make a distinction for the executable names (i.e. extensions - `.exe`).  Do you have thoughts on how to handle this?  I'm leaning towards adding a freestanding function:

```swift
func executable(_ name: String) -> String {
#if os(Windows)
  guard name.suffix(from: name.index(name.endIndex, offsetBy: -4))) == ".exe" else {
    return "\(name).exe"
  }
  return name
#else
  return name
#endif
}
```

compnerd:
@swift-ci please test Darwin platform

compnerd:

compnerd:
> > @artemcm - the big problem with this change is that there are a few places where I cannot reliably make a distinction for the executable names (i.e. extensions - `.exe`). Do you have thoughts on how to handle this? I'm leaning towards adding a freestanding function:
> > ```swift
> > func executable(_ name: String) -> String {
> > #if os(Windows)
> >   guard name.suffix(from: name.index(name.endIndex, offsetBy: -4))) == ".exe" else {
> >     return "\(name).exe"
> >   }
> >   return name
> > #else
> >   return name
> > #endif
> > }
> > ```
> 
> I think this is an entirely reasonable approach.

Okay; just wanted to make sure that someone doesn't have a better idea (I can propose and implement and still dislike my idea right? :-D).  There is some duplication - I'm like to move the function to a different file if I could get a bit of guidance as to where such a utility function should reside.

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -166,10 +180,12 @@ extension Toolchain {
...
     } else if let path = lookupExecutablePath(filename: executable, searchPaths: searchPaths) {
       return path
-    } else if executable == "swift-frontend" {
+    } else if executable == tool("swift-frontend") {
       // Temporary shim: fall back to looking for "swift" before failing.
-      return try lookup(executable: "swift")
+      return try lookup(executable: tool("swift"))
     } else if fallbackToExecutableDefaultPath {
+      // TODO(compnerd) correct this for Windows - %SystemDrive%\Developer\Toolchains\....xctoolchain\usr\bin
+      // To do this we need access to the Driver as we can check Driver.hostTriple.isWindows
```
cltnschlosser, line 188: Not pretty, but `self is WindowsToolchain`

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -131,6 +142,9 @@ extension Toolchain {
   private func envVarName(for toolName: String) -> String {
     let lookupName = toolName
       .replacingOccurrences(of: "-", with: "_")
+      // FIXME(compnerd) we should extract the extension for generating the
+      // toolname rather than assuming that we can convert the tool name blindly
+      .replacingOccurrences(of: ".", with: "_")
```
cltnschlosser, line 147: Yeah, I guess it would be nice if the environment variables were consistent across platforms. So windows doesn't always have an extra `_EXE` in its variables.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -14,6 +14,17 @@ import TSCUtility
...
+private func executable(_ name: String) -> String {
+#if os(Windows)
+  if name.suffix(from: name.index(name.endIndex, offsetBy: -4)) == ".exe" {
+    return name
+  }
+  return "\(name).exe"
+#else
+  return name
+#endif
+}
```
cltnschlosser, line 26: Could make a WindowsExtensions.swift or something and make this internal so it doesn't need to be duplicated.

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -104,7 +115,7 @@ public protocol Toolchain {
 
 extension Toolchain {
   public var searchPaths: [AbsolutePath] {
-    getEnvSearchPaths(pathString: env["PATH"], currentWorkingDirectory: fileSystem.currentWorkingDirectory)
+    getEnvSearchPaths(pathString: ProcessEnv.path, currentWorkingDirectory: fileSystem.currentWorkingDirectory)
```
cltnschlosser, line 118: Does this break tests or something? Where an (arbitrary) env might have been injected in.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
```diff
@@ -0,0 +1,145 @@
...
+    // variable.
+
+    // TODO(compnerd): replicate the SPM processing of the SDKInfo.plist
+    if let SDKROOT = env["SDKROOT"] {
+      return AbsolutePath(SDKROOT)
+    } else if let DEVELOPER_DIR = env["DEVELOPER_DIR"], let target = target {
+      // FIXME(compnerd) we should capitalise the OS name, e.g. windows ->
+      // Windows; we get away with this for now as Windows has a
+      // case-insensitive file system.
+      let os: String = target.os?.rawValue ?? "none"
```
cltnschlosser, line 124: You can probably capitalize the rawValue of this enum for Windows.

Is "none" a better value than just returning nil here?

cltnschlosser (commented):
For the most part looks good, obviously lots of TODOs/FIXMEs for the future.

BTW is there a good "Beginners guide to Swift on Windows"?

Sources/SwiftDriver/Driver/Driver.swift
compnerd: That works for me!

Sources/SwiftDriver/Toolchains/Toolchain.swift
compnerd: Yeah, it was breaking both tests as well as use of the driver.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
compnerd: Ah, I think I see what you are proposing. I think that the returning `nil` actually could work and be better.  I'll make that change. The capitalization though I don't know how to handle :-(.

Sources/SwiftDriver/Toolchains/Toolchain.swift
artemcm: I'm not convinced this won't potentially break downstream driver clients. `Toolchain.env` comes all the way from a `Driver.init` argument, which defaults to `ProcessEnv.vars`, but can also be invoked with a custom environment when the driver is used as a library.

Can we instead continue using `env` and disambiguate `"Path"` or `"PATH"` based on the platform like [TSC does](https://github.com/apple/swift-tools-support-core/blob/main/Sources/TSCBasic/ProcessEnv.swift#L61)?

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -166,11 +169,27 @@ extension Toolchain {
...
     } else if let path = lookupExecutablePath(filename: executable, searchPaths: searchPaths) {
       return path
-    } else if executable == "swift-frontend" {
+    } else if executable == exe("swift-frontend") {
       // Temporary shim: fall back to looking for "swift" before failing.
-      return try lookup(executable: "swift")
+      return try lookup(executable: exe("swift"))
     } else if fallbackToExecutableDefaultPath {
-      return AbsolutePath("/usr/bin/" + executable)
+      // TODO(compnerd) correct this for Windows - %SystemDrive%\Developer\Toolchains\....xctoolchain\usr\bin
```
artemcm, line 176: I think this comment is now stale?

artemcm (changes requested):
I have just one concern about the Toolchain's use of environment variables, but other than that this is great!

Sources/SwiftDriver/Driver/WindowsExtensions.swift
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+internal func exe(_ name: String) -> String {
```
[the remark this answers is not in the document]
compnerd: I'm fine with `executableName`.  I originally tried `executable` but ran into a conflict.

Sources/SwiftDriver/Toolchains/Toolchain.swift
compnerd: Oh, right, it is.

PR #884, opened 2021-10-21
SwiftDriverTests: conditionalise `exitStatus`

compnerd:
Windows does not have a concept of signals.  As such, we cannot have an
exit status of `signalled`.  This allows the test suite to build, though
this test will currently fail.


compnerd:
@swift-ci please test

cltnschlosser:
Does it make sense to just skip the whole test? `testSignalledBatchMessages`

PR #893, opened 2021-10-28
Improve diagnostics to explain why an identifier isn't a valid

amomchilov:
This is a WIP, just changed the test cases to gauge the interest.

This spawned from my confusion in https://forums.swift.org/t/is-this-a-swiftc-bug-error-module-name-foo-is-not-a-valid-identifier/53119/5

I think we should offer more detail as to what exactly makes an identifier invalid.

Thoughts?

### DIFF
--- a/Sources/SwiftDriver/Utilities/Diagnostics.swift
+++ b/Sources/SwiftDriver/Utilities/Diagnostics.swift
@@ -63,16 +63,17 @@ extension Diagnostic.Message {
 
   static func error_bad_module_name(
     moduleName: String,
-    explicitModuleName: Bool
+    explicitModuleName: Bool,
+		validationFailureReason reason: String
   ) -> Diagnostic.Message {
     let suffix: String
     if explicitModuleName {
       suffix = ""
     } else {
-      suffix = "; use -module-name flag to specify an alternate name"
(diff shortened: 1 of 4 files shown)


artemcm:
@amomchilov, I think this would be a really great addition.

Do you know of which code in the compiler does similar verification that we can be consistent with?

amomchilov:
@artemcm No clue! This is my first time ever making a contribution to the Swift projects. I didn't dive much on the implementation side yet, uncertain if this is something people would ever want

Sources/SwiftDriver/Utilities/Diagnostics.swift
```diff
@@ -63,16 +63,17 @@ extension Diagnostic.Message {
...
     let suffix: String
     if explicitModuleName {
       suffix = ""
     } else {
-      suffix = "; use -module-name flag to specify an alternate name"
+      suffix = " Use -module-name flag to specify an alternate name"
     }
 
-    return .error("module name \"\(moduleName)\" is not a valid identifier\(suffix)")
+    return .error("module name \"\(moduleName)\" is not a valid Swift identifier, because \(reason).\(suffix)")
```
CodaFi, line 76: The existing style of having separate clauses is intentional as it matches clang.

PR #894, opened 2021-10-29
Move the check for the swiftmodule SDK behind an env var

xymus:
Set the env var `ENABLE_RESTRICT_SWIFTMODULE_SDK` to enable passing down the SDK name to the compiler for it to ensure that swiftmodule are only loaded with the SDK used to build them.


xymus:
@swift-ci Please test

PR #895, opened 2021-11-02
Android: make lld the default linker, since the Android NDK just removed all other linkers

buttaface:
This is just bringing over [the one C++ Driver change from a larger compiler pull](https://github.com/apple/swift/pull/39921/files#diff-55fd9483c4660c88c6c3e942cdc38a2cae2647839c581e322c7bf292775e5c43), so I don't have to [change the linker manually all the time](https://github.com/buttaface/swift-android-sdk/blob/main/swift-android-trunk.patch#L112).

### DIFF
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -1585,6 +1585,14 @@ final class SwiftDriverTests: XCTestCase {
       XCTAssertTrue(cmd.contains(.flag("-lotherlib")))
     }
 
+    do {
+      // The Android NDK only uses the lld linker now
+      var driver = try Driver(args: commonArgs + ["-emit-library", "-target", "aarch64-unknown-linux-android24"], env: env)
+      let plannedJobs = try driver.planBuild().removingAutolinkExtractJobs()
+      let lastJob = plannedJobs.last!
+      XCTAssertTrue(lastJob.tool.name.contains("clang"))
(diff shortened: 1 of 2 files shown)


artemcm (approved):
Could you please add a test-case under `testLinking` that confirms this is the executable used for an Android target triple?

artemcm:
@swift-ci please test

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1585,6 +1585,18 @@ final class SwiftDriverTests: XCTestCase {
       XCTAssertTrue(cmd.contains(.flag("-lotherlib")))
     }
 
+    do {
+      // The Android NDK only uses the lld linker now
+      var driver = try Driver(args: commonArgs + ["-emit-library", "-target", "aarch64-unknown-linux-android24"], env: env)
+      let plannedJobs = try driver.planBuild().removingAutolinkExtractJobs()
+      let lastJob = plannedJobs.last!
+      #if os(macOS)
+      XCTAssertTrue(lastJob.tool.name.contains("lld"))
```
buttaface, line 1594: Added a test but I have not tried this on macOS; the other case was tested natively on Android and passes.

buttaface: Removed this Mac-specific line as it failed on the Mac CI, guess it uses this Unix toolchain config instead based on the target triple.

artemcm:
@swift-ci please test

PR #901, opened 2021-11-10
[Explicit Modules] Do not pass module map files or lazy modules

bnbarham:
Explicit module build's pass -fno-implicit-modules and
`-fno-implicit-module-maps` and *all* modules are provided via
`-fmodule-file`. The module files themselves will contain all necessary
information, so there's no reason to load their module map files and
then lazily load the modules.

Remove `-fmodule-map-file` and use `-fmodule-file=<path>` rather than
`-fmodule-file=<name>=<path>` so that each module is eagerly loaded,
avoiding the need to parse the module maps entirely.

Resolves rdar://85238419


Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -288,9 +288,7 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
         TypedVirtualPath(file: moduleArtifactInfo.moduleMapPath.path,
                          type: .clangModuleMap)
       commandLine.appendFlags("-Xcc", "-Xclang", "-Xcc",
-                              "-fmodule-file=\(moduleArtifactInfo.moduleName)=\(clangModulePath.file.description)")
-      commandLine.appendFlags("-Xcc", "-Xclang", "-Xcc",
-                              "-fmodule-map-file=\(clangModuleMapPath.file.description)")
+                              "-fmodule-file=\(clangModulePath.file.description)")
       inputs.append(clangModulePath)
       inputs.append(clangModuleMapPath)
```
bnbarham, line 293: Should the module map file still be an input? I wasn't sure if it was used for incremental builds, though I suppose if the module map *is* updated it would only matter if the resulting module was updated 🤔. Thoughts?

bnbarham:
@swift-ci please test

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
jansvoboda11: I'm not sure how dependencies are modeled in Swift driver, but my guess would be that only specifying the PCM file as input should be enough and removing the modulemap from inputs should be fine.

I assume that the rule that produces `clangModulePath` has `clangModuleMapPath` as an input. If the `clangModuleMapPath` changes, `clangModulePath` will be compiled which should trigger recompile of module identified by `moduleId`, right?

artemcm: If we do not expect the `.modulemap` file to be read by this job, which it sounds like we do not, then it should not be an input, no. 👍🏼

bnbarham: Sounds good, I'll make that change - thanks :)

bnbarham:
@swift-ci please test

TestInputs/ExplicitModuleBuilds/CHeaders/module.modulemap
```diff
@@ -1,34 +0,0 @@
-module A {
```
bnbarham, line 1: It seems like it would be a Good Thing to test that module map files from dependencies aren't being added as an input. Unfortunately that means this has to be split, as otherwise each dependency will have the same module map file - which *will* be an input.

TestInputs/ExplicitModuleBuilds/CHeaders/Bridging.h
```diff
@@ -1,3 +0,0 @@
-#include "BridgingOther.h"
```
bnbarham, line 1: These don't appear to be used and none of the ExplicitModuleBuild tests fail without them 🤷

jansvoboda11 (approved):
LGTM.

bnbarham:
Testing this on some large projects before merging.

bnbarham:
@swift-ci please test

jansvoboda11:
Note that there are some issues with loading PCMs eagerly, so we're currently loading them lazily with Clang.

allevato:
`-fmodule-file=NAME=PATH` is also useful if you need to relocate modules that weren't built with `-fmodule-file-home-is-cwd`, since without the latter flag the PCMs will embed absolute paths to their dependencies.

bnbarham:
> Note that there are some issues with loading PCMs eagerly, so we're currently loading them lazily with Clang.

> -fmodule-file=NAME=PATH is also useful if you need to relocate modules that weren't built with -fmodule-file-home-is-cwd, since without the latter flag the PCMs will embed absolute paths to their dependencies.

Ah, interesting. That's somewhat unfortunate, but hopefully we can revisit this later. Going to close this PR for now, I think a year is long enough :).

PR #903, opened 2021-11-10
Pass-down -downgrade-typecheck-interface-error to interface verification job

nkcsgexi:
The interface verification phase should downgrade errors to warnings as an
intermediate step to stage it in. This won't fail the build if interface errors
occur but it should help users verify project-side fixes.

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -452,6 +452,8 @@ extension Driver {
                               default: false)
     else { return }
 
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

xymus:
Can we pass it down only when the verification is not explicitly requested as discussed earlier today?

nkcsgexi:
@swift-ci please test

nkcsgexi:
@xymus sure, additional check added.

nkcsgexi:
@swift-ci please test

xymus (approved):

xymus:
Thanks for the env var support!

PR #904, opened 2021-11-10
Ensure that the separate emit-module job does not emit `.d.` outputs.

artemcm:
If we have both individual source files and the emit-module file emit .d files, we
are risking collisions in output filenames.

In cases where other compile jobs exist, they will produce dependency outputs already.
There are currently no cases where this is the only job because even an `-emit-module`
driver invocation currently still involves partial compilation jobs.
When partial compilation jobs are removed for the `compilerOutputType == .swiftModule`
case, this will need to be changed here.

Workaround for rdar://85253406


artemcm:
@swift-ci please test

xymus (approved):
Thanks!

artemcm:
@swift-ci please test

PR #906, opened 2021-11-11
Tests: adjust condition for skipped tests

compnerd:
These tests should be skipped with ObjC as a heuristic that the ObjC
runtime enabled Foundation is available.  This is working around a
Foundation issue, and that should be fixed.  This simply makes the
workaround more portable and precise.


compnerd:
@swift-ci please test

cltnschlosser (approved):

artemcm (commented):
The condition makes sense to me, while we figure out the foundation issues.

PR #907, opened 2021-11-11
Tests: drop `testAutolinkOutputPath`

compnerd:
This somehow uses the wrong Toolchain.  Using `GenericUnixToolchain` on
Windows or Darwin is not guaranteed to work.  This is required to enable
the test suite on Windows.

### DIFF
--- a/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
+++ b/Tests/SwiftDriverTests/IncrementalCompilationTests.swift
@@ -135,6 +135,10 @@ extension IncrementalCompilationTests {
   /// autolink job.
   /// Much of the code below is taking from testLinking(), but uses the output file map code here.
   func testAutolinkOutputPath() {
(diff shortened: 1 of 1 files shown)


Tests/SwiftDriverTests/IncrementalCompilationTests.swift
```diff
@@ -135,6 +135,9 @@ extension IncrementalCompilationTests {
   /// autolink job.
   /// Much of the code below is taking from testLinking(), but uses the output file map code here.
   func testAutolinkOutputPath() {
+    XCTSkip("Does not select the host toolchain")
```
artemcm, line 138: Could we please make the skip target-specific for Windows?

artemcm (approved):

compnerd:
Actually, @artemcm was able to help isolate where the problem is getting introduced and we can keep this enabled!

PR #908, opened 2021-11-11
Allow users to specify additional driver flags via env var

nkcsgexi:


nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

artemcm (approved):

jakepetroules:
This seems potentially dangerous due to the hidden nature. Can we have the driver also emit a remark or note diagnostic when this happens, in order to make build logs debuggable and obvious that additional flags are being passed from outside?

PR #912, opened 2021-11-17
Introduce module aliasing to Driver

elsh:
Support -module-alias in driver 
Parse input and store aliases in ModuleOutputInfo
Validate input aliases and emit diagnostics 
Resolves rdar://85523248

PR #913, opened 2021-11-18
Module aliasing: use real module names to resolve graph during dependency scan

elsh:
Use real module names during dependency scan / construct deps graph and imports
Add tests using driver dependency scanner and fallback frontend scanner
Make isFramework optional in SwiftModuleDetails and SwiftPrebuiltExternalModuleDetails (else throws JSON Decoding error)

Resolves rdar://85525296

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
@@ -313,7 +313,7 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
           let isFramework: Bool
           swiftModulePath = .init(file: dependencyInfo.modulePath.path,
                                   type: .swiftModule)
-          isFramework = try dependencyGraph.swiftModuleDetails(of: dependencyId).isFramework
+          isFramework = try dependencyGraph.swiftModuleDetails(of: dependencyId).isFramework ?? false
           // Accumulate the required information about this dependency
           // TODO: add .swiftdoc and .swiftsourceinfo for this module.
           swiftDependencyArtifacts.append(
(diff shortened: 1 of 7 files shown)


elsh:
Depends on https://github.com/apple/swift-driver/pull/912

elsh:
@swift-ci  test

elsh:
@swift-ci test

beccadax (approved):
You should wait for a review from Artem to check that threading a dictionary parameter through is the best way to structure this within the driver (as opposed to making it a property somewhere), but otherwise this LGTM.

artemcm:
The existing changes look good to me, but I think there are a couple of things missing:
- What happens when dependencies are queried on the fallback path of invoking `swift-frontend -scan-dependencies`? 
In places where we get the graph via:
```
dependencyGraph = try self.executor.execute(job: scannerJob, ...
```
You may need to mutate the graph we get using the module information. It would be nice to have a test for that as well. 
- As you mentioned offline, the import-rescan flow needs the same kind of `moduleAliases` param and name resolution logic.

elsh:
Depends on https://github.com/apple/swift/pull/40384

artemcm (approved):

elsh:
> The existing changes look good to me, but I think there are a couple of things missing:
> 
> * What happens when dependencies are queried on the fallback path of invoking `swift-frontend -scan-dependencies`?
>   In places where we get the graph via:
> 
> ```
> dependencyGraph = try self.executor.execute(job: scannerJob, ...
> ```
> 
> You may need to mutate the graph we get using the module information. It would be nice to have a test for that as well.
> 
> * As you mentioned offline, the import-rescan flow needs the same kind of `moduleAliases` param and name resolution logic.

Thanks for the pointers; addressed feedback / tests pending https://github.com/apple/swift/pull/40384

elsh:
@swift-ci test

elsh:
Pending https://github.com/apple/swift/pull/40412

elsh:
@swift-ci test

elsh:
@swift-ci test

elsh:
@swift-ci test

PR #914, opened 2021-11-19
[Incremental Builds] Skip `jobsBeforeCompiles` if no compilation jobs are to be run.

artemcm:
Otherwise, for example, when using the `emit-module-separately` mode, the `emit-module` job is *always* mandatory, even on a null build.

Resolves rdar://85554508

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
@@ -121,14 +121,25 @@ extension IncrementalCompilationState.FirstWaveComputer {
         : compileGroups[input]
     }
 
-    let mandatoryJobsInOrder = try
-      jobsInPhases.beforeCompiles +
-      batchJobFormer.formBatchedJobs(
-        mandatoryCompileGroupsInOrder.flatMap {$0.allJobs()},
-        showJobLifecycle: showJobLifecycle)
+    let batchedCompilationJobs = try batchJobFormer.formBatchedJobs(
+      mandatoryCompileGroupsInOrder.flatMap {$0.allJobs()},
+      showJobLifecycle: showJobLifecycle)
 
+    // In the case where there are no compilation jobs to run on this build (no source-files were changed),
+    // we can skip running `beforeCompiles` jobs if we also ensure that none of the `afterCompiles` jobs
+    // have any dependencies on them.
+    let skipAllJobs = batchedCompilationJobs.isEmpty ? !afterCompileJobsDependOnBeforeCompileJobs() : false
+    let mandatoryJobsInOrder = skipAllJobs ? [] : jobsInPhases.beforeCompiles + batchedCompilationJobs
     return (initiallySkippedCompileGroups: initiallySkippedCompileGroups,
             mandatoryJobsInOrder: mandatoryJobsInOrder)
+  }
+
+  /// Determine if any of the jobs in the `afterCompiles` group depend on outputs produced by jobs in
+  /// `beforeCompiles` group.
+  private func afterCompileJobsDependOnBeforeCompileJobs() -> Bool {
+      let beforeCompileJobOutputs = jobsInPhases.beforeCompiles.reduce(into: Set<TypedVirtualPath>(),
+                                                                       { (pathSet, job) in pathSet.formUnion(job.outputs) })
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
```diff
@@ -121,12 +121,12 @@ extension IncrementalCompilationState.FirstWaveComputer {
...
-      jobsInPhases.beforeCompiles +
-      batchJobFormer.formBatchedJobs(
-        mandatoryCompileGroupsInOrder.flatMap {$0.allJobs()},
-        showJobLifecycle: showJobLifecycle)
+    let batchedCompilationJobs = try batchJobFormer.formBatchedJobs(
+      mandatoryCompileGroupsInOrder.flatMap {$0.allJobs()},
+      showJobLifecycle: showJobLifecycle)
 
+    let mandatoryJobsInOrder = batchedCompilationJobs.isEmpty ? [] : jobsInPhases.beforeCompiles +
+                                                                     batchedCompilationJobs
```
davidungar, line 129: Two thoughts:
1. a comment explaining the "why" would be good
2. I like the use of the ?:
3. What if there are post-compilation jobs? Might the pre-compilation jobs still be needed then? Should there be an assert ensuring no post-compilation jobs? Or a check?
4. I guess that was more than two thoughts.

davidungar (changes requested):
My worry is that there may be post-compile jobs depending upon the pre-compile jobs, even if there are no compilations. Is there some sort of check that could be added somewhere, or if it already exists elsewhere then point to it from the new logic?
Otherwise, it's fine.

artemcm:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
artemcm: Thanks for the comments @davidungar!
Post-compile job dependences is a valid concern that I didn't think of.  

I've added a check to make sure `beforeCompiles` jobs do not produce any such dependencies to make sure, before we decide to not run them.

davidungar: Great, thank you. I'll take a look.

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
```diff
@@ -121,16 +121,29 @@ extension IncrementalCompilationState.FirstWaveComputer {
...
   }
 
+  /// Determine if any of the jobs in the `afterCompiles` group depend on outputs produced by jobs in
+  /// `beforeCompiles` group.
+  private func afterCompileJobsDependOnBeforeCompileJobs() -> Bool {
+    let beforeCompileJobOutputs = jobsInPhases.beforeCompiles.reduce(into: Set<TypedVirtualPath>(),
+                                                                     { (pathSet, job) in pathSet.formUnion(job.outputs) })
+    let afterCompileJobInputs = jobsInPhases.afterCompiles.reduce(into: Set<TypedVirtualPath>(),
+                                                                  { (pathSet, job) in pathSet.formUnion(job.inputs) })
+    return !beforeCompileJobOutputs.intersection(afterCompileJobInputs).isEmpty
```
davidungar, line 144: Looks good. If you need an optimization, you could redo the previous three lines:
```
return jobsInPhases.afterCompiles.contains {postCompileJob in
  postCompileJob.inputs.contains(beforeCompileJobOutputs.contains)}
```
I'm not sure which is the best order for the iteration, but the key idea would be to stop looking as soon as the first intersection is found.
Just something to consider.

davidungar (approved):
Thanks, certainly above threshold now. Two thoughts you could consider:
1. The `afterCompileJobsDependOnBeforeCompileJobs` routine could be written to be faster, at the cost of some clarity. I don't know what the right tradeoff is, but consider projects with lots of .o files.
2. AFAICT, you didn't add a test to ensure that the aforementioned code is working. Again, not critical, IMO. Just a thought.

But it's fine to land as-is.

artemcm:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
artemcm: I like your suggestion and don't think it sacrifices much clarity. Thank you.

PR #916, opened 2021-11-21
Add a reverse index to Multidictionary

kbongort:
Add a reverse index to Multidictionary to make value removal more efficient in large dictionaries.

In a module with several tens of thousands of definitions in the dependency graph, disappearing nodes from the graph becomes very expensive as removeOccurences(of:) is an O(number_of_keys) operation. This change makes removeOccurrences(of:) O(number_of_keys_containing_the_value), which can be tens of thousands of times faster.

The cost of this change is that insertValue(:forKey:) and removeValue(:forKey:) are approximately twice as expensive, and the dictionary will use about twice as much memory. For small graphs, this difference should be trivial. For my several-tens-of-thousands-of-definitions graph, it takes about four seconds to load rather than two. The driver doesn't seem to use significantly more memory overall. Without this change removeOccurences(of:) is so exceedingly expensive (about 0.5 seconds per call) that the increase in load time is well worth it.

This change could be further optimized so that the `reverseIndex` is only built the first time removeOccurences(of:) is called. Then there would be no negative performance impact at all for many builds. However, the negative impact seems to be quite small, so I am not sure the extra complexity of such an optimization is worthwhile.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
@@ -141,11 +141,6 @@
     if self.reverseIndex[v] != nil {
       return false
     }
-    for vals in self.forwardDictionary.values {
-      if vals.contains(v) {
-        return false
-      }
-    }
-    return true
+    return !self.forwardDictionary.values.contains(where: { $0.contains(v) })
   }
 }
--- a/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
@@ -89,7 +89,7 @@
   ///                this dictionary.
   /// - Returns: The set of keys associated with the given value.
   public func keysContainingValue(_ v: Value) -> Set<Key> {
-    return self.reverseIndex[v, default: []]
+    return self.reverseIndex[v] ?? []
   }
 
   /// Inserts the given value in the set of values associated with the given key.
--- a/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
@@ -20,6 +20,8 @@
   public typealias Element = (Key, Set<Value>)
   public typealias Index = Dictionary<Key, Set<Value>>.Index
 
+  public init() {}
+
   /// The number of key-value pairs in this multi-dictionary.
   public var count: Int {
     self.dictionary.count
@@ -123,9 +125,22 @@
   ///
   /// - Parameter v: The value to remove.
   public mutating func removeOccurrences(of v: Value) {
-    for k in self.reverseIndex[v, default: []] {
+    for k in self.reverseIndex.removeValue(forKey: v) ?? [] {
       self.dictionary[k]!.remove(v)
     }
-    self.reverseIndex.removeValue(forKey: v)
+    assert(expensivelyCheckThatValueIsRemoved(v))
+  }
+
+  /// For assertions. Returns true `v` is removed from all `dictionary` values.
+  private func expensivelyCheckThatValueIsRemoved(_ v: Value) -> Bool {
+    if self.reverseIndex[v] != nil {
+      return false
+    }
+    for vals in self.dictionary.values {
+      if vals.contains(v) {
+        return false
+      }
+    }
+    return true
   }
 }
--- a/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
@@ -136,7 +136,7 @@
     assert(expensivelyCheckThatValueIsRemoved(v))
   }
 
-  /// For assertions. Returns true `v` is removed from all `dictionary` values.
+  /// For assertions. Returns true `v` is removed from all `forwardDictionary` values.
   private func expensivelyCheckThatValueIsRemoved(_ v: Value) -> Bool {
     if self.reverseIndex[v] != nil {
       return false
--- a/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
@@ -13,6 +13,9 @@
 /// A collection that associates keys with one or more values.
 @_spi(Testing) public struct Multidictionary<Key: Hashable, Value: Hashable>: Collection, Equatable {
   private var dictionary: Dictionary<Key, Set<Value>> = [:]
+
+  /// Reverse index used to make value removal more efficient.
+  private var reverseIndex: Dictionary<Value, Set<Key>> = [:]
   
   public typealias Element = (Key, Set<Value>)
   public typealias Index = Dictionary<Key, Set<Value>>.Index
@@ -84,12 +87,7 @@
   ///                this dictionary.
   /// - Returns: The set of keys associated with the given value.
   public func keysContainingValue(_ v: Value) -> Set<Key> {
-    return self.dictionary.reduce(into: Set<Key>()) { acc, next in
-      guard next.value.contains(v) else {
-        return
-      }
-      acc.insert(next.key)
-    }
+    return self.reverseIndex[v, default: []]
   }
 
   /// Inserts the given value in the set of values associated with the given key.
@@ -101,6 +99,7 @@
   ///            other values for the given key. Else, returns `false.
   @discardableResult
   public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
+    self.reverseIndex[v, default: []].insert(key)
     return self.dictionary[key, default: []].insert(v).inserted
   }
 
@@ -112,6 +111,7 @@
   /// - Returns: The removed element, if any.
   @discardableResult
   public mutating func removeValue(_ v: Value, forKey key: Key) -> Value? {
+    self.reverseIndex[v]?.remove(key)
     return self.dictionary[key, default: []].remove(v)
   }
 
@@ -123,8 +123,9 @@
   ///
   /// - Parameter v: The value to remove.
   public mutating func removeOccurrences(of v: Value) {
-    for k in self.dictionary.keys {
+    for k in self.reverseIndex[v, default: []] {
       self.dictionary[k]!.remove(v)
     }
+    self.reverseIndex.removeValue(forKey: v)
   }
 }
--- a/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
@@ -101,8 +101,10 @@
   ///            other values for the given key. Else, returns `false.
   @discardableResult
   public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
-    self.reverseIndex[v, default: []].insert(key)
-    return self.dictionary[key, default: []].insert(v).inserted
+    let inserted1 = self.reverseIndex[v, default: []].insert(key).inserted
+    let inserted2 = self.dictionary[key, default: []].insert(v).inserted
+    assert(inserted1 == inserted2)
+    return inserted1
   }
 
   /// Removes the given value from the set of values associated with the given key.
@@ -113,8 +115,11 @@
   /// - Returns: The removed element, if any.
   @discardableResult
   public mutating func removeValue(_ v: Value, forKey key: Key) -> Value? {
-    self.reverseIndex[v]?.remove(key)
-    return self.dictionary[key, default: []].remove(v)
+    let removedKey = self.reverseIndex[v]?.remove(key)
+    let removedVal = self.dictionary[key, default: []].remove(v)
+    assert((removedKey == nil && removedVal == nil) ||
+           (removedKey != nil && removedVal != nil))
+    return removedVal
   }
 
   /// Removes all occurrences of the given value from all entries in this
--- a/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
@@ -116,7 +116,7 @@
   @discardableResult
   public mutating func removeValue(_ v: Value, forKey key: Key) -> Value? {
     let removedKey = self.reverseIndex[v]?.remove(key)
-    let removedVal = self.dictionary[key, default: []].remove(v)
+    let removedVal = self.dictionary[key]?.remove(v)
     assert((removedKey == nil && removedVal == nil) ||
            (removedKey != nil && removedVal != nil))
     return removedVal
--- a/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
@@ -12,7 +12,7 @@
 
 /// A collection that associates keys with one or more values.
 @_spi(Testing) public struct Multidictionary<Key: Hashable, Value: Hashable>: Collection, Equatable {
-  private var dictionary: Dictionary<Key, Set<Value>> = [:]
+  private var forwardDictionary: Dictionary<Key, Set<Value>> = [:]
 
   /// Reverse index used to make value removal more efficient.
   private var reverseIndex: Dictionary<Value, Set<Key>> = [:]
@@ -24,18 +24,18 @@
 
   /// The number of key-value pairs in this multi-dictionary.
   public var count: Int {
-    self.dictionary.count
+    self.forwardDictionary.count
   }
 
   /// The position of the first element in a nonempty multi-dictionary.
   public var startIndex: Index {
-    self.dictionary.startIndex
+    self.forwardDictionary.startIndex
   }
 
   /// The collection’s “past the end” position—that is, the position one greater
   /// than the last valid subscript argument.
   public var endIndex: Index {
-    self.dictionary.endIndex
+    self.forwardDictionary.endIndex
   }
 
   /// Returns the index for the given key.
@@ -44,7 +44,7 @@
   /// - Returns: The index for key and its associated value if key is in the
   ///            multi-dictionary; otherwise, nil.
   public func index(forKey key: Key) -> Dictionary<Key, Set<Value>>.Index? {
-    self.dictionary.index(forKey: key)
+    self.forwardDictionary.index(forKey: key)
   }
 
   /// Computes the position immediately after the given index.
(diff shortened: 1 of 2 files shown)


kbongort:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
```diff
@@ -123,8 +123,9 @@
   ///
   /// - Parameter v: The value to remove.
   public mutating func removeOccurrences(of v: Value) {
-    for k in self.dictionary.keys {
+    for k in self.reverseIndex[v, default: []] {
       self.dictionary[k]!.remove(v)
     }
+    self.reverseIndex.removeValue(forKey: v)
```
davidungar, line 129: IIIRC, `removeValue` returns the old value if any. So this function only needs one probe of the `reverseIndex` dictionary.

davidungar (changes requested):
Nicely done! But, there is now an invariant requiring the `reverseIndex` to be consistent with the forward one. I'm worried about this invariant being completely tacit and unchecked. How about something like adding an assertion on every (forward or backward) probe, checking the consistency?
The code is fairly short and clear now, but over time it may not be.

kbongort:
> Nicely done! But, there is now an invariant requiring the `reverseIndex` to be consistent with the forward one. I'm worried about this invariant being completely tacit and unchecked. How about something like adding an assertion on every (forward or backward) probe, checking the consistency? The code is fairly short and clear now, but over time it may not be.

Can I make a counter-offer of unit tests for this data structure? I think tests would be preferable to asserts, since they'd aid future developers and add no run-time cost (if only for development builds).

davidungar:
> > Nicely done! But, there is now an invariant requiring the `reverseIndex` to be consistent with the forward one. I'm worried about this invariant being completely tacit and unchecked. How about something like adding an assertion on every (forward or backward) probe, checking the consistency? The code is fairly short and clear now, but over time it may not be.
> 
> Can I make a counter-offer of unit tests for this data structure? I think tests would be preferable to asserts, since they'd aid future developers and add no run-time cost (if only for development builds).

Interesting thought, but I'm not ready to accept it at this time. Here's why:
1. Unit tests do not make the invariant explicit in the code
2. I suspect that adding an assert can be done with minimal overhead. Either one Dictionary-of-Sets probe in each direction after modification, or a verify pass at the end of the run. I'd feel better going that route.
3. A pitfall of unit tests--which I've fallen into myself from time to time--is that by testing specific inputs, a unit test can miss cases an assert would catch.

Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
```diff
@@ -101,6 +99,7 @@
   ///            other values for the given key. Else, returns `false.
   @discardableResult
   public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
+    self.reverseIndex[v, default: []].insert(key)
```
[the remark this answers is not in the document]
kbongort: Ack - yes I'm surprised too! Makes for very nice syntax in the tests.

kbongort:
> > > Nicely done! But, there is now an invariant requiring the `reverseIndex` to be consistent with the forward one. I'm worried about this invariant being completely tacit and unchecked. How about something like adding an assertion on every (forward or backward) probe, checking the consistency? The code is fairly short and clear now, but over time it may not be.
> > 
> > 
> > Can I make a counter-offer of unit tests for this data structure? I think tests would be preferable to asserts, since they'd aid future developers and add no run-time cost (if only for development builds).
> 
> Interesting thought, but I'm not ready to accept it at this time. Here's why:
> 
> 1. Unit tests do not make the invariant explicit in the code
> 2. I suspect that adding an assert can be done with minimal overhead. Either one Dictionary-of-Sets probe in each direction after modification, or a verify pass at the end of the run. I'd feel better going that route.
> 3. A pitfall of unit tests--which I've fallen into myself from time to time--is that by testing specific inputs, a unit test can miss cases an assert would catch.

Okay, I added an assert to check that removeOccurrences(of:) removes all occurrences from self.dictionary. I'm having trouble translating your direction from (2) into code that makes sense. I thought about the following:
```
  public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
    self.reverseIndex[v, default: []].insert(key)
    let inserted = self.dictionary[key, default: []].insert(v).inserted
    assert(self.reverseIndex[v]?.contains(key) == true)
    assert(self.dictionary[key]?.contains(v) == true)
    return inserted
  }
```
But it seems silly to assert that the inserts or removals worked as expected. Let me know what you think.

Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
kbongort: Done

davidungar:
> > > > Nicely done! But, there is now an invariant requiring the `reverseIndex` to be consistent with the forward one. I'm worried about this invariant being completely tacit and unchecked. How about something like adding an assertion on every (forward or backward) probe, checking the consistency? The code is fairly short and clear now, but over time it may not be.
> > > 
> > > 
> > > Can I make a counter-offer of unit tests for this data structure? I think tests would be preferable to asserts, since they'd aid future developers and add no run-time cost (if only for development builds).
> > 
> > 
> > Interesting thought, but I'm not ready to accept it at this time. Here's why:
> > 
> > 1. Unit tests do not make the invariant explicit in the code
> > 2. I suspect that adding an assert can be done with minimal overhead. Either one Dictionary-of-Sets probe in each direction after modification, or a verify pass at the end of the run. I'd feel better going that route.
> > 3. A pitfall of unit tests--which I've fallen into myself from time to time--is that by testing specific inputs, a unit test can miss cases an assert would catch.
> 
> Okay, I added an assert to check that removeOccurrences(of:) removes all occurrences from self.dictionary. I'm having trouble translating your direction from (2) into code that makes sense. I thought about the following:
> 
> ```
>   public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
>     self.reverseIndex[v, default: []].insert(key)
>     let inserted = self.dictionary[key, default: []].insert(v).inserted
>     assert(self.reverseIndex[v]?.contains(key) == true)
>     assert(self.dictionary[key]?.contains(v) == true)
>     return inserted
>   }
> ```
> 
> But it seems silly to assert that the inserts or removals worked as expected. Let me know what you think.

Thank you. Sorry about being unclear. I'll take a look and see if I can clarify.

davidungar:
> > > > > Nicely done! But, there is now an invariant requiring the `reverseIndex` to be consistent with the forward one. I'm worried about this invariant being completely tacit and unchecked. How about something like adding an assertion on every (forward or backward) probe, checking the consistency? The code is fairly short and clear now, but over time it may not be.
> > > > 
> > > > 
> > > > Can I make a counter-offer of unit tests for this data structure? I think tests would be preferable to asserts, since they'd aid future developers and add no run-time cost (if only for development builds).
> > > 
> > > 
> > > Interesting thought, but I'm not ready to accept it at this time. Here's why:
> > > 
> > > 1. Unit tests do not make the invariant explicit in the code
> > > 2. I suspect that adding an assert can be done with minimal overhead. Either one Dictionary-of-Sets probe in each direction after modification, or a verify pass at the end of the run. I'd feel better going that route.
> > > 3. A pitfall of unit tests--which I've fallen into myself from time to time--is that by testing specific inputs, a unit test can miss cases an assert would catch.
> > 
> > 
> > Okay, I added an assert to check that removeOccurrences(of:) removes all occurrences from self.dictionary. I'm having trouble translating your direction from (2) into code that makes sense. I thought about the following:
> > ```
> >   public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
> >     self.reverseIndex[v, default: []].insert(key)
> >     let inserted = self.dictionary[key, default: []].insert(v).inserted
> >     assert(self.reverseIndex[v]?.contains(key) == true)
> >     assert(self.dictionary[key]?.contains(v) == true)
> >     return inserted
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But it seems silly to assert that the inserts or removals worked as expected. Let me know what you think.
> 
> Thank you. Sorry about being unclear. I'll take a look and see if I can clarify.

OK, after spending some time staring at this, here are my thoughts:
First, I would suggest renaming the member `dictionary` to `forwardDictionary` for the sake of clarity.
Then, instead of the assertion you added, I think it would suffice to just check consistency after every mutation.
That means adding an assertion to the end of `insertValue` and another to the end of `removeValue`.
Here's a sketch, I haven't actually tried these:: ```
 public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
    let reverseInserted = self.reverseIndex[v, default: []].insert(key)
    let forwardInserted = self.dictionary[key, default: []].insert(v).inserted
    assert(forwardInserted = reverseInstered)
    return forwardInserted
  }
```
Removal is a bit trickier: ```
  public mutating func removeValue(_ v: Value, forKey key: Key) -> Value? {
    let reverseOldKey = self.reverseIndex[v]?.remove(key)
    let forwardOldValue = self.dictionary[key, default: []].remove(v)
    assert(isConsistentAfterRemoval(key: key, value: v, oldKey: reverseOldKey, oldValue: reverseOldValue))
    return forwardOldValue
  }
  
private func isConsistentAfterRemoval(key: Key, value: Value, oldKey: Key?, oldValue: Value?))
    switch (oldKey, oldValue) {
      case (nil, nil): return true
      case let (oldKey?, oldValue?) where oldKey == key, oldValue == value: return true
      default: return false
    }
  }
```
Please take a look and see what you think. Thanks!

kbongort:
> > > > > > Nicely done! But, there is now an invariant requiring the `reverseIndex` to be consistent with the forward one. I'm worried about this invariant being completely tacit and unchecked. How about something like adding an assertion on every (forward or backward) probe, checking the consistency? The code is fairly short and clear now, but over time it may not be.
> > > > > 
> > > > > 
> > > > > Can I make a counter-offer of unit tests for this data structure? I think tests would be preferable to asserts, since they'd aid future developers and add no run-time cost (if only for development builds).
> > > > 
> > > > 
> > > > Interesting thought, but I'm not ready to accept it at this time. Here's why:
> > > > 
> > > > 1. Unit tests do not make the invariant explicit in the code
> > > > 2. I suspect that adding an assert can be done with minimal overhead. Either one Dictionary-of-Sets probe in each direction after modification, or a verify pass at the end of the run. I'd feel better going that route.
> > > > 3. A pitfall of unit tests--which I've fallen into myself from time to time--is that by testing specific inputs, a unit test can miss cases an assert would catch.
> > > 
> > > 
> > > Okay, I added an assert to check that removeOccurrences(of:) removes all occurrences from self.dictionary. I'm having trouble translating your direction from (2) into code that makes sense. I thought about the following:
> > > ```
> > >   public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
> > >     self.reverseIndex[v, default: []].insert(key)
> > >     let inserted = self.dictionary[key, default: []].insert(v).inserted
> > >     assert(self.reverseIndex[v]?.contains(key) == true)
> > >     assert(self.dictionary[key]?.contains(v) == true)
> > >     return inserted
> > >   }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > But it seems silly to assert that the inserts or removals worked as expected. Let me know what you think.
> > 
> > 
> > Thank you. Sorry about being unclear. I'll take a look and see if I can clarify.
> 
> OK, after spending some time staring at this, here are my thoughts: First, I would suggest renaming the member `dictionary` to `forwardDictionary` for the sake of clarity. Then, instead of the assertion you added, I think it would suffice to just check consistency after every mutation. That means adding an assertion to the end of `insertValue` and another to the end of `removeValue`. Here's a sketch, I haven't actually tried these:
> 
> ```
>  public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
>     let reverseInserted = self.reverseIndex[v, default: []].insert(key)
>     let forwardInserted = self.dictionary[key, default: []].insert(v).inserted
>     assert(forwardInserted = reverseInstered)
>     return forwardInserted
>   }
> ```
> 
> Removal is a bit trickier:
> 
> ```
>   public mutating func removeValue(_ v: Value, forKey key: Key) -> Value? {
>     let reverseOldKey = self.reverseIndex[v]?.remove(key)
>     let forwardOldValue = self.dictionary[key, default: []].remove(v)
>     assert(isConsistentAfterRemoval(key: key, value: v, oldKey: reverseOldKey, oldValue: reverseOldValue))
>     return forwardOldValue
>   }
>   
> private func isConsistentAfterRemoval(key: Key, value: Value, oldKey: Key?, oldValue: Value?))
>     switch (oldKey, oldValue) {
>       case (nil, nil): return true
>       case let (oldKey?, oldValue?) where oldKey == key, oldValue == value: return true
>       default: return false
>     }
>   }
> ```
> 
> Please take a look and see what you think. Thanks!

Well, I'm still not sold on having assertions at all, but if we must: I don't think removing the assertion that I added is a good idea, because it covers the most complex mutation that can happen. For the post-removal assertion, I'm not sure it makes sense to do equality checks between the removed value and the value-to-be-removed. That's essentially testing the Set API. So I think we could pare it down to simply check that both removed values are nil or both are non-nil.

Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
```diff
@@ -123,8 +130,22 @@
...
+    }
+    assert(expensivelyCheckThatValueIsRemoved(v))
+  }
+
+  /// For assertions. Returns true `v` is removed from all `forwardDictionary` values.
+  private func expensivelyCheckThatValueIsRemoved(_ v: Value) -> Bool {
+    if self.reverseIndex[v] != nil {
+      return false
+    }
+    for vals in self.forwardDictionary.values {
```
davidungar, line 144: Not a big deal, but if you would like a more concise version, you could code it this way:
```
return !self.forwardDictionaryValues.contains {$0.contains(v)}
```

davidungar (approved):
Looks good!

davidungar:
> > > > > > > Nicely done! But, there is now an invariant requiring the `reverseIndex` to be consistent with the forward one. I'm worried about this invariant being completely tacit and unchecked. How about something like adding an assertion on every (forward or backward) probe, checking the consistency? The code is fairly short and clear now, but over time it may not be.
> > > > > > 
> > > > > > 
> > > > > > Can I make a counter-offer of unit tests for this data structure? I think tests would be preferable to asserts, since they'd aid future developers and add no run-time cost (if only for development builds).
> > > > > 
> > > > > 
> > > > > Interesting thought, but I'm not ready to accept it at this time. Here's why:
> > > > > 
> > > > > 1. Unit tests do not make the invariant explicit in the code
> > > > > 2. I suspect that adding an assert can be done with minimal overhead. Either one Dictionary-of-Sets probe in each direction after modification, or a verify pass at the end of the run. I'd feel better going that route.
> > > > > 3. A pitfall of unit tests--which I've fallen into myself from time to time--is that by testing specific inputs, a unit test can miss cases an assert would catch.
> > > > 
> > > > 
> > > > Okay, I added an assert to check that removeOccurrences(of:) removes all occurrences from self.dictionary. I'm having trouble translating your direction from (2) into code that makes sense. I thought about the following:
> > > > ```
> > > >   public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
> > > >     self.reverseIndex[v, default: []].insert(key)
> > > >     let inserted = self.dictionary[key, default: []].insert(v).inserted
> > > >     assert(self.reverseIndex[v]?.contains(key) == true)
> > > >     assert(self.dictionary[key]?.contains(v) == true)
> > > >     return inserted
> > > >   }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > But it seems silly to assert that the inserts or removals worked as expected. Let me know what you think.
> > > 
> > > 
> > > Thank you. Sorry about being unclear. I'll take a look and see if I can clarify.
> > 
> > 
> > OK, after spending some time staring at this, here are my thoughts: First, I would suggest renaming the member `dictionary` to `forwardDictionary` for the sake of clarity. Then, instead of the assertion you added, I think it would suffice to just check consistency after every mutation. That means adding an assertion to the end of `insertValue` and another to the end of `removeValue`. Here's a sketch, I haven't actually tried these:
> > ```
> >  public mutating func insertValue(_ v: Value, forKey key: Key) -> Bool {
> >     let reverseInserted = self.reverseIndex[v, default: []].insert(key)
> >     let forwardInserted = self.dictionary[key, default: []].insert(v).inserted
> >     assert(forwardInserted = reverseInstered)
> >     return forwardInserted
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Removal is a bit trickier:
> > ```
> >   public mutating func removeValue(_ v: Value, forKey key: Key) -> Value? {
> >     let reverseOldKey = self.reverseIndex[v]?.remove(key)
> >     let forwardOldValue = self.dictionary[key, default: []].remove(v)
> >     assert(isConsistentAfterRemoval(key: key, value: v, oldKey: reverseOldKey, oldValue: reverseOldValue))
> >     return forwardOldValue
> >   }
> >   
> > private func isConsistentAfterRemoval(key: Key, value: Value, oldKey: Key?, oldValue: Value?))
> >     switch (oldKey, oldValue) {
> >       case (nil, nil): return true
> >       case let (oldKey?, oldValue?) where oldKey == key, oldValue == value: return true
> >       default: return false
> >     }
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please take a look and see what you think. Thanks!
> 
> Thank you for illustrating your suggestion!
> 
> While I'm still not sold on having assertions at all, if we must: I don't think removing the assertion that I added is a good idea, because it covers the most complex mutation that can happen. For the post-removal assertion, I'm not sure it makes sense to do equality checks between the removed value and the value-to-be-removed. That's essentially testing the Set API. So I think we could pare it down to simply check that both removed values are nil or both are non-nil.
> 
> I've updated the PR accordingly.

Yes, I agree with both leaving the longer assertion in and the nil/non-nil check on the other assertion. 
Thanks for your patience!

Sources/SwiftDriver/IncrementalCompilation/Multidictionary.swift
kbongort: Thanks for the suggestion, done.

kbongort:
Thank you for the review David! This is my first PR to the project so I'm not familiar with the process – please let me know if there is any further action I need to take.

cltnschlosser:
@swift-ci please test

kbongort:
<friendly bump> Can we merge this now?

davidungar:
> Can we merge this now?

Fine by me. Meant to imply that by my approval. Go for it!

kbongort:
I don't think I have the appropriate level of permission to do it myself.

kbongort:
(nudge) I don't seem to be able to merge this myself, so if someone else could do so that would be great – thanks!

davidungar:
Done!

kbongort:
Thank you! 🥳

PR #917, opened 2021-12-01
Enable emitting module files separately in whole-module builds

xymus:
When passed the flag `-emit-module-separately-wmo`, schedule a separate optimized emit-module job to output the module files even when building in whole-module mode (`-wmo`). In this mode, the classic wmo job generates only the object file and the emit-module job generates the module files concurrently. With enough available cores, this will generate the files needed to unblock building dependencies early, while the classic wmo job is still building the binary file.

The flag should only be passed down to the compiler when the target has dependencies that will be built in the same run, otherwise if there is no dependencies, this mode mostly duplicates work and doesn't provide any gain.

rdar://85428222


xymus:
@swift-ci Please smoke test

xymus:
@swift-ci Please test

artemcm (approved):
This is great! Excited to see what kinds of improvements this results in.

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -226,6 +226,12 @@ extension Driver {
     // Ensure that only one job emits the module files and insert a verify swiftinterface job
     var jobCreatingSwiftModule: Job? = nil
     func addPostModuleFilesJobs(_ emitModuleJob: Job) throws {
+      let emitsSwiftInterface =
+        emitModuleJob.outputs.contains(where: { out in out.type == .swiftInterface })
+      if !emitsSwiftInterface {
```
cltnschlosser, line 231: Nit: guard

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -613,6 +614,13 @@ public struct Driver {
       debugInfoLevel: debugInfo.level, diagnosticsEngine: diagnosticEngine,
       workingDirectory: self.workingDirectory)
 
+    // Should we schedule a separate emit-module job?
+    self.emitModuleSeparately = Self.computeEmitModuleSeparately(parsedOptions: &parsedOptions,
+                                                                compilerMode: compilerMode,
```
cltnschlosser, line 619: Indentation

cltnschlosser (approved):
Looks cool. Is the precursor to this a main factor in the multi core improvements in the latest Xcode beta?

xymus:
Fixed the duplicated emit of ABI descriptor files and style issues.

@swift-ci Please test

xymus:
@swift-ci Please test

PR #919, opened 2021-12-01
Tests: make `ExplicitModuleBuildTests.*` pass on Windows

compnerd:
This fixes the ExplicitmoduleBuildTests to pass on Windows.  The
original port to Windows had setup the stdlib paths incorrectly.
Unfortunately, that is a larger change that needs to wait until other
stuff is complete.  Namely, we currently structure the SDK as:

~~~
SDKROOT/usr/lib/swift/[OS]/[Arch]/[Module].swiftmodule/[Triple].swift{module,doc,info}
~~~

The directory form of the swiftmodule was adopted to allow
multi-architecture variants of the modules and does allow the
intermediate architecture directory to be removed.  The architecture
specific directory would still contain the import libraries (TBD-like
content) which is not currently multi-architecture friendly (due to
tooling).  So while the directory needs to remain, the swiftmodules
could be hoisted a level.  This would need to be coordinated along the
driver, the frontend, and the installer.  So it is best to hold off on
this change for the time being.

Correcting the path computation and then wiring that along to the tests
allows the ExplicitModuleBuildTests to now pass on Windows.

Special thanks to @artemcm for pointing out that we weren't including
those paths in the search.

### DIFF
--- a/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
+++ b/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
@@ -652,7 +652,15 @@ final class ExplicitModuleBuildTests: XCTestCase {
         return (sdkroot
                   .appending(component: "usr")
                   .appending(component: "lib")
-                  .appending(component: "swift"),
+                  .appending(component: "swift")
+                  .appending(component: "windows")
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

artemcm (approved):
Awesome, thank you!

PR #921, opened 2021-12-02
Tests: account for EOL marker differences

compnerd:
Windows uses a different EOL marker (CR+LF) compared to Unix (LF).  This
accounts for that in the output enabling the
HideAndShowFuncInStrucAndExtensionTest to now pass on Windows.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #922, opened 2021-12-02
Turn on emit-module-separately for whole-module builds by default

xymus:
Scheduling a separate emit-module job for whole-module builds, combined with the integrated driver in the build system, will allow to parallelize what used to be essentially sequential builds. This mode unblocks building dependents early by duplicating some work in two processes, it requires more cpu work for a faster wall clock build time. Another advantage of this mode is the resulting swiftmodule is lighter from a classic whole-module build.

rdar://85976762

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2102,10 +2102,9 @@ extension Driver {
       // An option has been passed which requires a module, but the user hasn't
       // requested one. Generate a module, but treat it as an intermediate output.
       moduleOutputKind = .auxiliary
-    } else if compilerMode != .singleCompile &&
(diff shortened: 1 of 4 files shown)


xymus:
I put this up as is to run tests. I expect this to break a lot of tests, here and in the driver.

@swift-ci Please test

xymus:
@swift-ci Please test

xymus:
@swift-ci Please test

xymus:
More tests are failing this time, it's a good sign. The first version scheduled the separate emit-module job only when the `-emit-module` flag was passed explicitly, not for other module-wide outputs.

xymus:
@swift-ci Please test

xymus:
@swift-ci Please test

xymus:
The tests look good here and on the compiler side. I think this is ready for review!

nkcsgexi (approved):

artemcm:
:ship:

PR #923, opened 2021-12-03
Tests: make ParsableMessageTests pass on Windows

compnerd:
This adjusts the test suite to account for the differences on Windows.

We must construct the absolute path to the working directory, which
includes the drive letter.  Furthermore, construct the paths for the
checks as relative to the working directory.  This makes the test more
explicit about it's expectations and makes it support Windows.

Windows does not support signals, even for unexpected termination.
Instead the process exits with a particular error code which encodes the
"exception".  This adjusts the test expectation accordingly.

Adjust the diagnostic expectations which do not match due to the EOL
specifier (CR+LF vs LF).

The ParsableMessageTests now pass on Windows.

### DIFF
--- a/Tests/SwiftDriverTests/ParsableMessageTests.swift
+++ b/Tests/SwiftDriverTests/ParsableMessageTests.swift
@@ -12,6 +12,7 @@
 import XCTest
 import TSCBasic
 import TSCUtility
+import Foundation
 
 @_spi(Testing) import SwiftDriver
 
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

artemcm (approved):
Awesome!

compnerd:
@swift-ci please test

compnerd:
This is slightly annoying as the `escaped` does not touch the path since it is in use in different places.  This might be the cleanest approach.

PR #924, opened 2021-12-06
Toolchains: adjust environment variable generation on Windows

compnerd:
Avoid injecting a `_EXE` into environment variables for tool
replacements on Windows.  This allows us to use the same variable across
all the platforms.


compnerd:
@swift-ci please test

cltnschlosser (approved):

artemcm (approved):

PR #925, opened 2021-12-06
Driver: be more careful about extension handling

compnerd:
We would previously try to strip the extension without care about the
length.  This can potentially cause an out-of-bounds access (which
fortunately an assert helped caught this!).  Add a guard to avoid an
invalid access.  This is important to enable a subsequent change that
makes finding tools more flexible for cross-platform testing.


compnerd:
@swift-ci please test

artemcm (approved):
Thank you.

cltnschlosser (approved):

PR #926, opened 2021-12-06
Fix build break with an ambiguous contextual base on older compilers

artemcm:
Slightly older compilers may error out here with:
```
error: cannot infer contextual base in reference to member 'replacingOccurrences'
           .replacingOccurrences(of: "-", with: "_")
           ~^~~~~~~~~~~~~~~~~~~~
```

Resolves rdar://86127658


artemcm:
@swift-ci please test

compnerd (approved):
Oops!  Didn't know that it was still being built with older compilers.  Thanks for fixing this.

PR #927, opened 2021-12-07
Toolchain: be more aggressive about adding extensions

compnerd:
This more aggressively adds the `.exe` extension to the tools on
Windows.  This allows a greater number of tools to be found on Windows
as the proper name is now used.


compnerd:
@swift-ci please test

cltnschlosser:
If lookup is called with `swift.exe` will this look for `swift.exe.exe`? Or does executableName make sure not to do that?

compnerd:
> If lookup is called with `swift.exe` will this look for `swift.exe.exe`? Or does executableName make sure not to do that?

Nope, it will look for `swift.exe`: https://github.com/apple/swift-driver/blob/main/Sources/SwiftDriver/Driver/WindowsExtensions.swift#L17-L20

cltnschlosser (approved):

PR #928, opened 2021-12-07
Stop launching REPL when there aren't inputs

bitjammer:
Now that there is a more user-friendly help intro that shows by
default, stop launching the REPL when there are no input files.
Launch the REPL explicitly with the `repl` subcommand.


bitjammer:
@swift-ci Please test

artemcm (approved):

PR #931, opened 2021-12-14
Respect -use-ld= on Darwin

benlangmuir:
Use the custom ld path rather than the system 'ld'.

### DIFF
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -1244,6 +1244,19 @@ final class SwiftDriverTests: XCTestCase {
       XCTAssertFalse(cmd.contains(.flag("-shared")))
     }
 
+    do {
+      // -use-ld
+      var driver = try Driver(args: commonArgs + ["-use-ld=/bin/my-ld"], env: env)
+      let plannedJobs = try driver.planBuild()
+
+      XCTAssertEqual(3, plannedJobs.count)
+      XCTAssertFalse(plannedJobs.contains { $0.kind == .autolinkExtract })
+
+      let linkJob = plannedJobs[2]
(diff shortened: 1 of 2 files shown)


benlangmuir:
@swift-ci please test

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1244,6 +1244,19 @@ final class SwiftDriverTests: XCTestCase {
...
+      // -use-ld
+      var driver = try Driver(args: commonArgs + ["-use-ld=/bin/my-ld"], env: env)
+      let plannedJobs = try driver.planBuild()
+
+      XCTAssertEqual(3, plannedJobs.count)
+      XCTAssertFalse(plannedJobs.contains { $0.kind == .autolinkExtract })
+
+      let linkJob = plannedJobs[2]
+      XCTAssertEqual(linkJob.kind, .link)
+      XCTAssertEqual(linkJob.tool.name, "/bin/my-ld")
```
compnerd, line 1257: Could you please add a test case for the suffix form as well?

benlangmuir: Sorry, what is the "suffix form" in this context?

compnerd: `-use-ld=lld` should use `ld64.lld`.

benlangmuir: @compnerd  Ohh, thanks. I didn't realize that `-[f]use-ld` was designed for choosing a "variant" like this and that the version taking a path was deprecated.  It turns out the thing I wanted is spelled `--ld-path` in clang, and it doesn't exist at all in swiftc.  I'll just close this for now.

PR #933, opened 2021-12-16
Driver: wire through the actual error from OFM loading

compnerd:
The deserialisation process may fail for reasons other than unable to
load the file.  This was occurring on Windows and was giving an
erroneous error message that the file could not be loaded when the issue
was the contents of the file.  This threads that error through the
wrapped error type rather than just dropping the error on the floor.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -27,7 +27,7 @@ public struct Driver {
     case integratedReplRemoved
     case cannotSpecify_OForMultipleOutputs
     case conflictingOptions(Option, Option)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

cltnschlosser (approved):
I assume that error actually has a nice localizedDescription?

compnerd:
For some definition of "nice".  Windows is being a bit of a pain:
```
error "unable to load output file map 'C:\Users\compnerd\AppData\Local\Temp\TemporaryDirectory.OKyOxN\ofm.json': The operation could not be completed. The data isnΓÇÖt in the correct format."
```

PR #934, opened 2021-12-16
Tests: properly emit the output file map

compnerd:
This adjusts the output file map emission to ensure that the path is
properly escaped.  This was causing problems on Windows, where the path
separator (`\`) was not escaped and would therefore generate invalid
JSON.  This allows the CrossModuleIncrementalBuildTests to now pass on
Windows.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #935, opened 2021-12-16
Tests: rename variable (NFC)

compnerd:
`dynamicLinker` is an entirely inappropriate name for this variable.
This is not the "dynamic linker" (a synonym for the loader), but rather
the "static linker" (aka linker) on Darwin or the linker driver on
non-Darwin targets.  `linkerOrLinkerDriver` seems overly verbose, and
since the linker is rarely meant to be invoked directly by the user (as
traditional Unix linking semantics imply meaning to the order of
arguments), name the variable `linkerDriver`.  This is purely an
aesthetic change, but one that caused me significant amount of
discomfort.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #937, opened 2021-12-17
Tests: be more careful about executable names

compnerd:
Some platforms expect an executable to have a certain suffix.  This
allows a hook for decorating the name accordingly.  This allows a number
of additional asserts to now pass on Windows.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #938, opened 2021-12-17
SwiftDriver: implement response file handling for Windows

compnerd:
This adds an initial pass for computing the command line length for the
purposes of response file triggering.  This is complicated as the
computation requires constructing the quoted string parameter to be
passed and proper command line quoting in Windows is odd.

### DIFF
--- a/Sources/SwiftDriver/Utilities/System.swift
+++ b/Sources/SwiftDriver/Utilities/System.swift
@@ -46,6 +46,64 @@ func commandLineFitsWithinSystemLimits(path: String, args: [String]) -> Bool {
   }
   return commandLineLength < effectiveArgMax
 }
+#elseif os(Windows)
+func commandLineFitsWithinSystemLimits(path: String, args: [String]) -> Bool {
+  func flattenWindowsCommandLine(_ arguments: [String]) -> String {
+    func argNeedsQuoting(_ argument: String) -> Bool {
+      if argument.isEmpty { return false }
+      let chars: Set<Character> = Set("\t \"&'()*<>\\`^|\n")
+      return argument.firstIndex(where: { chars.contains($0) }) != argument.endIndex
+    }
+
+    func quote(_ argument: String) -> String {
+      var unquoted: Substring = argument[...]
+      var quoted: String = #"""#
+      while !unquoted.isEmpty {
+        guard let firstNonBS = unquoted.firstIndex(where: { $0 != "\\" }) else {
+          // The rest of the string is backslashes. Escape all of them and exit.
+          (0 ..< (2 * unquoted.count)).forEach { _ in quoted += "\\" }
+          break
+        }
+
+        let bsCount = unquoted.distance(from: unquoted.startIndex, to: firstNonBS)
+        if unquoted[firstNonBS] == #"""# {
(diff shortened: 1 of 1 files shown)


Sources/SwiftDriver/Utilities/System.swift
```diff
@@ -46,6 +46,62 @@ func commandLineFitsWithinSystemLimits(path: String, args: [String]) -> Bool {
...
+func commandLineFitsWithinSystemLimits(path: String, args: [String]) -> Bool {
+  func flattenWindowsCommandLine(_ arguments: [String]) -> String {
+    func argNeedsQuoting(_ argument: String) -> Bool {
+      let chars: Set<Character> = Set("\t \"&'()*<>\\`^|\n")
+      return argument.count > 0 && argument.firstIndex(where: { chars.contains($0) }) != argument.endIndex
+    }
+
+    func quote(_ argument: String) -> String {
+      var unquoted: Substring = argument[...]
+      var quoted: String = "\""
```
SeanROlszewski, line 59: We'll likely find the code easier to read if we use raw strings in places like here, since we're escaping so many characters:

```swift
var quoted: String = #"#
```

compnerd:
@swift-ci please test

Sources/SwiftDriver/Utilities/System.swift
```diff
@@ -46,6 +46,64 @@ func commandLineFitsWithinSystemLimits(path: String, args: [String]) -> Bool {
...
+      } else {
+        quoted += arg
+      }
+      quoted += " "
+    }
+    return quoted
+  }
+
+  let arguments: [String] = [path] + args
+  let commandLine = flattenWindowsCommandLine(arguments)
```
cltnschlosser, line 100: My first thought was: shouldn't we already have something that does this? Since we need to have code for creating processes.

My next thought was: why don't we have to account for escaping characters on unix platforms? Do we have to here for windows?

And then lastly: If none of those above matter, then we don't actually need to build up the quoted string, just a count.
Ex:
```swift
(0 ..< bsCount).forEach { _ in quoted += "\\" }
quoted += "\(unquoted[firstNonBS])"
```
becomes:
```swift
totalCount += bsCount * 2 // (0 ..< bsCount).forEach { _ in quoted += "\\" }
totalCount += 1 // quoted += "\(unquoted[firstNonBS])"
// Could shorten that into one line :shrug:
```

compnerd: Yes, we do - LLVMSupport :).  You don't need to account for escaping characters on Unix platforms because they handle command lines entirely differently.  Yes, we need to account for it here.  If you have a command line that when quoted exceeds the limit (32K), you will fail.  This is extremely common due to the source code layout enforced by SPM and the fact that absolute paths will be used in some cases.

I do agree that we don't need to build up the command line, AFAICT.  I want to leave the building in place for now for two reasons:
1. this matches the code in LLVMSupport (its a transliteration)
2. this might prove to be useful for testing the tokenization which will need to be implemented as well (yes, tokenization is different as a result)

compnerd:
@artemcm @cltnschlosser - ping

cltnschlosser (approved):

artemcm (approved):

PR #939, opened 2021-12-18
Tests: prefer `XCAssertEqual` over `XCAssertTrue(... = ...)`

compnerd:
This makes it easier to diagnose issues as we can immediately tell what
the two values are.

### DIFF
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -3206,7 +3206,7 @@ final class SwiftDriverTests: XCTestCase {
                                      "foo.swift"])
       let frontendJobs = try driver.planBuild()
       XCTAssertTrue(frontendJobs.count == 2)
-      XCTAssertTrue(frontendJobs[1].tool.absolutePath!.pathString == ld.pathString)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -3206,7 +3206,7 @@ final class SwiftDriverTests: XCTestCase {
                                      "foo.swift"])
       let frontendJobs = try driver.planBuild()
       XCTAssertTrue(frontendJobs.count == 2)
-      XCTAssertTrue(frontendJobs[1].tool.absolutePath!.pathString == ld.pathString)
+      XCTAssertEqual(frontendJobs[1].tool.absolutePath!.pathString, ld.pathString)
```
cltnschlosser, line 3209: Seeing the line above, there are probably a lot of these

cltnschlosser (approved):

Tests/SwiftDriverTests/SwiftDriverTests.swift
compnerd: Likely; while I'd love to wholistically go through the tests, the file is already large, and I have about 15 more changes to get through to get this part of the test suite fully passing on Windows.  Doing a pass over the entire file as a separate change sounds like a good idea to me.

PR #940, opened 2021-12-18
Tests: improve native path representation

compnerd:
Rather than hardcoding the paths and the path separator, we now compute
the actual locations to ensure that we can correctly represent the path
representation natively on the target.  This enables the correct
representation on Windows.

### DIFF
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -27,6 +27,14 @@ private func executableName(_ name: String) -> String {
 #endif
 }
 
+private func rebase(_ arc: String, at base: AbsolutePath) -> String {
+  base.appending(component: arc).pathString.nativePathString()
+}
+
+private func rebase(_ arcs: String..., at base: AbsolutePath) -> String {
+  base.appending(components: arcs).pathString.nativePathString()
+}
+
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -27,6 +27,14 @@ private func executableName(_ name: String) -> String {
 #endif
 }
 
+private func rebase(_ arc: String, at base: AbsolutePath) -> String {
```
compnerd, line 30: If you have a better name for this function, please suggest it.  I'm not tied to it, but this was the best I could come up with.

compnerd:
@cltnschlosser @artemcm - I'd like to get this merged as it unblocks another set of changes.

cltnschlosser (approved):
This change is needed because, Windows needs the drive letter for it to be a valid absolute path?

artemcm (approved):

compnerd:
> This change is needed because, Windows needs the drive letter for it to be a valid absolute path?

Correct.  The Unix absolute path `/foo/bar` is interpreted as `\foo\bar` which is a drive relative path.  We compute the absolute path by taking the current drive and appending that relative path to it.  This gives us an absolute path.

PR #941, opened 2021-12-18
Tests: avoid some warnings on Windows (NFCI)

compnerd:
Process away the body of the tests on Windows to avoid the warning
during the build.  This is hopefully temporary as the tests should
not be skipped, but require further investigation and skipping the
tests enables running the test suite completely on Windows.

### DIFF
--- a/Tests/SwiftDriverTests/APIDigesterTests.swift
+++ b/Tests/SwiftDriverTests/APIDigesterTests.swift
@@ -176,8 +176,10 @@ class APIDigesterTests: XCTestCase {
   }
 
   func testBaselineGenerationEndToEnd() throws {
+#if true
     // rdar://82302797
     throw XCTSkip()
+#else
     try withTemporaryDirectory { path in
       try localFileSystem.changeCurrentWorkingDirectory(to: path)
       let source = path.appending(component: "foo.swift")
@@ -214,6 +216,7 @@ class APIDigesterTests: XCTestCase {
         XCTAssertTrue((json?["children"] as? [Any])!.count >= 1)
       }
     }
+#endif
   }
 
   func testComparisonOptionValidation() throws {
@@ -261,8 +264,10 @@ class APIDigesterTests: XCTestCase {
   }
 
   func testAPIComparisonEndToEnd() throws {
+#if true
     // rdar://82302797
     throw XCTSkip()
+#else
     try withTemporaryDirectory { path in
       try localFileSystem.changeCurrentWorkingDirectory(to: path)
       let source = path.appending(component: "foo.swift")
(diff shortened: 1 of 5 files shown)


compnerd:
@swift-ci please test

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
```diff
@@ -136,9 +136,9 @@ extension IncrementalCompilationTests {
   /// Much of the code below is taking from testLinking(), but uses the output file map code here.
   func testAutolinkOutputPath() {
 #if os(Windows)
-    XCTSkip("Driver.init fatalError")
+    // XCTSkip("Driver.init fatalError")
     return
```
cltnschlosser, line 140: `throw XCTSkip` would be better than silently returning.

Tests/SwiftDriverTests/APIDigesterTests.swift
```diff
@@ -176,8 +176,10 @@ class APIDigesterTests: XCTestCase {
   }
 
   func testBaselineGenerationEndToEnd() throws {
+#if true
     // rdar://82302797
     throw XCTSkip()
```
cltnschlosser, line 181: This approach seems fine to me.

https://developer.apple.com/documentation/xctest/3726077-xctexpectfailure is an option too if we have that on all of the platforms and versions that we need to support.

cltnschlosser (approved):
Should fix warnings on all OS :)

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
compnerd: I agree ... except the test isn't marked as throwing, so I cannot throw in the context.  If that is preferred, I can mark the test as `throws`.  I don't have an issue with that.

Tests/SwiftDriverTests/APIDigesterTests.swift
compnerd: Hmm, I wouldn't be against that; however, it does change the behaviour - this is skipping the test.  Perhaps @artemcm has thoughts on this as well.

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
cltnschlosser: I think changing it to `throws` is fine. The error is handled by XCTest.

compnerd:
@swift-ci please test

compnerd:
Going to merge this for now.  If @artemcm has thoughts on the `#if true`, I can clean that up in a follow up.

PR #942, opened 2021-12-18
Don't unnecessarily constrain ifdefs to macOS

jakepetroules:
They should reflect the semantics of what they intend to represent,
which is "Darwin-based platform".

rdar://85918430


artemcm (approved):

CodaFi (approved):
⛵

PR #943, opened 2021-12-19
SwiftDriver: address TODO about abnormal process termination

compnerd:
This adds the handling for abnormal process termination on Windows.
With this, we can now track processes which exit abnormally, indicating
failure.  This was identified by the saveTemps test which now passes on
Windows.  Add an additional test case for the new message type to
differentiate the abnormal exit due to an exception from a signalled
exit which is limited to Unixish platforms.

### DIFF
--- a/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
+++ b/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
@@ -91,14 +91,15 @@ import Glibc
 
     buildRecordInfo?.jobFinished(job: job, result: result)
 
-    // FIXME: Currently, TSCBasic.Process uses NSProcess on Windows and discards
-    // the bits of the exit code used to differentiate between normal and abnormal
-    // termination.
-    #if !os(Windows)
+#if os(Windows)
+    if case .abnormal = result.exitStatus {
+      anyJobHadAbnormalExit = true
+    }
+#else
     if case .signalled = result.exitStatus {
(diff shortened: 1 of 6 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
```diff
@@ -91,14 +91,15 @@ import Glibc
 
     buildRecordInfo?.jobFinished(job: job, result: result)
 
-    // FIXME: Currently, TSCBasic.Process uses NSProcess on Windows and discards
-    // the bits of the exit code used to differentiate between normal and abnormal
-    // termination.
-    #if !os(Windows)
+#if os(Windows)
+    if case .abnormal = result.exitStatus {
```
cltnschlosser, line 95: So was there a change in TSC to support this?

cltnschlosser (approved):

Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
compnerd: Ugh, yes :-(.  There is a small follow up fix that I am going to be pushing for that shortly.  It isn't merged yet, but the associated change is: apple/swift-tools-support-core#271

PR #944, opened 2021-12-19
Tests: generalise the handling for ELF vs non-ELF targets

compnerd:
When targeting ELF, we use the `swift-autolink-extract` tool to extract
extra arguments for autolinking as ELF does not have a blessed universal
mechanism for doing so.  This adjusts the tests to account for that.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #945, opened 2021-12-19
Tests: add an explicit triple for Darwin SDK tests

compnerd:
This is required as the Windows toolchain uses `-sdk` for locating the
SDK which is required for loading the standard library.  Without this,
we would target the default target (Windows), but use the stub SDK,
which does not contain the standard library and thus fail to compile.
This allows the SwiftDriverTests.SwiftDriverTests/testDarwinSDKTooOld
test to pass on Windows.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #949, opened 2021-12-21
Tests: correct environment testing for Windows

compnerd:
The immediate mode execution will not set an environment variable (i.e.
`DYLD_LIBRARY_PATH` or `LD_LIBRARY_PATH`) as the runtime components are
searched for by `Path` or by the explicit path that a manifest may
provide for a UWP application.

### DIFF
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -2833,12 +2833,28 @@ final class SwiftDriverTests: XCTestCase {
       }
 
       XCTAssertFalse(job.commandLine.contains(.flag("--")))
-      // On darwin, swift ships in the OS. Immediate mode should use that runtime.
-      #if os(macOS)
-      XCTAssertFalse(job.extraEnvironment.keys.contains("\(driver.targetTriple.isDarwin ? "DYLD" : "LD")_LIBRARY_PATH"))
-      #else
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -2833,12 +2833,28 @@ final class SwiftDriverTests: XCTestCase {
...
-      XCTAssertFalse(job.extraEnvironment.keys.contains("\(driver.targetTriple.isDarwin ? "DYLD" : "LD")_LIBRARY_PATH"))
-      #else
-      XCTAssertTrue(job.extraEnvironment.keys.contains("\(driver.targetTriple.isDarwin ? "DYLD" : "LD")_LIBRARY_PATH"))
-      #endif
+
+      let envVar: String
+      if driver.targetTriple.isDarwin {
+        envVar = "DYLD_LIBRARY_PATH"
+      } else if driver.targetTriple.isWindows {
+        envVar = "Path"
```
cltnschlosser, line 2841: We don’t have any assertions for this case. If I understand the comment below, there isn’t anything to assert?

compnerd: Correct; there is nothing to really assert on - `Path` is always part of the environment.  So the assertion would check absolutely nothing.

cltnschlosser (approved):

artemcm (approved):

PR #950, opened 2021-12-21
Tests: handle differences in quoting on Windows

compnerd:
Windows does not use the same quoting strategy at Unix platforms.
Adjust the expectations on Windows to match.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #952, opened 2021-12-21
SwiftDriver: pass along linker options to the linker driver

compnerd:
Correct the handling for non-`-l` linker options on Windows.  We now
correctly forward the user-specified linker search paths to the linker
through the linker driver.


compnerd:
@swift-ci please test

artemcm (approved):

PR #915, opened 2021-11-20
begin to integrate SwiftSystem

tomerd:
motivation: SwiftSystem provides infra that can replace some bespoke implementations in TSC and, improving the cross-platform support. as such we are adding it as a dependency to TSC

changes: add SwiftSystem to CMake setup so it can be included in CMake based builds

PR #953, opened 2021-12-21
Install `swift-system` into the toolchain during CMake build's `install` action

artemcm:
https://github.com/apple/swift-driver/pull/915 added a CMake dependency on swift-system.
Since then, the CMake build of the driver builds swift-system, but when the `install` action is triggered, the dependency's module and shared library are not installed into the toolchain. As a result, built toolchains are unable to invoke swift-driver due to the missing dylib.

This ensures that much like other driver dependencies (TSC, ArgumentParser), we also install SwiftSystem.

Resolves rdar://86754758

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -254,8 +254,9 @@ def install_swiftdriver(args, build_dir, prefix, targets) :
 
   # Binary Swift Modules:
   # swift-driver: SwiftDriver.swiftmodule, SwiftOptions.swiftmodule
-  # TODO: swift-argument-parser: ArgumentParser.swiftmodule (disabled until needed)
   # swift-tools-support-core: TSCUtility.swiftmodule, TSCLibc.swiftmodule, TSCBasic.swiftmodule
+  # swift-argument-parser: ArgumentParser.swiftmodule (disabled until needed)
+  # swift-system: SystemPackage.swiftmodule
   install_binary_swift_modules(args, build_dir, install_lib, targets)
 
   # Modulemaps for C Modules:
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci please test

tomerd (approved):

Utilities/build-script-helper.py
```diff
@@ -254,8 +254,9 @@ def install_swiftdriver(args, build_dir, prefix, targets) :
 
   # Binary Swift Modules:
   # swift-driver: SwiftDriver.swiftmodule, SwiftOptions.swiftmodule
-  # TODO: swift-argument-parser: ArgumentParser.swiftmodule (disabled until needed)
   # swift-tools-support-core: TSCUtility.swiftmodule, TSCLibc.swiftmodule, TSCBasic.swiftmodule
+  # swift-argument-parser: ArgumentParser.swiftmodule (disabled until needed)
+  # swift-system: SystemPackage.swiftmodule
```
compnerd, line 259: Why is the swiftmodule required?  Do you expect to be building against the installed version of swift-driver?

artemcm: Potential clients of `libSwiftDriver` would need it if they try to use the toolchain's driver library instead of depending on it as a package.

Utilities/build-script-helper.py
```diff
@@ -333,10 +334,17 @@ def install_libraries(args, build_dir, universal_lib_dir, toolchain_lib_dir, tar
     install_library(args, build_dir, package_subpath, lib,
                     universal_lib_dir, toolchain_lib_dir, 'swift-tools-support-core', targets)
 
+  # Instal the swift-system shared library into the toolchain lib
```
cltnschlosser, line 337: Typo copied 3 times: `instal`

Utilities/build-script-helper.py
tomerd: I think its an interesting question to understand who such clients are and if static vs dynamic linking of the dependencies is the right strategy.

tomerd: for now 👍 to unblock the toolchain tho

PR #954, opened 2021-12-22
Tests: disable TSAN tests on Windows

compnerd:
TSAN is not supported on Windows at the current time.  This excludes the
tests on Windows as they cannot be supported.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -1962,9 +1962,22 @@ extension Driver {
         isShared: sanitizer != .fuzzer
       )
 
-      // TSan is explicitly not supported for 32 bits.
-      if sanitizer == .thread && !targetTriple.arch!.is64Bit {
-        sanitizerSupported = false
+      if sanitizer == .thread {
+        // TSAN is unavailable on Windows
+        if targetTriple.isWindows {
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

cltnschlosser:
Should windows have a unique diagnostic for when tsan is enabled? Then we can test for that.

compnerd:
Honestly, I'm not sure.  Although TSAN isn't supported currently, there is no reason to assume that it will be unavailable forever.  I suppose that we could add a generic `?san is currently unavailable on '<triple>'`.

compnerd:
@artemcm - thoughts on this?

artemcm:
I like the idea of the driving emitting an error on these options:
> `Tsan is currently unavailable on '<triple>'.`

And even for disabling the tests, it might be better to skip them explicitly with:
```
#if os(Windows)
  throw XCTSkip("Skipping on Windows: Tsan is currently unavailable.")
#endif
```
To make it easier to hunt them down in the future when we may want to re-enable them.

compnerd:
Introduced the new error; I've also adjusted the tests to enable as much as possible rather than skipping them all.

compnerd:
@swift-ci please test

artemcm (approved):
Looks great, thank you.

PR #955, opened 2021-12-22
Tests: explicitly pass a SDK for environment inference tests

compnerd:
This explicitly passes a SDK to the driver invocation as on Windows, the
default SDK does have a SDKSettings.json which does not match the macOS
one.  This results in a spurious warning which causes the test to fail.
Pass the SDKChecks SDK root as the SDK for test.  This allows
SwiftDriverTests.SwiftDriverTests/testEnvironmentInferenceWarning to
pass on Windows.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #957, opened 2021-12-23
Tests: remove an unnecessary target host check

compnerd:
The difference between the work graph is due to the use of the autolink
extract step which is ELF specific, not host specific.  This configures
the path to check the output format instead.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #958, opened 2021-12-23
Tests: correct output file map entries

compnerd:
Correct the path separator for the output file map entries.  This
has no impact on the test coverage but does make it easier to copy-paste
values and to work with the values on Windows.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #929, opened 2021-12-08
[5.6] Update package dependencies to use the 'release/5.6' branch

abertelrud:
Update `swift-llbuild` and `swift-tools-support-core` dependencies to use the 'release/5.6' branch instead of `main`.  This is needed before we can switch over SwiftPM to depend on 'release/5.6'.

PR #960, opened 2021-12-23
[5.6] Turn on emit-module-separately for whole-module builds by default

xymus:
Scheduling a separate emit-module job for whole-module builds, combined with the integrated driver in the build system, will allow to parallelize what used to be essentially sequential builds. This mode unblocks building dependents early by duplicating some work in two processes, it requires more cpu work for a faster wall clock build time. Another advantage of this mode is the resulting swiftmodule is lighter from a classic whole-module build.

Cherry-pick from #922.

rdar://85976762

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2102,10 +2102,9 @@ extension Driver {
       // An option has been passed which requires a module, but the user hasn't
       // requested one. Generate a module, but treat it as an intermediate output.
       moduleOutputKind = .auxiliary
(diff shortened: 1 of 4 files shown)


xymus:
@swift-ci Please test

abertelrud:
The test failure is the same as in https://github.com/apple/swift-driver/pull/929, which is that there is no downloadable Swift 5.6 toolchain yet, and `@swift-ci test` doesn't do a full build in this repository.

xymus:
@swift-ci Please test

xymus:
@swift-ci Please test

artemcm (approved):

PR #961, opened 2021-12-24
[5.6] SwiftDriver: address TODO about abnormal process termination

compnerd:
This adds the handling for abnormal process termination on Windows.
With this, we can now track processes which exit abnormally, indicating
failure.  This was identified by the saveTemps test which now passes on
Windows.  Add an additional test case for the new message type to
differentiate the abnormal exit due to an exception from a signalled
exit which is limited to Unixish platforms.

### DIFF
--- a/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
+++ b/Sources/SwiftDriver/Driver/ToolExecutionDelegate.swift
@@ -91,14 +91,15 @@ import Glibc
 
     buildRecordInfo?.jobFinished(job: job, result: result)
 
-    // FIXME: Currently, TSCBasic.Process uses NSProcess on Windows and discards
-    // the bits of the exit code used to differentiate between normal and abnormal
-    // termination.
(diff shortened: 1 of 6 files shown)


compnerd:
@swift-ci please test

compnerd:
This is needed to repair the build on Windows, but there is no 5.6 snapshot yet which prevents testing.  Could we still get this merged @artemcm?  (In theory there is a circular dependency here - we need this to build for Windows, which is needed to get a snapshot).

artemcm (approved):
This change LGTM as a cherry-pick of https://github.com/apple/swift-driver/pull/943. 
@compnerd is right in that we will need this change merged in order for the 5.6 package bots to succeed once they come online.

compnerd:
I've tested this locally and verified that this repairs the build.  The associated change for apple/swift-tools-support-core#272 was merged without this change.  I really dropped the ball on this, sorry about the trouble.

DougGregor (approved):

PR #962, opened 2021-12-27
tests: enable IncrementalCompilationTests.testAutolinkOutputPath on W…

compnerd:
…indows

This enables a previously disabled test on Windows.  This was disabled
to get a full running test suite to make it easier to improve the
swift-driver on Windows.  As we now reach the point where there are two
areas of the driver (incremental compilation, incremental importing)
which remain problematic, we can now enable this test as the work of
addressing the test failures allows this to pass.


compnerd:
@swift-ci please test

artemcm (approved):

PR #963, opened 2022-01-04
[Build Script Helper] Install `libArgumentParserToolInfo.dylib` into the toolchain

artemcm:
This is now a dependency of the main `libArgumentParser.dylib` and not having it installed casuses `Library not loaded` errors when using the toolchain.

Resolves rdar://86859779


artemcm:
@swift-ci please test

artemcm:
We should transition to linking statically to avoid this kind of brittleness in the future, but for the time being this is needed to get `main` snapshot toolchains operational.

xymus (approved):

PR #964, opened 2022-01-04
[5.6] Update `5.5`-specific manifest dependencies to `5.6`

artemcm:
This brings `Package@swift-5.5.swift` up to parity with recent change in https://github.com/apple/swift-driver/pull/929

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -14,8 +14,8 @@
         "package": "llbuild",
         "repositoryURL": "https://github.com/apple/swift-llbuild.git",
         "state": {
-          "branch": "main",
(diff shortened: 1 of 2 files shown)


abertelrud:
Thanks, @artemcm!

abertelrud:
One more thought:  since Package.resolved is checked in, should that update be included also in this diff?

abertelrud (approved):

artemcm:
> One more thought: since Package.resolved is checked in, should that update be included also in this diff?

Yeah, it makes sense to update that here also. Now included in the PR.

PR #965, opened 2022-01-05
[CMake Build] Build and link driver dependencies as static libraries

artemcm:
This will allow us to not install shared libraries of TSC, ArgumentParser, etc. into the toolchain when installing the driver.

This would also help avoid having to do things like:
https://github.com/apple/swift-driver/pull/963
https://github.com/apple/swift-driver/pull/953

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,8 @@ endif()
 
 find_package(dispatch QUIET)
 find_package(Foundation QUIET)
+find_package(Threads)
+find_package(SQLite3 REQUIRED)
 find_package(Yams CONFIG REQUIRED)
 find_package(ArgumentParser CONFIG REQUIRED)
 find_package(SwiftSystem CONFIG REQUIRED)
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -249,20 +249,6 @@ def install_swiftdriver(args, build_dir, prefix, targets) :
   # swift-driver and swift-help
   install_executables(args, build_dir, bin_dir, install_bin, targets)
 
-  # libSwiftDriver and libSwiftDriverExecution and libSwiftOptions
-  install_libraries(args, build_dir, lib_dir, install_lib, targets)
-
-  # Binary Swift Modules:
-  # swift-driver: SwiftDriver.swiftmodule, SwiftOptions.swiftmodule
-  # swift-tools-support-core: TSCUtility.swiftmodule, TSCLibc.swiftmodule, TSCBasic.swiftmodule
-  # swift-argument-parser: ArgumentParser.swiftmodule (disabled until needed)
-  # swift-system: SystemPackage.swiftmodule
-  install_binary_swift_modules(args, build_dir, install_lib, targets)
-
-  # Modulemaps for C Modules:
-  # TSCclibc
-  install_c_module_includes(args, build_dir, install_include)
-
 # Install universal binaries for swift-driver and swift-help into the toolchain bin
 # directory
 def install_executables(args, build_dir, universal_bin_dir, toolchain_bin_dir, targets):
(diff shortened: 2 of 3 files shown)


Utilities/build-script-helper.py
```diff
@@ -307,47 +307,16 @@ def install_libraries(args, build_dir, universal_lib_dir, toolchain_lib_dir, tar
...
-  # Install the swift-argument-parser shared libraries into the toolchain lib
-  package_subpath = os.path.join(args.configuration, 'dependencies', 'swift-argument-parser')
-  for lib in ['libArgumentParser', 'libArgumentParserToolInfo']:
-      install_library(args, build_dir, package_subpath, lib,
-                      universal_lib_dir, toolchain_lib_dir,'swift-argument-parser', targets)
-
   # Install the llbuild core shared libraries into the toolchain lib
   package_subpath = os.path.join(args.configuration, 'dependencies', 'llbuild')
-  for lib in ['libllbuildSwift', 'libllbuild']:
+  for lib in ['libllbuild']:
```
artemcm, line 319: Despite specifying `-DBUILD_SHARED_LIBS=OFF` on the LLBuild CMake invocation we are still getting a `libllbuild.dylib` so I am still looking into getting that to be a static library as well.

artemcm:
@swift-ci please test

artemcm:
Cross-repo tests in https://github.com/apple/swift/pull/40754 are green, including toolchain builds.
This PR is now ready for review.

tomerd:
not a CMake expert, but seems good to me 👍

CMakeLists.txt
```diff
@@ -60,6 +60,8 @@ endif()
 
 find_package(dispatch QUIET)
 find_package(Foundation QUIET)
+find_package(Threads QUIET)
```
compnerd, line 63: Hmm, any reason that this should be quiet?

Utilities/build-script-helper.py
```diff
@@ -475,7 +329,8 @@ def build_llbuild_using_cmake(args, target, swiftc_exec, build_dir, base_cmake_f
         '-DCMAKE_C_COMPILER:=clang',
         '-DCMAKE_CXX_COMPILER:=clang++',
         '-DCMAKE_CXX_FLAGS=-target %s' % target,
-        '-DLLBUILD_SUPPORT_BINDINGS:=Swift'
+        '-DLLBUILD_SUPPORT_BINDINGS:=Swift',
```
compnerd, line 332: This looks like a typo

```suggestion
        '-DLLBUILD_SUPPORT_BINDINGS=Swift',
```

Utilities/build-script-helper.py
```diff
@@ -496,9 +351,7 @@ def build_system_using_cmake(args, target, swiftc_exec, build_dir, base_cmake_fl
   print('Building Swift Driver dependency: Swift System')
   system_source_dir = os.path.join(os.path.dirname(args.package_path), 'swift-system')
   system_build_dir = os.path.join(build_dir, 'swift-system')
-  flags = [
-      # requried due to swift-autolink-extract bug ("The file was not recognized as a valid object file")
-      "-DBUILD_SHARED_LIBS=YES"]
+  flags = ['-DBUILD_SHARED_LIBS=OFF']
```
compnerd, line 354: I prefer `NO` but it's the same.  The underlying issue here was fixed a while ago.

compnerd (commented):
One thing that this would do is prevent the use of the build-script-helper on windows (static linking is not available on Windows, though that is something that I would like to fix).

neonichu (approved):

artemcm:
@swift-ci please test

artemcm:
> One thing that this would do is prevent the use of the build-script-helper on windows (static linking is not available on Windows, though that is something that I would like to fix).

Is it currently being used to build swift-driver? If it is, should we preserve the ability to follow the old (pre-this-PR) code-path?

compnerd:
No, I don't believe that the python script is being used, so it should be okay to merge this and fix that at some later date.

compnerd (approved):

PR #971, opened 2022-01-11
Throw error for invalid double dash argument

keith:
Previously `swiftc --invalid foo.swift` would be interpreted as
`swiftc -- foo.swift`. Now we fail in this case since `--invalid` is
likely to be a typo regardless of whether you meant to do `-- invalid`
or `--something-valid`.

### DIFF
--- a/Sources/SwiftOptions/OptionParsing.swift
+++ b/Sources/SwiftOptions/OptionParsing.swift
@@ -133,6 +133,10 @@ extension OptionTable {
         index += 1
 
       case .remaining:
+        if argument != option.spelling {
+          throw OptionParseError.unknownOption(
+            index: index - 1, argument: argument)
+        }
         parsedOptions.addOption(.DASHDASH, argument: .single("--"))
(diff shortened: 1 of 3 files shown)


cltnschlosser:
@swift-ci please test

cltnschlosser:
Makes complete sense to me, but do you happen to know how the c++ driver behaved in this situation?

keith:
It seems to handle it the same way that this change does:

```
% swiftc -disallow-use-new-driver --invalid foo.swift
<unknown>:0: error: unknown argument: '--invalid'
```

keith:
Updated the test here that was previously invalid

cltnschlosser (approved):

cltnschlosser:
@swift-ci please test

artemcm (approved):

keith:
Looks like CI is happy here, can someone merge this?

keith:
Thanks!

PR #973, opened 2022-01-12
Tests: inject a sleep for test modifications

compnerd:
This allows the source modification during build tests to pass on
Windows where nanosecond precision is not available for file
modification times.


compnerd:
@swift-ci please test

artemcm (approved):

PR #975, opened 2022-01-12
Tests: adjust the paths for output map resolution

compnerd:
Use the TSC path types to compute the path strings for the output file
map resolution as `/foo_root` is not an absolute path on all platforms.
This adjustment now allows this test to pass on Windows.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #980, opened 2022-01-13
Kill the process after cleanup on interrupt

BenchR267:
This is identical behavior on bash but when spawning the process using posix_spawn (like llbuild does) the exit code of (SIGINT | 0x80) == 130 gets encoded as an exit code rather than the uncaught signal.

rdar://77055430


BenchR267:
@swift-ci please test

artemcm (approved):

PR #981, opened 2022-01-13
Tests: correct output file map JSON emission

compnerd:
The previous emission did not take into account separator escaping
(Windows uses `\` which must be escaped when emitted).  This corrects
the emission of the JSON content.

Once we do so, the loading of the JSON succeeds, but the tests fail as
the paths are now the Windows native paths and must be adjusted
accordingly.


compnerd:
@swift-ci please test

artemcm (approved):

PR #982, opened 2022-01-13
Tests: add some sleeps for file system modifications

compnerd:
This adds a couple more strategic `sleep` invocations to ensure that
filesystems without nanosecond precision on file timestamps can be
accommodated.  This allows the incremental compilation tests to now pass
on Windows.


compnerd:
@swift-ci please test

artemcm (approved):
🎉

PR #989, opened 2022-01-24
Hide intro for `swiftc --help`

stevapple:
Users know what they're doing by using `swiftc` instead of `swift`. Subcommand info is useless for `swiftc` users and should be eliminated.


bitjammer:
Makes sense.

bitjammer (approved):

bitjammer:
@swift-ci Please test

PR #992, opened 2022-01-26
build: install the executable targets

compnerd:
This is the first and last step towards enabling swift-driver on
Windows.  The current swift-driver test suite passes on Windows, and
with the changes now merged, `swift build --use-integrated-swift-driver`
works.  This change will extract the tools so that we can package them
up.  This sets the stage for the last change needed in SPM and allows
experimenting with packaging now.


artemcm (approved):
🎉

compnerd:
@swift-ci please test

PR #993, opened 2022-01-26
Driver: add support for tokenizing a Windows-style rsp

compnerd:
This adds support for handling Windows style response files.  These are
important to get correct as they are consumed by both swift and clang.
The differences will result in the `\` path separator getting discarded,
yielding invalid paths.


compnerd:
@swift-ci please test

artemcm (approved):

PR #994, opened 2022-01-26
Loosen restriction for building deprecated iOS targets

cyndyishida:
This patch loosens the restriction for building deprecated iOS targets
to only when the user attempts to compile a binary or emit IR.
This allows the compiler to continue building swift modules and
interfaces for newer deployment targets.

The error message is also slightly modified to allow the same message to
apply to different platforms.

resolves: <rdar://87898177>

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -143,7 +143,7 @@ public final class DarwinToolchain: Toolchain {
   public enum ToolchainValidationError: Error, DiagnosticData {
     case osVersionBelowMinimumDeploymentTarget(String)
     case argumentNotSupported(String)
-    case iOSVersionAboveMaximumDeploymentTarget(Int)
+    case invalidDeploymentTargetForIR(String, String)
     case unsupportedTargetVariant(variant: Triple)
     case darwinOnlySupportsLibCxx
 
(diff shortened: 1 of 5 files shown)


xymus (approved):
This looks good, thank you!

xymus:
I launched the tests from the compiler side because there's an equivalent check in the C++ driver so there could be tests affected there too.

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -3133,14 +3133,22 @@ final class SwiftDriverTests: XCTestCase {
...
                                            "foo.swift"])) { error in
-      guard case DarwinToolchain.ToolchainValidationError.iOSVersionAboveMaximumDeploymentTarget(12) = error else {
+      guard case DarwinToolchain.ToolchainValidationError.invalidDeploymentTargetForIR("iOS 11") = error else {
         XCTFail()
         return
       }
     }
 
+    XCTAssertThrowsError(try Driver(args: ["swiftc", "-emit-module", "-c", "-target", 
+                                           "armv7-apple-ios12.0", "foo.swift"])) { error in
```
xymus, line 3145: Should we test a version without the `-c` where there's no error?

cyndyishida:
@swift-ci please test

cyndyishida:
@swift-ci please test

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -143,16 +143,17 @@ public final class DarwinToolchain: Toolchain {
...
 
     public var description: String {
       switch self {
       case .osVersionBelowMinimumDeploymentTarget(let target):
         return "Swift requires a minimum deployment target of \(target)"
-      case .iOSVersionAboveMaximumDeploymentTarget(let version):
-        return "iOS \(version) does not support 32-bit programs"
+      case .invalidDeploymentTargetForIR(let target):
+        return
+          "\(target) and above does not support emitting binaries or IR for 32-bit programs"
```
jakepetroules, line 156: Could we pass through and print the specific architecture name in place of "32-bit" in order to be more generalized?

lorentey (approved):

cyndyishida:
@swift-ci please test

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
cyndyishida: This is updated to emit "iOS 11 and above does not support emitting binaries or IR for armv7" for `armv7-apple-ios12`

cyndyishida:
@swift-ci please test

PR #995, opened 2022-01-26
Only emit ABI descriptors when Library Evolution is enabled

artemcm:
These descriptors are not otherwise meaningful so this avoid unnecessary work when we do not require it.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -363,6 +363,9 @@ public struct Driver {
     guard let moduleOutput = moduleOutputInfo.output else {
       return nil
     }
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci please test

xymus (approved):
Looks good!

nkcsgexi (changes requested):
Discussed with Artem about this offline, we plan to still emit the abi descriptor for non-library-evolution modules because it's an extensible format which may carry other semantic information for other purposes. 

Instead, we decided to drop ABI descriptor emission for non-Darwin platforms.

artemcm:
Closing in favour of https://github.com/apple/swift-driver/pull/998

PR #997, opened 2022-01-27
Skip currently-failing `testModuleAliasingPrebuiltWithScanDeps`.

artemcm:
This test does not fail with recent toolchain snapshots, but seems to fail when using older (shipping) toolchains.


artemcm:
@swift-ci please test

xedin (approved):

PR #999, opened 2022-01-28
SwiftDriver: enable response files on LinkJobs

compnerd:
The non-Darwin targets use clang as the linker driver.  As a result, the
non-Darwin targets implicitly support response files (clang will expand
it if the linker does not support response files or file lists).
Unfortunately, the Darwin target manually invokes ld, which requires
re-implementing the handling there.  To complicate things further, older
ld64 does not support response files, but does support linker file
lists, while newer ld64 supports both.  Ideally, Darwin would also use
clang as the linker driver to homogenise the paths.

This is required for Windows support as the command line limit is much
lower on Windows than on Linux - nearing 4k which is possible to exceed
when building a larger project (e.g. swift-driver).

### DIFF
--- a/Sources/SwiftDriver/Jobs/LinkJob.swift
+++ b/Sources/SwiftDriver/Jobs/LinkJob.swift
@@ -64,7 +64,6 @@ extension Driver {
       targetInfo: frontendTargetInfo
     )
 
-    // TODO: some, but not all, linkers support response files.
     return Job(
       moduleName: moduleOutputInfo.name,
       kind: .link,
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

artemcm (approved):

Sources/SwiftDriver/Jobs/LinkJob.swift
```diff
@@ -73,7 +72,10 @@ extension Driver {
       displayInputs: inputs,
       inputs: inputs,
       primaryInputs: [],
-      outputs: [.init(file: outputFile.intern(), type: .image)]
+      outputs: [.init(file: outputFile.intern(), type: .image)],
+      // FIXME: newer ld64 supports response files as well, though really,
```
keith, line 76: doesn't this mean this could always be enabled at this point? clang now defaults to on for all new versions since xcode 13

compnerd: I don't know what deployments that this code would run on, and we don't have enough details about what ld64 is in use, so it is safe to limit it to non-Darwin targets.  Ideally, we would use clang as the linker driver.

keith: Yea I guess I'm just thinking a new driver version + old xcode tools version is likely considered unsupported? But yea it seems minor enough for now

PR #1004, opened 2022-02-01
remove unnecessary `TSCLibc` dependency

compnerd:
These modules did not use any of the `TSCLibc` interfaces but relied on
the module's re-exported libc import.  Simply inline that and remove the
dependency on `TSCLibc`.  On Linux and Apple platforms, this does not
have much impact, however, Windows does not support static linking of
Swift content yet.  As a result, this avoids an unnecessary dynamic
library being linked and increasing the dirty memory of the process.


compnerd:
@swift-ci please test

PR #1005, opened 2022-02-02
SwiftDriver: explicitly indicate the TSCUtility interfaces

compnerd:
This explicitly identifies the TSCUtility interfaces that SwiftDriver
depends on.  This helps identify a "burn down list" of interfaces that
remain in TSC(Utility) which are in use.  As these interfaces are
replaced, we can easily monitor the remaining interfaces that are in
use.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -10,9 +10,11 @@
 //
 //===----------------------------------------------------------------------===//
 import TSCBasic
-import TSCUtility
 import Foundation
 import SwiftOptions
+
+import enum TSCUtility.Diagnostics
+import struct TSCUtility.Version
 
 /// The Swift driver.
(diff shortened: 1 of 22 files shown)


compnerd:
CC: @tomerd @artemcm

cltnschlosser (approved):

artemcm (approved):
This is super useful!

tomerd:
this is great. I would love to see us deprecate TSC over time

one specific area worth noting is that SwiftPM is no longer using TSC's DiagnosticsEngine and considers it deprecated, so the only library that uses it now is Driver. if Driver wants to continue and use it as-is it can be moved into driver as an internal impl details, or a different  abstraction can be exposed at the Driver level. in any case the idea of DiagnosticsEngine as an API for composition between libraries is something we want to do away with.

compnerd:
@swift-ci please test

PR #1007, opened 2022-02-09
Tests: introduce `AbsolutePath.nativePathString(escaped:)`

compnerd:
This adds a new helper function to create the native path string and
escape it in a single operation.  This reduces some of the verbosity
at the call sites.  The desire is to eventually remove the extension
on `String` and be able to simplify all the call sites.


compnerd:
@swift-ci please test

artemcm (approved):
Very nice improvement, thank you.

PR #1008, opened 2022-02-09
Tests: introduce `VirtualPath.nativePathString(escaped:)`

compnerd:
This adds a new helper function to create the native path string and
escape it in a single operation.  This reduces some of the verbosity
at the call sides.  The desire is to eventually remove the extension
on `String` and be able to simplify the call sites.


compnerd:
@swift-ci please test

PR #1014, opened 2022-02-15
Darwin: introduce a global override for debug prefix map entries 

adrian-prantl:
This patch adds a new Darwin Swift driver environment variable in the spirit of
RC_DEBUG_OPTIONS, called RC_DEBUG_PREFIX_MAP, which allows a meta build tool to
add one additional -fdebug-prefix-map entry without the knowledge of the build
system.

See also https://reviews.llvm.org/D119850

rdar://85224717

### DIFF
--- a/Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
@@ -130,6 +130,8 @@ extension WindowsToolchain.ToolchainValidationError {
     !env["RC_DEBUG_OPTIONS", default: ""].isEmpty
   }
 
+  public var globalDebugPathRemapping: String? { nil }
+    
   public func runtimeLibraryName(for sanitizer: Sanitizer, targetTriple: Triple,
                                  isShared: Bool) throws -> String {
     // TODO(compnerd) handle shared linking
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -488,6 +488,16 @@ final class SwiftDriverTests: XCTestCase {
         XCTAssertTrue(jobs[0].commandLine.contains(.flag("qux=")))
     }
 
+    do {
+      var env = ProcessEnv.vars
+      env["SWIFT_DRIVER_TESTS_ENABLE_EXEC_PATH_FALLBACK"] = "1"
+      env["RC_DEBUG_PREFIX_MAP"] = "old=new"
+      var driver = try Driver(args: ["swiftc", "-c", "-target", "arm64-apple-macos12", "foo.swift"], env: env)
+      let jobs = try driver.planBuild()
+      XCTAssertTrue(jobs[0].commandLine.contains(.flag("-debug-prefix-map")))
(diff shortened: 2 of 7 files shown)


adrian-prantl:
See also https://github.com/apple/swift/pull/41388

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
```diff
@@ -126,9 +126,9 @@ extension WindowsToolchain.ToolchainValidationError {
     return nil
   }
 
-  public var shouldStoreInvocationInDebugInfo: Bool {
-    !env["RC_DEBUG_OPTIONS", default: ""].isEmpty
```
artemcm, line 130: @compnerd was this a carryover from the Darwin toolchain or is this actually needed/correct/used?

artemcm (approved):
Other than the Windows toolchain question above, the change looks good to me.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
compnerd: I did carry it over from the Darwin toolchain, but did so intentionally.  I wanted to reserve the ability to use that to trigger the behavior.

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -488,6 +488,12 @@ final class SwiftDriverTests: XCTestCase {
         XCTAssertTrue(jobs[0].commandLine.contains(.flag("qux=")))
     }
 
+    try assertNoDriverDiagnostics(args: "swiftc", "-target", "arm64-apple-macos12", "foo.swift", env: ["RC_DEBUG_PREFIX_MAP": "old=new"]) { driver in
```
compnerd, line 491: Pretty sure that this is going to regress the windows build.  Please drop the explicit triple.

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
adrian-prantl: The RC_* environment variables are very specific to Darwin, I assumed this was a copy & paste error, because it isn't tested and the C++ implementation also only implements the feature on Darwin.

Tests/SwiftDriverTests/SwiftDriverTests.swift
adrian-prantl: I can't, the feature is gated on a Darwin toolchain. But this is also only testing the _driver_, so I would have assumed it works on any platform.

adrian-prantl:
@swift-ci test

adrian-prantl:
@swift-ci test

Sources/SwiftDriver/Toolchains/WindowsToolchain.swift
adrian-prantl: I'll put it back in, but this is a divergence from the C++ implementation.

adrian-prantl:
@swift-ci test

Tests/SwiftDriverTests/SwiftDriverTests.swift
adrian-prantl: I rewrite the test by copy&pasting from another test that also uses an explicit target, so I think this should be good now.

PR #1016, opened 2022-02-18
Construct the job description string in a way that doesn't retain trailing spaces if the rhs is nil

jakepetroules:


jakepetroules:
@swift-ci please test

artemcm (approved):

PR #1019, opened 2022-02-21
fix cross-module-optimization builds and add the option `-disable-cmo`

eeckstein:
The driver splits the compile job into two jobs: one for creating the object file, one for emitting the swiftmodule.
For the swiftmodule-job the driver adds -experimental-skip-non-inlinable-function-bodies-without-types.
This breaks CMO, because in CMO, the inlinable-decision is not derived from the AST but done in the optimizer.

As a quick hack we could disable -experimental-skip-non-inlinable-function-bodies-without-types.
But for CMO, the two-job approach is highly problematic, because it risks the two compile jobs to get out of sync, e.g. by different command line options, indeterministic behavior, etc. This can result in unresolved symbol errors, which are very hard to debug.

This change also adds a new option `-disable-cmo` to disable cross-module-optimization. It can be used to go back to the old behavior.

rdar://89223981

### DIFF
--- a/Sources/SwiftOptions/Options.swift
+++ b/Sources/SwiftOptions/Options.swift
@@ -61,6 +61,7 @@ extension Option {
   public static let continueBuildingAfterErrors: Option = Option("-continue-building-after-errors", .flag, attributes: [.frontend, .doesNotAffectIncrementalBuild], helpText: "Continue building, even after errors are encountered")
   public static let coveragePrefixMap: Option = Option("-coverage-prefix-map", .separate, attributes: [.frontend], metaVar: "<prefix=replacement>", helpText: "Remap source paths in coverage info")
   public static let CrossModuleOptimization: Option = Option("-cross-module-optimization", .flag, attributes: [.helpHidden, .frontend], helpText: "Perform cross-module optimization")
+  public static let disableCrossModuleOptimization: Option = Option("-disable-cmo", .flag, attributes: [.helpHidden, .frontend], helpText: "Disable cross-module optimization")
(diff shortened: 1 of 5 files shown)


eeckstein:
@swift-ci test

Sources/SwiftOptions/Options.swift
```diff
@@ -61,6 +61,7 @@ extension Option {
   public static let continueBuildingAfterErrors: Option = Option("-continue-building-after-errors", .flag, attributes: [.frontend, .doesNotAffectIncrementalBuild], helpText: "Continue building, even after errors are encountered")
   public static let coveragePrefixMap: Option = Option("-coverage-prefix-map", .separate, attributes: [.frontend], metaVar: "<prefix=replacement>", helpText: "Remap source paths in coverage info")
   public static let CrossModuleOptimization: Option = Option("-cross-module-optimization", .flag, attributes: [.helpHidden, .frontend], helpText: "Perform cross-module optimization")
+  public static let disableCrossModuleOptimization: Option = Option("-disable-cmo", .flag, attributes: [.helpHidden, .frontend], helpText: "Disable cross-module optimization")
```
artemcm, line 64: Please update `Options.swift` according to the procedure described in:

artemcm (approved):

xymus (approved):

Sources/SwiftOptions/Options.swift
eeckstein: ok, I'll do this in a separate PR

PR #1020, opened 2022-02-24
Toolchains: introduce `clangxx` tool (NFC)

compnerd:
This introduces a new `clangxx` tool which maps to `clang++`.  This is
required to ensure that we are able to support linking against the C++
runtime with C++ interop is enabled.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1023, opened 2022-02-24
SwiftDriver: use `clang++` as linker driver with C++ Interop

compnerd:
This adjusts the linker job construction to switch the linker driver to
the C++ compiler driver as the linker driver when C++ interop is
enabled.  This allows us to ensure that we properly link against the C++
runtime which should be assumed to be needed with C++ interop.  This is
required for further exploration of C++ interop.


compnerd:
@swift-ci please test

compnerd:
CC: @hyp @zoecarver

compnerd:
@swift-ci please test

artemcm (commented):
This looks good to me in terms of the changes and being gated on `enableExperimentalCxxInterop`, but I'd really like this PR to get a review from @hyp or @zoecarver.

hyp (approved):
LGTM

PR #1029, opened 2022-03-08
[Incremental Builds] Produce an empty first wave if only job to run is verify module interface job.

artemcm:
Existing check decides to skip all compilation if there are no compile jobs and no post-compile job depends on before-compile jobs' outputs. This change refines the check to also ensure we do not run before-compile jobs whose only dependents are interface verification jobs. Otherwise, we would have an emit-module job, followed by verify-module-interface jobs on a null incremental build.

Resolves rdar://89913538


artemcm:
@swift-ci please test

xymus (approved):
This will make a lot of people happy. Thank you!

PR #1030, opened 2022-03-11
Using lld as default linker.

cntrump:
`lld` is Swift toolchain builtin linker, but not used by default. 

If I want to use `lld`, I must install `gold` first and add `-use-ld=lld` flag.

`gold` is included in package `binutils` on Debian/Ubuntu, it is not installed by default.

On a minimal Linux OS, We can just install Swift toolchain.

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -22,12 +22,12 @@ extension GenericUnixToolchain {
     case .arm, .aarch64, .armeb, .thumb, .thumbeb:
       // BFD linker has issues wrt relocation of the protocol conformance
       // section on these targets, it also generates COPY relocations for
-      // final executables, as such, unless specified, we default to gold
+      // final executables, as such, unless specified, we default to lld
       // linker.
-      return "gold"
+      return "lld"
     case .x86, .x86_64, .ppc64, .ppc64le, .systemz:
       // BFD linker has issues wrt relocations against protected symbols.
-      return "gold"
+      return "lld"
     default:
       // Otherwise, use the default BFD linker.
       return ""


compnerd:
I don't have any objections to `lld` being used as a default on Linux, however, this is not a "builtin linker".  Additionally, it is unclear why you would need to install gold first to use the toolchain, using `-use-ld=lld` should be sufficient.

cntrump:
> I don't have any objections to `lld` being used as a default on Linux, however, this is not a "builtin linker". Additionally, it is unclear why you would need to install gold first to use the toolchain, using `-use-ld=lld` should be sufficient.

On Ubuntu Server 20.04:

Install Swift 5.6:

```bash
wget https://download.swift.org/swift-5.6.1-release/ubuntu2004/swift-5.6.1-RELEASE/swift-5.6.1-RELEASE-ubuntu20.04.tar.gz
tar xvf swift-5.6.1-RELEASE-ubuntu20.04-aarch64.tar.gz
sudo mv swift-5.6.1-RELEASE-ubuntu20.04-aarch64 /opt/swift
PATH=/opt/swift/usr/bin:$PATH
```

Create a test project:

```bash
mkdir test && cd test
swift package init --type executable
```

Build test project using `-use-ld=lld`:

```bash
swift build -c release -Xswiftc -use-ld=lld
```

I got error:

```
'test': error: invalidManifestFormat("error: link command failed with exit code 1 (use -v to see invocation)\nclang-13: error: invalid linker name in argument \'-fuse-ld=gold\'", diagnosticFile: nil)
```

compnerd:
Ah, that is a fun issue.  That is something that should be fixed irrespective of the default.  The problem is that SPM does not have a way to control the flags at a global level - you are passing the flag just to the compilation, not the manifest generation.  I suspect that `swift build -c release -Xswiftc -use-ld=lld -Xmanifest -use-ld=lld` would work without gold.

cntrump:
> Ah, that is a fun issue. That is something that should be fixed irrespective of the default. The problem is that SPM does not have a way to control the flags at a global level - you are passing the flag just to the compilation, not the manifest generation. I suspect that `swift build -c release -Xswiftc -use-ld=lld -Xmanifest -use-ld=lld` would work without gold.

Thanks, `-Xmanifest -use-ld=lld` works for me!

PR #1031, opened 2022-03-12
Add potential member edge incremental test

keith:
This case tests that adding, and then removing, an association between
types resets the incremental state to the same as before the association
was made.


keith:
This test currently fails, but I wanted to submit it to validate CI fails as well, before ideally trying to find a fix

keith:
@swift-ci please test

Tests/IncrementalImportTests/PotentialMemberEdgeRemoveTest.swift
```diff
@@ -0,0 +1,112 @@
...
+      andWhenDisabled: [])
+
+    let steps = [
+      Step(                          building: modules, .expecting(modules.allSourcesToCompile)),
+      Step(                          building: modules, .expecting(.none)),
+      Step(adding: "updateConstant", building: modules, .expecting(whenUpdatingConstant)),
+      Step(                          building: modules, .expecting(whenUpdatingConstant)),
+      Step(adding: "privateLet",     building: modules, .expecting(whenAddRemovePrivateLet)),
+      Step(                          building: modules, .expecting(whenAddRemovePrivateLet)),
+      Step(adding: "updateConstant", building: modules, .expecting(whenUpdatingConstant)),
```
cltnschlosser, line 105: So looks like this was the first step to fail. If you added a none step right before this one, does that step succeed? I could be reading the test result wrong, but just wondering if it’s the updateConstant that’s failing or the privateLet before it. (It seemed like updateConstant worked the first time)

cltnschlosser: Up to you. I *think* that proves the issue isn't caused by the `privateLet` step, so maybe it's good to have, but I'm new to this area of the driver. Seems like an interesting bug.

artemcm:
Interesting catch!
@swift-ci please test

keith:
It should still fail, I haven't had the chance to go digging to try and fix just yet

artemcm:
Screen Shot 2022-03-15 at 4 59 18 PM

Maybe confirming the obvious, but still: the issue seems to be the fact that the first `whenAddRemovePrivateLet` compile adds an edge to the module dependency graph of `A` (for example) from the interface of `C` to source-file of `A`. Which is expected. But after the second `whenAddRemovePrivateLet` compile that undoes the change, dependency graph of A still contains this edge, which is causing the extra recompiles.

Screenshot top-down are module dependency graphs of `A`: 
- After step 3 (second `whenUpdatingConstant` compile)
- After step 5 (first `whenAddRemovePrivateLet` compile)
- After step 6 (second `whenAddRemovePrivateLet` compile)

PR #1034, opened 2022-03-18
Pass-down -empty-abi-descriptor if evolution is disabled

nkcsgexi:
We should also be able to emit ABI descriptors for modules without evolution. However, doing so leads us to deserialize more contents from binary modules, exposing more deserialization issues as a result.


nkcsgexi:
@swift-ci please test

PR #1043, opened 2022-03-30
Always emitting private swift interfaces if public interfaces are emitted

nkcsgexi:
With the introduction of features like @_spi_available, we may print public and private interfaces differently even from the same codebase without @_spi. For this reason, we should always print private interfaces so that we don’t mix the public interfaces with private Clang modules.

rdar://91061334

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -767,7 +767,7 @@ public struct Driver {
         emitModuleSeparately: emitModuleSeparately,
         outputFileMap: self.outputFileMap,
         moduleName: moduleOutputInfo.name)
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

xymus:
Should we do this in the build system instead? Even if the compiler produces the private swiftinterface, the build system needs to know to install it.

nkcsgexi:
I think driver-side change is fine for it because we don't need to change every build system about it.

nkcsgexi:
To clarify, we could do both. For build systems where install logic is needed, we can certainly also pass down the private interface flag.

xymus (approved):
We can still merge this if you see it being useful.

PR #1046, opened 2022-04-01
Track serialized diagnostics of emit-module jobs separately.

artemcm:
When serialized diagnostics are requested by the build system, it is expected to provide an output-file-map to specify per-file diagnostic output paths. The build system may also request serialized diagnostics module-wide in WMO compilation by specifying an output-file-map entry of module scope (`""`).

Emit-module jobs are always run at module scope, which means that if we want to get serialized diagnostics from the emit-module job, in WMO we do not know which path to use, because the module-scope path is specified for the WMO compilation job.

This change introduces a separate output path for the emit-module serialized diagnostics to be specified either:
- On the command-line with `-emit-module-serialize-diagnostics-path`
- In the output file-map using key `"emit-module-diagnostics": "…/…emit-module.dia"` at global (module) scope.

Resolves rdar://83110325

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -278,6 +278,9 @@ public struct Driver {
   /// Path to the serialized diagnostics file.
   let serializedDiagnosticsFilePath: VirtualPath.Handle?
 
+  /// Path to the serialized diagnostics file of the emit-module task.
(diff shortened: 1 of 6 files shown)


artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

artemcm:
@swift-ci please test

BenchR267 (approved):

artemcm:
@swift-ci please test

PR #1045, opened 2022-04-01
Add support for `-enable-regex-literals` flag

artemcm:
Part of rdar://91119995

PR #1047, opened 2022-04-02
Revert "Add support for `-enable-regex-literals` flag"

beccadax:
Reverts apple/swift-driver#1045 due to failures on several bots. (It looks like the test needs to be more flexible to accommodate non-Darwin platforms.)

Workaround for rdar://91190206.


beccadax:
Force-merging this revert to unblock PR tests.

PR #1051, opened 2022-04-09
AdopterAnalysis: diagnose several issues within .swiftinterface

nkcsgexi:
1. missing `-enable-library-evolution`
2. missing a library level flag
3. wrong library level flag

### DIFF
--- a/Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
+++ b/Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
@@ -13,17 +13,43 @@ import TSCBasic
 import SwiftOptions
 import Foundation
 
-@_spi(Testing) public func isIosMacInterface(_ path: VirtualPath) throws -> Bool {
+func getModuleFlags(_ path: VirtualPath, _ ignorable: Bool) throws -> [String] {
   let data = try localFileSystem.readFileContents(path).cString
   let myStrings = data.components(separatedBy: .newlines)
-  let prefix = "// swift-module-flags: "
+  let prefix = ignorable ? "// swift-module-flags-ignorable: " : "// swift-module-flags: "
   if let argLine = myStrings.first(where: { $0.hasPrefix(prefix) }) {
-    let args = argLine.dropFirst(prefix.count).components(separatedBy: " ")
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
```diff
@@ -250,6 +276,12 @@ public struct PrebuiltModuleInput {
   }
 }
 
+public enum AdopterIssueKind: String, Codable {
+  case LibraryEvolutionDisabled
+  case LibraryLevelMissing
+  case LibraryLevelWrong
```
cltnschlosser, line 282: Should these be lowercased? With custom uppercase raw values if that’s needed.

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
```diff
@@ -271,6 +304,27 @@ public class SwiftAdopter: Codable {
...
+      issues.append(.LibraryLevelMissing)
+    }
+    if libLevel == .spi && !isPrivate {
+      issues.append(.LibraryLevelWrong)
+    }
+    if libLevel == .api && isPrivate {
+      issues.append(.LibraryLevelWrong)
+    }
+    if !flags.contains("-enable-library-evolution") {
+      issues.append(.LibraryEvolutionDisabled)
```
cltnschlosser, line 325: Is it always the case that library evolution should be enabled for this? Not super familiar with this area.

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
nkcsgexi: We totally should. Fixed.

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
nkcsgexi: Yeah, emitting textual interface and enabling library-evolution are tied together. Using one without the other should raise an alarm.

nkcsgexi:
@swift-ci please test

cltnschlosser (approved):

PR #1052, opened 2022-04-10
README: document how to build swift-driver on Windows

compnerd:
Although the toolchain distribution builds use CMake to build on Windows
(for bootstrapping the toolchain distribution), the development of
swift-driver on Windows requires building with swift-package-manager.
Update the documentation to explain how to build on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,29 @@ The `swift-driver` project is a new implementation of the Swift compiler driver 
 
 **Note:** Currently, swift-driver is only compatible with trunk development snapshots from [swift.org](https://swift.org/download/#snapshots).
 
-The preferred way to build `swift-driver` is to use the Swift package manager:
+The preferred way to build `swift-driver` is to use the Swift package manager.
+
+On most platforms you can build using:
 
 ```
 $ swift build
 ```
+
+However, on Windows, some additional work must be done by the developer.
+
+Due to the default version of swift-tools-support-core that `Package.resolved` references, we must first update the package dependencies.
+
+```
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

cltnschlosser (approved):
Awesome, I’ll have to try this out sometime. I gave up trying to build the toolchain on windows when I tried a few months back.

does `swift test` with those additional parameters, also work?

compnerd:
Ah, alas I forgot that I never documented this in the README :-(.  The swift driver bring up was all done with swift-package-manager (the only time to use CMake is when you are first building yourself a toolchain).  Yes, swift-test works with the additional arguments ... in theory.

Currently, there have been some regressions (some of my own doing 😬).  I would like to fix up the issues which I have caused with the recent rework of `Path` from swift-tools-support-core.  There were a few other failures that I found due to the toolchain packaging having issues which I was unaware of.  https://github.com/apple/swift-installer-scripts/pull/116 should help resolve the missing swift-api-digester which is required for the test suite.  After that, there are ~14 failures due to path mismatches which I need to fix up to make swift-driver work with the latest swift-tools-support-core.  Subsequently, there are 2 failures that I do not yet understand in `SwiftDriverTests.testSupportedFeatureJson`.

I would like to add that all of the 24 failures are regressions and that previously `swift test` with the documented flags would have given you a clean bill of health.

Hopefully @shahmishal will be able to help get CI enabled for Windows here.

artemcm (approved):

PR #1053, opened 2022-04-10
Tests: correct invalid tests

compnerd:
These tests specified the Darwin triple unconditionally, resulting in
the test always being run in Darwin mode.  However, the ABI output is
not emitted on all platforms, which resulted in an out-of-bounds access
in the output.  Remove the explicit target triple, repairing the tests
on Windows.


compnerd:
@swift-ci please test

cltnschlosser (approved):

cltnschlosser:
Looks like you need to filter out the auto link extract step for tests.

compnerd:
I missed that locally - Windows doesn't add in the auto-link-extract job either.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #1054, opened 2022-04-10
SwiftDriver: unlearn about `DEVELOPER_DIR`

compnerd:
The toolchain should not really consult `DEVELOPER_DIR`.  While this was
initially useful, as the toolchain distribution grows and matures, it is
desirable to restore the ability to have disjoint and re-rooted installs
on Windows.  As a result, `DEVELOPER_DIR` would no longer be as valuable
for identifying the locations as the toolchain may be installed separate
from the SDKs.  This is inline with swift-package-manager which now also
has unlearnt this variable. `SDKROOT` remains as a means for the user to
specify the SDK implicitly through the environment.


compnerd:
@swift-ci please test

cltnschlosser (approved):

artemcm (approved):

PR #1060, opened 2022-04-13
swiftinterface verification: disable verification for modules with generated headers

xymus:
Swiftinterface verification is incompatible with modules using generated headers. Let's disable the by default verification in such a case.

---

Also fix some indentation issues and clarify the many conditions for the verification to be active.


xymus:
@swift-ci Please test

artemcm (approved):

tshortli (approved):

PR #1061, opened 2022-04-13
Tests: simplify test condition (NFCI)

compnerd:
Rather than iterating over all the jobs (even if there are only two) to
find a matching job, extract the module emission job and verify that the
assertions hold over the job.


compnerd:
@swift-ci please test

cltnschlosser (approved):

PR #1062, opened 2022-04-13
[5.7] swiftinterface verification: disable verification for modules with generated headers

xymus:
Swiftinterface verification is incompatible with modules using generated headers. Let's disable the by default verification in such a case.

Cherry-pick of #1060
rdar://91711963


xymus:
@swift-ci Please test

artemcm (approved):

nkcsgexi (approved):

PR #1063, opened 2022-04-18
Tests: address TODO for Windows

compnerd:
Now that the frontend properly consults the architecture independent
path for the Swift modules, we can address the TODO that was left to
be addressed.  This requires apple/swift#42419 and
apple/swift-installer-scripts#119


compnerd:
@swift-ci please test

artemcm (approved):

PR #1072, opened 2022-04-21
Add and propagate the -strict-concurrency flag 

DougGregor:
Fixes the driver part of rdar://91930849


DougGregor:
@swift-ci please test

PR #1074, opened 2022-04-21
Driver: correct tokenisation on Windows

compnerd:
We would tokenise the entire file as a single line rather than
tokenising each line.  Match the semantics of the function.


artemcm (approved):

compnerd:
@swift-ci please test

PR #1075, opened 2022-04-21
Driver: correct tokenisation on Windows

compnerd:
We would tokenise the entire file as a single line rather than
tokenising each line.  Match the semantics of the function.


compnerd:
@swift-ci please test

artemcm (approved):
This is a useful and safe change to get into the release.

compnerd:
Nominating this change for 5.7.

This is a relatively low risk change for Windows that improves the response file tokenisation.  This is on a Windows specific path so the risk to the other platforms is zero as the code is gated on an `os` check.  The change is already on main, but improves the handling of response files which are quoted properly.

PR #1078, opened 2022-04-26
Set up an externally installed JSON config file for blocking driver features for given module names

nkcsgexi:


nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -438,6 +439,46 @@ extension Driver {
...
+    }
+    let toolchains = "Toolchains"
+    guard swiftPath.components.contains(toolchains) else {
+      return nil
+    }
+    while swiftPath.basename != toolchains  {
+      swiftPath = swiftPath.parentDirectory
+    }
+    assert(swiftPath.basename == toolchains)
+    return swiftPath.appending(component: "XcodeDefault.xctoolchain")
```
compnerd, line 455: Hmm, this would unintentionally return a non-existent path on Windows.  Does it make sense to just stop one level Egor and take the toolchain name from the path?  It also would be safer on macOS with a custom toolchain.

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -438,6 +439,46 @@ extension Driver {
...
+      .appending(component: "usr")
+      .appending(component: "local")
+      .appending(component: "lib")
+      .appending(component: "swift")
+      .appending(component: "adopter_configs.json")
```
compnerd, line 460: Is this something that is new in the Swift repo?

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -438,6 +439,46 @@ extension Driver {
     return try mergeModuleJob(inputs: moduleInputs, inputsFromOutputs: moduleInputsFromJobOutputs)
   }
 
+  func getDefaultAdopterConfigPath() -> AbsolutePath? {
```
artemcm, line 442: Nit: I think it would be nice for the name of this method to indicate that this is a default **Xcode** config path.

Sources/SwiftDriver/Jobs/Planning.swift
artemcm: I agree with @compnerd, I think that by-default this ought to work with the toolchain the driver is configured against, including the swift.org snapshots, for example.

nkcsgexi: I've added new tests for non-existent paths. These non-existent paths should work fine.

artemcm (approved):

Sources/SwiftDriver/Jobs/Planning.swift
artemcm, line 442: Nit: If we do target this specifically to Xcode toolchain, could we please make the name reflect this?

nkcsgexi: Makes sense. Will fix.

PR #1080, opened 2022-04-26
Add a utility function for collecting modules under a given config key. NFC

nkcsgexi:


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -479,6 +479,13 @@ extension Driver {
     return []
   }
 
+  @_spi(Testing) public static func getModule(withConfigKey: String, _ configs: [AdopterConfig]) -> Set<String> {
```
artemcm, line 482: Given that this is part of Driver public interface, could we give this method a more-descriptive name please?
Something like `getAllConfiguredModules(with key: String, ...`.

artemcm (approved):

Sources/SwiftDriver/Jobs/Planning.swift
nkcsgexi: Great point. Fixed.

nkcsgexi:
@swift-ci please test

PR #1082, opened 2022-04-27
[5.7 Sooner] Set up an externally installed config file to specify block list for certain driver features

nkcsgexi:
Instead of embedding a block list of module names to disable a certain feature, we should set up an externally installed config file for that purpose. Therefore, we don't need to rebuild swift-driver when the block list needs an update.

rdar://92407381


nkcsgexi:
@swift-ci please test

artemcm (approved):

PR #1083, opened 2022-04-27
Use python3 not legacy python

llvm-beanz:
On most systems python is python2, and on macOS this fails if you don't have a python2 install or a python3 install that is overriding python.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -1,6 +1,4 @@
-#!/usr/bin/env python
-
-from __future__ import print_function
+#!/usr/bin/env python3
 
 import argparse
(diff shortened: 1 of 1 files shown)


artemcm (approved):
Thanks! 
Maybe overkill, but just to make sure we don't affect PR and toolchain bots, could you please run a cross-repo test?
https://github.com/apple/swift-driver#testing-against-swift-compiler-trunk

dduan:
done in https://github.com/apple/swift-driver/pull/1087

artemcm:
The just-merged https://github.com/apple/swift-driver/pull/1087 seems to be a super-set of this change, so I believe this PR is no longer needed.

PR #1084, opened 2022-04-28
Tests: adjust tests for improvements to path normalization in TSC

compnerd:
When building with apple/swift-tools-support-core#310, we normalize more
paths which results in test failures.  Update the tests to accommodate
the new behaviour.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1085, opened 2022-04-28
build: re-resolve dependencies

compnerd:
This re-resolves the package, which updates the unpinned
swift-tools-support-core and swift-llbuild repositories.  This should
have no impact on other platforms, but enables building on Windows
without the need to explicitly re-resolve the package.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1087, opened 2022-04-29
Update build-script-helper to python3

keith:
This was just a run of 2to3 on this script.

Fixes https://github.com/apple/swift/issues/58528

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -1,6 +1,4 @@
-#!/usr/bin/env python
-
-from __future__ import print_function
+#!/usr/bin/env python3
 
 import argparse
 from distutils import file_util
@@ -66,10 +64,10 @@ def swiftpm(action, swift_exec, swiftpm_args, env=None):
   subprocess.check_call(cmd, env=env)
 
 def swiftpm_bin_path(swift_exec, swiftpm_args, env=None):
-  swiftpm_args = list(filter(lambda arg: arg != '-v' and arg != '--verbose', swiftpm_args))
+  swiftpm_args = [arg for arg in swiftpm_args if arg != '-v' and arg != '--verbose']
   cmd = [swift_exec, 'build', '--show-bin-path'] + swiftpm_args
   print(' '.join(cmd))
-  return subprocess.check_output(cmd, env=env).strip()
+  return subprocess.check_output(cmd, env=env).strip().decode("UTF-8")
 
 def get_swiftpm_options(args):
   swiftpm_args = [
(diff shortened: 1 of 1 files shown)


artemcm:
Please run a cross-repo test with this change:
https://github.com/apple/swift-driver#testing-against-swift-compiler-trunk

keith:
https://github.com/apple/swift/pull/58535

keith:
@swift-ci please test

dduan (approved):

keith:
@swift-ci please test

keith:
@swift-ci please test

Utilities/build-script-helper.py
```diff
@@ -153,12 +151,7 @@ def add_rpath(rpath, binary, verbose):
     print(stdout)
 
 def should_test_parallel():
-  if platform.system() == 'Linux':
-    distro = platform.linux_distribution()
-    if distro[0] != 'Ubuntu':
-      # Workaround hang in Process.run() that hasn't been tracked down yet.
-      return False
-  return True
```
keith, line 161: On my linux 20.04 machine, `platform.linux_distribution()[0]` actually returns `debian`, so I don't think this was doing what it intended to. At least not for newer ubuntu versions.

dduan: Also this is a deprecated Python API. Probably should use something like the external command `lsb_release -i`

keith: Yea that was my last attempt, but then I realized I'd have to hardcode the path to it because it failed with a FileNotFound, and so I thought this might be acceptable given that I don't think it was doing the right thing up front

dduan (approved):

dduan:
@swift-ci test

keith:
All green on the integration pr now. @artemcm what do you think about the Linux conditional removal? Do you know if removing that entirely and doing an integration PR would show if the issue was fixed? Or is it more of a subtle race?

artemcm:
> All green on the integration pr now. @artemcm what do you think about the Linux conditional removal? Do you know if removing that entirely and doing an integration PR would show if the issue was fixed? Or is it more of a subtle race?

I really don't know the context here, this snippet seems to have been in place since the day this script was written. 

Since we are switching Python versions, I'd be tempted to at least check if we can remove this restriction altogether? If it passes integration testing I think it's worth a shot at removing it, then we can keep an eye on the bots for a bit.

artemcm (approved):

keith:
I'll merge this to unblock folks and submit another one to test

keith:
https://github.com/apple/swift-driver/pull/1092

PR #1056, opened 2022-04-12
Create discovered dependencies file for emit-module jobs

BenchR267:
Emit-module jobs might have different discovered dependencies and run in a different context where they need to get invalidated separately from the compile jobs.
With the new key "emit-module-dependencies" in the output file map the build system can specify that it expects a makefile-style dependency file at the given path.

rdar://91574616

PR #1088, opened 2022-05-01
Driver: provide a more suitable default for `emit-module.d`

compnerd:
Provide a fallback default for the `emit-module.d` file.  It has been
observed on Windows and Android that builds subsequent to #1056 would
emit files into the root of the source tree (specifically, `pwd`). If
there is no explicit location specified for the new output, and given
that there is no primary output associated with the command, the path
that defaults is simply a singular file name component, emitting that
file into the location that the driver was executed from (which comes
out to the root of the package for commandline invocations), dirtying
the source tree.

Rather than sinking the knowledge for the default into
`existingOutputForSingleInput`, where we have no access to the module
name, emit the logic inline and derive a name based on the Swift module
and place it as a peer.  This should help ensure that the serialized
diagnostics do not end up committed accidentally.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -3011,10 +3011,21 @@ extension Driver {
     
     // Emit-module discovered dependencies are always specified as a single-output
     // file
-    if type == .emitModuleDependencies,
-      let singleOutputPath = outputFileMap?.existingOutputForSingleInput(
-           outputType: type) {
-      return singleOutputPath
+    if type == .emitModuleDependencies {
+      if let path = outputFileMap?.existingOutputForSingleInput(outputType: type) {
+        return path
+      }
+
+      // If an explicit path is not provided by the output file map, attempt to
+      // synthesize a path from the master swift dependency path.  This is
+      // important as we may other emit this file at the location where the
+      // driver was invoked, which is normally the root of the package.
+      if let path = outputFileMap?.existingOutputForSingleInput(outputType: .swiftDeps) {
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @buttaface

artemcm (approved):
Thanks!

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -3011,10 +3011,21 @@ extension Driver {
     
     // Emit-module discovered dependencies are always specified as a single-output
     // file
-    if type == .emitModuleDependencies,
-      let singleOutputPath = outputFileMap?.existingOutputForSingleInput(
-           outputType: type) {
-      return singleOutputPath
+    if type == .emitModuleDependencies {
```
BenchR267, line 3014: It seems that we're running into the same problem for `.emitModuleDiagnostics` here. Can we get that added in this check to do something similar?

compnerd: Hmm, I think that this is one of those cases of CS has 3 numbers: 0, 1, N.  I think that if we are going to generalise this further, I'd like to run through a bit more testing.  I think I prefer to do that in a follow up change.

buttaface:
@compnerd, confirmed that this is fixed in the latest trunk snapshot, please merge into 5.7 too.

compnerd:
Thanks for the verification, and a double thank you for the reminder - I completely forgot about the 5.7 PR 🤦‍♂️

PR #1089, opened 2022-05-01
Tests: ensure that the path will be relative

compnerd:
There is no guarantee that the path between the temporary directory and
current working directory can be represented with a relative path.  This
is particularly important on Windows where the paths may reside on
different volumes, which have separate roots.  Alter the two instances
in the test to use the current directory for the temporary storage, and
ensure that we clean up afterwards as we are likely in the source tree.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1090, opened 2022-05-02
Driver: use a proper error rather than `fatalError`

compnerd:
`fatalError` is more akin to an assertion rather than a "condition
failed, please exit".  The improper use here would, upon a typo on
Windows, induce the Doctor to try to categorise the fault and then
emit a minidump.  This is both frustrating and unnecessary.  Add a
new error case and use it to emit the message and exit gracefully.


compnerd:
@swift-ci please test

artemcm (approved):

cltnschlosser (approved):

PR #1093, opened 2022-05-03
Driver: further generalise #1088

compnerd:
It seems that a similar problem has impacted the generation of
serialised diagnostics for modules.  Follow the adage of CS inherently
having 3 numbers: 0, 1, N and generalise the path to a penultimate
fallback.  We should always have the swift dependency information on
hand to locate the build directory as that is the "master dependency".
This should hopefully prevent accidental cases of files leaking into the
current directory.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -3014,21 +3014,9 @@ extension Driver {
     
     // Emit-module discovered dependencies are always specified as a single-output
     // file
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

cltnschlosser (approved):
Since this is only for supplementary outputs, it should be safe to use that location as a fallback for all of them.

compnerd:
@artemcm @nkcsgexi - is this something we should consider pulling into 5.7?

PR #1095, opened 2022-05-04
[5.7] Driver: use a proper error rather than `fatalError`

compnerd:
`fatalError` is more akin to an assertion rather than a "condition
failed, please exit".  The improper use here would, upon a typo on
Windows, induce the Doctor to try to categorise the fault and then
emit a minidump.  This is both frustrating and unnecessary.  Add a
new error case and use it to emit the message and exit gracefully.


compnerd:
Nominating this change for the 5.7 release.  It is a super low risk change - the change is specifically to the error handling path, which is meant to result in death.  This is important as Windows is planning on moving entirely to the swift-driver in 5.7 and this results in a debug prompt if the user typos a command.

artemcm (approved):
LGTM, thanks @compnerd.
cc @nkcsgexi

PR #1097, opened 2022-05-05
Parallel tests has a use-after-free in swift-test

etcwilde:
swift-test has a use-after-free/segfault when run in parallel mode
causing crashes. This disables parallel testing.

This results in frequent spurious failures on https://ci.swift.org/job/oss-swift-package-centos-7/


etcwilde:
@swift-ci please test

artemcm:
Let's use this instead:
https://github.com/apple/swift-driver/pull/1098
which will only disable parallel tests on Linux at least.

etcwilde:
Going with https://github.com/apple/swift-driver/pull/1098 instead.

PR #1101, opened 2022-05-11
[5.7] ParsedOptions: save unknown driver flags that are recognizable for swift-frontend

nkcsgexi:
This is to prevent driver being lagging behind swift-frontend. For any unknown
driver flags that are recognizable to the swift compiler, we should save these flags
and pass them down to the compiler because the compiler is ready to take them.


nkcsgexi:
@swift-ci please test

artemcm (approved):

owenv (approved):

PR #1102, opened 2022-05-13
Remove Noisy Remark

CodaFi:
This remark is basically tripped by every swiftmodule in the SDK for
most projects. As it doesn't appear in tests anywhere and it provides
very little signal, best to strike it for now. If we discover it has
some diagnostic utility we can always revert this and add some tests.


CodaFi:
@swift-ci test

artemcm (approved):

PR #1108, opened 2022-05-23
[5.7-04182022] Provide a more-suitable default for emit-module auxiliary files

artemcm:
Cherry-pick of https://github.com/apple/swift-driver/pull/1088 & https://github.com/apple/swift-driver/pull/1093
----------------------------------------------------------------------

- Explanation: Provide a fallback default for the `emit-module.d` and `emit-module.dia` files. If there is no explicit location specified for the new output, and given that there is no primary output associated with the command, the path that defaults is simply a singular file name component, emitting that file into the location that the driver was executed from (which comes out to the root of the package for commandline invocations), dirtying the source tree. Rather than sinking the knowledge for the default into `existingOutputForSingleInput`, where we have no access to the module name, emit the logic inline and derive a name based on the Swift module and place it as a peer. This should help ensure that the serialized diagnostics do not end up committed accidentally.
- Scope of issue: Users of SwiftPM will see these spurious build artifacts left over in their working (e.g. source) directory. 
- Risk: Low

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -3017,10 +3017,21 @@ extension Driver {
     
     // Emit-module discovered dependencies are always specified as a single-output
     // file
-    if type == .emitModuleDependencies,
-      let singleOutputPath = outputFileMap?.existingOutputForSingleInput(
-           outputType: type) {
-      return singleOutputPath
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci please test

nkcsgexi (approved):

artemcm:

airspeedswift (approved):

PR #1111, opened 2022-06-12
Switch to using clang as linker on darwin

keith:
This allows us to remove lots of duplicated logic from the clang driver.


keith:
@swift-ci please test

keith:
This isn't done, but I just want to see how CI feels about it vs my local test

keith:
@swift-ci please test

keith:
@swift-ci Please test source compatibility

artemcm:
There's no source-compat testing on this repo, it needs to be requested from `swift` with cross-repo testing.

keith:
thanks, doing on https://github.com/apple/swift/pull/59414

compnerd:
Thanks for doing this!  I had mentioned to @artemcm that I would like to see this done but wanted to ensure that Apple was okay with the change.  Unifying the behaviour and relying on clang to drive the linker is definitely going to help simplify the code base, but also help the C++ interop functionality as well IMO.

keith:
@swift-ci please test

keith:
source compatibility on https://github.com/apple/swift/pull/59414 looks good, realistically im not entirely sure how we'll be 100% confident here thought

keith:
@artemcm can you take a look? 🙏🏻

artemcm (approved):
@Keith this change looks good to me, and this is a huge improvement, thank you! 
Let's merge this now and keep an eye out for any possible regressions.

artemcm:
@swift-ci please test

keith:
@swift-ci please test

Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
```diff
@@ -354,43 +183,44 @@ extension DarwinToolchain {
...
 
     // Add the SDK path
     if let sdkPath = targetInfo.sdkPath?.path {
-      commandLine.appendFlag("-syslibroot")
+      commandLine.appendFlag("--sysroot")
       commandLine.appendPath(VirtualPath.lookup(sdkPath))
     }
 
     commandLine.appendFlags(
+      "-fobjc-link-runtime",
```
artemcm, line 213: @Keith, what was the reason for always adding this flag? 
It appears it does not match previous behavior and causes `clang` to always import Foundation, whether it is needed or not.

keith: Hrm, I don't recall since it has been a while, but I agree I don't see where it clearly came from. I checked clang as well in case I was trying to mirror that and don't see it there either. I suppose I could have pulled it from a `-v` invocation of what this was doing before, but if that flag is present there I don't see where it's coming from.

Note that it does look like https://github.com/apple/llvm-project/blob/c10fcd3e640b3736a7b2442f00db8feca674648c/clang/lib/Driver/ToolChains/Darwin.cpp#L745-L746 relies on this being passed, so maybe it was required for that case?

compnerd: This really should be gated on `-enable-objc-interop` I think.  `-lSystem` is also rather odd.  This should no longer be needed as we now rely on `clang` as the driver which will add `-lSystem` as appropriate.

PR #1112, opened 2022-06-12
Handle _StringProcessing module in ExplicitModuleBuildTests

rxwei:
Handle _StringProcessing module the same way it handles _Concurrency.

To be merged with apple/swift#42611.

PR #1115, opened 2022-06-21
[5.7] Handle _StringProcessing module in ExplicitModuleBuildTests

bnbarham:
This cherry-picks bbb8e2a39abeeec9ddbadd3ce48e2588f002933d (https://github.com/apple/swift-driver/pull/1112), a small test fix for `_StringProcessing`.

-----

Handle _StringProcessing module the same way it handles _Concurrency.

To be merged with apple/swift#42611.

### DIFF
--- a/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
+++ b/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
@@ -207,6 +207,11 @@ final class ExplicitModuleBuildTests: XCTestCase {
                                             moduleId: .swift("_Concurrency"),
                                             dependencyGraph: moduleDependencyGraph,
                                             pcmFileEncoder: pcmFileEncoder)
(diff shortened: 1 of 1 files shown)


bnbarham:
@rxwei is string processing enabled by default in 5.7? I couldn't find a PR for it, but `_StringProcessing` is definitely an added module - https://ci.swift.org/job/pr-swift-driver-macos/138.

nkcsgexi (approved):
This is a test-only change.

rxwei (approved):

bnbarham:
Ah. This is definitely not needed - I checked the corresponding option in 5.7 and it isn't enabled by default. It looks like we're using the wrong toolchain for 5.7 (ie. the one from main).

bnbarham:
@rxwei looking at this change, it seems like it'd be safe to merge regardless of the swift change. Should we merge now to unblock things?

PR #1116, opened 2022-06-23
Prohibit IR Generation for 32bit watchos when deployment target > 9.0

cyndyishida:
watchOS 9 doesn't support 32bit architectures, however we should still
allow the compiler to generate swift modules for these target triples
for back deployment support.


xymus (approved):
This looks good!

cyndyishida:
@swift-ci please test

lorentey (approved):
👍🎉

PR #1117, opened 2022-06-27
Fix some correctness issues in target triple validation and improve test coverage

jakepetroules:
Fixes "off by one" type error in watchOS validation - 32-bit archs should be blocked for watchOS >= 9.0, not >= 8.7. Fixes an incorrect fallthrough where invalid tvOS archs would issue errors about 32-bit iOS archs not being supported on iOS 11 and later. Fixes labeling of "macOS" as "OS X" in an error. Fixes an issue where arm64_32 would incorrectly be blocked for watchOS >= 8.7. Generally simplifies the validation code. Improves the test coverage to cover all of these scenarios.

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -154,19 +154,18 @@ public final class DarwinToolchain: Toolchain {
   }
 
   public enum ToolchainValidationError: Error, DiagnosticData {
-    case osVersionBelowMinimumDeploymentTarget(String)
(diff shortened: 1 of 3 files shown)


artemcm:
I'd like @cyndyishida to comment on the watchOS changes though, since she's the expert here.

jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please test

jakepetroules:
@swift-ci please test

jakepetroules:
> I'd like @cyndyishida to comment on the watchOS changes though, since she's the expert here.

We worked on this together so I'm also confident in the changes in this area. Since she's OOO, I want to fix the regression from the previous PR right away.

PR #1119, opened 2022-06-28
Explicitly copy defsIDependUpon into an appropriately sized array when initializing SourceFileDependencyGraph.Node

owenv:
Previously, the removeAll call below the initializer would trigger CoW. If the scratch buffer was larger than the count, this would result in an unnecessarily large allocation. In a very large source file, the source-file-provides-implementation node may have multiple orders of magnitude more edges compared to any other node in the graph, which could lead to excessive memory usage and a large slowdown deallocating the array storage when tearing down the graph.

I tested this using the swiftdeps from an artificially generated swiftdeps with 50k empty struct declarations. Before this change, it took 31s to load and destroy, after this change it only took 0.25s.

rdar://95648282


owenv:
@swift-ci please test

owenv:
@swift-ci please test

artemcm (approved):
Awesome catch!

PR #1120, opened 2022-06-28
fix SIGINT handler in swift-frontend

tomerd:
motivation: fix regression where SIGINT would not terminate frontend processes

changes:
* call DispatchSourceSignal::resume
* call signal(SIGINT, SIG_IGN)

rdar://95862616


tomerd:
@swift-ci test

artemcm (approved):

PR #1126, opened 2022-06-28
Reapply "Explicitly copy defsIDependUpon into an appropriately sized array when initializing SourceFileDependencyGraph.Node"

owenv:
Reapplying the reverted PR with an extra type annotation which will hopefully allow it to compile with hosttools.

Previously, the removeAll call below the initializer would trigger CoW. If the scratch buffer was larger than the count, this would result in an unnecessarily large allocation. In a very large source file, the source-file-provides-implementation node may have multiple orders of magnitude more edges compared to any other node in the graph, which could lead to excessive memory usage and a large slowdown deallocating the array storage when tearing down the graph.

I tested this using the swiftdeps from an artificially generated swiftdeps with 50k empty struct declarations. Before this change, it took 31s to load and destroy, after this change it only took 0.25s.

rdar://95648282

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/SourceFileDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/SourceFileDependencyGraph.swift
@@ -227,10 +227,14 @@ extension SourceFileDependencyGraph {
       private mutating func finalizeNode() throws {
         guard let key = key else {return}
 
+          let defsIDependUpon: [Int] = Array(unsafeUninitializedCapacity: defsNodeDependUpon.count) { destinationBuffer, initializedCount in
+            _ = destinationBuffer.initialize(from: defsNodeDependUpon)
+            initializedCount = defsNodeDependUpon.count
+        }
         let node = try Node(key: key,
                             fingerprint: fingerprint?.intern(in: internedStringTable),
                             sequenceNumber: nodeSequenceNumber,
-                            defsIDependUpon: defsNodeDependUpon,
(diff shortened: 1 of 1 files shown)


owenv:
Cross-repo testing here: https://github.com/apple/swift/pull/59758

artemcm (approved):

Sources/SwiftDriver/IncrementalCompilation/SourceFileDependencyGraph.swift
```diff
@@ -227,10 +227,14 @@ extension SourceFileDependencyGraph {
       private mutating func finalizeNode() throws {
         guard let key = key else {return}
 
+          let defsIDependUpon: [Int] = Array(unsafeUninitializedCapacity: defsNodeDependUpon.count) { destinationBuffer, initializedCount in
+            _ = destinationBuffer.initialize(from: defsNodeDependUpon)
+            initializedCount = defsNodeDependUpon.count
+        }
```
cltnschlosser, line 233: Is this the easiest way of forcing a copy?

cltnschlosser (approved):

Sources/SwiftDriver/IncrementalCompilation/SourceFileDependencyGraph.swift
owenv: AFAIK, this is the best way to force a copy into an array of the exact size we want (it won't be mutated after the node is initialized)

PR #1128, opened 2022-07-11
Update incremental build hash to encode option arguments and ordering

owenv:
Otherwise, changing the argument to an option which affects incremental builds, or the relative ordering of multiple options affecting incremental builds won't correctly rebuild all of the sources.

rdar://96556787


owenv:
@swift-ci please test

artemcm (approved):
Thank you!

PR #1137, opened 2022-07-14
Remove non-fatal test error conditions in Linux testing

artemcm:
In Linux testing (e.g. in our CI) these currently generate non-fatal 'error:' messages. This makes human scanning of CI logs more difficult than it should be


artemcm:
@swift-ci please test

owenv (approved):

PR #911, opened 2021-11-12
Add `-print-target-info` libSwiftScan entry-points

artemcm:
This change adds the relevant API entry-points to `SwiftScan`.
It does not yet transition the driver's current use-sites where we shell out a `swift-frontend` process, but should make it easy to do so.

The library itself is gaining these entry-points here:
https://github.com/apple/swift/pull/40160

PR #1138, opened 2022-07-15
Revert "Add `-print-target-info` libSwiftScan entry-points"

artemcm:
Putting up a revert in case this is the culprit behind failures like:
https://ci.swift.org/job/swift-PR-Linux-smoke-test/2536/

Reverts apple/swift-driver#911


artemcm:
@swift-ci please test

PR #1143, opened 2022-07-20
Switch to using clang as linker on darwin

DougGregor:
This allows us to remove lots of duplicated logic from the clang driver.


keith:
Thanks for fixing and re-reverting!

DougGregor:
@swift-ci please test

artemcm (approved):

PR #1144, opened 2022-07-20
Add supplementary output path handling for extracted compile-time-known values

artemcm:
Driver equivalent to the compiler change in https://github.com/apple/swift/pull/60139.


artemcm:
@swift-ci please test

xymus (approved):

PR #1151, opened 2022-08-07
SwiftDriverTests: adjust module count on Windows

compnerd:
A recent refactoring for Windows increased the module count for the
system modules and this resulted in the test suite failing due to the
difference.  This is technically backwards incompatible and expects that
the most recent toolchain is in use but it seems unclear how to account
for that easily.


compnerd:
@swift-ci please test

compnerd:

compnerd:
apple/swift-corelibs-foundation#4621

compnerd:
apple/swift-corelibs-foundation#4622

DougGregor (approved):

DougGregor:
We could make this a looser check (between 10 and 15 or whatever), but this fix to unblock the refactoring LGTM.

compnerd:
Yeah, I suppose that we could create the window, but I like the specificity - it ensures that we do not accidentally slip in modules (or drop modules).  Stronger guarantees vs easier maintenance tradeoffs 🤷‍♂️

PR #1152, opened 2022-08-11
SwiftDriver: introduce `-nostartfiles` option

compnerd:
This replicates the `-nostartfiles` option from clang and MSVC.  This
allows the registrar from being linked against, which is important for
the case where the swift driver is used for non-Swift targets (i.e.
C/C++).


compnerd:
@swift-ci please test

compnerd:
@DougGregor ... turns out that I had a small mixup in the spelling, but this is roughly the thought that I had.

artemcm (approved):

PR #1154, opened 2022-08-16
[Explicit Modules] Pass '-explicit-interface-module-build' to interface build jobs.

artemcm:
Companion change to https://github.com/apple/swift/pull/60556. 

This allows us to gather all the required flags for compilation of a given interface during the dependency scan, and then use those flags directly during the build, instead of once again reading out flags from the interface file and instantiating a new compiler subInstance when building said interface.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -15,7 +15,7 @@
         "repositoryURL": "https://github.com/apple/swift-llbuild.git",
         "state": {
           "branch": "main",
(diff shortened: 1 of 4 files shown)


artemcm:
@swift-ci test

artemcm:
Cross-repo tests are looking green.

xymus (approved):

PR #1155, opened 2022-08-18
Re-enable 'testModuleAliasingPrebuiltWithScanDeps'.

artemcm:
It appears the test setup resulted in us being unable to find modules when passing them as explicit dependencies to depending compiler invocations. Setting an explicit module cache path helps with that.

Resolves rdar://88073675


artemcm:
@swift-ci test

nkcsgexi (approved):

PR #1156, opened 2022-08-18
Separate two kinds of emit-module jobs into different categories: source emit-module and explicit dependency interface build

artemcm:
`.emitModule` is re-used for two kinds of jobs:
- Build all sources of a target into a binary module
- In Explicit Module Builds, compile a dependency module interface into a binary module

It is worthwhile to treat these separate kinds of jobs separately.

Resolves rdar://98842129


artemcm:
@swift-ci test

owenv (approved):

PR #1157, opened 2022-08-18
On interrupt, after handling the compilation process set termination, restore the SIGINT default action.

artemcm:
Otherwise after https://github.com/apple/swift-driver/pull/1120 we are now falling through to `fatalError` on interrupt.  

Resolves rdar://98584872


artemcm:
@swift-ci test

owenv (approved):
thanks!

PR #1153, opened 2022-08-14
Work around lld 13+ issue with --gc-sections for ELF by adding -z nostart-stop-gc

buttaface:
This is a Swift version of apple/swift#60544, which will fix apple/swift#60406. I ran the testsuite on linux x86_64 with this pull without a problem.

PR #1158, opened 2022-08-19
[5.7] Work around lld 13+ issue with --gc-sections for ELF by adding -z nostart-stop-gc

buttaface:
Cherrypick of #1153

__Explanation:__ [lld 13 changed how `--gc-sections` works last year](https://reviews.llvm.org/D96914), which was then [flipped on by default](https://reviews.llvm.org/rG6d2d3bd0a61f5fc7fd9f61f48bc30e9ca77cc619), and causes all Swift code on ELF platforms not to link with lld 13 and 14 anymore if stripping is enabled.

__Scope:__ Only affects lld for these ELF platforms

__SR Issue:__ apple/swift#60406

__Risk:__ low, since it only affects ELF platforms for this one linker

__Testing:__ I [just built the last August 2 source snapshot of the 5.7 toolchain for Android AArch64](https://github.com/buttaface/termux-packages/actions/runs/2886129294), where lld 14 is the default linker, with [this patch](https://github.com/buttaface/termux-packages/blob/swift-devel/packages/swift/swift-lld-gc.patch) and it works well.
 
__Reviewer:__ @compnerd

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -84,6 +84,17 @@ extension GenericUnixToolchain {
         #else
         commandLine.appendFlag("-fuse-ld=\(linker)")
         #endif
+        // Starting with lld 13, Swift stopped working with the lld
+        // --gc-sections implementation for ELF, unless -z nostart-stop-gc is
+        // also passed to lld:
+        //
+        // https://reviews.llvm.org/D96914
+        if linker == "lld" || linker.hasSuffix("ld.lld") {
+          commandLine.appendFlag(.Xlinker)
+          commandLine.appendFlag("-z")
+          commandLine.appendFlag(.Xlinker)
+          commandLine.appendFlag("nostart-stop-gc")
+        }
       }
 
       // Configure the toolchain.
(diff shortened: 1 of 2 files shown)


keith:
@swift-ci test

Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
```diff
@@ -84,6 +84,17 @@ extension GenericUnixToolchain {
         #else
         commandLine.appendFlag("-fuse-ld=\(linker)")
         #endif
+        // Starting with lld 13, Swift stopped working with the lld
```
DougGregor, line 87: Is this safe to do with earlier versions of `lld`, or do we need to perform a version check?

buttaface: It is not, [as I mentioned on the bug report a couple weeks ago](https://github.com/apple/swift/issues/60406#issuecomment-1211429665), "that will break using lld 12 and earlier, since they don't know the nostart-stop-gc flag and error, but it's been almost a year since lld 13 came out, so should be fine." I don't know of a way to do a version check short of having Swift invoke `lld --version` first, but I figured a year since lld 12 was obsoleted was okay.

DougGregor: Alright, I'm okay with the ~1 year for lld users. If we run into configuration problems, then we can revisit the idea of calling `lld --version`. Thank you!

DougGregor (approved):

PR #1164, opened 2022-08-29
Pass down SDK name to the compiler by default

xymus:
The compiler uses the -target-sdk-name flag information to restrict which swiftmodules can be loaded and which should be ignored. This avoids preventable crash when swiftmodules are copied between mismatching SDKs.

rdar://83104265


xymus:
@swift-ci Please test

artemcm (approved):

PR #1165, opened 2022-08-29
Add ability to override/disable response file use in ArgsResolver

artemcm:
This change deprecates `ArgsResolver` API that takes a boolean `forceResponseFiles` in favor of a new API that takes an `ResponseFileHandling` enum. This enum allows the previous behaviors of `.forced` and `.heuristic` (`true` and `false`, respectively of `forceResponseFiles`), and adds an option to override disable use of response files completely (`.disabled`).

This is useful in the example of dependency scanning when it is done via a C interface to a compiler library: even for exceedingly-long command-line argument lists, using response files is not necessary since we just pass in an address of an already-allocated memory buffer containing the argument list, directly.

Resolves rdar://99296444

### DIFF
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -1485,7 +1485,7 @@ final class SwiftDriverTests: XCTestCase {
       XCTAssertTrue(jobs.count == 1 && jobs[0].kind == .interpret)
       let interpretJob = jobs[0]
       let resolver = try ArgsResolver(fileSystem: localFileSystem)
-      let resolvedArgs: [String] = try resolver.resolveArgumentList(for: interpretJob, forceResponseFiles: false)
+      let resolvedArgs: [String] = try resolver.resolveArgumentList(for: interpretJob)
       XCTAssertTrue(resolvedArgs.count == 2)
       XCTAssertEqual(resolvedArgs[1].first, "@")
       let responseFilePath = try AbsolutePath(validating: String(resolvedArgs[1].dropFirst()))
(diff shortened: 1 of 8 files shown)


artemcm:
@swift-ci test

owenv (approved):

nkcsgexi (approved):

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -1494,14 +1494,27 @@ final class SwiftDriverTests: XCTestCase {
...
+    // Needs response file + disable override
+    do {
+      var driver = try Driver(args: ["swift"] + manyArgs + ["foo.swift"])
+      let jobs = try driver.planBuild()
+      XCTAssertTrue(jobs.count == 1 && jobs[0].kind == .interpret)
+      let interpretJob = jobs[0]
+      let resolver = try ArgsResolver(fileSystem: localFileSystem)
+      let resolvedArgs: [String] = try resolver.resolveArgumentList(for: interpretJob, useResponseFiles: .disabled)
+      XCTAssertTrue(resolvedArgs.count > 2)
+      XCTAssertFalse(resolvedArgs.map { $0.hasPrefix("@") }.reduce(false){ $0 || $1 })
```
bnbarham, line 1507: I realize this is a copy paste from below, couldn't this just be `XCTAssertFalse(resolvedArgs.contains { $0.hasPrefix("@") })` though? I'd probably also extract these blocks into a local function but that doesn't really matter.

bnbarham (approved):

Tests/SwiftDriverTests/SwiftDriverTests.swift
artemcm: Cleaned up, thanks!

Issue #1167, opened 2022-09-03
`libCYaml.a` has wrong arch on M1 Mac


azarovalex:
As part of the toolchain build process on M1 mac, the [`build-toolchain` script](https://github.com/apple/swift/blob/main/utils/build-toolchain) runs the next 2 commands: 

`build-script-helper.py build` and `build-script-helper.py install`

<details>
  <summary>Full commands</summary>

  `swift-driver/Utilities/build-script-helper.py build --package-path /Users/azarovalex/Developer/OpenSource/swift-project/swift-driver --build-path /Users/azarovalex/Developer/OpenSource/swift-project/build/buildbot_osx/swiftdriver-macosx-arm64 --configuration release --toolchain /Users/azarovalex/Developer/OpenSource/swift-project/build/buildbot_osx/intermediate-install/macosx-arm64/Library/Developer/Toolchains/swift-LOCAL-2022-09-03-a.xctoolchain/usr --ninja-bin /Users/azarovalex/Developer/OpenSource/swift-project/build/buildbot_osx/ninja-build/ninja --cmake-bin /opt/homebrew/bin/cmake --cross-compile-hosts macosx-x86_64`

  `swift-driver/Utilities/build-script-helper.py install --package-path /Users/azarovalex/Developer/OpenSource/swift-project/swift-driver --build-path /Users/azarovalex/Developer/OpenSource/swift-project/build/buildbot_osx/swiftdriver-macosx-arm64 --configuration release --toolchain /Users/azarovalex/Developer/OpenSource/swift-project/build/buildbot_osx/intermediate-install/macosx-arm64/Library/Developer/Toolchains/swift-LOCAL-2022-09-03-a.xctoolchain/usr --ninja-bin /Users/azarovalex/Developer/OpenSource/swift-project/build/buildbot_osx/ninja-build/ninja --cmake-bin /opt/homebrew/bin/cmake --cross-compile-hosts macosx-x86_64 --prefix /Users/azarovalex/Developer/OpenSource/swift-project/build/buildbot_osx/intermediate-install/macosx-arm64/Library/Developer/Toolchains/swift-LOCAL-2022-09-03-a.xctoolchain/usr
`
  
</details>

Both these commands contain a `--cross-compile-hosts macosx-x86_64` flag, so I expect artifacts for both arm64 and x86 to be built. 
The arm one builds as expected, but x86 build crashes with an error: 

```
Undefined symbols for architecture x86_64:
[...]

PR #1168, opened 2022-09-05
Clean `swift_flags` array between targets in build-script-helper.py

azarovalex:
Script `Utilities/build-script-helper.py` doesn't clean `swift_flags` array while iterating on different targets, which results in the wrong `-target` flag passed on the second iteration.

Same with `base_cmake_flags` array, it should not be shared between different targets.

This bug currently blocks swift toolchain cross-compiling on M1 macs.

PR #1169, opened 2022-09-06
Set explicitly the desired Darwin architecture for all CMake builds

edymtt:
This would be needed to compile on Apple Silicon, so that CMake does not
infer the wrong architecture when we want to build for Intel (see
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5291)

Addresses rdar://99587478

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -425,6 +425,7 @@ def build_using_cmake(args, toolchain_bin, build_dir, targets):
     swift_flags.append('-target %s' % target)
     if platform.system() == 'Darwin':
       base_cmake_flags.append('-DCMAKE_OSX_DEPLOYMENT_TARGET=%s' % macos_deployment_target)
+      base_cmake_flags.append('-DCMAKE_OSX_ARCHITECTURES=%s' % target.split('-')[0])
 
     # Target directory for build artifacts
     # If building for a local compiler build, use the build directory directly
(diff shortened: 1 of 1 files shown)


edymtt:
@swift-ci please test

azarovalex:
@edymtt big thanks for trying to fix the toolchain build👍

I tried a similar fix while was trying to fix https://github.com/apple/swift-driver/issues/1167 but for some reason, CMake didn't allow passing `CMAKE_OSX_ARCHITECTURES` flag to the project containing Swift sources, maybe something was wrong with my setup, I've just retested your solution and it works like a charm, I can close #1167.

Please note that this script has another problem I addressed here: https://github.com/apple/swift-driver/pull/1168, this is not yet merged.

artemcm:
@edymtt the CI here does not run the CMake build, so please verify it independently either locally or via cross-repo testing.

artemcm (approved):

edymtt:
Did the verification both locally and at https://github.com/apple/swift/pull/60901

Checked that the generated toolchain is still building swift-driver fat

```
% lipo -archs Library/Developer/Toolchains/swift-PR-60901-291.xctoolchain/usr/bin/swift-driver
x86_64 arm64
% lipo -archs Library/Developer/Toolchains/swift-PR-60901-291.xctoolchain/usr/bin/swift-help  
x86_64 arm64
```

PR #1171, opened 2022-09-07
Replace Foundation.Date with TimePoint

CodaFi:
A TimePoint functions like a Duration, but for products that cannot yet use the niceties in Swift 5.7 like the Swift driver. This type carries a 64-bit second value and a 32-bit nanosecond value. Like Duration, it does not represent any one OS clock value, but is rather just a "point in time".

Using this type, we can simplify and clean up quite a few places that were trafficking in Foundation.Date and prepare for a future world where we have Duration too. The driver uses time point values to compare file modification times to each other - hence why we can get away with ignoring the epoch. The one oddity is the build record, which serializes time point values. But build records, like the rest of the incremental state, are not actually meant to leave a given user's machine.

### DIFF
--- a/Sources/SwiftDriver/Utilities/DateAdditions.swift
+++ b/Sources/SwiftDriver/Utilities/DateAdditions.swift
@@ -10,27 +10,127 @@
 //
 //===----------------------------------------------------------------------===//
 
-import struct Foundation.Date
-import protocol Foundation.LocalizedError
-import func Foundation.floor
+#if os(Windows)
+import WinSDK
+#elseif os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
+import Darwin
+#else
+import Glibc
+#endif
 
-public extension Date {
-  init(legacyDriverSecsAndNanos secsAndNanos: [Int]) throws {
-  enum Errors: LocalizedError {
-    case needSecondsAndNanoseconds
+/// Represents a time point value with nanosecond precision.
+///
+/// - Warning: The accuracy of measured `TimePoint` values is an OS-dependent property.
+///            `TimePoint` does not correct for OS-level differences in e.g.
+///             clock epochs. This makes it unsuitable for serialization in
+///             products that are expected to transit between machines.
+public struct TimePoint: Equatable, Comparable, Hashable {
+  public var seconds: UInt64
+  public var nanoseconds: UInt32
+
+  public init(seconds: UInt64, nanoseconds: UInt32) {
+    self.seconds = seconds
+    self.nanoseconds = nanoseconds
   }
-  guard secsAndNanos.count == 2 else {
-    throw Errors.needSecondsAndNanoseconds
+
+  public static func < (lhs: TimePoint, rhs: TimePoint) -> Bool {
+    return (lhs.seconds, lhs.nanoseconds) < (rhs.seconds, rhs.nanoseconds)
+  }
+}
+
+extension TimePoint {
+  public static func seconds(_ value: Int) -> TimePoint {
+    precondition(value >= 0,
+                 "Duration value in seconds is \(value), but cannot be negative")
+    return TimePoint(seconds: UInt64(value), nanoseconds: 0)
+  }
+
+  public static func nanoseconds(_ value: Int) -> TimePoint {
+    precondition(value >= 0,
+                 "Duration value in nanoseconds is \(value), but cannot be negative")
+    let (seconds, nanos) = value.quotientAndRemainder(dividingBy: TimePoint.nanosecondsPerSecond)
+    return TimePoint(seconds: UInt64(seconds),
+                     nanoseconds: UInt32(nanos))
+  }
+}
+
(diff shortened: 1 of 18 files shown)


CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test Windows

artemcm (approved):
Fantastic, thank you!

Sources/SwiftDriver/Utilities/DateAdditions.swift
```diff
@@ -10,27 +10,132 @@
...
+    // Add raw operands
+    var seconds = lhs.seconds + rhs.seconds
+    var nanos = lhs.nanoseconds + rhs.nanoseconds
+    // Normalize nanoseconds
+    if nanos >= TimePoint.nanosecondsPerSecond {
+      nanos -= UInt32(TimePoint.nanosecondsPerSecond)
+      seconds += 1
+    }
+    return TimePoint(seconds: seconds,
+                    nanoseconds: nanos)
```
compnerd, line 72: nit: alignment seems odd

Sources/SwiftDriver/Utilities/DateAdditions.swift
```diff
@@ -10,27 +10,132 @@
...
+    #if os(Windows)
+    var stTime: SYSTEMTIME = SYSTEMTIME()
+    var ftTime: FILETIME = FILETIME()
+
+    GetSystemTime(&stTime)
+    SystemTimeToFileTime(&stTime, &ftTime)
+
+    let result: UInt64 = (UInt64(ftTime.dwLowDateTime) << 0)
+                       + (UInt64(ftTime.dwHighDateTime) << 32)
+    // Windows ticks in 100 nanosecond intervals.
```
compnerd, line 104: This makes me think that you intended to use `GetSystemTimePreciseAsFileTime` rather than `GetSystemTime` and converting that to a `FILEITME`.

Sources/SwiftDriver/Utilities/DateAdditions.swift
```diff
@@ -10,27 +10,132 @@
...
+    return .seconds(result / 10_000_000)
```
compnerd, line 105: This seems to need an explicit conversion ...
```suggestion
    return .seconds(Int(result / UInt64(10_000_000)))
```

Sources/SwiftDriver/Utilities/DateAdditions.swift
CodaFi: Great idea! I cribbed this from CF - perhaps they ought to look into using this same API, though TimeInterval isn't exactly a useful type when you start talking about <1us resolution precision.

compnerd: Yeah, CF, dispatch, Swift, everyone has a slightly different implementation.  There is a slight cost associated with the higher precision which is why it isn't as pervasively used.

CodaFi:
@swift-ci test

compnerd:
Eek; this seems to regress the test suite pretty badly:

```
Executed 1 test, with 34 failures (0 unexpected) in 129.017 (129.017) seconds
```

owenv (approved):

compnerd (changes requested):
Seems to regress WIndows

CodaFi:
My bad, I need to correct for the Windows epoch difference in the file modification time function.

CodaFi:
@swift-ci test

CodaFi:
@swift-ci test Windows

CodaFi:
@swift-ci test Windows

compnerd (approved):
Thanks, this seems to resolve the Windows issues.

CodaFi:
apple/swift-package-manager#5759

@swift-ci test

Sources/SwiftDriver/Utilities/DateAdditions.swift
```diff
@@ -10,27 +10,127 @@
...
+    // compare against.
+    //
+    // However, it is important to note that Foundation's value cannot be both
+    // fixed in time AND correct because of leap seconds and other calendrical
+    // oddities.
+    //
+    // Other candidates include std::chrono's std::time_point::max - which is
+    // about 9223372036854775807 - enough to comfortably bound the age of the
+    // universe.
+    return .seconds(64_092_211_200)
```
buttaface, line 126: @CodaFi, it appears this line [broke building this repo for armv7](https://github.com/buttaface/swift-android-sdk/runs/8302044832?check_suite_focus=true#step:6:7000), because `seconds(value:)` takes an `Int` above:
```
swift-driver/Sources/SwiftDriver/Utilities/DateAdditions.swift:126:21: error: integer literal '64092211200' overflows when stored into 'Int'
    return .seconds(64_092_211_200)
```
Maybe it should take an explicit `Int64` instead?

PR #1174, opened 2022-09-08
[Explicit Modules] Query and then emit diagnostics from the dependency scanner

artemcm:
The driver invokes the scanning action via a `libSwiftScan` API call. This PR adds the diagnostic query API which the driver uses to get the scanner's diagnostics after a scan and emits them for its clients.

The API entry-points added in https://github.com/apple/swift/pull/60999


artemcm:
@swift-ci test

nkcsgexi (approved):

PR #1175, opened 2022-09-08
Add a Dot serializer for the inter-module dependency graph.

artemcm:
It can be nice to be able to visualize all of a module's dependencies.

A followup change will wire it up to a driver flag to allow an arbitrary driver invocation to emit this to a specified path. 
(Such flags must first be added in the main `swift` repo)

### DIFF
--- a/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
+++ b/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
@@ -1167,6 +1167,69 @@ final class ExplicitModuleBuildTests: XCTestCase {
     }
   }
 
+  func testDependencyGraphDotSerialization() throws {
+      let (stdlibPath, shimsPath, toolchain, hostTriple) = try getDriverArtifactsForScanning()
+      let dependencyOracle = InterModuleDependencyOracle()
+      let scanLibPath = try Driver.getScanLibPath(of: toolchain,
+                                                  hostTriple: hostTriple,
+                                                  env: ProcessEnv.vars)
+      guard try dependencyOracle
+              .verifyOrCreateScannerInstance(fileSystem: localFileSystem,
+                                             swiftScanLibPath: scanLibPath) else {
+        XCTFail("Dependency scanner library not found")
+        return
+      }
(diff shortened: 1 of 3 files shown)


artemcm:
@swift-ci test

nkcsgexi (approved):

Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
```diff
@@ -1167,6 +1167,70 @@ final class ExplicitModuleBuildTests: XCTestCase {
...
+          try! dependencyOracle.getDependencies(workingDirectory: path,
+                                                commandLine: scannerCommand)
+        let serializer = DOTModuleDependencyGraphSerializer(dependencyGraph)
+        
+        let outputFile = path.appending(component: "dependency_graph.dot")
+        var outputStream = try! ThreadSafeOutputByteStream(LocalFileOutputByteStream(outputFile))
+        serializer.writeDOT(to: &outputStream)
+        outputStream.flush()
+        let contents = try localFileSystem.readFileContents(outputFile).description
+        print(contents)
```
bnbarham, line 1224: Are you purposefully texting writing out here, or can this just be an in-memory string?

Also there's a couple of `try!` here but this is just a test and the method already throws, so they don't really seem needed.

bnbarham (approved):
Nice :). Love me some dependency graphs!

Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
artemcm: The primary use for this will be writing out to some kind of output stream so it doesn't hurt to use as such here.

Removed the `try!`, thanks!

artemcm:
@swift-ci test

Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
bnbarham: I was imagining you'd test that with the option when it's added. But I don't really mind either way 🤷

PR #1179, opened 2022-09-12
Revert "[Explicit Modules] Query and then emit diagnostics from the dependency scanner"

artemcm:
Reverts apple/swift-driver#1174


artemcm:
@swift-ci test

PR #1181, opened 2022-09-12
Add support for '-explicit-dependency-graph-format=' flag

artemcm:
This flag allows printing of the inter-module dependency graph using either JSON (default) or DOT formats. 
Added in https://github.com/apple/swift/pull/61056.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -634,6 +634,7 @@ public struct Driver {
                                fileSystem: fileSystem,
                                workingDirectory: workingDirectory,
                                diagnosticEngine: diagnosticEngine)
+    Self.validateEmitDependencyGraphArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
     Self.validateParseableOutputArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
     Self.validateCompilationConditionArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
     Self.validateFrameworkSearchPathArgs(&parsedOptions, diagnosticEngine: diagnosticEngine)
@@ -2646,6 +2647,28 @@ extension Driver {
     }
   }
 
+  static func validateEmitDependencyGraphArgs(_ parsedOptions: inout ParsedOptions,
+                                              diagnosticEngine: DiagnosticsEngine) {
+    // '-print-explicit-dependency-graph' requires '-explicit-module-build'
+    if parsedOptions.hasArgument(.printExplicitDependencyGraph) &&
+        !parsedOptions.hasArgument(.driverExplicitModuleBuild) {
+      diagnosticEngine.emit(Error.optionRequiresAnother(Option.printExplicitDependencyGraph.spelling,
+                                                        Option.driverExplicitModuleBuild.spelling))
+    }
+    // '-explicit-dependency-graph-format=' requires '-print-explicit-dependency-graph'
+    if parsedOptions.hasArgument(.explicitDependencyGraphFormat) &&
+        !parsedOptions.hasArgument(.printExplicitDependencyGraph) {
+      diagnosticEngine.emit(Error.optionRequiresAnother(Option.explicitDependencyGraphFormat.spelling,
+                                                        Option.printExplicitDependencyGraph.spelling))
+    }
+    // '-explicit-dependency-graph-format=' only supports values 'json' and 'dot'
+    if let formatArg = parsedOptions.getLastArgument(.explicitDependencyGraphFormat)?.asSingle {
+      if formatArg != "json" && formatArg != "dot" {
+        diagnosticEngine.emit(.error_unsupported_argument(argument: formatArg,
+                                                          option: .explicitDependencyGraphFormat))
+      }
+    }
+  }
 
   static func validateProfilingArgs(_ parsedOptions: inout ParsedOptions,
                                     fileSystem: FileSystem,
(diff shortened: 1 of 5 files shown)


artemcm:
@swift-ci test

artemcm:
@swift-ci test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2646,6 +2647,28 @@ extension Driver {
...
 
+  static func validateEmitDependencyGraphArgs(_ parsedOptions: inout ParsedOptions,
+                                              diagnosticEngine: DiagnosticsEngine) {
+    // '-print-explicit-dependency-graph' requires '-explicit-module-build'
+    if parsedOptions.hasArgument(.printExplicitDependencyGraph) &&
+        !parsedOptions.hasArgument(.driverExplicitModuleBuild) {
+      diagnosticEngine.emit(Error.optionRequiresAnother(Option.printExplicitDependencyGraph.spelling,
+                                                        Option.driverExplicitModuleBuild.spelling))
+    }
+    // '-explicit-dependency-graph-format=' requires '-print-explicit-dependency-graph'
```
bnbarham, line 2658: Could `-print-explicit-dependency-graph` just take a format, or would that break all the things? (ie. do we need to be backwards compatible?)

Sources/SwiftOptions/Options.swift
```diff
@@ -236,6 +236,7 @@ extension Option {
   public static let embedTbdForModule: Option = Option("-embed-tbd-for-module", .separate, attributes: [.frontend], helpText: "Embed symbols from the module in the emitted tbd file")
   public static let emitAbiDescriptorPath: Option = Option("-emit-abi-descriptor-path", .separate, attributes: [.frontend, .noDriver], metaVar: "<path>", helpText: "Output the ABI descriptor of current module to <path>")
   public static let emitAssembly: Option = Option("-emit-assembly", .flag, attributes: [.frontend, .noInteractive, .doesNotAffectIncrementalBuild], helpText: "Emit assembly file(s) (-S)", group: .modes)
+  public static let emitAst: Option = Option("-emit-ast", .flag, alias: Option.dumpAst, attributes: [.frontend, .noInteractive, .doesNotAffectIncrementalBuild])
```
bnbarham, line 239: Crazy how many options we add. Didn't I only just run this? This also means that these were only added for the old driver :(

Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
```diff
@@ -1240,6 +1240,91 @@ final class ExplicitModuleBuildTests: XCTestCase {
...
+      try localFileSystem.writeFileContents(main) {
+        $0 <<< "import C;"
+        $0 <<< "import E;"
+        $0 <<< "import G;"
+      }
+      let cHeadersPath: AbsolutePath = testInputsPath.appending(component: "ExplicitModuleBuilds").appending(component: "CHeaders")
+      let swiftModuleInterfacesPath: AbsolutePath = testInputsPath.appending(component: "ExplicitModuleBuilds").appending(component: "Swift")
+      let sdkArgumentsForTesting = (try? Driver.sdkArgumentsForTesting()) ?? []
+
+      let baseCommandLine = ["swiftc",
```
bnbarham, line 1255: Not particularly relevant for this review, but IMO it'd be really nice if we could go from typed argument -> string. Ie. instead of `-explicit-module-build` just `.explicitModuleBuild.str` (or something similar). Though I suppose there's an argument to be made for making renaming arguments fairly difficult (ie. would have to rename in tests) so that it's not done accidentally 🤷

Tests/SwiftDriverTests/ParsableMessageTests.swift
```diff
@@ -15,23 +15,37 @@ import TSCBasic
...
-    let errorStream = stderrStream
-    let errorOutputFile = path.appending(component: prefix)
-    TSCBasic.stderrStream =
-      try! ThreadSafeOutputByteStream(LocalFileOutputByteStream(errorOutputFile))
-    try body(errorOutputFile)
-    TSCBasic.stderrStream.flush()
-    // Restore the error stream to what it was
-    TSCBasic.stderrStream = errorStream
-  }
+internal func withHijackedBufferedErrorStream(
```
bnbarham, line 18: Is there a reason we write to files here rather than just passing back a string? Also, could we pass the stream instead? eg.

```
@discardableResult
func withHijacked(stream: ThreadSafeOutputByteStream, body: () throws -> Void) rethrows -> String { ... }
```

Or if the body needs the stream, pass in the stream instead. Though from a quick look at uses, that argument is either ignored or used in a bunch of asserts at the end of the body (which IMO should be out of the body).

Up to you whether you think that's all worth cleaning up.

bnbarham (approved):

Sources/SwiftDriver/Driver/Driver.swift
artemcm: We don't need it to be backwards compatible, I don't know of anyone who uses this other than me but the ergonomics of this seemed much more obvious by not conflating two things: emit graph, and emit format.

Sources/SwiftOptions/Options.swift
artemcm: My observation is that most of these are `.frontend` flags, so at least they do not need the logic modified here.

Tests/SwiftDriverTests/ParsableMessageTests.swift
artemcm: I've cleaned this up to return a `String` instead, that's much cleaner indeed, thank you! 
I'll leave the two streams as separate for now, I ran into some trouble when trying to pass in a stream that I will look into later.

artemcm:
@swift-ci test

Sources/SwiftDriver/Driver/Driver.swift
bnbarham: Is it conflating two things 🤔? If we're emitting something we're emitting it in some format, so that seems relevant to me. I don't particularly mind here, I just personally prefer the one.

Tests/SwiftDriverTests/ParsableMessageTests.swift
bnbarham: Nice :). It might be better as separate anyway, ie. I don't have to know the stream name this way 🤷

Sources/SwiftDriver/Driver/Driver.swift
artemcm: Gah, sorry I missed this comment. 
That's fair, let me add to my previous comment: my intent was to separate the common case of wanting to emit this graph (just dump the complete JSON) from having to specify a more-specific output format. And I don't know of a way to have a `separate` or `joined` flag that **optionally** takes a parameter is there one?

bnbarham: I don't think there is. My point was more I think the format is always relevant and thus I'd be fine with specifying `json` all the time for the "default" case. But I can see the pros and cons, so it doesn't really matter :).

PR #1184, opened 2022-09-15
Make the new TimePoint.seconds() input explicitly Int64

buttaface:
else it breaks when compiling for 32-bit armv7 with large inputs.

I [recently hit this when cross-compiling trunk SPM for Android armv7 on my daily CI](https://github.com/buttaface/swift-android-sdk/runs/8302044832?check_suite_focus=true#step:6:7000), as `distantFuture` overflows `Int` on 32-bit platforms. I had to add this patch to get it to build again, buttaface/swift-android-sdk@043df19.

@CodaFi, you added `TimePoint`, let me know what you think of this tweak.

### DIFF
--- a/Sources/SwiftDriver/Utilities/DateAdditions.swift
+++ b/Sources/SwiftDriver/Utilities/DateAdditions.swift
@@ -39,7 +39,7 @@ public struct TimePoint: Equatable, Comparable, Hashable {
 }
 
 extension TimePoint {
-  public static func seconds(_ value: Int) -> TimePoint {
+  public static func seconds(_ value: Int64) -> TimePoint {
     precondition(value >= 0,
                  "Duration value in seconds is \(value), but cannot be negative")
     return TimePoint(seconds: UInt64(value), nanoseconds: 0)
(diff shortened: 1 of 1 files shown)


Sources/SwiftDriver/Utilities/DateAdditions.swift
```diff
@@ -97,7 +97,7 @@ extension TimePoint{
     let result: UInt64 = (UInt64(ftTime.dwLowDateTime) << 0)
                        + (UInt64(ftTime.dwHighDateTime) << 32)
     // Windows ticks in 100 nanosecond intervals.
-    return .seconds(Int(result / 10_000_000))
+    return .seconds(Int64(result / 10_000_000))
```
buttaface, line 100: @stevapple, I didn't want this change to cause unnecessary narrowing on Windows x86, but maybe this `Int64` conversion is not needed at all now: let me know what you think.

stevapple: `FILETIME` consists of two `UInt32`s (no matter 32- or 64-bit OS), so I believe the current way of handling in `UInt64` and converting to a `Int64` is still necessary.

buttaface:
@artemcm, would you run the CI on this? [I built the Sep. 10 source snapshot for Android armv7 a couple days ago](https://github.com/buttaface/termux-packages/actions/runs/3072535020) and it worked well, so this pull gets the 32-bit toolchain building again.

artemcm:
@swift-ci test

buttaface:
Ping @CodaFi, would like your input on this tiny pull.

CodaFi:
Good to avoid the mistakes of the DispatchTime API here. Thanks

PR #1185, opened 2022-09-15
Remove extra checks for no emitted diagnostics in 'testPrintingExplicitDependencyGraph'

artemcm:
These do not actually have much to do with the test. If dependency scanner produces an error or a warning, for example, all this test needs to make sure the right data is emitted to the output. Other tests verify the actual integrity of the scanning action's output.

Resolves rdar://99977956


artemcm:
@swift-ci test

bnbarham (approved):

artemcm:
@swift-ci test Windows Platform

PR #1188, opened 2022-09-17
Add “swift -e” support

beccadax:
“swift -e” can be used to run code specified directly on the command line rather than passed through stdin or a source file. The values of the “-e” switches are each treated as one line of source code, which are concatenated together into a temporary main.swift file and passed to the frontend jobs.

Fixes rdar://17258540.


beccadax:
With apple/swift#61171

@swift-ci test

beccadax:
Tagging Artem because the relevant radar is assigned to him.

artemcm (approved):

PR #1189, opened 2022-09-18
Remove Minor Version Constraints

SDGGiesbrecht:
Several dependencies are marked `upToNextMinor` even though they no longer point at 0‐prefixed versions. They are not kept in sync across toolchain packages very well over time, and it causes issues for clients attempting to use them together.


SDGGiesbrecht:
Closing for same reason as apple/swift-package-manager#5769.

PR #1190, opened 2022-09-19
SwiftDriverTests: exclude explicit triple

compnerd:
The explicit triple sets an OS, which then requires that the value for `-sdk` match.  Because Windows uses a `-sdk` parameter, this will fail on Windows due to a mismatch for the SDKs.  By excluding the triple, a host specific value is used which allows this to match and pass.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1192, opened 2022-09-19
Fix accidental use of new feature in -e

beccadax:
The change in #1188 accidentally used a 5.7 feature (multi-statement closure result type inference), but it needs to support at least 5.6 for the earlyswiftdriver pass. Add an explicit type to avoid this problem.


beccadax:
@swift-ci please test

artemcm (approved):

jakepetroules:
Thank you!

PR #1195, opened 2022-09-21
Update the name of the swiftinterface verification job to name the target file

xymus:
This new diagnostic should distinguish between the verification of the public and the private swiftinterface. It should look like this:
```
Verifying emitted module interface MyFramework.private.swiftinterface
```

When building for two archs at a time, the same message will still appear twice (down from 4 times the same message). This appears to be common to all jobs named from the driver side, I'm not sure if we could improve this from the driver side or if it should be from the driver client side.

I didn't see tests for these so please let me know if I missed a testing opportunity.


xymus:
@swift-ci Please test

xymus:
I was forgetting about this PR. Could you give this a quick review @artemcm?

artemcm (approved):

PR #1196, opened 2022-09-21
Enable Explicit Module Build tests on Linux

artemcm:
They were waiting for the dependency scanner to learn how to query Clang dependency scanner without relying on Objective-C. This is now possible.


artemcm:
@swift-ci test

PR #1197, opened 2022-09-25
Allow response files to be special files

stephank:
In NixOS, we wrap toolchain executables to inject search paths (among other things). These wrappers are bash scripts that build an often large array of arguments, then provide them as a response file via redirect like so:

```
swift-driver @<(printf "%q\n" --driver-mode=swiftc -print-target-info)
```

The above is a simple test case that currently fails. Bash turns this into:

```
swift-driver @/dev/fd/63
```

But the device node used there doesn't pass the `isFile` check. This PR broadens that check by using `isReadable` instead.


artemcm:
Thank you @stephank.

Could you please add a test for this (with a command like in the PR description) to the test suite?

stephank:
I added a test case and checked the before and after. I'm not sure if this sort of thing works on Windows, I don't have a machine to test.

artemcm:
@swift-ci test

artemcm:

artemcm (approved):

stephank:
That CI error is a little weird:

```
...\sourcekit-lsp\Sources\LanguageServerProtocol\SupportTypes\DocumentURI.swift:15:13: error: no such module 'TSCBasic'
import func TSCBasic.resolveSymlinks
            ^
```

But that line doesn't have that import; it appears to be in `swift-driver/Sources/SwiftDriver/Utilities/VirtualPath.swift` instead. Seems unrelated?

Regardless, when I wrote the comment about Windows, I was thinking file descriptors, but I suppose the whole `/dev/fd` namespace is just not a thing. I can wrap it in a `#if !os(Windows)` if you want?

artemcm:
> Regardless, when I wrote the comment about Windows, I was thinking file descriptors, but I suppose the whole `/dev/fd` namespace is just not a thing. I can wrap it in a `#if !os(Windows)` if you want?

Makes sense for this particular test, I think.
cc @compnerd

compnerd:
https://github.com/apple/sourcekit-lsp/pull/640 should help with the LSP issue.

compnerd:
https://github.com/apple/sourcekit-lsp/pull/640

artemcm:
> Please test with following PRs: [apple/sourcekit-lsp#640](https://github.com/apple/sourcekit-lsp/pull/640)
> 
> @swift-ci please test Windows platform

I *think* this repository does not support cross-repo testing. :(

compnerd:
Unfortunately, the CI currently is build only; please do not merge this without testing on Windows, I am worried that this might regress the test suite.

stephank:
Oh hey, Windows is green. But I'm looking at the logs, and can't really find a test run for swift-driver in there, is that correct?

compnerd:
Yeah, that is the problem - the CI doesn't actually run the test suite yet.  That is something that would be nice to have, but needs work.  In the mean time, tests need to be run locally.

PR #1201, opened 2022-10-06
SwiftDriver: migrate LLVM bitcode support from tools-support-core

compnerd:
The swift-driver is the only consumer of the bitcode implementation in TSCUtility.  Migrate the code to the single user to allow us to deprecate the API in tools-support-core.  While there are no practical changes to the implementation, this does take the opportunity to split up the bitcode implementation into multiple files.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/Bitcode/Bitstream.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/Bitcode/Bitstream.swift
@@ -1,0 +1,179 @@
+//===--------------- Bitstream.swift - LLVM Bitstream routines -----------===//
+//
+// This source file is part of the Swift.org open source project
+//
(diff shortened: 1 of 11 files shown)


compnerd:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/Bitcode/Bitstream.swift
```diff
@@ -0,0 +1,168 @@
+
```
artemcm, line 1: All the newly-added files are missing the headers:
```
//===------- Bitstream.swift ----------------------------------===//
```
etc.

artemcm:
The new files also must be added to the `CMakeLists`, otherwise we will break the build.

compnerd:
@artemcm hah, thanks!  I forgot about that.  Thats why I wanted to get an early test in.

compnerd:

PR #1203, opened 2022-10-12
Re-enable swiftinterface verification with compatibility headers

xymus:
Lift the workaround that disabled scheduling the swiftinterface verification task when a compatibility header is generated. This is an issue only when the header is also installed, we now expect the build system to disable the verification when it installs the header.

For more context, this issue lead to errors on missing headers and failures to build the underlying clang module at verification.

rdar://100989723


xymus:
@swift-ci Please test

artemcm (approved):

tshortli (approved):

PR #1209, opened 2022-10-29
Use the versioned triple outside of Darwin.

3405691582:
Some platforms require the normal, versioned triple to be supplied for the target flag, as Swift modules use this versioned triple as a subdirectory. If the unversioned triple is used instead, the standard library won't be found on these platforms.

Instead, use the unversioned triple only on Darwin. This also should be a noop on Linux. This behavior matches apple/swift-package-manager#3576 and apple/swift-tools-support-core#229 which contain additional context.

### DIFF
--- a/Utilities/build-script-helper.py
+++ b/Utilities/build-script-helper.py
@@ -601,7 +601,11 @@ def get_build_target(swiftc_path, args):
                                                    stderr=subprocess.PIPE,
                                                    universal_newlines=True).strip()
         args.target_info = json.loads(target_info_json)
-        return args.target_info['target']['unversionedTriple']
+        triple = args.target_info['target']['triple']
+        # Windows also wants unversionedTriple, but does not use this.
+        if platform.system() == 'Darwin':
+          triple = args.target_info['target']['unversionedTriple']
+        return triple
     except Exception as e:
         error(str(e))
 


Utilities/build-script-helper.py
```diff
@@ -601,7 +601,10 @@ def get_build_target(swiftc_path, args):
                                                    stderr=subprocess.PIPE,
                                                    universal_newlines=True).strip()
         args.target_info = json.loads(target_info_json)
-        return args.target_info['target']['unversionedTriple']
+        triple = args.target_info['target']['triple']
+        if platform.system() == 'Darwin':
+          triple = args.target_info['target']['unversionedTriple']
+        return triple
```
compnerd, line 607: We don't use this on windows but windows also expects the unversioned triple.  At the very least, I think that should be in a comment.

3405691582: Added a comment.

compnerd (approved):

compnerd:
@swift-ci please test

artemcm (approved):

MaxDesiatov:
@artemcm do you mind if this is merged?

artemcm:
@swift-ci please test

MaxDesiatov:
@swift-ci test Windows

MaxDesiatov:
Since the PR has been approved and CI is passing I'm assuming this can be merged.

PR #1210, opened 2022-10-29
swift-system is now a dependency.

3405691582:
This was introduced as a dependency in #915, so mention it.


compnerd (approved):

compnerd:
@swift-ci please test

PR #1216, opened 2022-11-03
If need be, bring back not verifying modules with a compatibility header

xymus:
Revert of 58c83b43d18f23da4730b3da36e82cfee5d861f1 and fixes the forced verification mode.


xymus:
@swift-ci Please test

PR #1223, opened 2022-11-11
Adapt to dependency scanner outputting canonical Clang PCM output paths

artemcm:
This change adapts to the `modulePath` of Clang modules in the dependency scanner output being *the* output path for such PCMs. Implemented in https://github.com/apple/swift/pull/62067.

The change temporarily handles the previous case to ensure the driver CI can work with both kinds of compilers/scanners.


artemcm:
@swift-ci test

artemcm:
@swift-ci test

nkcsgexi (approved):

PR #1224, opened 2022-11-15
Only add '-fobjc-link-runtime' to the linker invocation when the driver is invoked with '-link-objc-runtime'

artemcm:
Otherwise, every single compiled program will end up linking it, and Foundation on top of it.

### DIFF
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -214,9 +214,14 @@ extension DarwinToolchain {
       commandLine.appendPath(VirtualPath.lookup(sdkPath))
     }
 
+    // -link-objc-runtime also implies -fobjc-link-runtime
+    if parsedOptions.hasFlag(positive: .linkObjcRuntime,
+                             negative: .noLinkObjcRuntime,
+                             default: false) {
+      commandLine.appendFlag("-fobjc-link-runtime")
+    }
+
     commandLine.appendFlags(
-      "-fobjc-link-runtime",
-      "-lobjc",
       "-lSystem"
     )
 
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -1653,6 +1653,8 @@ final class SwiftDriverTests: XCTestCase {
 
       XCTAssertFalse(cmd.contains(.flag("-static")))
       XCTAssertFalse(cmd.contains(.flag("-shared")))
+      // Handling of '-lobjc' is now in the Clang linker driver.
+      XCTAssertFalse(cmd.contains(.flag("-lobjc")))
     }
 
     do {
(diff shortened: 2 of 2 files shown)


DougGregor (approved):

artemcm:
@swift-ci test

Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
```diff
@@ -214,8 +214,14 @@ extension DarwinToolchain {
       commandLine.appendPath(VirtualPath.lookup(sdkPath))
     }
 
+    // -link-objc-runtime also implies -fobjc-link-runtime
+    if parsedOptions.hasFlag(positive: .linkObjcRuntime,
```
compnerd, line 218: Is this an existing option?  Why not gate this on `-enable-objc-interop`?

artemcm: It is an existing option and seems to very explicitly map to Clang's equivalent. 

I am not sure if there are cases where `-enable-objc-interop` is set where we wouldn't want to do this though, explicit override with `-no-link-obcj-runtime` aside...

compnerd: The only question then is, is there ever a case where you want to link the ObjC runtime (and Foundation) when ObjC is explicitly disabled?  If not, I feel like `-enable-objc-interop` is nicer.

artemcm: `-enable-objc-interop` is actually a frontend-only flag (`.noDriver`) that the driver sets to be on by-default for Darwin platforms, always. 

So we cannot use it as an input switch to enable/disable this. And if we did, because it is enabled by the driver by-default, it would essentially be equivalent to what things were like before this change. I may lack some historical context here, but the existence of `-link-objc-runtime` really indicates that it ought to be used to control this behavior.

compnerd: Yeah, as a short term thing, I don't think that it is terrible.  But it seems like having an extra flag that is unnecessary and we should figure out how to deprecate that IMO.  Having the language runtime be linked in properly based on the build type seems like the right approach.

Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
```diff
@@ -214,8 +214,14 @@ extension DarwinToolchain {
...
+                             negative: .noLinkObjcRuntime,
+                             default: false) {
+      commandLine.appendFlag("-fobjc-link-runtime")
+    }
+
     commandLine.appendFlags(
-      "-fobjc-link-runtime",
       "-lobjc",
```
compnerd, line 225: Should `-lobjc` be dropped now?  This should be handled by `-fobjc-link-runtime` in the clang driver.

compnerd (changes requested):

Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
artemcm: Yup. https://github.com/apple/llvm-project/blob/cached/stable/20220421/clang/lib/Driver/ToolChains/Darwin.cpp#L751

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
Catfish-Man: It's not that there are cases where you want this without ObjC, but there are cases where you want ObjC without this. For example, if you're in a binary which Foundation itself links.

Catfish-Man: (and then you manually link the runtime of course. I'm not aware of a situation where you want ObjC support but not the ObjC runtime, it's just Foundation that makes things tricky)

DougGregor:
This might be causing CI failures in LLDB. Speculatively reverting to do more testing

PR #1225, opened 2022-11-15
SwiftDriver: avoid `-lSystem` on Darwin

compnerd:
Now that we use the proper linker driver (`clang`), we no longer need to handle the explicit standard library linking.  Delegate this to the linker driver which will properly link to libSystem instead.  This simplifies some of the logic and further harmonises the beahviour across the various platforms.


compnerd:
@swift-ci please test

artemcm (approved):
Thanks @compnerd

PR #1230, opened 2022-11-18
[Explicit Module Builds] Serialize and re-use the inter-module dependency graph as incremental build state

artemcm:
Building on top of https://github.com/apple/swift-driver/pull/621, on an incremental build, load a prior dependency graph, check if it is still valid:
1. import set of the current module has not changed
2. input files of every module in the prior graph are older than the output and if it is, re-use it, skipping invoking the dependency scanner again.

More concretely, computation of the initial incremental state in `computeIncrementalStateForPlanning` will attempt to read out and validate a prior serialized inter-module dependency graph. If validated, such graph will become a part of the initial incremental state itself.

During build planning, at the stage of computing inter-module dependencies, the dependency scanning action will be skipped if the initial incremental state contains a still-valid prior inter-module dependency graph. 

In both cases, if a prior graph is being re-used, or if a new graph was computed via a dependency scanning action, the graph will become a component of the incremental build state to be serialized after the build is complete. 

This change does not yet result in skipping the actual module build jobs in the planning output, which can be done with this, but this is left to a followup PR. 

There is also some notable overlap between this and what is already done for detecting changes in input files to the current compilation:
- Suppose a Swift module interface of a module dependency has changed. This means that, according to the above mechanism, a dependency scanning action must be performed on an incremental build. This will result in a new dependency graph and a subsequent re-build of the dependency. But, moreover, there is an existing mechanism in place to ensure that source-files also get re-built on changes to previously-recorded inputs to compilation so the source-files will also be re-built when a dependency changes. 

Once explicit builds become the default, we can consolidate the two approaches into one.

Partially resolves rdar://66801475

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -235,22 +235,6 @@ public struct Driver {
   /// Only present when the driver will be writing the record.
   /// Only used for reading when compiling incrementally.
   @_spi(Testing) public let buildRecordInfo: BuildRecordInfo?
-
-  /// A build-record-relative path to the location of a serialized copy of the
-  /// driver's dependency graph.
-  ///
-  /// FIXME: This is a little ridiculous. We could probably just replace the
-  /// build record outright with a serialized format.
-  var driverDependencyGraphPath: VirtualPath? {
-    guard let recordInfo = self.buildRecordInfo else {
-      return nil
-    }
-    let filename = recordInfo.buildRecordPath.basenameWithoutExt
-    return recordInfo
-      .buildRecordPath
-      .parentDirectory
-      .appending(component: filename + ".priors")
-  }
   
   /// Whether to consider incremental compilation.
   let shouldAttemptIncrementalCompilation: Bool
@@ -1283,10 +1267,6 @@ extension Driver {
   public mutating func run(
     jobs: [Job]
   ) throws {
-    if parsedOptions.hasArgument(.v) {
-      try printVersion(outputStream: &stderrStream)
-    }
-
     let forceResponseFiles = parsedOptions.contains(.driverForceResponseFiles)
 
     // If we're only supposed to print the jobs, do so now.
@@ -1450,6 +1430,15 @@ extension Driver {
           /// Ensure that a bogus dependency graph is not used next time.
           buildRecordInfo?.removeBuildRecord()
           return
+      }
+      do {
+        try incrementalCompilationState.writeInterModuleDependencyGraph(buildRecordInfo)
+      }
+      catch {
+        diagnosticEngine.emit(
+          .warning("next compile must run a full dependency scan; could not write inter-module dependency graph: \(error.localizedDescription)"))
+        buildRecordInfo?.removeInterModuleDependencyGraph()
+        return
       }
     }
     buildRecordInfo?.writeBuildRecord(
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
@@ -41,6 +41,116 @@ extension Diagnostic.Message {
   }
 }
 
+@_spi(Testing) public extension Driver {
+  /// Scan the current module's input source-files to compute its direct and transitive
+  /// module dependencies.
+  mutating func gatherModuleDependencies()
+  throws -> InterModuleDependencyGraph {
+    var dependencyGraph = try performDependencyScan()
+
+    if parsedOptions.hasArgument(.printPreprocessedExplicitDependencyGraph) {
+      try stdoutStream <<< dependencyGraph.toJSONString()
+      stdoutStream.flush()
+    }
+
+    if let externalTargetDetails = externalTargetModuleDetailsMap {
+      // Resolve external dependencies in the dependency graph, if any.
+      try dependencyGraph.resolveExternalDependencies(for: externalTargetDetails)
+    }
+
+    // Re-scan Clang modules at all the targets they will be built against.
+    // This is currently disabled because we are investigating it being unnecessary
+    // try resolveVersionedClangDependencies(dependencyGraph: &dependencyGraph)
+
+    // Set dependency modules' paths to be saved in the module cache.
+    try resolveDependencyModulePaths(dependencyGraph: &dependencyGraph)
+
+    if parsedOptions.hasArgument(.printExplicitDependencyGraph) {
+      let outputFormat = parsedOptions.getLastArgument(.explicitDependencyGraphFormat)?.asSingle
+      if outputFormat == nil || outputFormat == "json" {
+        try stdoutStream <<< dependencyGraph.toJSONString()
+      } else if outputFormat == "dot" {
+        DOTModuleDependencyGraphSerializer(dependencyGraph).writeDOT(to: &stdoutStream)
+      }
+      stdoutStream.flush()
+    }
+
+    return dependencyGraph
+  }
+
+  /// Update the given inter-module dependency graph to set module paths to be within the module cache,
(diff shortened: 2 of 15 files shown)


Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
```diff
@@ -41,6 +41,118 @@ extension Diagnostic.Message {
   }
 }
 
+@_spi(Testing) public extension Driver {
+  /// Scan the current module's input source-files to compute its direct and transitive
+  /// module dependencies.
+  mutating func gatherModuleDependencies()
```
artemcm, line 47: This code just moved from `Planning.swift` it is unchanged.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -235,22 +235,6 @@ public struct Driver {
   /// Only present when the driver will be writing the record.
   /// Only used for reading when compiling incrementally.
   @_spi(Testing) public let buildRecordInfo: BuildRecordInfo?
-
-  /// A build-record-relative path to the location of a serialized copy of the
-  /// driver's dependency graph.
-  ///
-  /// FIXME: This is a little ridiculous. We could probably just replace the
-  /// build record outright with a serialized format.
-  var driverDependencyGraphPath: VirtualPath? {
```
artemcm, line 244: This was duplicated in BuildRecordInfo and this instance was not used anywhere.

Sources/SwiftDriver/IncrementalCompilation/BuildRecordInfo.swift
```diff
@@ -263,6 +295,15 @@ import class Dispatch.DispatchQueue
       .appending(component: filename + ".priors")
   }
 
+  /// A build-record-relative path to the location of a serialized copy of the
+  /// driver's inter-module dependency graph.
+  var interModuleDependencyGraphPath: VirtualPath {
```
artemcm, line 300: Alternatively we can also make this a part of the `.swiftdeps` file, but for now it seemed simpler to keep it separate.

artemcm:
@swift-ci test

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
```diff
@@ -1160,10 +1343,15 @@ extension IncrementalCompilationTests {
 // MARK: - Incremental test perturbation helpers
 extension IncrementalCompilationTests {
   private func touch(_ name: String) {
-    Thread.sleep(forTimeInterval: 1)
     print("*** touching \(name) ***", to: &stderrStream); stderrStream.flush()
-    let (path, contents) = try! XCTUnwrap(inputPathsAndContents.filter {$0.0.pathString.contains(name)}.first)
-    try! localFileSystem.writeFileContents(path) { $0 <<< contents }
```
artemcm, line 1166: I changed this to actually just re-write the file's contents with what they already are, so it actually just updates the timestamp. Previously `touch` was overwriting file contents with a pre-determined initial state for that input, instead of what the test may have modified it to be at that point.

nkcsgexi (approved):
Discussed with Artem about this offline. LGTM as the first step for handling incremental explicit module build!

Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
```diff
@@ -92,10 +100,116 @@ extension IncrementalCompilationState {
...
+      case .swift(let swiftDetails):
+        if module.key.moduleName == graph.mainModuleName {
+          continue
+        }
+        guard let outputModTime = try? buildRecordInfo.fileSystem.lastModificationTime(for: VirtualPath.lookup(module.value.modulePath.path)) else {
+          reporter?.report("Unable to 'stat' \(module.value.modulePath.description)")
+          return false
+        }
+        if let moduleInterfacePath = swiftDetails.moduleInterfacePath {
+          let inputModTime = try buildRecordInfo.fileSystem.lastModificationTime(for: VirtualPath.lookup(moduleInterfacePath.path))
```
tshortli, line 160: There's a mix of error handling approaches in this method; above `try?` is used with explicit handling of `nil` and then here the error actually propagates. Should these be consistent, or is there a reason some errors propagate directly?

Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
```diff
@@ -41,6 +41,118 @@ extension Diagnostic.Message {
...
+      }
+      // TODO: Remove this once toolchain is updated
+      else if case .clang(_) = moduleId {
+        let modulePath = VirtualPath.lookup(moduleInfo.modulePath.path)
+        // Use VirtualPath to get the OS-specific path separators right.
+        let modulePathInCache =
+        try VirtualPath(path: moduleCachePath)
+          .appending(component: modulePath.basename)
+        dependencyGraph.modules[moduleId]!.modulePath =
+        TextualVirtualPath(path: modulePathInCache.intern())
```
tshortli, line 150: Nit: Should the continued lines be indented? I find it a bit hard to read without

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState+Extensions.swift
```diff
@@ -406,6 +425,48 @@ extension IncrementalCompilationState {
...
+    guard let recordInfo = buildRecordInfo else {
+      throw WriteDependencyGraphError.noBuildRecordInfo
+    }
+    guard let interModuleDependencyGraph = self.upToDateInterModuleDependencyGraph else {
+      throw WriteInterModuleDependencyGraphError.noDependencyGraph
+    }
+    do {
+      let data = try JSONEncoder().encode(interModuleDependencyGraph)
+      if let json = try? JSONSerialization.jsonObject(with: data, options: .mutableContainers),
+         let jsonData = try? JSONSerialization.data(withJSONObject: json, options: .prettyPrinted) {
```
tshortli, line 453: `JSONEncoder` natively supports pretty printing via `encoder.outputFormatting = .prettyPrinted`, is there a separate reason that round-tripping through Foundations `JSONSerialization` is needed?

Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
artemcm: Consolidated the logic to a helper and made consistent. Thanks!

Sources/SwiftDriver/IncrementalCompilation/BuildRecordInfo.swift
```diff
@@ -245,6 +254,29 @@ import class Dispatch.DispatchQueue
...
+      reporter?.report("Incremental compilation did not attempt to read inter-module dependency graph.")
+      return nil
+    }
+
+    let decodedGraph: InterModuleDependencyGraph
+    do {
+      let contents = try fileSystem.readFileContents(interModuleDependencyGraphPath).cString
+      decodedGraph = try JSONDecoder().decode(InterModuleDependencyGraph.self,
+                                              from: Data(contents.utf8))
+     } catch {
```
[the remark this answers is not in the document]
artemcm: Sharp eye! 👀
Fixed, thanks.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState+Extensions.swift
artemcm: Nope. Simplified to avoid JSONSerialization. 👍🏼

artemcm:
@swift-ci test

artemcm:
Thanks @tshortli

tshortli (approved):

PR #1233, opened 2022-12-01
Rephrase use of 'guard let' not supported by older compilers.

artemcm:
'guard let X else {}' works with latest Swift compilers (such as one used in swift-driver CI) but isn't supported in some older compilres used in CI and other bots.


artemcm:
@swift-ci test

etcwilde (approved):
LGTM

PR #1238, opened 2022-12-12
Make `-e` flag 'separate', instead of 'joinedOrSeparate'

artemcm:
Update `Options.swift` to bring in a change to make `-e` 'separate', instead of 'joinedOrSeparate'.
From: https://github.com/apple/swift/pull/62526


artemcm:
@swift-ci test

beccadax:
I'm out and cannot properly review this, but I'm not conceptually opposed to making `-e` separate-only if it's interfering with other options. (Is the fact that it's interfering a bug in the argument-processing code, though? I'd expect that to use a longest-prefix match.)

artemcm:
Even if we can improve our argument parsing logic in the short term, it's worth landing this change for now to fix a broken behavior.

xymus (approved):

owenv (approved):

PR #1240, opened 2022-12-13
[Dependency Scanning] Query scanner whether discovered binary Swift modules are frameworks

artemcm:
This info is required so that it can be specified as input to compilation that uses such module as a dependency, in order to decide whether or not to insert an auto-linking directive for it.

Companion change to https://github.com/apple/swift/pull/62540

Part of rdar://102824777

### DIFF
--- a/Sources/CSwiftScan/include/swiftscan_header.h
+++ b/Sources/CSwiftScan/include/swiftscan_header.h
@@ -130,6 +130,8 @@ typedef struct {
   (*swiftscan_swift_binary_detail_get_module_doc_path)(swiftscan_module_details_t);
   swiftscan_string_ref_t
   (*swiftscan_swift_binary_detail_get_module_source_info_path)(swiftscan_module_details_t);
+  bool
+  (*swiftscan_swift_binary_detail_get_is_framework)(swiftscan_module_details_t);
 
   //=== Swift Placeholder Module Details query APIs -------------------------===//
   swiftscan_string_ref_t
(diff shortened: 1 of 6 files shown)


artemcm:
@swift-ci test

nkcsgexi (approved):
LGTM! Shouldn't linking be explicit too given that we have all the dependency info?

artemcm:
> LGTM! Shouldn't linking be explicit too given that we have all the dependency info?

It absolutely should. But I'd like to get the existing flow working first, to avoid changing too many systems at-once, and then followup with the explicit linking work.

owenv (approved):

artemcm:

artemcm:
```
C:\Users\swift-ci\jenkins\workspace\swift-driver-PR-windows\sourcekit-lsp\Sources\LanguageServerProtocol\Messages\.swift:59:3: error: cannot find 'OpenInterfaceRequest' in scope
```
This is a new one and is not related to the contents of this PR. @compnerd have we seen this before?

compnerd:
Don't think we have.  Could be a recent change for LSP?  Should check that, but at the very least, we should check the driver build doesn't fail.

artemcm:

artemcm:
Ohh, I think the CMake build is broken, but not the package build.

compnerd:
That doesn't make sense: https://github.com/apple/sourcekit-lsp/commit/c31dff2ae64da1dea3633c2b582b21e42d5b3f8d#diff-9e9107c03ddddab7789c4b0e4591aa6a0c569955404bda08039056526263391dR59 seems that it does update the CMake build.

artemcm:
> That doesn't make sense: [apple/sourcekit-lsp@c31dff2#diff-9e9107c03ddddab7789c4b0e4591aa6a0c569955404bda08039056526263391dR59](https://github.com/apple/sourcekit-lsp/commit/c31dff2ae64da1dea3633c2b582b21e42d5b3f8d#diff-9e9107c03ddddab7789c4b0e4591aa6a0c569955404bda08039056526263391dR59) seems that it does update the CMake build.

https://github.com/apple/sourcekit-lsp/pull/678

artemcm:

artemcm:
Now Windows CI is broken for a new reason. 😥
Waiting on https://github.com/apple/llvm-project/pull/5792 to merge to unblock.

artemcm:

PR #1243, opened 2022-12-30
Consign the YAML Build Record to the Legacy Incremental Build Path

CodaFi:
This PR fixes a pretty harrowing set of Linux bugs we've observed on and off over the years with the incremental build.

The build record and the module dependency graph are currently emitted by "cross-module incremental builds" alongside a "swiftdeps" YAML file that summarizes the compilation session that occurred before this one. This has a number of downsides
- We have a hard dependency on a YAML library for just this one thing
- With two input files comes a quad-state of cases to handle (if priors desyncs from build record and vice versa... bad things happen)
- Moreover, we have observed that mod times returned to us on ext4 on older Ubuntus have a pretty high (probably << 1/100) chance of telling us that the priors file was somehow emitted _before_ the compilation session ended.

There's no reason to emit two files when we can just emit one. To start, bump the serialized dependency graph format and absorb the information from the build record into it. Then, teach all the things that instantiate a ModuleDependencyGraph to also hand it a build record. Where they get that build record from is their business, but we can also make our own now.

-------

Time Traveling priors occur when a file system cache lies to us about the state of the incremental build. Since I'd rather not deal with three operating system's worth of file systems abstractions to fix that particular problem, let's make sure this can never happen again by relegating one of the priors files to the legacy incremental build path.

The idea is that the "incremental imports" flag controls things here:

Incremental Imports Enabled: Read and write Module.priors bitstream files
Incremental Imports Disabled: Read and write Module.swiftdeps YAML files

The only notable thing needed to make this work is the old fallback that used to "disable" incremental compilation by bailing out of the entire incremental build now instead constructs an empty module dependency graph and fires off a "clean" incremental build. This should be a pretty significant optimization as it now means that on the happy path we can reach a steady state in the incremental build in one build instead of two.

I've also cleaned up and fixed a number of incremental build tests that happened to be passing in configurations where it was somehow acceptable to read stale, or in the worst case incorrect, priors files paired with the correct build record files. Finally, I've deleted tests that had anything to do with the time traveling priors phenomenon.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -1424,30 +1424,69 @@ extension Driver {
     // In case the write fails, don't crash the build.
     // A mitigation to rdar://76359678.
     // If the write fails, import incrementality is lost, but it is not a fatal error.
-    if let incrementalCompilationState = self.incrementalCompilationState {
+    guard
+      let buildRecordInfo = self.buildRecordInfo,
+      let absPath = buildRecordInfo.buildRecordPath.absolutePath
+    else {
+      return
+    }
+
+    let buildRecord = buildRecordInfo.buildRecord(
+      jobs, self.incrementalCompilationState?.blockingConcurrentMutationToProtectedState{
+        $0.skippedCompilationInputs
+      })
+
+    if
+      let incrementalCompilationState = self.incrementalCompilationState,
+      incrementalCompilationState.info.isCrossModuleIncrementalBuildEnabled
+    {
       do {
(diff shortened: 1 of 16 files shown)


CodaFi:
@harlanhaskins Has suggested we might be able to drop the Yams dependency even sooner - since YAML is a strict superset of JSON - by making `BuildRecord` conform to `Codable`.

CodaFi:
@swift-ci test

artemcm (approved):
I think this is a really great simplification and improvement.
I can't wait till we can unify things to just this code-path. 
Thank you @CodaFi

PR #1248, opened 2023-01-05
IncrementalDependencyAndInputSetup.swift: fix mutability warning

MaxDesiatov:
This is a warning that's currently caused by `priors` using a `var` binding:
```swift
IncrementalCompilation/IncrementalDependencyAndInputSetup.swift:281:17: warning: variable 'priors' was never mutated; consider changing to 'let' constant
      guard var priors = computeGraphAndInputsInvalidatedByExternals() else {
            ~~~ ^
            let
```


MaxDesiatov:
@swift-ci test Linux

PR #1250, opened 2023-01-09
Enable '-clang-target' by-default

artemcm:
This will result in the driver passing down '-clang-target' flag to the frontend, which will then ensure that all downstream module (Swift and Clang) compilation compiler sub-invocations are consistent with respect to the target triple used by their underlying instances of Clang. Resulting in all Clang module dependencies being built against the main module's target triple.


artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
Cross-repo tests in https://github.com/apple/swift/pull/60581 are also green.

nkcsgexi (approved):
So excited to see this happening by default!

PR #1251, opened 2023-01-09
[5.8] Enable '-clang-target' by-default

artemcm:
Cherry-pick of https://github.com/apple/swift-driver/pull/1250
This will result in the driver passing down '-clang-target' flag to the frontend, which will then ensure that all downstream module (Swift and Clang) compilation compiler sub-invocations are consistent with respect to the target triple used by their underlying instances of Clang. Resulting in all Clang module dependencies being built against the main module's target triple.


artemcm:
@swift-ci test

artemcm:
@swift-ci test macOS platform

nkcsgexi (approved):

PR #1259, opened 2023-01-12
Specify "-rtlib=compiler-rt" on Linux as required for some of Swift's numeric conversion symbols.

artemcm:
Resolves rdar://104134160

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -291,6 +291,10 @@ extension GenericUnixToolchain {
         }
       }
 
+      // Use compiler-rt for Swift's numeric conversion symbols
+      commandLine.appendFlag("-rtlib=compiler-rt")
+      commandLine.appendFlag("-lunwind")
+
       // Run clang++ in verbose mode if "-v" is set
       try commandLine.appendLast(.v, from: &parsedOptions)
 
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -1993,6 +1993,7 @@ final class SwiftDriverTests: XCTestCase {
       XCTAssertTrue(commandContainsTemporaryPath(cmd, "bar.o"))
       XCTAssertTrue(commandContainsTemporaryResponsePath(cmd, "Test.autolink"))
       XCTAssertEqual(linkJob.outputs[0].file, try VirtualPath(path: "libTest.so"))
+      XCTAssertTrue(cmd.contains(.flag("-rtlib=compiler-rt")))
 
       XCTAssertFalse(cmd.contains(.flag("-dylib")))
       XCTAssertFalse(cmd.contains(.flag("-static")))


artemcm:
@swift-ci test

eeckstein (approved):
lgtm

artemcm:
Looks like it may be causing a test failure on Linux on main:
https://ci.swift.org/job/swift-PR-Linux/5421/console
```
/home/build-user/build/buildbot_linux/none-swift_package_sandbox_linux-x86_64/usr/lib/swift_static/linux/libswiftCore.a(Errors.cpp.o):Errors.cpp:function swift::withCurrentBacktrace(std::function<void (void**, int)>): error: undefined reference to '_Unwind_Backtrace'
/home/build-user/build/buildbot_linux/none-swift_package_sandbox_linux-x86_64/usr/lib/swift_static/linux/libswiftCore.a(Errors.cpp.o):Errors.cpp:function swift::printCurrentBacktrace(unsigned int): error: undefined reference to '_Unwind_Backtrace'
/home/build-user/build/buildbot_linux/none-swift_package_sandbox_linux-x86_64/usr/lib/swift_static/linux/libswiftCore.a(Errors.cpp.o):Errors.cpp:function SwiftUnwindFrame(_Unwind_Context*, void*): error: undefined reference to '_Unwind_GetIP'
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)

error: command failed with exit status: 1
```

eeckstein:
@mikeash When linking with the clang rt, can this be solved by removing the `#elif defined(__ELF__)` case in `withCurrentBacktraceImpl`?

mikeash:
I think so. /cc @al45tair who would know better than I.

al45tair:
Does that not just mean that we also need `-lunwind`?

bnbarham:
Yeah, changing from `libgcc` means that we don't get the default `libgcc` unwind. In saying that though, could we instead make this change when building Clang? Ie. set `CLANG_DEFAULT_RTLIB` and `CLANG_DEFAULT_UNWINDLIB` (which both default to `libgcc` but we could instead have `compiler-rt` and `libunwind` respectively).

artemcm:
> Could we instead make this change when building Clang? Ie. set `CLANG_DEFAULT_RTLIB` and `CLANG_DEFAULT_UNWINDLIB` (which both default to `libgcc` but we could instead have `compiler-rt` and `libunwind` respectively).

If this achieves the same thing (`clang` linker driver defaulting to `rtlib=compiler-rt` then I'd be in favor of this approach also. But I'll let others weigh in.

bnbarham:
> If this achieves the same thing (clang linker driver defaulting to rtlib=compiler-rt) then I'd be in favor of this approach also. But I'll let others weigh in.

@al45tair / @mikeash any thoughts?

al45tair:
I have no problem with it either way, as long as it works :-)

I think it might be slightly cleaner to configure Clang, if that way will work, not least because then we won't have oddities like things built with our Clang linking with `libgcc` but things built with Swift itself linking with `libcompiler-rt` and `libunwind`.

compnerd:
Accidentally linking libgcc would be a problem - libgcc is an unwinder, and mixing unwinders will lead to crashes as the context is implementation defined.

artemcm:
@swift-ci test

artemcm:
Re-running cross-repo tests in https://github.com/apple/swift/pull/60581, but from what I saw last time, adding `-lunwind` does not help with failing test-cases in main CI if this is merged. So this PR cannot be merged in its current form.

PR #1269, opened 2023-01-25
[Explicit Modules] Use external dependency info to determine if discovered binary dependenceis are frameworks.

artemcm:
The driver library's client may supply information about the current module's dependencies, incl. whether or not they are frameworks. With the dependency scanner no longer re-using source module information from a prior scan (https://github.com/apple/swift/pull/63201), such modules will be reported as prebuilt binary dependencies.

Moreover, add a check to avoid resolving the main module's source module information with a binary dependency, covering an edge case of a module having a dependency with the same name.


artemcm:
@swift-ci test

artemcm:
@swift-ci test

owenv (approved):

nkcsgexi (approved):

PR #1270, opened 2023-01-25
swift-build-sdk-interfaces: fix deprecated AbsolutePath.init

MaxDesiatov:
As `AbsolutePath.init(_:)` is deprecated, we should use throwing `AbsolutePath.init(validating:)` instead to fix the warnings.


MaxDesiatov:
@swift-ci test

artemcm (approved):

PR #1275, opened 2023-01-27
SwiftDriver: remove implicit dependency on TSCUtility

compnerd:
We were relying on the implicit import of the `DiagnosticsEngine` extension to emit the diagnostics.  Explicitly inline the casting to avoid the dependency.


compnerd:
@swift-ci please test

cltnschlosser (approved):

artemcm (approved):
Great catch, thank you @compnerd

PR #1277, opened 2023-01-27
SwiftDriver: avoid using `toSwiftString` for target info

compnerd:
The C-String to Swift-String conversion would corrupt the path on Windows, where the path separator would get dropped.  This would result in the inability to use the driver due to the failure to load the standard library as the SDK would not be consulted.

Directly convert the raw buffer to a `Data` for the foreign representation.  Note that this also makes the code clearer as we explicitly see the copying as opposed to the use of the deprecated `String(bytesNoCopy:length:encoding:freeWhenDone:)` function.

Thanks to @artemcm for the help with this!


compnerd:
@swift-ci please test

artemcm (approved):
Thanks @compnerd !

PR #1278, opened 2023-01-28
SwiftDriver: avoid use of `fatalError` in the driver

compnerd:
Present a diagnostic for the failure rather than invoking `fatalError`. This ensures that the user sees a proper diagnostic for the failure and also enables the test suite to run to completion in case of an error.


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

owenv (approved):

PR #1281, opened 2023-01-31
Utilities: update build-script-helper.py

compnerd:
Remove the reference to `libTSCLibc.dylib` as the module is now static and will not be present on the file system as a dylib as it is compacted into TSCBasic.


compnerd:
@swift-ci please smoke test

artemcm:
@swift-ci test

artemcm (approved):

PR #1283, opened 2023-02-07
Add -package-name to driver

elsh:
Add -package-name to driver
Add packageName to ModuleOutputInfo
Add package name validation and diagnostics
If package name is empty or invalid, fall back to module name
Resolves rdar://105119573.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -243,8 +243,11 @@ public struct Driver {
   /// The debug information to produce.
   @_spi(Testing) public let debugInfo: DebugInfo
 
-  // The information about the module to produce.
+  /// The information about the module to produce.
   @_spi(Testing) public let moduleOutputInfo: ModuleOutputInfo
+
+  /// Name of the package containing a target module or file.
+  @_spi(Testing) public let packageName: String?
 
   /// Info needed to write and maybe read the build record.
   /// Only present when the driver will be writing the record.
@@ -636,6 +639,13 @@ public struct Driver {
 
     // Compute debug information output.
     self.debugInfo = Self.computeDebugInfo(&parsedOptions, diagnosticsEngine: diagnosticEngine)
+
+    // Validate package name; if package name is nil, it will be checked
+    // in the frontend during type check on `package` symbols
+    self.packageName = parsedOptions.getLastArgument(.packageName)?.asSingle
+    if let packageName = packageName, !packageName.sd_isSwiftIdentifier {
+      diagnosticsEngine.emit(.error_bad_package_name(packageName))
+    }
 
     // Determine the module we're building and whether/how the module file itself will be emitted.
     self.moduleOutputInfo = try Self.computeModuleInfo(
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -689,6 +689,29 @@ final class SwiftDriverTests: XCTestCase {
     try assertNoDriverDiagnostics(args: "swiftc", "foo-bar.swift")
   }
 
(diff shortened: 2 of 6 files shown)


elsh:
@swift-ci smoke test

artemcm:
@elsh could you please update `Options.swift` with these instructions:
https://github.com/apple/swift-driver#rebuilding-optionsswift ?

This should bring in `-package-name` and any other newly added/modified flags.

artemcm:
Oh and this repo does not have a smoke test, just `@swift-ci test`.

elsh:
@swift-ci test

artemcm (approved):

nkcsgexi (approved):

elsh:
@swift-ci test macOS

elsh:
@swift-ci test macOS

elsh:
@swift-ci test

elsh:
@swift-ci test macOS

elsh:
@swift-ci test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2427,9 +2427,30 @@ extension Driver {
     // Retrieve and validate module aliases if passed in
     let moduleAliases = moduleAliasesFromInput(parsedOptions.arguments(for: [.moduleAlias]), with: moduleName, onError: diagnosticsEngine)
 
+    // Validate package name and fall back to module name or module alias if it exists
+    var fallback = true
```
artemcm, line 2431: Nit: Could you please make this variable describe which fallback it's referring to?

artemcm (approved):
LGTM!

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -689,6 +689,28 @@ final class SwiftDriverTests: XCTestCase {
...
+    try assertNoDriverDiagnostics(args: "swiftc", "file.swift", "bar.swift", "-module-name", "MyModule", "-package-name", "mypkg", "-emit-module", "-emit-module-path", "../../path/to/MyModule.swiftmodule") { driver in
+        XCTAssertEqual(driver.moduleOutputInfo.packageName,  "mypkg")
+        XCTAssertEqual(driver.moduleOutputInfo.output, .topLevel(try VirtualPath.intern(path: "../../path/to/MyModule.swiftmodule")))
+    }
+  }
+
+  func testPackageNameFallbacks() throws {
+    try assertDriverDiagnostics(args: ["swiftc", "file.swift"]) { // -package-name is nil
+      $1.expect(.warning("package name is nil; will fall back to module name \"file\""))
+    }
```
xymus, line 702: I don't think the warning on no package name is necessary. If we write a custom driver command it's usually for tests or simple scripts, we may just not care about package visibility in that case and we don't want a warning at that point.

If no package is defined I would just make it a warning or error in the code that uses that `package` access level. So if you don't use `package` and don't define a package name then you won't see any diagnostics.

xymus (approved):
I think it's worth removing the warning when not using package at all. Otherwise this looks good!

Tests/SwiftDriverTests/SwiftDriverTests.swift
artemcm: Oh! Agreed with @xymus, we should not warn here.

Sources/SwiftDriver/Driver/Driver.swift
elsh: no more fallback; now emits an error instead

Tests/SwiftDriverTests/SwiftDriverTests.swift
elsh: updated

elsh:
@swift-ci test

elsh:
@swift-ci test macOS

PR #1286, opened 2023-02-10
Add support for macro frontend options 

DougGregor:
Pass through options for macros, especially compiler plugin search
paths. Add a default search path into `lib/swift/host/plugins`.

Driver side of rdar://105095761


DougGregor:
@swift-ci please test

artemcm (approved):

PR #1287, opened 2023-02-14
Pass down an optimization level to the Clang linker invocation

artemcm:
Resolves rdar://105261863

### DIFF
--- a/Sources/SwiftDriver/Toolchains/Toolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/Toolchain.swift
@@ -325,6 +325,30 @@ extension Toolchain {
       supportsResponseFiles: tool.supportsResponseFiles(in: self)
     )
   }
+
+  /// Maps an optimization level swiftc arg to a corresponding flag for the Clang linker driver invocation
+  internal func mapOptimizationLevelToClangArg(from parsedOptions: inout ParsedOptions) -> String? {
+    guard let opt = parsedOptions.getLast(in: .O) else {
+      return nil
+    }
+    let clangArg: String?
+    switch opt.option {
+    case .Oplayground:
+      fallthrough
+    case .Onone:
+      clangArg = "-O0"
+    case .O:
+      fallthrough
+    case .Ounchecked:
+      clangArg = "-O3"
+    case .Osize:
+      clangArg = "-Os"
+    default:
+      clangArg = nil
+      assert(false, "Unhandled Optimization Mode: \(opt.description)")
+    }
+    return clangArg
+  }
 }
 
 @_spi(Testing) public enum ToolchainError: Swift.Error {
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -191,6 +191,11 @@ extension DarwinToolchain {
       commandLine.appendFlag("-fapplication-extension")
     }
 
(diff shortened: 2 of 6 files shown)


artemcm:
cc @dmcyk.

artemcm:
@swift-ci test

dcci (approved):

dcci:
Does this allow arbitrary `-Oxxx` ? (e.g. `-O`, `-O1`, `-Oz`)?
Looks like it, but I want to double check.

artemcm:
> Does this allow arbitrary `-Oxxx` ? (e.g. `-O`, `-O1`, `-Oz`)?
> Looks like it, but I want to double check.

It doesn't, because Swift (driver or frontend) does not have such flags.
Swift only has: 
```
-O                   - Compile with optimizations
-Osize               - Compile with optimizations and target small code size
-Ounchecked          - Compile with optimizations and remove runtime safety checks
-Oplayground         - Compile with optimizations appropriate for a playground
```
So the mapping that myself and @dmcyk came up with is:
```
-O and -Ounchecked -> -O3
-Onone             -> -O0
-Osize             -> -Os
-Oplayground       -> -O0 (that's what it maps to in swift-frontend) 
```

dcci:
sounds good.

artemcm:
Users can override this with a more fine-grained Clang flag if they wish to, adding to their `swiftc` command:
```
-Xclang-linker -O2
```

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -325,6 +325,29 @@ extension Toolchain {
       supportsResponseFiles: tool.supportsResponseFiles(in: self)
     )
   }
+
+  /// Maps an optimization level swiftc arg to a corresponding flag for the Clang linker driver invocation
+  internal func mapOptimizationLevelToClangArg(from parsedOptions: inout ParsedOptions) -> String? {
+    guard let opt = parsedOptions.getLast(in: .O) else {
+      return nil
```
dmcyk, line 332: Is a lack of any -O* option equivalent to -Onone? If so, would it make sense to map to -O0 to make it explicit?

Sources/SwiftDriver/Toolchains/Toolchain.swift
```diff
@@ -325,6 +325,29 @@ extension Toolchain {
...
+    case .Onone:
+      clangArg = "-O0"
+    case .O:
+      fallthrough
+    case .Ounchecked:
+      clangArg = "-O3"
+    case .Osize:
+      clangArg = "-Os"
+    default:
+      clangArg = nil
```
dmcyk, line 347: nit; maybe worth adding an assert so if a new opt level is introduced we won't forget to add the mapping?

dmcyk (commented):
Left two minor comments, but otherwise LGTM!

Sources/SwiftDriver/Toolchains/Toolchain.swift
dmcyk: On a second thought I think this is correct, we probably shouldn't make an assumption.
e.g., clang can't make assumption about the optimization level when not explicitly specified, because build systems tend to not pass the level to clang linker invocation.

Sources/SwiftDriver/Toolchains/Toolchain.swift
artemcm: Added, thanks.

artemcm:
@swift-ci test

artemcm:

PR #1288, opened 2023-02-15
[Swift Explain] Add a simple tool to break down and annotate Swift compile commands

artemcm:


Sources/swift-explain/main.swift
```diff
@@ -0,0 +1,132 @@
...
+         string.hasSuffix("clang") ||
+         string.hasSuffix("libtool") ||
+         string.hasSuffix("dsymutil") ||
+         string.hasSuffix("clang") ||
+         string.hasSuffix("clang++") ||
+         string.hasSuffix("swift-autolink-extract") ||
+         string.hasSuffix("lldb") ||
+         string.hasSuffix("dwarfdump") ||
+         string.hasSuffix("swift-help") ||
+         string.hasSuffix("swift-api-digester")
```
compnerd, line 41: Do we need to worry about `.exe` extensions on these?

PR #1290, opened 2023-02-15
Recover functionality of passing down unrecognized frontend-only options

artemcm:
Once '-e' got introduced, we started failing to detect when the driver was specified a frontend-only option and passing it down to the frontend, if such frontend-only option started with '-e...'. This is because we do a prefix match on the options trie and treat an unrecognized suffix as the argument to it, and later detect whether or not such an option is recognized.

We can restore the functionality for flags that the driver is aware of as being frontend-only flags, by not skipping such flags when constructing the trie and detecting them explicitly (`.noDriver`) and processing them as any other unknown flag to be checked whether or not it can be passed down to the frontend.

And reword the diagnostic a bit to indicate to the user that `-Xfrontend` can get rid of the warning.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -689,7 +689,7 @@ public struct Driver {
       Driver.isOptionFound($0, allOpts: supportedFrontendFlagsLocal)
     }
     self.savedUnknownDriverFlagsForSwiftFrontend.forEach {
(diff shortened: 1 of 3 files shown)


artemcm:
@swift-ci test

owenv:
Let's say there's a driver flag called -do-something and a frontend flag -do-somethingwithfrontend. Would this change cause `-do-somethingwithfrontend` to get parsed incorrectly if -do-something supported a joined argument? I know it seems unlikely, but also hard to debug if it ever happens

PR #1292, opened 2023-02-17
Be more strict about invalid -module-name arguments

keith:
In the case this argument is passed explicitly and invalid, it was previous ignored and fell back to 'main' even in cases where you might not want that like compiling object files. It doesn't seem necessary to allow invalid cases like this through regardless of what object type is being built.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2290,32 +2290,6 @@ extension Driver {
     return ""
   }
 
-  /// Whether we are going to be building an executable.
-  ///
-  /// FIXME: Why "maybe"? Why isn't this all known in advance as captured in
-  /// linkerOutputType?
-  private static func maybeBuildingExecutable(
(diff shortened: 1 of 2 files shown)


keith:
@swift-ci please test

artemcm (approved):
This change LGTM, thanks @keith!
Whatever this breaks is probably worth breaking to see what's going on...

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2290,32 +2290,6 @@ extension Driver {
     return ""
   }
 
-  /// Whether we are going to be building an executable.
-  ///
-  /// FIXME: Why "maybe"? Why isn't this all known in advance as captured in
-  /// linkerOutputType?
-  private static func maybeBuildingExecutable(
```
artemcm, line 2297: Happy to see this gone.

artemcm:

keith:
Thanks!

PR #1293, opened 2023-02-19
Utilities: adjust install rules for apple/swift-llbuild#852

compnerd:
The libllbuild library has been converted into a static product and interalised into libllbuildSwift.  This avoids the distribution issues here.  Remove the library from the install set.


compnerd:
https://github.com/apple/swift-llbuild/pull/852

@swift-ci please test

compnerd:
There are two other changes that need to be submitted in tandem.

PR #1295, opened 2023-02-21
Remove support for linking arclite

artemcm:
Equivalent to https://github.com/apple/swift/pull/63662
--------------------------------------------------
Darwin no longer uses arclite and it's no longer distributed in the macOS SDKs.

This leaves the options '-link-objc-runtime' and '-no-link-objc-runtime' in place, but strips out all the logic that actually used them.

Part of: rdar://105406972


mikeash (approved):

artemcm:

artemcm:

PR #1296, opened 2023-02-22
[Dependency Scanning] Pass the bridging header path to scanning action invocations

artemcm:


artemcm:
@swift-ci test

artemcm:
Suspect this may need https://github.com/apple/swift/pull/63831/files to land in the toolchain picked up by CI.

artemcm:
We can also do better here by only specifying bridging header's own dependencies as inputs to the GeneratePCH job. Instead of all of the dependencies of the main module. That way this job will be able to run potentially much sooner than the main compilation. Looking into that as well.

nkcsgexi (approved):

artemcm:
@swift-ci test

artemcm:

artemcm:
@swift-ci test

artemcm:
@swift-ci test

cachemeifyoucan:
> We can also do better here by only specifying bridging header's own dependencies as inputs to the GeneratePCH job. Instead of all of the dependencies of the main module. That way this job will be able to run potentially much sooner than the main compilation. Looking into that as well.

That would so super welcomed too. Currently I need to figure out how to pass the PCH from within one `swift-frontend` invocation to clang Importer. If we can move PCH to an individual job that can run much sooner, it can be a cleaner solution for caching

PR #1299, opened 2023-02-24
Remove `package-collection` and `package-registry` from top-level command help

bitjammer:
This help should include top-level commands–these should be treated as sub-commands of `swift package`.

rdar://104603198


bitjammer:
@swift-ci Please test

neonichu (approved):

PR #1303, opened 2023-03-02
Attempt to recover from in-process Target Info and Supported Features query failures.

artemcm:
Catch thrown exceptions and fallback to the previous mechanism of shelling out a `swift-frontend -print-target-info` and `swift-frontend -emit-supported-features` tasks. We are seeing spurious failures here and this will help potential users while we investigate the root cause.

Workaround for rdar://105559904


artemcm:
@swift-ci test

artemcm:
@swift-ci test

nkcsgexi (approved):
It would be nice to figure the root cause for the query failure, but this short term fix makes sense to me!

owenv (approved):

PR #1307, opened 2023-03-07
Add `-gcc-toolchain` flag to swift-driver

etcwilde:
GCC toolchains are useful when cross-compiling to Linux with a given gcc toolchain. The gcc toolchain contains tools for linking, libraries for linking against, and the headers that map to those libraries.

This patch adds a `-gcc-toolchain` flag to the driver so that Swift can pass that information down to both the clang importer and clang linker appropriately.

I went with the double-dash, equals format passed to clang because maskray keeps poking at the clang flags and that format is the only one that goes all of the way back (to 2013). I'm also passing the path down directly instead of reinterpreting them as an absolute path. The driver seems to use the `-working-directory` path to resolve most of these instead of the current working directory path, which I believe is inconsistent with Clang. Letting clang deal with the paths will, by definition, be consistent with clang. Yes, little Bobby Tables is waiting to make a mess of things.

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -13,6 +13,7 @@
 import class TSCBasic.LocalFileOutputByteStream
 import class TSCBasic.TerminalController
 import struct TSCBasic.RelativePath
+import struct TSCBasic.AbsolutePath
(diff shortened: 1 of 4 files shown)


etcwilde:
@swift-ci please test

etcwilde:
https://github.com/apple/swift/pull/65108

@swift-ci please test

artemcm (approved):

PR #1308, opened 2023-03-08
Compute and reuse bridging header pch explicitly in explicit module b…

cachemeifyoucan:
…uild

When `-pch-output-dir` is used during explicit module build, compute the output pch path and explicitly reuse that to downstream job via `-import-objc-header output.pch`.

rdar://106430264

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -323,7 +323,12 @@ extension Driver {
       commandLine.appendFlag(.importObjcHeader)
       if bridgingHeaderHandling == .precompiled,
           let pch = bridgingPrecompiledHeader {
-        if parsedOptions.contains(.pchOutputDir) {
+        // For explicit module build, we directly pass the compiled pch as
+        // `-import-objc-header`, rather than rely on swift-frontend to locate
+        // the pch in the pchOutputDir and can start an implicit build in case
+        // of a lookup failure.
+        if parsedOptions.contains(.pchOutputDir) &&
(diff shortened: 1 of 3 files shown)


cachemeifyoucan:
After this patch, there is a warning:
```
<unknown>:0: warning: argument unused during compilation: '-fmodules-cache-path=PATH'
```

That needs to fixed on swift-frontend/scanner side maybe because we don't need access to module cache anymore for explicit build

artemcm:
> After this patch, there is a warning:
> 
> ```
> <unknown>:0: warning: argument unused during compilation: '-fmodules-cache-path=PATH'
> ```
> 
> That needs to fixed on swift-frontend/scanner side maybe because we don't need access to module cache anymore for explicit build

That's actually already there before this patch. 👍🏼

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -323,7 +323,8 @@ extension Driver {
       commandLine.appendFlag(.importObjcHeader)
       if bridgingHeaderHandling == .precompiled,
           let pch = bridgingPrecompiledHeader {
-        if parsedOptions.contains(.pchOutputDir) {
+        if parsedOptions.contains(.pchOutputDir) &&
+           !parsedOptions.contains(.driverExplicitModuleBuild) {
```
artemcm, line 327: @cachemeifyoucan could you please add a brief comment explaining why just an output dir won't do for explicit modules and we need direct known paths?

artemcm (approved):

cachemeifyoucan:
@swift-ci please test

PR #1306, opened 2023-03-07
[cxx-interop] Re-run `makeOptions` to generate `-cxx-compatibility-mode`.

zoecarver:
Refs https://github.com/apple/swift/pull/64088

PR #1302, opened 2023-02-28
Add Swift compilation flags to enable Clang's validate-per-build-session module behavior

artemcm:
Compiler flags added in: https://github.com/apple/swift/pull/63975

Driver portion of rdar://105982120

PR #1309, opened 2023-03-10
SwiftOptions: regenerate the swift options

compnerd:
Regenerate the swift options from the C++ driver to pick up changes for the new Windows options that are being introduced to control the SDK and Visual C++ tools that we use.


compnerd:
@swift-ci please test

artemcm (approved):
Worth adding logic that only adds these flags to frontend invocations when they support the new flags? 
e.g. https://github.com/apple/swift-driver/pull/1302/files

compnerd:
I like that idea!  I can do that as part of the change that forwards the argument.

zoecarver:
Thanks Artem, I guess that's why #1306 is also failing. I will fix that test in that PR but will likely cause merge conflicts over here :/ Sorry, Saleem.

compnerd:

PR #1313, opened 2023-03-21
SwiftDriver: add support for baremetal targets (to reflect https://github.com/apple/swift/pull/35970)

kubamracek:


Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -2892,12 +2892,16 @@ extension Driver {
 
 extension Triple {
   func toolchainType(_ diagnosticsEngine: DiagnosticsEngine) throws -> Toolchain.Type {
+    guard let os else {
+      diagnosticsEngine.emit(.error_unknown_target(triple))
+      throw Driver.ErrorDiagnostics.emitted
+    }
```
compnerd, line 2898: Why the guard over the default case?

Sources/SwiftDriver/Utilities/Triple+Platforms.swift
```diff
@@ -282,9 +282,10 @@ extension Triple {
   ///   identified as just `darwin` instead of by individual platform names.
   ///   Defaults to `false`.
   public func platformName(conflatingDarwin: Bool = false) -> String? {
-    switch os {
-    case nil:
+    guard let os else {
       fatalError("unknown OS")
+    }
```
compnerd, line 287: Why not sink the `fatalError` into the `default` case?

Sources/SwiftDriver/Driver/Driver.swift
rauhul: I'm guessing it's to avoid future reader ambiguity between Optional.none and OS.none.

compnerd: That would be helpful as a comment if that is the case.  Doubly so since the handling is identical.

Sources/SwiftDriver/Utilities/Triple+Platforms.swift
MaxDesiatov: I'd assume if `default` case is added to this `switch`, that would silence future compiler errors in this `switch` when a new case is added to `Triple.OS`.

compnerd: Sure, but you can handle that with `.some` and `.none` :)

kubamracek:
So apparently the "none" case was at least confusing but I'd say misleading. Changed it to "noneOS" to avoid the clash with Optional.none. Added tests.

kubamracek:
@swift-ci please test

compnerd (approved):
Thank you for porting the work from the c++ driver!  I know I'd been slow to finish up the work :(

MaxDesiatov:
Is it worth retrying the Windows job? Couldn't find the exact failure reason for it.

compnerd:
Seems to have timed out.  I think that the machine sometimes is under higher load.

compnerd:

PR #1317, opened 2023-03-26
Allow clients of `Driver` to pass `DiagnosticsHandler`

MaxDesiatov:
Uses of `DiagnosticsEngine` are [deprecated in SwiftPM](https://github.com/apple/swift-package-manager/blob/2c9fa974bc6ed9662e9399b734a0567952fb3e1d/Sources/Basics/Observability.swift#L530), since it uses a different logging infrastructure based on its own `ObservabilityScope` type. Resolving that deprecation warning is possible if `Driver.init` is able to take a diagnostics handler directly.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -143,6 +143,12 @@ public struct Driver {
     }
   }
 
+  /// Specific implementation of a diagnostics output type that can be used when initializing a new `Driver`.
+  public enum DiagnosticsOutput {
+    case engine(DiagnosticsEngine)
+    case handler(DiagnosticsEngine.DiagnosticsHandler)
+  }
+
   /// The set of environment variables that are visible to the driver and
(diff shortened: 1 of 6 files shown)


MaxDesiatov:
@swift-ci test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -476,7 +482,7 @@ public struct Driver {
   public init(
     args: [String],
     env: [String: String] = ProcessEnv.vars,
-    diagnosticsEngine: DiagnosticsEngine = DiagnosticsEngine(handlers: [Driver.stderrDiagnosticsHandler]),
```
owenv, line 479: Could we keep a deprecated copy of the old initializer which forwards to this one to avoid breaking clients?

MaxDesiatov:
https://github.com/apple/swift-package-manager/pull/6333

@swift-ci test

owenv (approved):

MaxDesiatov:
@swift-ci test windows

PR #1320, opened 2023-03-29
[Macros] Set -external-plugin-path when the toolchain is not Xcode

rintaro:
rdar://108525615

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -275,6 +275,10 @@ extension Driver {
       commandLine.appendPath(localPluginPath)
     }
 
+    if isFrontendArgSupported(.externalPluginPath) {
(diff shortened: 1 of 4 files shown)


Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -280,6 +280,35 @@ extension Driver {
       commandLine.appendPath(localPluginPath)
     }
 
+    // Pass -external-plugin-path if the current toolchain is not a Xcode
+    // default toolchain.
+    if
+      isFrontendArgSupported(.externalPluginPath),
+      let darwinToolchain = toolchain as? DarwinToolchain,
```
artemcm, line 287: Can we instead move this to `addPlatformSpecificCommonFrontendOptions`, please?

artemcm (approved):
Could we please add a macOS-only test for this? 
Other than that and moving to code to a toolchain-specific place, LGTM!

rintaro:
@swift-ci Please test

PR #1323, opened 2023-04-06
[Dependency Scanning] Open `libSwiftScan` without `RTLD_FIRST` 

artemcm:
`RTLD_FIRST` seems to be causing us to use the first-opened `,dylib` in a given process, ignoring hte fact that subsequent calls to `dlopen` may be opening a different library.

### DIFF
--- a/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
+++ b/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
@@ -100,7 +100,7 @@ internal extension swiftscan_diagnostic_severity_t {
     #if os(Windows)
     self.dylib = try Loader.load(path.pathString, mode: [])
     #else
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci test

owenv (approved):

akyrtzi:
What I have seen is that `dlopen` loads a different library than specified in path when `DYLD_LIBRARY_PATH` is involved, which is actually documented:

> When path contains a slash (i.e. a full path or a partial path) dlopen()
     searches the following the following until it finds a compatible Mach-O
     file: $DYLD_LIBRARY_PATH (with leaf name from path ), current working
     directory (for partial paths), $DYLD_FALLBACK_LIBRARY_PATH (with leaf
     name from path ).

artemcm:
This fix I do not believe is correct. Thank you @akyrtzi !

PR #1325, opened 2023-04-06
Propagate sanitizer arguments to the clang-linker-driver invocations for dynamic libraries

artemcm:
Unless we are building a static library, we need to propagate `-fsanitizer=` flag to the clang linker driver, otherwise sanitizers libraries will not be found at link-time.

Resolves rdar://107733898

### DIFF
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -199,7 +199,7 @@ extension DarwinToolchain {
     // Linking sanitizers will add rpaths, which might negatively interact when
     // other rpaths are involved, so we should make sure we add the rpaths after
     // all user-specified rpaths.
-    if linkerOutputType == .executable && !sanitizers.isEmpty {
+    if linkerOutputType != .staticLibrary && !sanitizers.isEmpty {
       let sanitizerNames = sanitizers
         .map { $0.rawValue }
         .sorted() // Sort so we get a stable, testable order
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
```diff
@@ -199,7 +199,7 @@ extension DarwinToolchain {
     // Linking sanitizers will add rpaths, which might negatively interact when
     // other rpaths are involved, so we should make sure we add the rpaths after
     // all user-specified rpaths.
-    if linkerOutputType == .executable && !sanitizers.isEmpty {
+    if linkerOutputType != .staticLibrary && !sanitizers.isEmpty {
```
artemcm, line 202: @Keith do you, by any chance, remember why we had this restriction for `.executable`? 

`!= .staticLibrary` seems like the right thing to do here.

artemcm:
@swift-ci test

etcwilde (approved):
LGTM

Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
keith: hrm im not sure, maybe I got it from this https://github.com/apple/swift-driver/blob/7a96ffdd68dda2049cc2da005493b872ea43e3bc/Sources/SwiftDriver/Jobs/GenericUnixToolchain%2BLinkerSupport.swift#L272, I think you're right though

keith: besides previously there was a check for excluding the fuzzer sanitizer for executables, so you might want that back https://github.com/apple/swift-driver/commit/a557d591d7a2dab74611ac268d343384dd6d3292#diff-d44ec7a906797a39be2e42ecb334167ae217537ad09a726d1f905c4ba9130200L365

PR #1326, opened 2023-04-06
[5.9] Propagate sanitizer arguments to the clang-linker-driver invocations for dynamic libraries

artemcm:
Cherry-pick of: https://github.com/apple/swift-driver/pull/1325
------------------------------
Unless we are building a static library, we need to propagate `-fsanitizer=` flag to the clang linker driver, otherwise sanitizers libraries will not be found at link-time.

Resolves rdar://107733898


artemcm:
@swift-ci test

nkcsgexi (approved):

PR #1333, opened 2023-04-18
[5.9 🍒][Explicit Module Builds] Remove logic for Swift-client-specific Clang module dependency jobs

artemcm:
Cherry-pick of #1332
------------------------------ 
We have moved towards a build model where the output of the dependency scanner is the canonical collection of modules that must be built. The driver's logic to generate potentially multiple examples of a given Clang dependency, with different command sets, as required per Swift client module in the build graph, is no longer valid - even module output paths are no longer determined by the driver. Remaining logic that is removed in this PR may result in mistakenly producing duplicate jobs for the same Clang PCM build job.

Change affects explicit module builds only and does not have any impact on implicit build code-path.


artemcm:
@swift-ci test

nkcsgexi (approved):
Nice clean-ups!

PR #1343, opened 2023-04-28
Use system-appropriate method when quoting path arguments on command-lines

artemcm:
Particularly, we have a special utility to quote paths on Windows, where the approach of adding `'` prefix and postfix is not valid.


compnerd (approved):
Awesome!  Thank you for fixing this @artemcm

artemcm:
@swift-ci test

compnerd:
Going to merge this to unblock the windows toolchains.

PR #1345, opened 2023-04-30
Don't try to link private CYaml in CMake

stephank:
As of https://github.com/jpsim/Yams/pull/343 and Yams 5.0.0, CYaml is now a private static lib, and we no longer have to link it.

I think we also have to fix this in `release/5.9` and `main`? Do we need separate PRs for those?

### DIFF
--- a/Sources/SwiftDriver/CMakeLists.txt
+++ b/Sources/SwiftDriver/CMakeLists.txt
@@ -131,7 +131,6 @@ target_link_libraries(SwiftDriver PUBLIC
   TSCUtility
   SwiftOptions)
 target_link_libraries(SwiftDriver PRIVATE
-  CYaml
   Yams
   CSwiftScan)
 


compnerd:
https://github.com/apple/swift/blob/main/utils/update_checkout/update-checkout-config.json#L121 shows that we are on 5.0.1 which seems like it does mean that we shouldn't be linking in CYaml here.  Thanks for fixing this @stephank!

compnerd:
@swift-ci please test

artemcm (approved):
Thanks @stephank

artemcm:
> I think we also have to fix this in `release/5.9` and `main`? Do we need separate PRs for those?

Please open a new PR with a cherry-pick of this commit against the `release/5.9` branch.

artemcm:
This repo's CI system does not exercise the CMake build here. So I am running a cross-repo test from https://github.com/apple/swift/pull/60581, just for sanity. Please don't merge until that's green.

airspeedswift:
Is there a strong motivator to change this in 5.8, versus just in main and 5.9?

compnerd:
https://github.com/apple/swift/blob/main/utils/update_checkout/update-checkout-config.json#L251 ... seems to indicate that we _should_ fix it on 5.8 as well.  Not sure how this escaped the 5.8 release.

stephank:
I believe there's an argument to be made this is Nixpkgs specific. CYaml is probably still available in the Yams _build tree_ in during the normal Swift SDK build process, but in order to leverage existing CMake infra in Nixpkgs, we split up each dependency into its own build,  and also perform separate install steps. CYaml is no longer available when you install Yams separately.

On the other hand, CYaml is now a private detail of Yams, and in theory they could pull the rug on you. :) (Though unlikely, I guess, especially as long as the version is pinned.)

PR #1346, opened 2023-05-01
In WMO, make sure the dummy "primary" compilation input is a Swift source file

artemcm:
Otherwise, if the first argument is anything else (e.g. an object file) we will use it as a key and the compiler won't be too happy about that.

Resolves rdar://108741675

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -585,7 +585,10 @@ extension Driver {
         // To match the legacy driver behavior, make sure we add the first input file
         // to the output file map if compiling without primary inputs (WMO), even
         // if there aren't any corresponding outputs.
-        entries[inputFiles[0].fileHandle] = [:]
+        guard let firstSourceInputHandle = inputFiles.first(where:{ $0.type == .swift })?.fileHandle  else {
+          fatalError("Formulating swift-frontend invocation without any input .swift files")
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -585,7 +585,10 @@ extension Driver {
         // To match the legacy driver behavior, make sure we add the first input file
         // to the output file map if compiling without primary inputs (WMO), even
         // if there aren't any corresponding outputs.
-        entries[inputFiles[0].fileHandle] = [:]
+        guard let firstSourceInputHandle = inputFiles.first(where:{ $0.type == .swift })?.fileHandle  else {
+          fatalError("Formulating swift-frontend invocation without any input .swift files")
```
etcwilde, line 589: I think I can hit this `fatalError`, though honestly, if we're even formulating FrontendJobs when we don't have any source inputs, something has probably gone wrong before this point.

artemcm: My thinking is that, if anything, this will help catch such issues early. This code-path should not otherwise be taken if we don't have any source inputs, I don't think.

etcwilde (approved):
Yep, this combined with https://github.com/apple/swift-driver/pull/1347 looks good.
Thank you for looking into this.

PR #1347, opened 2023-05-01
Do not attempt to create compilation jobs when no Swift source files are specified as inputs

artemcm:
Doing so means downstream code is likely to error or crash, because it expects Swift source (or SIL) inputs. 

Resolves rdar://108057797


etcwilde (approved):
Alright, let's go!
Thank you!

PR #1350, opened 2023-05-02
In WMO, make sure the dummy "primary" compilation input is a Swift source file

artemcm:
Attempting to land https://github.com/apple/swift-driver/pull/1346 again
-------------------------------------------------------
Otherwise, if the first argument is anything else (e.g. an object file) we will use it as a key and the compiler won't be too happy about that.

Resolves rdar://108741675


artemcm:
The original PR for this and https://github.com/apple/swift-driver/pull/1347 got tested at the same time and passed testing; but, the change in https://github.com/apple/swift-driver/pull/1347 wasn't quite right, which got exposed when this was merged. I've amended this change now to fix the logic error. 
cc @etcwilde

artemcm:

etcwilde (approved):

PR #1353, opened 2023-05-04
build-script-helper.py: remove unused `distutils` import

MaxDesiatov:
`distutils` was deprecated in Python 3.10 and will be removed in Python 3.12. We should not depend on it and this import looks unused anyway.


MaxDesiatov:
@swift-ci test

PR #1354, opened 2023-05-05
[Dependency Scanning] Unify path-escaping handling using TSC's spm_shellEscaped

artemcm:
Command lines generated for `libSwiftScan` (either for actual dependency scanning queries, or for target-info queries) do not go through the shell, and are instead tokenized at the entry-points (inside `libSwiftScan`). In order for them to be correctly tokenized, we must ensure that all the various file paths are escaped in case they contain whitespace characters, to ensure they get treated as a whole path, rather than multiple strings.

The driver previously relied on a separate mechanism to do this for libSwiftScan command lines, by getting the `ArgsResolver` to always escape `.path` arguments. This turned out to be insufficient, because it happens to miss a whole class of paths specified on the command-line that the driver cannot/doesn't recognize as paths: arguments forwarded to tools, such as `-Xcc` or `-Xfrontend` or `-Xclang-linker`. As a result, it is possible for such paths to end-up unescaped and fail to get tokenized correctly by `libSwiftScan`.

Instead, this change switches the formulation of these command-lines to use the existing `spm_shellEscaped` mechanism from TSC which is a robust way to achieve exactly the desired behavior: produce shell-escaped command-line arguments by detecting flags/arguments that contain characters that would prevent correct tokenization: whitespaces, etc, and only quote-escaping them, and doing so in a platform-appropriate manner (e.g. using `"` instead of `'` on Windows)

Resolves rdar://108971395

### DIFF
--- a/Sources/SwiftDriver/Execution/ArgsResolver.swift
+++ b/Sources/SwiftDriver/Execution/ArgsResolver.swift
@@ -57,17 +57,16 @@ public final class ArgsResolver {
     }
   }
 
-  public func resolveArgumentList(for job: Job, useResponseFiles: ResponseFileHandling = .heuristic,
-                                  quotePaths: Bool = false) throws -> [String] {
-    let (arguments, _) = try resolveArgumentList(for: job, useResponseFiles: useResponseFiles,
-                                                 quotePaths: quotePaths)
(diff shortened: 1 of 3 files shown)


artemcm:
@swift-ci test

nkcsgexi (approved):
Thank you!

PR #1358, opened 2023-05-09
Specify encoding when running a ninja invocation

ahoppen:
This makes the error messages show up properly in the logs instead of as a binary string.


artemcm (approved):
Nice, thank you!

PR #1356, opened 2023-05-08
Disable adding '-external-plugin-path' in integrated driver mode

rintaro:
rdar://109060298

PR #1359, opened 2023-05-09
Revert "[Macros] Set -external-plugin-path when the toolchain is not Xcode"

rintaro:
Revert #1320 and #1356 . Doing this in Driver was not a good idea.


rintaro:
@swift-ci Please test

artemcm (approved):

owenv (approved):

PR #1361, opened 2023-05-12
Add support for compatibility shims that we don't want to -force-load

slavapestov:
The CompatibilityPacks shim doesn't install any override hooks; it just exports a pair of symbols. We don't want to link the shim if we're not using the symbols.

This is part of https://github.com/apple/swift/pull/65884.


slavapestov:
@swift-ci Please test

artemcm (approved):

slavapestov:
@swift-ci Please test

PR #1363, opened 2023-05-18
Add option `-explain-module-dependency` to explain a given direct or transitive module dependency

artemcm:
Adds support for '-explain-module-dependency <MODULE_NAME>', which causes the driver to run a dependency scan, and find all dependency paths that cause the main module to depend, directly or transitively, on a given module dependency.


nkcsgexi:
🎉

artemcm:
@swift-ci test

nkcsgexi (approved):

PR #1365, opened 2023-05-19
[Dependency Scanning] Add support for Swift Overlay dependencies as a separate dependency details field

artemcm:
Instead of assuming such dependencies are contained in the 'directDependencies' field of the module info, support querying them from a details of a textual module.

Adds support for dependency scanner changes added in:
https://github.com/apple/swift/pull/66031


nkcsgexi (approved):
Makes sense to me!

PR #1371, opened 2023-06-07
[cxx-interop] make -emit-clang-header-path a driver flag

hyp:


artemcm:
@hyp, just to sanity check, was `Options.swift` generated using the automation described in:
?

artemcm (approved):

PR #1376, opened 2023-06-13
SwiftDriver: mark bitcode interfaces as `internal`

compnerd:
We do not vend the bitcode reading as an API that others may use.  Mark this as `internal` as the interfaces are used strictly within the `IncrementalCompilation` module.  If we need, we can extend in the further to use `@package` visibility.


compnerd:
@swift-ci please test

artemcm (approved):
This is good, thank you!

PR #1382, opened 2023-06-20
Add `repl` to swift-help topics

bitjammer:
Most just run `swift repl` here, but also reference the possibility of passing Swift Frontend arguments.

rdar://109411493


bitjammer:
@swift-ci Please test

artemcm (approved):

PR #1383, opened 2023-06-20
Driver: allow static executables when using Musl

MaxDesiatov:
This error diagnostic doesn't make sense when linking on Linux with Musl, which allows fully statically linked executables.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -660,7 +660,9 @@ public struct Driver {
 
     self.lto = Self.ltoKind(&parsedOptions, diagnosticsEngine: diagnosticsEngine)
     // Figure out the primary outputs from the driver.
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci test

al45tair (commented):
Seems reasonable to me; I wonder whether `Triple` should grow a `supportsStaticExecutables` property though, rather than explicitly testing for `.musl` in here?

MaxDesiatov:
@swift-ci test

al45tair (approved):
LGTM.

MaxDesiatov:
@swift-ci test windows

rauhul:
> Seems reasonable to me; I wonder whether `Triple` should grow a `supportsStaticExecutables` property though, rather than explicitly testing for `.musl` in here?

+100, I was planning on adding a `supportsDynamicExecutables` to handle `os(none)`

artemcm (approved):

PR #1390, opened 2023-07-04
[Driver] For immediate mode prefer using the default SDK over the `SDKROOT` variable

akyrtzi:
Avoid using `SDKROOT` because that may be set to a compilation platform (like iOS) that is different than the host. For immediate mode the host SDK is what needs to be used.


akyrtzi:
@swift-ci Please test

artemcm (approved):
Makes sense.

PR #1392, opened 2023-07-05
Set Explicit Incremental Module Build tests to use own module cache

artemcm:
With the timeline of the module cache being tied to the current run of the test suite. To try and avoid possible conflicts with stale cached modules.


artemcm:
@swift-ci test

bnbarham (approved):

PR #1393, opened 2023-07-05
Support Swift CompileJob Caching

cachemeifyoucan:
Add swift-driver support for compiler caching.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -74,6 +74,8 @@ public struct Driver {
     case missingContextHashOnSwiftDependency(String)
     case dependencyScanningFailure(Int, String)
     case missingExternalDependency(String)
+    // Compiler Caching Failures
+    case unsupportedConfigurationForCaching(String)
 
     public var description: String {
       switch self {
@@ -135,6 +137,8 @@ public struct Driver {
         return "unable to load output file map '\(path)': \(error)"
       case .missingExternalDependency(let moduleName):
         return "Missing External dependency info for module: \(moduleName)"
+      case .unsupportedConfigurationForCaching(let reason):
+        return "unsupported configuration for -cache-compile-job: \(reason)"
       case .baselineGenerationRequiresTopLevelModule(let arg):
         return "generating a baseline with '\(arg)' is only supported with '-emit-module' or '-emit-module-path'"
       case .optionRequiresAnother(let first, let second):
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
@@ -71,6 +71,11 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
     self.enableCAS = enableCAS
   }
 
+  /// Supports resolving bridging header pch command from swiftScan.
+  public func supportsBridgingHeaderPCHCommand() throws -> Bool {
+    return try swiftScanOracle.supportsBridgingHeaderPCHCommand()
+  }
+
   /// Generate build jobs for all dependencies of the main module.
   /// The main module itself is handled elsewhere in the driver.
   ///
@@ -241,7 +246,12 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
       inputs.append(TypedVirtualPath(file: dependencyModule.modulePath.path,
                                      type: .swiftModule))
 
-      for headerDep in dependencyModule.prebuiltHeaderDependencyPaths ?? [] {
+      let prebuiltHeaderDependencyPaths = dependencyModule.prebuiltHeaderDependencyPaths ?? []
+      if enableCAS && !prebuiltHeaderDependencyPaths.isEmpty {
+        throw Driver.Error.unsupportedConfigurationForCaching("module \(dependencyModule.moduleName) has prebuilt header dependency")
+      }
+
+      for headerDep in prebuiltHeaderDependencyPaths {
         commandLine.appendFlags(["-Xcc", "-include-pch", "-Xcc"])
         commandLine.appendPath(VirtualPath.lookup(headerDep.path))
         inputs.append(TypedVirtualPath(file: headerDep.path, type: .pch))
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -687,6 +687,14 @@ extension Driver {
     try dependencyPlanner.resolveBridgingHeaderDependencies(inputs: &inputs, commandLine: &commandLine)
   }
 
+  /// If explicit dependency planner supports creating bridging header pch command.
+  public func supportsBridgingHeaderPCHCommand() throws -> Bool {
+    guard let dependencyPlanner = explicitDependencyBuildPlanner else {
+      return false
+    }
+    return try dependencyPlanner.supportsBridgingHeaderPCHCommand()
+  }
+
   /// In Explicit Module Build mode, distinguish between main module jobs and intermediate dependency build jobs,
   /// such as Swift modules built from .swiftmodule files and Clang PCMs.
   public func isExplicitMainModuleJob(job: Job) -> Bool {
--- a/Sources/SwiftDriver/Jobs/VerifyModuleInterfaceJob.swift
+++ b/Sources/SwiftDriver/Jobs/VerifyModuleInterfaceJob.swift
@@ -25,6 +25,12 @@ extension Driver {
                                                                     input: mainInput.fileHandle)
   }
 
+  @_spi(Testing)
+  public func supportExplicitModuleVerifyInterface() -> Bool {
+    // swift-frontend that has -input-file-key option can support explicit module build for verify interface.
+    return isFrontendArgSupported(.inputFileKey)
+  }
+
   mutating func verifyModuleInterfaceJob(interfaceInput: TypedVirtualPath, emitModuleJob: Job, optIn: Bool) throws -> Job {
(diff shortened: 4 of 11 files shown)


Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -80,10 +80,13 @@ extension Driver {
     // May also be used for generation of the dependency graph itself in ExplicitModuleBuild mode.
     if (parsedOptions.contains(.driverExplicitModuleBuild) &&
           moduleDependencyGraphUse == .computed) {
-      if isGeneratePCH {
+      switch kind {
+      case .generatePCH:
         try addExplicitPCHBuildArguments(inputs: &inputs, commandLine: &commandLine)
-      } else {
+      case .compile, .emitModule, .interpret:
```
cachemeifyoucan, line 86: This might break some JobKind that are not covered in the test suite. Looks like `.verifyModuleInterface` should also work with explicit module build. I need to get that working with CAS as well.

cachemeifyoucan:
New changes that requires https://github.com/apple/swift/pull/67263

Sources/CSwiftScan/include/swiftscan_header.h
```diff
@@ -77,6 +77,18 @@ typedef struct {
...
 typedef void *swiftscan_scanner_t;
 
+//=== CAS/Caching Specification -------------------------------------------===//
+typedef struct swiftscan_cas_s *swiftscan_cas_t;
+
+typedef enum {
+  SWIFTSCAN_OUTPUT_TYPE_OBJECT = 0,
+  SWIFTSCAN_OUTPUT_TYPE_SWIFTMODULE = 1,
+  SWIFTSCAN_OUTPUT_TYPE_SWIFTINTERFACE = 2,
+  SWIFTSCAN_OUTPUT_TYPE_SWIFTPRIAVEINTERFACE = 3,
```
artemcm, line 87:
```suggestion
  SWIFTSCAN_OUTPUT_TYPE_SWIFTPRIVATEINTERFACE = 3,
```
Same typo in:
https://github.com/apple/swift/blob/6e47a3680871089934abff12a028b8ee2dc37d07/include/swift-c/DependencyScan/DependencyScan.h#L443C40-L443C40

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -409,14 +432,34 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
...
-    let dependencyFile =
-    try serializeModuleDependencies(for: mainModuleId,
-                                    swiftDependencyArtifacts: swiftDependencyArtifacts,
-                                    clangDependencyArtifacts: clangDependencyArtifacts)
-    commandLine.appendFlag("-explicit-swift-module-map-file")
-    commandLine.appendPath(dependencyFile)
-    inputs.append(TypedVirtualPath(file: dependencyFile.intern(),
-                                   type: .jsonSwiftArtifacts))
+    // Return if depscanner provided build commands.
+    if enableCAS, let scannerPCHArgs = mainModuleDetails.bridgingPchCommandLine {
```
artemcm, line 436: Why don't we want `scannerPCHArgs` in the general case?

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -433,6 +476,22 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
...
+                                           clangDependencyArtifacts: [ClangModuleArtifactInfo]
+  ) throws -> String {
+    // The module dependency map in CAS needs to be stable.
+    // Sort the dependencies by name.
+    let allDependencyArtifacts: [ModuleDependencyArtifactInfo] =
+      swiftDependencyArtifacts.sorted().map {ModuleDependencyArtifactInfo.swift($0)} +
+      clangDependencyArtifacts.sorted().map {ModuleDependencyArtifactInfo.clang($0)}
+    let encoder = JSONEncoder()
+    // Use sorted key to ensure the order of the keys is stable.
+    encoder.outputFormatting = [.prettyPrinted, .sortedKeys]
```
artemcm, line 490: Can we common out the gathering and sorting of the dependencies to be common between here and `serializeModuleDependencies` so that the only difference is the last step?

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -433,6 +476,22 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
...
+    let contents = try encoder.encode(allDependencyArtifacts)
+    return try swiftScanOracle.store(data: contents)
```
artemcm, line 492: Hmm, it seems strange to have such side-effects here, because otherwise the planner does not mutate the rest of driver state. Not sure if I see a better option though.

Sources/CSwiftScan/include/swiftscan_header.h
cachemeifyoucan: https://github.com/apple/swift/pull/67321

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -80,10 +80,13 @@ extension Driver {
...
     if (parsedOptions.contains(.driverExplicitModuleBuild) &&
           moduleDependencyGraphUse == .computed) {
-      if isGeneratePCH {
+      switch kind {
+      case .generatePCH:
         try addExplicitPCHBuildArguments(inputs: &inputs, commandLine: &commandLine)
-      } else {
+      case .compile, .emitModule, .interpret, .verifyModuleInterface:
         try addExplicitModuleBuildArguments(inputs: &inputs, commandLine: &commandLine)
+      default:
```
artemcm, line 88: Should we make this exhaustive so that we don't miss newly-added job kinds?

Sources/SwiftDriver/Jobs/VerifyModuleInterfaceJob.swift
```diff
@@ -26,6 +26,21 @@ extension Driver {
...
+        commandLine.appendFlag("-input-file-key")
+        commandLine.appendFlag(key)
+      }
+      if let key = privateSwiftInterfaceCacheKey, interfaceInput.type == .privateSwiftInterface {
+        commandLine.appendFlag("-input-file-key")
+        commandLine.appendFlag(key)
+      }
+      // Need to create an output file for swiftmodule output. Currently put it next to the swift interface.
+      let moduleOutPath = try interfaceInput.file.appendingToBaseName(".verified.swiftmodule")
+      commandLine.appendFlags("-o", moduleOutPath.name)
```
artemcm, line 41: As far as I can tell we currently do not have an output at all, we just run a type-check. Adding serialization to the verification job is undesirable, I think.

Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
```diff
@@ -374,6 +375,134 @@ final class ExplicitModuleBuildTests: XCTestCase {
     }
   }
 
+  /// Test generation of explicit module build jobs for dependency modules when the driver
+  /// is invoked with -explicit-module-build, -verify-emitted-module-interface and -enable-library-evolution.
+  func testExplicitModuleVerifyInterfaceJobs() throws {
```
artemcm, line 380: Thank you!

Sources/SwiftDriver/Jobs/VerifyModuleInterfaceJob.swift
cachemeifyoucan: It DOES have an output right now: https://github.com/apple/swift/blob/main/lib/Frontend/ModuleInterfaceLoader.cpp#L1772-L1787

It is just in the clang module cache and it should be the same entry that the user of the module will look up when using it from implicit module (I guess?)

I can try to match the behavior, or maybe I can teach `swift-frontend` just send the output to void using the `NullOutputBackend`.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -204,6 +204,14 @@ public struct Driver {
   /// The working directory for the driver, if there is one.
   let workingDirectory: AbsolutePath?
 
+  /// CacheKey for bridging header
+  var bridgingHeaderCacheKey: String? = nil
```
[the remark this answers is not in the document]
cachemeifyoucan: Actually, updating the patch to remove all of them. Instead of putting a mutable state in `Driver` class, I will just pass the `Job` that produces the key for the output into the function that need to use the key and compute that on the fly.

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
cachemeifyoucan: Currently, only cas returning parameters for pch. I can drop that check and we can teach non-cas depscanner to return bridging header command-line too.

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
cachemeifyoucan: This is writing to CAS (just like writing the file system). 

The concern here might be the file is added to inputs but CASID cannot be added to build dependencies. It can be an issue if someone purge the CAS between runs but this is just a general problem for all cached builds (including clang). We need to figure out the solution in build system if that needs to be tracked.

artemcm (approved):

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

PR #1322, opened 2023-03-30
[5.9][Explicit Modules] Have `-clang-target` on Darwin platforms set to the SDK version by-default

artemcm:
Cherry-pick of https://github.com/apple/swift-driver/pull/1274
-------------------------

PR #1395, opened 2023-07-05
[5.9 🍒][Explicit Modules] Handle re-mapped platform versions when specifying SDK-aligned `-clang-target` on Darwin platforms

artemcm:
Cherry-pick of https://github.com/apple/swift-driver/pull/1394
--------------------------------------------
**• Release**: 5.9
**• Explanation**: https://github.com/apple/swift-driver/pull/1322 introduced functionality to the driver where `-clang-target` for all Clang module dependencies is set to the target SDK deployment target on Darwin platforms. The original implementation missed the case of targeting platforms with an SDK-specified OS version re-map, causing incorrect versions to be used in those cases, in Explicit Module Builds.
**• Reviewed by**: @xymus 
**• Scope of Issue**: Explicitly-built module dependencies of builds targeting platforms with a version remap (e.g. MacCatalyst) use incorrect/unexpected target OS versions for Clang module dependencies. 
**• Risk**: Low. This change modifies code that does not run on default (Implicit Modules) compilation code-path, and the change itself is straight-forward and small, for affected code-paths. 
**• Testing**: Automated tests added to the driver test suite.

Resolves rdar://108288193

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -415,7 +415,9 @@ public final class DarwinToolchain: Toolchain {
       if let explicitClangTripleArg = driver.parsedOptions.getLastArgument(.clangTarget)?.asSingle {
         clangTargetTriple = explicitClangTripleArg
       } else {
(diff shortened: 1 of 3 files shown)


artemcm:
@swift-ci test

nkcsgexi (approved):

PR #1396, opened 2023-07-07
Downgrade priors version mismatch warning

keith:
As a user when you're changing between versions of these files there's nothing you could do to avoid invalidation so this warning feel superfluous

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
@@ -344,8 +344,7 @@ extension IncrementalCompilationState.IncrementalDependencyAndInputSetup {
       graphIfPresent = try ModuleDependencyGraph.read(from: dependencyGraphPath, info: self)
     }
     catch let ModuleDependencyGraph.ReadError.mismatchedSerializedGraphVersion(expected, read) {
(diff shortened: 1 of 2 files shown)


artemcm:
Could you please replace the warning with an incremental build remark, instead of removing it? 
Something like:
```
reporter?.report("Will not do cross-module incremental builds, wrong version of priors; expected \(expected) but read \(read) at '\(dependencyGraphPath)'")
```

keith:
@swift-ci please test

keith:
Updated to use that option instead! FWIW I do think that even a remark (if the user sees that) might be a bit more noise that desired (related thread: https://forums.swift.org/t/incremental-compilation-has-been-disabled-it-is-not-compatible-with-whole-module-optimization/66092). Maybe there should be a `--verbose` mode that implies these instead?

keith:

artemcm:
> Updated to use that option instead! FWIW I do think that even a remark (if the user sees that) might be a bit more noise that desired (related thread: https://forums.swift.org/t/incremental-compilation-has-been-disabled-it-is-not-compatible-with-whole-module-optimization/66092). Maybe there should be a `--verbose` mode that implies these instead?

The remarks are really only meant for all of us to debug what's going on with incremental builds, so I think a bit of verbosity of this kind is just fine. Still though, very good to move it away from being an actual warning.

keith:
> The remarks are really only meant for all of us to debug what's going on with incremental builds, so I think a bit of verbosity of this kind is just fine. Still though, very good to move it away from being an actual warning.

Yea that makes sense, but I guess users do still end up seeing these, which might be confusing especially for ones that are this low level

PR #1398, opened 2023-07-19
[5.9] Downgrade priors version mismatch warning

keith:
As a user when you're changing between versions of these files there's
nothing you could do to avoid invalidation so this warning feel
superfluous

(cherry picked from commit e98ddbdab89730d039a461ca134f0e5ec20e55dc / https://github.com/apple/swift-driver/pull/1396)


keith:
@swift-ci please test

PR #1399, opened 2023-07-19
[5.9 🍒][Explicit Module Builds] Add support for header dependencies of binary Swift module dependencies

artemcm:
Cherry-pick of 
https://github.com/apple/swift-driver/pull/1375
--------------------------------------------------------------------
**• Release**: 5.9
**• Explanation**: When we encounter a pre-built Swift binary module dependency (without an interface file), such module may have been built with a bridging header, which must still be present and is referenced by the binary `.swiftmodule` as either a `.h` or, more-likely, a pre-built `.pch` in a fully-explicit build.

Clients must be able to know about such header dependencies in order to be able to import this binary module, because this binary module may be referencing types brought in via its bridging header. The build-system client (`swift-driver`) will then ensure these header dependencies are fed as inputs to all requiring compilation tasks.

This adds support to the driver to query such header dependencies and feed them as inputs to all requiring compilation tasks.
**• Reviewed by**: @nkcsgexi @cachemeifyoucan 
**• Scope of Issue**: Explicitly-built targets with binary module dependencies which were built with a PCH are unable to resolve references to said-PCH-defined symbols.
**• Risk**: Low. This change modifies code that does not run on default (Implicit Modules) compilation code-path, and the change itself is straight-forward and small, for affected code-paths.
**• Testing**: Automated tests added to the driver test suite.

Resolves rdar://108288193

### DIFF
--- a/Sources/CSwiftScan/include/swiftscan_header.h
+++ b/Sources/CSwiftScan/include/swiftscan_header.h
@@ -132,6 +132,8 @@ typedef struct {
   (*swiftscan_swift_binary_detail_get_module_doc_path)(swiftscan_module_details_t);
   swiftscan_string_ref_t
   (*swiftscan_swift_binary_detail_get_module_source_info_path)(swiftscan_module_details_t);
+  swiftscan_string_set_t *
+  (*swiftscan_swift_binary_detail_get_header_dependencies)(swiftscan_module_details_t);
   bool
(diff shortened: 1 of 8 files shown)


artemcm:
@swift-ci test

nkcsgexi (approved):

PR #1400, opened 2023-07-20
[Macros on Darwin] Use device platform paths when building for the simulator

DougGregor:
To avoid having duplicated macro implementations in both the device and
simulator platforms, when building for a simulator platform, pass
`-external-plugin-path` that points into the *device* platform. There
is no need to have macro implementations in the simulator platforms.

Implements rdar://112563633


DougGregor:
@swift-ci please test

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -6883,7 +6889,7 @@ final class SwiftDriverTests: XCTestCase {
     let pluginAIndex = job.commandLine.firstIndex(of: .path(VirtualPath.relative(.init("PluginA"))))
     XCTAssertNotNil(pluginAIndex)
 
-    let pluginBIndex = job.commandLine.firstIndex(of: .path(VirtualPath.relative(.init("PluginB#Bexe"))))
+    let pluginBIndex = job.commandLine.firstIndex(of: .path(VirtualPath.relative(.init("Plugin~B#Bexe"))))
```
bnbarham, line 6892: IIUC we're only shell escaping the SDK paths, not the ones that were already passed in. Could we just rename `PlatformChecks` to eg. `Platform Checks`?

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -6869,8 +6869,14 @@ final class SwiftDriverTests: XCTestCase {
...
   func testPluginPaths() throws {
-    let sdkRoot = testInputsPath.appending(component: "SDKChecks").appending(component: "iPhoneOS.sdk")
-    var driver = try Driver(args: ["swiftc", "-typecheck", "foo.swift", "-sdk", VirtualPath.absolute(sdkRoot).name, "-plugin-path", "PluginA", "-external-plugin-path", "PluginB#Bexe", "-load-plugin-library", "PluginB2", "-plugin-path", "PluginC"])
+    try pluginPathTest(platform: "iPhoneOS", searchPlatform: "iPhoneOS")
+    try pluginPathTest(platform: "iPhoneSimulator", searchPlatform: "iPhoneOS")
+  }
+
+  func pluginPathTest(platform: String, searchPlatform: String) throws {
+    let sdkRoot = testInputsPath.appending(
+      components: ["PlatformChecks", "\(platform).platform", "Developer", "SDKs", "\(platform).sdk"])
```
bnbarham, line 6878: Doesn't the SDK name normally contain the version?

bnbarham (approved):

Tests/SwiftDriverTests/SwiftDriverTests.swift
DougGregor: Sure, that's a good idea.

Tests/SwiftDriverTests/SwiftDriverTests.swift
DougGregor: It does. That shouldn't matter, but just in case... I'll update it.

bnbarham: Agreed, there's no where that matters currently but... 🤷

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Issue #1404, opened 2023-07-27
swift help build doesn't work when it's documented to


tristanlabelle:
The `swift help` says:

> Use `swift help <subcommand>` for more information about a subcommand.

But it only responds to `swift <subcommand> --help`, at least for the `build` subcommand.

![image](https://github.com/apple/swift-driver/assets/2314287/ed314921-83ed-41f5-8f16-77004567e757)

PR #1405, opened 2023-07-27
swift-help: use the executable suffix on Windows

compnerd:
Windows relies on the suffix for differentiating files.  The use of `.exe` as the executable suffix is nearly universal on Windows.  Ensure that we use that when targeting Windows, repairing the use of certain subcommands, e.g. `swift help build`.

Fixes: #1404


compnerd:
@swift-ci please test

owenv (approved):

artemcm (approved):

PR #1407, opened 2023-08-03
[ELF] Don't pass ELF shared objects to autolink-extract

drodriguez:
This is a version of apple/swift#30465 for the new driver.

Since shared objects are classified as `.object`, they end up getting passed to the autolink-extract job.

For ELF platforms, this is unnecessary, since the autolink entries are not in the final shared object, so the autolink-extract process will not find any entries.

Additionally, in ELF platforms, `.so` files can actually be linker scripts (text files) with instructions for the actual linker, which the autolink-extract job will not understand and will fail.

For those two reasons, modify the autolink-extract inputs to skip over objects in ELF platforms that have the `.so` extension.

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -599,7 +599,13 @@ extension Driver {
     addJob: (Job) -> Void)
   throws
   {
-    let autolinkInputs = linkerInputs.filter { $0.type == .object }
+    let autolinkInputs = linkerInputs.filter { input in
+      // Shared objects on ELF platforms don't have a swift1_autolink_entries
+      // section in them because the section in the .o files is marked as
+      // SHF_EXCLUDE. They can also be linker scripts which swift-autolink-extract
+      // does not handle.
+      return input.type == .object && !(targetTriple.objectFormat == .elf && input.file.`extension` == "so")
+    }
     if let autolinkExtractJob = try autolinkExtractJob(inputs: autolinkInputs) {
       addJob(autolinkExtractJob)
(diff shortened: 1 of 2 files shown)


drodriguez:
@swift-ci please test

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -597,7 +597,13 @@ extension Driver {
     addJob: (Job) -> Void)
   throws
   {
-    let autolinkInputs = linkerInputs.filter { $0.type == .object }
+    let autolinkInputs = linkerInputs.filter { li in
```
compnerd, line 600: `input` might be a nicer name than `li`.

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -597,7 +597,13 @@ extension Driver {
...
+      // Shared objects on ELF platforms don't have a swift1_autolink_entries
+      // section in them because the section in the .o files is marked as
+      // SHF_EXCLUDE. They can also be linker scripts which swift-autolink-extract
+      // does not handle.
+      return li.type == .object && (targetTriple.objectFormat != .elf || li.file.`extension` != "so")
```
compnerd, line 605: I wish that we could do a better job of classification of the input (rather than relying on the file extension).  I personally find `!(target.objectFormat == .elf && input.file.extension == "so")` is easier to read.

compnerd (approved):

drodriguez:
@swift-ci please test

drodriguez:
I don't have permissions to merge into this repo. Can someone be kindly enough to do it for me? Thanks!

PR #1413, opened 2023-08-11
[Explicit Module Builds][Incremental Builds] Only re-build module dependencies which have changed or whose dependencies have changed.

artemcm:
Up until now, incremental explicit builds behaved on an all-or-nothing basis: if any of the dependencies have changed, or new dependencies were added, then the driver launches a full dependency scan, followed by re-building all dependency modules. 

Now, instead, we take a fine-grained approach to rebuilding module dependencies. The overall incremental build flow w.r.t. explicit module dependencies is now as-follows:
1. If there is a prior inter-module dependency graph and the source import set has not changed, and all dependencies are up-to-date, skip the dependency scan and re-use the prior graph.
2. When computing the first compilation wave's mandatory pre-compile jobs, examine each module dependency in the graph (either re-used prior or newly-computed). If any of the module's inputs are newer than the output, or if the output does not exist, this module is invalidated and must be re-compiled. Necessarily, every module in the graph between the root and the invalidated module must also be invalidated and re-compiled. 

(Note: that we do not need to concern ourselves with changes to the compile command of a given module changing from the prior build - if that happened, then the module output path will have changed to reflect the change in the context hash, and expected output will be missing at the new expected path.)

Resolves rdar://113638007


Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
```diff
@@ -136,6 +135,76 @@ extension IncrementalCompilationState.FirstWaveComputer {
...
+
+    // We must determine if any of the module dependencies require re-compilation
+    // Since we know that a prior dependency graph was not completely up-to-date,
+    // there must be at least *some* dependencies that require being re-built.
+    //
+    // If a dependency is deemed as requiring a re-build, then every module
+    // between it and the root (source module being built by this driver
+    // instance) must also be re-built.
+    //
+    // Because the inter-module dependency graph is a DAG, we can scan the
```
owenv, line 161: This approach is correct, but not optimal. If we have a graph like this:
image

One topological ordering is A,B,C,D,E

If E changes, this approach will rebuild everything, when only E,B,A actually require rebuild.

It's hard to say if this will matter in practice, because modules which have actually changed will tend to be near the root, always appearing before a large number of modules built from the SDK. But it might be better to run a postorder DFS which adds the path to the root to the list of invalidated modules each time it visits one whose inputs have changed

artemcm: Yes, I've been thinking about this as well, and debating the trade-off here of the simple thing (this PR) versus the optimal approach. One argument in favor of the latter is that today the ordering of nodes at the same "level" in the graph does not seem deterministic, which will make incremental builds, at times, behave differently. But perhaps that's also something that can be addressed. Appreciate the feedback.

artemcm:
@swift-ci test

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
artemcm: I ended up doing the precise invalidation algorithm and added some tests to reflect it now. Thanks.

artemcm:
@swift-ci test

owenv (approved):

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:

PR #1416, opened 2023-08-16
Forward --ld-path to linker driver

kabiroberai:
This PR introduces an `--ld-path` option, which is forwarded to the Clang linker-driver option with the same name. This change is required to support the `toolset.linker.path` option in [SE-0387](https://github.com/apple/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md). 

## See also

Upstream: https://github.com/apple/swift/pull/67956.
Downstream: https://github.com/apple/swift-package-manager/pull/6719.

### DIFF
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -181,6 +181,8 @@ extension DarwinToolchain {
       commandLine.appendFlag("-fuse-ld=\(arg.asSingle)")
     }
 
+    try commandLine.appendLast(.ldPath, from: &parsedOptions)
+
     let fSystemArgs = parsedOptions.arguments(for: .F, .Fsystem)
     for opt in fSystemArgs {
       commandLine.appendFlag(.F)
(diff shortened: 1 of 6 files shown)


MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

compnerd (commented):
I think that this should map to `-B`.  Why is `-tools-path` insufficient?

kabiroberai:
@compnerd I think mapping `-tools-path` to `-B` would be a good follow-up, as that'd handle `toolset.toolsetRootPath` correctly. But the toolset schema allows each individual tool to have a custom path, so we still need to handle the possibility of `toolset.linker.path` being explicitly set.

compnerd:
> @compnerd I think mapping `-tools-path` to `-B` would be a good follow-up, as that'd handle `toolset.toolsetRootPath` correctly. But the toolset schema allows each individual tool to have a custom path, so we still need to handle the possibility of `toolset.linker.path` being explicitly set.

Multiple `-B` options can be specified so that can still work.  That said, if we are exposing that, can we make all of these options "invisible" as I'm not sure if we want them to be user visible - the toolset schema can coordinate with the driver with internal flags.

artemcm (approved):

artemcm:
@swift-ci test

kabiroberai:
@compnerd fyi following up on your comment about `-tools-directory`, it looks like `GenericUnixToolchain` and `WindowsToolchain` indeed forward it to `clang -B` but this is missing from `DarwinToolchain` and `WebAssemblyToolchain`. Should I open a follow-up PR to patch this?

MaxDesiatov:
Patching those in a follow up PR makes sense to me.

MaxDesiatov:
@swift-ci test windows

PR #1417, opened 2023-08-16
[Explain Dependency] Fix 'testTraceDependency' to account for Swift Overlay dependencies being separate from direct import dependencies.

artemcm:


cachemeifyoucan (approved):
LGTM. How does this break?

artemcm:
> LGTM. How does this break?

Just omits some edges from the output of `-explain-module-dependency` because overlays moved to a different category.

PR #1424, opened 2023-08-21
Update scudo sanitizer runtime name

etcwilde:
The Scudo sanitizer runtime name was changed from `scudo-<arch>.{a,so}` to `scudo_standalone-<arch>.{a,so}` in https://reviews.llvm.org/D138157.

@artemcm; I'm not sure how to deal with the fact that the driver may need to span across multiple toolchain versions here as there won't be a consistent runtime name that works for older and newer toolchains.

This is part of the work for rebranch; rdar://114209408


artemcm:
> the fact that the driver may need to span across multiple toolchain versions here as there won't be a consistent runtime name

Perhaps this is something we could capture in the output of `-print-target-info` so that the compiler informs the driver.

PR #1427, opened 2023-08-25
[Caching] Fix caching tests on windows hosts

cachemeifyoucan:
For Windows host, the on-disk CAS is not yet enabled in libSwiftScan yet (due to the lack of support for Windows path in cached file system abstraction), so report caching is not supported on windows hosts.


cachemeifyoucan:
@swift-ci please test

compnerd (approved):
This is fine but insufficient.  There are a number of tests that are still failing with this change.

PR #1428, opened 2023-08-25
Jobs: support multiple lookup paths

compnerd:
Adjust the driver to include the additional library paths that the driver may want to use.  We would previously only add a single path which worked due to internal ordering assumptions that held up.  This makes the driver more resilient to changes to the target info.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1430, opened 2023-08-28
[ExplicitModule] Correctly sort modules using direct and overlay dependencies

cachemeifyoucan:
When topological sort the modules to compute transitive dependencies, it needs to use both direct and overlay dependencies. Otherwise, any dependencies that can only be discovered through overlay dependencies of the main module can end up sorted after main module and not scheduled in the build graph.


cachemeifyoucan:
@swift-ci please test

Issue #1431, opened 2023-08-30
Test suite broken on Windows due to CAS


compnerd:
The CAS changes resulted in the test suite entirely breaking on Windows.

cachemeifyoucan:
The crash is fixed in PR mentioned above. The remaining error is from clang side for explicit build with message:
```
error: file 'C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.37.32822\include\module.modulemap' has been modified since the module file 'C:\\Users\\steven\\AppData\\Local\\Temp\\TemporaryDirectory.PtValo\\ModuleCache\SwiftShims-6IUGRKYRXQPGMTRU9CJKMIH9A.pcm' was built: size changed (was 9551, now 10479)
```

PR #1433, opened 2023-09-04
SwiftDriver: speculative changes to support Windows macros

compnerd:
The test suite is currently entirely broken on Windows (awaiting @cachemeifyoucan to fix apple/swift-driver#1431).  This speculatively changes the driver to alter the search paths to match apple/swift#68322).

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -482,13 +482,21 @@ extension VirtualPath {
   // Given a virtual path pointing into a toolchain/SDK/platform, produce the
   // path to `swift-plugin-server`.
   fileprivate var pluginServerPath: VirtualPath {
+#if os(Windows)
+    self.appending(components: "bin", "swift-plugin-server.exe")
+#else
     self.appending(components: "bin", "swift-plugin-server")
+#endif
   }
 
   // Given a virtual path pointing into a toolchain/SDK/platform, produce the
   // path to the plugins.
   var pluginPath: VirtualPath {
+#if os(Windows)
+    self.appending(components: "bin")
+#else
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -482,13 +482,21 @@ extension VirtualPath {
   // Given a virtual path pointing into a toolchain/SDK/platform, produce the
   // path to `swift-plugin-server`.
   fileprivate var pluginServerPath: VirtualPath {
+#if os(Windows)
+    self.appending(components: "bin", "swift-plugin-server.exe")
```
DougGregor, line 486: This is only in the `DarwinToolchain`, so I don't think it kicks in at all for Windows

compnerd: Oh, didn't notice the fileprivate.  How does the `Toolchain` get the extension?

DougGregor: Oh, I'm SO sorry, I misread the code here! Despite being in `DarwinToolchain.swift`, this code is `internal` and used on all platforms [here](https://github.com/apple/swift-driver/blob/main/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift#L405-L415). So, your change will work.

compnerd: No worries (and this code is confusing, it took me a bit to figure it out too).

compnerd:
@DougGregor would be nice to get this in to allow cherry-picking for the 5.9 release

DougGregor (approved):
LGTM, thank you!

PR #1434, opened 2023-09-05
Improve swift-driver test cases to avoid fatal crash

cachemeifyoucan:
Step 1 to fix https://github.com/apple/swift-driver/issues/1431
* Remove unnecessary `try!`: if in the throwable context, just use `try`. There are two remaining in `concurrentPerform` that would be good to be removed in the future, the rest are in setup/generic tests, that are probably file.
* Fix a nil unwrap when using `--parallel`


cachemeifyoucan:
@swift-ci please test

compnerd (approved):

PR #1444, opened 2023-09-20
Replace deprecated uses of `<<<` with `send`

MaxDesiatov:
`<<<` operator has been deprecated in TSC, as it mostly duplicates string interpolation, and the latter should be used instead in most cases where `<<<` was used previously. Additionally, `<<<` can't be introduced to scope with a qualified import (which is the case for any operator).

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -33,7 +33,7 @@
         "repositoryURL": "https://github.com/apple/swift-tools-support-core.git",
         "state": {
           "branch": "main",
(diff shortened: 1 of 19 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
@swift-ci test windows

compnerd (approved):
I welcome our new non-operator overlords. Looking forward to the many warning messages disappearing!  This should also allow us to track our TSC dependencies better.

PR #1445, opened 2023-09-20
Add `.vscode/launch.json` to `.gitignore`

MaxDesiatov:
This file is automatically generated by VS Code, so there's little sense in tracking its addition. If one would like to add customized launch configurations, those can still be added with `git add -f`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 xcuserdata/
 .swiftpm
 .*.sw?
+.vscode/launch.json


MaxDesiatov:
@swift-ci test

compnerd (approved):
This can still be tracked if we need - having customised launch configurations can be useful.  This just makes it explicit `git add -f` would add it, `git add` would not.

PR #1446, opened 2023-09-20
[NFC] Remove unused constant in `FirstWaveComputer.swift`

MaxDesiatov:
Fixes `Initialization of immutable value 'pathModuleInfo' was never used; consider replacing with assignment to '_' or removing it` warning in `FirstWaveComputer.swift`.

The function's body of which result is unused looks to be free of side-effects, so removing this call should in turn have no effect.


MaxDesiatov:
@swift-ci test

artemcm (approved):

PR #1455, opened 2023-09-26
[embedded] A number of changes to support embedded Swift.

zoecarver:


zoecarver:
@swift-ci please smoke test

Sources/SwiftDriver/Jobs/CompileJob.swift
```diff
@@ -363,13 +363,24 @@ extension Driver {
...
+    let expirementalFeatures = parsedOptions.arguments(for: .enableExperimentalFeature)
+    let embeddedEnabled = expirementalFeatures.map(\.argument).map(\.asSingle).contains("Embedded")
+
     try commandLine.appendLast(.trackSystemDependencies, from: &parsedOptions)
     try commandLine.appendLast(.CrossModuleOptimization, from: &parsedOptions)
     try commandLine.appendLast(.ExperimentalPerformanceAnnotations, from: &parsedOptions)
-    try commandLine.appendLast(.disableAutolinkingRuntimeCompatibility, from: &parsedOptions)
-    try commandLine.appendLast(.runtimeCompatibilityVersion, from: &parsedOptions)
-    try commandLine.appendLast(.disableAutolinkingRuntimeCompatibilityDynamicReplacements, from: &parsedOptions)
-    try commandLine.appendLast(.disableAutolinkingRuntimeCompatibilityConcurrency, from: &parsedOptions)
```
zoecarver, line 372: Maybe we should drop this change as we probably still want to let users configure these options manually.

zoecarver: CC @kubamracek

kubamracek: All we need is to make sure these flags don't passed by default.

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -6288,6 +6288,36 @@ final class SwiftDriverTests: XCTestCase {
...
+      let diags = DiagnosticsEngine()
+      var driver = try Driver(args: ["swiftc", "-target", "arm64-apple-macosx10.13",  "test.swift", "-enable-experimental-feature", "Embedded", "-parse-as-library", "-wmo", "-o", "a.out", "-module-name", "main", "-enable-library-evolution"], diagnosticsEngine: diags)
+      _ = try driver.planBuild()
+      XCTAssertTrue(diags.diagnostics.first!.message.text == "Library evolution cannot be enabled with embedded Swift.")
+    } catch _ { }
+    do {
+      let diags = DiagnosticsEngine()
+      var driver = try Driver(args: ["swiftc", "-target", "arm64-apple-macosx10.13",  "test.swift", "-enable-experimental-feature", "Embedded", "-parse-as-library", "-o", "a.out", "-module-name", "main"], diagnosticsEngine: diags)
+      _ = try driver.planBuild()
+      XCTAssertTrue(diags.diagnostics.first!.message.text == "Whole module optimization (wmo) must be enabled with embedded Swift.")
```
MaxDesiatov, line 6311: Could this assert against the actual variable instead of hardcoded text? That way tests won't break if diagnostic text needs to change.
```suggestion
      XCTAssertTrue(diags.diagnostics.first!.message.text == error_need_wmo_embedded.message.text)
```

zoecarver:
@swift-ci please test

Sources/SwiftDriver/Jobs/CompileJob.swift
zoecarver: Sort of feel like we should pass `-runtime-compatibility-version none` by default. But this seems to work 🤷

kubamracek (approved):

MaxDesiatov (approved):
thanks!

zoecarver:
@MaxDesiatov I can't follow you test suggestion (even though I think it's better) because of access levels:
```
 error: 'error_no_objc_interop_embedded' is inaccessible due to 'internal' protection level
```

Seems strings are how other diags are tested.

zoecarver:
@swift-ci please test

MaxDesiatov:
> I can't follow you test suggestion (even though I think it's better) because of access levels:
Import that module with diagnostics as `@testable`, that will bring `internal` declarations into scope.

zoecarver:
Aha! Thank you, Max!

zoecarver:
@swift-ci please test

MaxDesiatov:
@swift-ci test windows

PR #1459, opened 2023-10-06
Ensure supplementary output file maps associate Swift Source inputs with expected outputs in WMO

artemcm:
When handling invocations with many input kinds, it is important that the Swift compiler is passed a supplementary output filemap that maps the first `.swift` input to the expected outputs. It is already just a convention that we use the first input for a WMO compilation task. In cases where we use the first input which happens to not be a `.swift` input (e.g. a `.o` input), the Swift compiler fails to map it to expected output kinds of Swift compilation.

Resolves rdar://115577783

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -686,7 +686,10 @@ extension Driver {
     if let input = input?.fileHandle, input != OutputFileMap.singleInputKey {
       entryInput = input
     } else {
-      entryInput = inputFiles[0].fileHandle
+      guard let firstSourceInputHandle = inputFiles.first(where:{ $0.type == .swift })?.fileHandle else {
+        fatalError("Formulating swift-frontend invocation without any input .swift files")
+      }
+      entryInput = firstSourceInputHandle
     }
     entries[entryInput, default: [:]][output.type] = output.fileHandle
   }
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci test

artemcm:
@swift-ci test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -686,7 +686,10 @@ extension Driver {
     if let input = input?.fileHandle, input != OutputFileMap.singleInputKey {
       entryInput = input
     } else {
-      entryInput = inputFiles[0].fileHandle
+      guard let firstSourceInputHandle = inputFiles.first(where:{ $0.type == .swift })?.fileHandle else {
```
etcwilde, line 689: This is more like an assert, right? The driver should never try to form a frontend invocation if there aren't swift files to compile, right? Do we have a test that tries passing `-wmo` and doesn't pass any Swift source files? I seem to recall that crashing the driver for a while, so I hope so?

`testWMOWithNonSourceInput` seems like it's trying, but isn't quite that:
```
func testWMOWithNonSourceInput() throws {
    var driver1 = try Driver(args: [
      "swiftc", "-whole-module-optimization", "danger.o", "foo.swift", "bar.swift", "wibble.swift", "-module-name", "Test",
      "-driver-filelist-threshold=0"
    ])
```

artemcm: `testNoInputs` seems to cover this well.

artemcm:
@swift-ci test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
etcwilde: I meant more like `swiftc -wmo danger.o` so there is an input. I just tried it with this change and it seems to work as expected.

etcwilde (approved):
🚀

PR #1474, opened 2023-10-26
Adjust 'makeOptions' to new TableGen format of Swift's 'Options.inc'

artemcm:


artemcm:
@swift-ci test

Sources/makeOptions/makeOptions.cpp
```diff
@@ -125,7 +141,7 @@ struct RawOption {
   }
 };
 
-#define PREFIX(NAME, VALUE) static const char *const NAME[] = VALUE;
+#define PREFIX(NAME, VALUE) static const llvm::StringLiteral NAME[] = VALUE;
```
cachemeifyoucan, line 144: Probably want to do `constexpr` for `llvm::StringLiteral`

Sources/makeOptions/makeOptions.cpp
```diff
@@ -17,6 +17,18 @@
 #include <string>
 #include <vector>
 
+#if __has_include("llvm/ADT/ArrayRef.h")
```
cachemeifyoucan, line 20: I prefer something like this if works:
```
#if __has_include("swift/Option/Options.inc")
static_assert(__has_include("llvm/ADT/ArrayRef.h", GOOD_ERROR_MESSAGE)
// same for StringRef.h
#include "llvm/ADT/ArrayRef.h"
#endif
```

artemcm: You can't do `__has_include` in a `static_assert` but I added a `static_assert` to the corresponding `#else` for LLVM headers, and kept the warning for `Options.inc`.

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
CI is busted for an unrelated reason:
```
/Users/ec2-user/jenkins/workspace/pr-swift-driver-macos/branch-main/swift-driver/.build/checkouts/swift-llbuild/lib/Basic/include/llbuild/Basic/Tracing.h:179:84: error: format specifies type 'int' but the argument has type 'EngineQueueItemKind' [-Werror,-Wformat]
```

Sources/makeOptions/makeOptions.cpp
cachemeifyoucan: 👍 Works for me. I think it is cleaner to have only one switch to decide if makeOption needs to build or not (from finding Options.inc) and it is a clean error message if the Options.inc is found but not ArrayRef and StringRef.

cachemeifyoucan (approved):

artemcm:
@swift-ci test

artemcm:
@swift-ci test

PR #1478, opened 2023-11-01
Add support for automatically emit API descriptor files

zixu-w:


artemcm (approved):

zixu-w:
@swift-ci please test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -957,14 +957,26 @@ public struct Driver {
...
 
-    self.apiDescriptorFilePath = try Self.computeSupplementaryOutputPath(
-        &parsedOptions, type: .jsonAPIDescriptor, isOutputOptions: [],
-        outputPath: .emitApiDescriptorPath,
-        compilerOutputType: compilerOutputType,
-        compilerMode: compilerMode,
-        emitModuleSeparately: emitModuleSeparately,
-        outputFileMap: self.outputFileMap,
-        moduleName: moduleOutputInfo.name)
+    let apiDescriptorDirectory: VirtualPath? = if let apiDescriptorDirectoryEnvVar = env["TAPI_SDKDB_OUTPUT_PATH"] {
```
tshortli, line 960: This should be refactored to avoid using `if` expressions to avoid a dependency on the Swift 5.9 compiler for now.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -957,14 +957,26 @@ public struct Driver {
...
+        try VirtualPath(path: apiDescriptorDirectoryEnvVar)
+    } else if let ldTraceFileEnvVar = env["LD_TRACE_FILE"] {
+        try VirtualPath(path: ldTraceFileEnvVar).parentDirectory.appending(component: "SDKDB")
+    } else { nil }
+
+    if let apiDescriptorDirectory {
```
tshortli, line 966:
```suggestion
    if let apiDescriptorDirectory = apiDescriptorDirectory {
```
Again, to avoid a dependency on Swift 5.9 language features.

tshortli (changes requested):
There are a couple uses of new Swift syntax in this PR that use new language features only available in the Swift 5.9 compiler. For now I think we expect to maintain compatibility with Swift 5.8 and so the syntax should be tweaked.

Sources/SwiftDriver/Driver/Driver.swift
zixu-w: Ah I didn't realize I was using the new syntax. Will look into refactoring

tshortli (approved):

zixu-w:
@swift-ci please test

zixu-w:
@swift-ci please test Windows

PR #1482, opened 2023-11-11
Support invoking known SwiftPM subcommands directly, instead of via symlink

owenv:
This will remove one obstacle towards distributing a single statically linked SwiftPM executable on all platforms. This change is standalone, but in order for it to be effective I need to also remove the subcommand dispatch which still exists in the c++ driver


neonichu (approved):

owenv:
@swift-ci please test

PR #1489, opened 2023-11-24
build: simplify the build rules (NFC)

compnerd:
The minimum version of CMake is set to 3.19.6.  Remove the workarounds for the older releases of CMake allowing us to simplify the build rules.


compnerd:
@swift-ci please test

etcwilde (approved):

PR #1496, opened 2023-12-01
build: enable building in Debug mode on Windows

compnerd:
Windows does not have an ABI stable C/C++ runtime, and is ABI incompatible between debug and release mode on the same version even. Ensure that we always use the release mode DLL variant of the C/C++ runtime (`/MD`) allowing us to build the client code in debug or release mode.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1498, opened 2023-12-01
[Caching] Fix incremental cache build with bridging header

cachemeifyoucan:
When planning a caching build with bridging header, the consumer of the pre-compiled bridging header needs to know the cache key for the bridging header in order to construct the correct build command, which requires visibility to the job that creates PCH.

Fix the incremental planning for bridging header users so those job can be correctly constructed even the bridging header job is skipped for the incremental build.

rdar://118143215


Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationProtectedState.swift
```diff
@@ -59,9 +62,13 @@ extension IncrementalCompilationState.ProtectedState {
     job finishedJob: Job
   ) throws -> [Job]? {
     mutationSafetyPrecondition()
+    // If a new pch job is finished, update the pch creating job.
+    if finishedJob.kind == .generatePCH {
+      jobCreatingPch = finishedJob
```
benlangmuir, line 67: Can this be mutated more than once?

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationProtectedState.swift
```diff
@@ -59,9 +62,13 @@ extension IncrementalCompilationState.ProtectedState {
     job finishedJob: Job
   ) throws -> [Job]? {
     mutationSafetyPrecondition()
+    // If a new pch job is finished, update the pch creating job.
```
artemcm, line 65: Can there be a new PCH job? We place PCH jobs into `jobsBeforeCompiles` and those are always just run once, so the job we specified to the original Incremental State should always be *the* PCH-producing job.

cachemeifyoucan: I guess if after running the depscanner and realize the pch job is never seen before, there should always be a new `ProtectedState` created?
I wasn't quite sure if that is the case, if that is the case, this can just a `let` variable.

cachemeifyoucan:
@swift-ci please test

artemcm (approved):

PR #1499, opened 2023-12-07
build: support building swift-driver outside of the toolchain

compnerd:
We can now vendor the dependencies allowing us to build a separate copy of the dependencies when building the package if it is not provided for. This enables building the package standalone for debugging in scenarios where SPM is unable to build the package (e.g. due to incorrect linking).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,19 +6,17 @@
 # See http://swift.org/LICENSE.txt for license information
 # See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 
+cmake_minimum_required(VERSION 3.19.6)
+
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:

compnerd:

artemcm (approved):
LGTM. 
Reminder that CI here does not exercise the CMake build, cross-repo testing from `swift` can help exercising the new CMake code.

compnerd:
Right, this actually is not going to be exercised in that path either.  This is in support of building this without the toolchain, which is motivated by the need to build and debug the LSP which cannot be built with SPM due to the inability to create libraries.

PR #1500, opened 2023-12-08
Port over missing debug info driver functionality from the C++ implem…

adrian-prantl:
…entation

PR #1502, opened 2023-12-09
Only specify -dwarf-version if the frontend supports it

tshortli:
Follow up to https://github.com/apple/swift-driver/pull/1500.

The `-dwarf-version` frontend option is still relatively new and in some build configurations, a newer driver may be used with an older frontend that does not understand the argument.


tshortli:
@swift-ci please test

artemcm (approved):

PR #1511, opened 2023-12-19
[CAS] Don't create the same CAS multiple times from the same Oracle

cachemeifyoucan:
If the oracle has been used to create a CAS for the path, reuse the CAS if it is asked to create the same CAS again. This not only save the work of creating the CAS, but also avoid a potential problem that the CAS closing synchronization doesn't work across threads in the same process due to its uses of file lock.

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/InterModuleDependencyOracle.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/InterModuleDependencyOracle.swift
@@ -165,11 +165,26 @@ public class InterModuleDependencyOracle {
     return diags.isEmpty ? nil : diags
   }
 
-  public func createCAS(pluginPath: AbsolutePath?, onDiskPath: AbsolutePath?, pluginOptions: [(String, String)]) throws -> SwiftScanCAS {
+  public func getOrCreateCAS(pluginPath: AbsolutePath?, onDiskPath: AbsolutePath?, pluginOptions: [(String, String)]) throws -> SwiftScanCAS {
     guard let swiftScan = swiftScanLibInstance else {
       fatalError("Attempting to reset scanner cache with no scanner instance.")
     }
-    return try swiftScan.createCAS(pluginPath: pluginPath?.pathString, onDiskPath: onDiskPath?.pathString, pluginOptions: pluginOptions)
+    // Use synchronized queue to avoid creating multiple OnDisk CAS at the same location as that will leave to synchronization issues.
+    return try queue.sync {
+      let casOpt = CASConfig(onDiskPath: onDiskPath, pluginPath: pluginPath, pluginOptions: pluginOptions)
+      if let cas = createdCASMap[casOpt] {
+        return cas
+      }
+      if let path = onDiskPath {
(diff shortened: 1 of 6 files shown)


cachemeifyoucan:
@swift-ci please test

artemcm (approved):
Makes sense to me.

Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/InterModuleDependencyOracle.swift
```diff
@@ -165,11 +165,23 @@ public class InterModuleDependencyOracle {
...
+  public func getCAS(pluginPath: AbsolutePath?, onDiskPath: AbsolutePath?, pluginOptions: [(String, String)]) throws -> SwiftScanCAS {
     guard let swiftScan = swiftScanLibInstance else {
       fatalError("Attempting to reset scanner cache with no scanner instance.")
     }
-    return try swiftScan.createCAS(pluginPath: pluginPath?.pathString, onDiskPath: onDiskPath?.pathString, pluginOptions: pluginOptions)
+    // Use synchronized queue to avoid creating multiple OnDisk CAS at the same location as that will leave to synchronization issues.
+    // Assuming plugin CAS will handle synchronization by itself.
+    return try queue.sync {
+      if let path = onDiskPath, pluginPath == nil {
+        if let cas = createdCASMap[path] {
```
akyrtzi, line 176: For completeness I'd suggest to include all options in the key (e.g. put them in a `Hashable` `struct`).

Alternatively, another approach would be to move the responsibility to the client to avoid creating multiple CAS instances for the same configuration.

cachemeifyoucan:
@swift-ci please test

Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/InterModuleDependencyOracle.swift
```diff
@@ -165,11 +165,27 @@ public class InterModuleDependencyOracle {
...
+          guard casConfig.pluginPath == pluginPath, casConfig.pluginOptions.elementsEqual(casConfig.pluginOptions, by: ==) else {
+            throw DependencyScanningError.casError("Cannot create two different CAS at the same OnDiskPath")
+          }
+          return casConfig.cas
+        }
+        let cas = try swiftScan.createCAS(pluginPath: pluginPath?.pathString, onDiskPath: path.pathString, pluginOptions: pluginOptions)
+        createdCASMap[path] = CASConfig(onDiskPath: onDiskPath, pluginPath: pluginPath, pluginOptions: pluginOptions, cas: cas)
+        return cas
+      }
+      // If onDiskPath is nil, just create a new CAS.
```
akyrtzi, line 186: Since the client will end up depending on `getOrCreateCAS` to avoid creating separate CAS instances for every Swift invocation, could you also share CAS instances even if `onDiskPath` is nil (in case passing nil `onDiskPath` becomes important in the future), using `pluginPath`+`pluginOptions` as key in a separate dictionary?

akyrtzi (approved):

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

PR #1512, opened 2023-12-21
[NFC] Add `Decodable` conformance to `Triple.Arch`

MaxDesiatov:
With retroactive conformances triggering warnings in Swift 5.11, it makes more sense to declare `Decodable` on the original declaration instead of adding it retroactively in modules where `Triple.Arch` is used.


MaxDesiatov:
@swift-ci test

artemcm (approved):

PR #1513, opened 2023-12-23
Allow enabling embedded Swift without WMO when not generating SIL

owenv:
This allows modes like -index-file to work the same way they do when not using embedded Swift


owenv:
Finally got back to getting this rebased

@swift-ci please test

owenv:
@swift-ci please test

owenv:

PR #1514, opened 2023-12-27
Add `m68k` to `Triple.swift`

MaxDesiatov:
This architecture is available in [LLVM's `Triple` implementation](https://github.com/llvm/llvm-project/blob/a01b58aef0e42fb1b52e358adf4c56678a884d37/llvm/include/llvm/TargetParser/Triple.h#L62C55-L63C51) and has [an experimental LLVM backend](https://m680x0.github.io/doc/build-from-source) available in upstream LLVM sources. It would be great to have this triple variation supported in the driver, as this architecture is relatively popular in retrocomputing enthusiast circles. It was used in Amiga, Atari ST, Sega Genesis, and of course Apple Lisa and Apple Macintosh. It is supported by Qemu and multiple Linux distributions, including Debian.

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple.swift
+++ b/Sources/SwiftDriver/Utilities/Triple.swift
@@ -438,6 +438,8 @@ extension Triple {
     case bpfeb
     /// Hexagon: hexagon
     case hexagon
+    // M68k: Motorola 680x0 family
+    case m68k
     /// MIPS: mips, mipsallegrex, mipsr6
     case mips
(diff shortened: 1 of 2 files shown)


rauhul (commented):
Nice! There's a list of triples in a test file somewhere and it would be good to add a 68k triple to that list to prevent regressions.

MaxDesiatov:
@swift-ci test

rauhul (approved):

artemcm (approved):

PR #1517, opened 2024-01-03
[Embedded] correct link paths when building for embedded to avoid stray response file paths

phausler:
When building for embedded the link phase invocations would contain `@/path/to/swift` which is a directory and not a valid link list. This corrects that logic to appropriately avoid that link file path.


phausler:
@swift-ci please test

MaxDesiatov (approved):

MaxDesiatov:
Thank you!

PR #1518, opened 2024-01-03
[Dependency Scanning] Upon scan failure, ensure scanner diagnostics are emitted

artemcm:
By catching the relevant errors as thrown from the `InterModuleDependencyOracle` and ensuring diagnostics are emitted before re-throwing them.


artemcm:
@swift-ci test

PR #1519, opened 2024-01-04
Pass down the -no-verify-emitted-module-interface flag to the frontend

xymus:
The flag `-no-verify-emitted-module-interface` tells the driver to skip verifying the swiftinterfaces emitted by the compiler. Let's pass it down to the frontend as well and so it can write it down in the skipped swiftinterface. This will help understand them from the reader side.

rdar://120445669


xymus:
@swift-ci test

tshortli (approved):

xymus:
@swift-ci test Windows

PR #1522, opened 2024-01-11
[Explicit Module Builds] Do not specify Clang dependency `module.modulemap` files as task inputs

artemcm:
These files exist outside the task execution graph and are not produced by any of the tasks involved, so we rely on trusting they are present. This is crucial because these paths may be pre-VFS-remap paths which means that the driver (not-VFS-aware) may not have access to their "external" path and thus unable to verify their actual existence.


artemcm:
@swift-ci test

owenv (approved):

PR #1523, opened 2024-01-14
Prefer `llvm-ar` over `ar` on WebAssembly

kateinoigakukun:
Static "ar"chivers not prefixed with "llvm-" rarely support WebAssembly object file formats, and the legacy driver uses "llvm-ar", so it would be better to keep the consistency.

This unblocks the use of swift-driver for WebAssembly target.


kateinoigakukun:
@swift-ci Please test

MaxDesiatov (approved):

PR #1527, opened 2024-01-22
Fix explicit modules incremental build failure when only the clang module of a mixed-source module requires rebuild

owenv:
After reading a prior intermodule dependency graph in an incremental explicit module build, the driver loops over each module in the graph to check if it is up to date. The check which is intended to skip the main (Swift) module was skipping a clang module with the same name as well, if it existed. If this clang module was the only one to be updated in an incremental compile, we wouldn't regenerate its pcm, and compilation would later fail since it wasn't up to date.

rdar://121352768


owenv:
@swift-ci please test

artemcm (approved):

PR #1532, opened 2024-01-25
When terminating because a subprocess failed, don't print `error: fatalError`

DougGregor:
We already have a way to say "a subprocess failed", use it. Fixes rdar://121627384.


DougGregor:
@swift-ci please test

artemcm (approved):
Thanks!

DougGregor:
@swift-ci please test windows

PR #1538, opened 2024-02-02
Fix wrong -Xlinker propagation to Clang with WebAssembly

kateinoigakukun:
The `-Xlinker` values should be passed to Clang with `-Xlinker` but it was not done for WebAssembly. This patch fixes it by factoring out the common logic for processing `-Xlinker` and `-Xclang-linker` options and using it in all toolchains.


kateinoigakukun:
@swift-ci Please test

artemcm (approved):
Thanks!

PR #1539, opened 2024-02-03
Port `--global-base` option for WebAssembly from legacy driver

kateinoigakukun:
Just port a missing part from legacy driver:  https://github.com/apple/swift/blob/387156ae1127983afdc4325ebedbb35d72997f34/lib/Driver/WebAssemblyToolChains.cpp#L190-L193


owenv (approved):

kateinoigakukun:
@swift-ci Please test macOS platform

PR #1540, opened 2024-02-04
Execution: special case the first two arguments in RSP mode

compnerd:
The second argument is often treated specially as a mode specifier for tools. This is particularly important for Windows where `-lib` or `/lib` passed to `link` or `ld.lld-link` will change the linker to the librarian. This argument may not be sunk into the response file as it is not treated as the mode specifier then. Special case the first two arguments to ensure that the mode switch parameter is always passed for any tools.


compnerd:
@artemcm, yes, this is dumb. I don't know of a good way to handle this without special casing. I'm open to alternative ideas.

artemcm (approved):

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

compnerd:
Going to merge this for now. If we come up with a better solution, we can revisit this.

PR #1211, opened 2022-11-02
SwiftDriver: on FreeBSD, use lld, not gold

mhjacobson:
This is equivalent to passing `-use-ld=lld` to the driver.

PR #1545, opened 2024-02-16
Update Generic Unix linker selection

etcwilde:
Not all generic-"unix" environments have the Gold linker available to them, and in some cases, the vendor of the toolchain may provide their own linker. In these cases, the driver should be internally consistent with the toolchain that it is shipped with.

Now that we have the clang-linker, we can lean on the linker selection in the clang-linker to determine a default linker. If the clang-linker, and thus, the swift compiler driver, are part of a specific toolchain, that clang-linker should be built for that platform with the appropriate linker defaults set. If someone overrides the linker with `-use-ld`, we should still honour that, but should otherwise be consistent with the appropriate toolchain linker.

This will enable building and linking on AmazonLinux 2023 without requiring folks to always pass additional flags. It also means we can produce a self-contained toolchain that is internally consistent using `lld` as the linker.

LLD removes the start/stop symbols as of https://reviews.llvm.org/D96914, which the Swift runtime used to register the runtime type metadata and other runtime metadata tables. I'm waiting to merge this until https://github.com/apple/swift/pull/72061 merges as we won't know when to pass `nostart-stop-gc` with this change.

Fixes: rdar://123061492

### DIFF
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -2368,20 +2368,11 @@ final class SwiftDriverTests: XCTestCase {
 
     do {
       // The Android NDK only uses the lld linker now
-      var driver = try Driver(args: commonArgs + ["-emit-library", "-target", "aarch64-unknown-linux-android24"], env: env)
+      var driver = try Driver(args: commonArgs + ["-emit-library", "-target", "aarch64-unknown-linux-android24", "-use-ld=lld"], env: env)
       let plannedJobs = try driver.planBuild().removingAutolinkExtractJobs()
       let lastJob = plannedJobs.last!
       XCTAssertTrue(lastJob.tool.name.contains("clang"))
-      XCTAssertTrue(lastJob.commandLine.contains(subsequence: [.flag("-fuse-ld=lld"),
-        .flag("-Xlinker"), .flag("-z"), .flag("-Xlinker"), .flag("nostart-stop-gc")]))
-    }
-
-    do {
-      var driver = try Driver(args: commonArgs + ["-emit-library", "-target", "x86_64-unknown-freebsd"], env: env)
-      let plannedJobs = try driver.planBuild().removingAutolinkExtractJobs()
-      let lastJob = plannedJobs.last!
-      XCTAssertTrue(lastJob.tool.name.contains("clang"))
-      XCTAssertTrue(lastJob.commandLine.contains(.flag("-fuse-ld=lld")))
+      XCTAssertTrue(lastJob.commandLine.contains(.flag("--fuse-ld=lld")))
     }
   }
 
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -176,8 +176,8 @@ extension DarwinToolchain {
       }
     }
 
-    if let arg = parsedOptions.getLastArgument(.useLd) {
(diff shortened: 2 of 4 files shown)


etcwilde:
https://github.com/apple/llvm-project/pull/8214
@swift-ci please test

etcwilde:
https://github.com/apple/llvm-project/pull/8214
@swift-ci please test

compnerd:

keith (approved):

artemcm (approved):

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -2337,20 +2337,17 @@ final class SwiftDriverTests: XCTestCase {
 
     do {
       // The Android NDK only uses the lld linker now
-      var driver = try Driver(args: commonArgs + ["-emit-library", "-target", "aarch64-unknown-linux-android24"], env: env)
+      var driver = try Driver(args: commonArgs + ["-emit-library", "-target", "aarch64-unknown-linux-android24", "-use-ld=lld"], env: env)
       let plannedJobs = try driver.planBuild().removingAutolinkExtractJobs()
       let lastJob = plannedJobs.last!
       XCTAssertTrue(lastJob.tool.name.contains("clang"))
-      XCTAssertTrue(lastJob.commandLine.contains(subsequence: [.flag("-fuse-ld=lld"),
-        .flag("-Xlinker"), .flag("-z"), .flag("-Xlinker"), .flag("nostart-stop-gc")]))
```
finagolfin, line 2345: Change this to `XCTAssertTrue(lastJob.commandLine.contains(.flag("-fuse-ld=lld")))` instead, so we have a test of overriding the linker with a Unix toolchain.

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -2337,20 +2337,17 @@ final class SwiftDriverTests: XCTestCase {
...
     }
 
     do {
       var driver = try Driver(args: commonArgs + ["-emit-library", "-target", "x86_64-unknown-freebsd"], env: env)
       let plannedJobs = try driver.planBuild().removingAutolinkExtractJobs()
       let lastJob = plannedJobs.last!
       XCTAssertTrue(lastJob.tool.name.contains("clang"))
-      XCTAssertTrue(lastJob.commandLine.contains(.flag("-fuse-ld=lld")))
```
finagolfin, line 2353: This test block can be removed altogether, as you're removing the FreeBSD logic it was added to test in #1211.

finagolfin (changes requested):
LGTM, but test nits.

Tests/SwiftDriverTests/SwiftDriverTests.swift
etcwilde: Fixed

Tests/SwiftDriverTests/SwiftDriverTests.swift
etcwilde: Fixed

etcwilde:
@swift-ci please test

finagolfin (approved):
Thanks, this is a good change that we probably should've been doing all along.

MaxDesiatov:
@swift-ci test windows

etcwilde:
@swift-ci please test

shahmishal (approved):

sebsto:
@MaxDesiatov @etcwilde This has been merged in `main` but was not cherry picked for `release/6.0.0` nor `release/6.0.2`
Toolchains built on Amazon Linux 2023 without this can only be used on machines where `ld.gold` is installed.

Can we include this change in one of the next release ?

PR #1547, opened 2024-02-20
Allow passing `-nostdlib` to `clang` linker driver

MaxDesiatov:
Embedded targets may need `-nostdlib` passed to `clang` linker driver in certain situations, see https://github.com/apple/swift-driver/pull/1534 for more details.

Currently, `swiftc` doesn't recognize `-nostdlib` at all, while `-Xlinker -nostdlib` passed to a `swiftc` link job from the build system will be passed verbatim to the linker and not recognized by the linker driver. We should support this flag in `swiftc` to fix this.


MaxDesiatov:
@swift-ci test

PR #1548, opened 2024-02-21
Only pass down the `-no-verify-emitted-module-interface` flag if the frontend accepts it

xymus:
The flag `-no-verify-emitted-module-interface` may still rejected by older frontends, make sure the local frontend accepts it before passing down the flag.

rdar://123345345


xymus:
@swift-ci Please test

tshortli (approved):

tshortli (approved):

artemcm (approved):

PR #1551, opened 2024-02-26
Passthrough `-tools-directory` as `-B` to clang linker for Wasm target

kateinoigakukun:
This change makes the clang linker driver for Wasm targets respect the `-tools-directory` option. This is necessary to use linker tools from Swift SDK toolset.


kateinoigakukun:
@swift-ci Please test

MaxDesiatov (approved):
Thanks!

kateinoigakukun:
@swift-ci test windows

PR #1554, opened 2024-03-05
Set `stdout` and `stderr` encodings to UTF-8 in `build-script-helper.py`

tshortli:
This will hopefully fix the following error that has been causing the https://ci.swift.org/job/oss-swift-RA-linux-ubuntu-18_04-long-test/ pipeline to fail for a few days:

```
UnicodeEncodeError: 'ascii' codec can't encode character '\u2502' in position 100311: ordinal not in range(128)
```

Resolves rdar://123809082

PR #1555, opened 2024-03-05
Avoid ascii/utf encoding errors on some systems

kastiglione:
An alternative attempt to https://github.com/apple/swift-driver/pull/1554. Some systems still use Python 3.6 where `reconfigure` is not available.

rdar://123809082


kastiglione:
@swift-ci test

tshortli (approved):
Thanks!

artemcm (approved):

PR #1556, opened 2024-03-08
Tests: repair test for Windows

compnerd:
The index was incorrectly `2` for non-Darwin targets. Windows does not have the autolink extract phase which results in a off-by-one index. Use the test helper to exclude the auto-link-extract job to support Linux and Windows similarly.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1557, opened 2024-03-09
Compute and append -file-compilation-dir when -g options are used

cachemeifyoucan:
SwiftDriver should compute `-file-compilation-dir` when debug info is used. This can avoid `swift-frontend` to compute its own compilation directory, which can cause `swift-frontend` to generate different code with the same command if current working directory changed.


cachemeifyoucan:
@swift-ci please test

adrian-prantl (approved):

PR #1558, opened 2024-03-10
Don't normalize `*-unknown-windows-coff` as `*-unknown-windows-msvc`

MaxDesiatov:
This fixes the inability to use `x86_64-unknown-windows-coff` and `i686-unknown-windows-coff` triples with Embedded Swift.

*Steps to reproduce:*

1. Install latest development snapshot off `main`, e.g. `swift-DEVELOPMENT-SNAPSHOT-2024-03-01-a-osx`.
2. Try to build with `swiftc -target i686-unknown-windows-coff -enable-experimental-feature Embedded -wmo -c test.swift`

*Expected result:*

Code compiles successfully or an actionable diagnostic message is produced.

*Actual result:*

<unknown>:0: error: could not find module 'Swift' for target 'i686-unknown-windows-msvc'

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple.swift
+++ b/Sources/SwiftDriver/Utilities/Triple.swift
@@ -126,7 +126,7 @@ public struct Triple {
         parser.components.resize(toCount: 4, paddingWith: "")
         parser.components[2] = "windows"
         if parsedEnv?.value.environment == nil {
-          if let objectFormat = parsedEnv?.value.objectFormat, objectFormat != .coff {
+          if let objectFormat = parsedEnv?.value.objectFormat {
             parser.components[3] = Substring(objectFormat.name)
           } else {
             parser.components[3] = "msvc"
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

al45tair:
What would `x86_64-unknown-windows-coff` be used for with Embedded Swift?

MaxDesiatov:
This would allow us to cross-compile to Windows from platforms other than Windows without requiring MSVC toolchain installed.

MaxDesiatov:
> but at least historically I would have been inclined to use the `-mingw` triple for cross compilation, not least because it supplies headers and import libraries that (in theory at least) are not subject to onerous license restrictions.

Parts of MinGW are licensed as GPL. I'm not interested in MinGW per se, only the approach it takes for cross-compilation to Windows.

> I'm not sure that `-msvc` as an _environment_ in a triple necessarily means that you're using Microsoft's compiler either;

This may be correct, but I don't think that justifies the current normalization approach where it defaults to `-msvc` when no environment is specified.

al45tair (approved):

al45tair:
> This may be correct, but I don't think that justifies the current normalization approach where it defaults to -msvc when no environment is specified.

Maybe. Maybe we should have a `-ucrt` environment (where use the Universal CRT), and maybe that should be the default. Maybe `-coff` as a triple should mean that you don't have a C runtime underneath you at all (that's certainly a reasonable thing for Swift to do, at least on Windows, even if we don't support it right now).

PR #1559, opened 2024-03-12
[Caching] Remap the path in `-file-compilation-dir` if needed

cachemeifyoucan:
A follow-up to #1557 that when caching is enabled and using the scanner prefix mapping option to canonicalize the path, make sure the path passed via `-file-compilation-dir` is remapped.

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -372,9 +372,11 @@ extension Driver {
       // TODO: Should we support -fcoverage-compilation-dir?
       commandLine.appendFlag(.fileCompilationDir)
       if let compilationDir = parsedOptions.getLastArgument(.fileCompilationDir)?.asSingle {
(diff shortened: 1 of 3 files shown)


cachemeifyoucan:
@swift-ci please test

adrian-prantl:
Is the scanner prefix mapping separate from the debug info prefix mapping? If not, it might be more consistent the have IRGenDebugInfo apply the mapping.

adrian-prantl:
Otherwise this LGTM

cachemeifyoucan:
> Is the scanner prefix mapping separate from the debug info prefix mapping? If not, it might be more consistent the have IRGenDebugInfo apply the mapping.

scannerPrefixMapping remaps all input source files so the rest of swift compiler will not even see the original path at all. It works on the CAS file system level to canonicalize path, so debug info remap should not be used in caching builds (same for clang).

This patch is to remap the path on the command-line, so all paths are consistent in this new remapped file system.

artemcm (approved):

PR #1573, opened 2024-03-28
Stop adding '-external-plugin-path' to SDK directories

rintaro:
Apple SDK plugins are placed under `{}.platform/Developer/usr` directories, not in the SDK directory. Adding nonexisting paths can be confusing.

rdar://121451763


rintaro:
@swift-ci Please test

DougGregor (approved):
These were never used; thanks for removing them.

adrian-prantl:
Thanks, this was very confusing in LLDB!

adrian-prantl:
@rintaro Is there a 6.0 branch for the driver this needs to be cherry-picked to?

PR #1576, opened 2024-04-04
Implement basic `xtensa` architecture triple parsing

MaxDesiatov:
This reflects corresponding architecture component handling in LLVM: https://github.com/apple/llvm-project/blob/c7d5769e8229c589f9de3d5a88e36ffbfec164fa/llvm/lib/TargetParser/Triple.cpp#L572.

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple.swift
+++ b/Sources/SwiftDriver/Utilities/Triple.swift
@@ -520,6 +520,8 @@ extension Triple {
     case renderscript32
     // 64-bit RenderScript
     case renderscript64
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

rauhul (approved):
Is this backend enabled in our llvm config?

MaxDesiatov:
Not that I'm aware of, but I don't think that should preclude triple parsing support, as it already has architectures that we don't enable in LLVM.

kubamracek (approved):

PR #1582, opened 2024-04-16
[Macros] Add support for wasm plugins

kabiroberai:
This PR implements the driver portion of support for Wasm Macros in Swift. The frontend requires the driver's help to locate `swift-wasm-plugin-server`, which we obtain in a similar fashion to the existing `swift-plugin-server`.

Counterpart to https://github.com/apple/swift/pull/73031 — will mark this as a draft until that PR lands but would appreciate input on the approach here (is explicitly passing `-wasm-plugin-server-path` the right approach or should we do something different?)

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -488,6 +488,11 @@ extension Driver {
 
       commandLine.appendFlag(.pluginPath)
       commandLine.appendPath(pluginPathRoot.localPluginPath)
+
+      if isFrontendArgSupported(.wasmPluginServerPath) {
+        commandLine.appendFlag(.wasmPluginServerPath)
+        commandLine.appendPath(pluginPathRoot.wasmPluginServerPath)
+      }
     }
   }
 
@@ -904,5 +909,17 @@ extension ParsedOptions {
       let experimentalFeatures = self.arguments(for: .enableExperimentalFeature)
       return experimentalFeatures.map(\.argument).map(\.asSingle).contains("Embedded")
     }
+  }
+}
+
+extension VirtualPath {
+  // Given a virtual path pointing into a toolchain/SDK/platform, produce the
+  // path to `swift-wasm-plugin-server`.
+  fileprivate var wasmPluginServerPath: VirtualPath {
+#if os(Windows)
+    self.appending(components: "bin", "swift-wasm-plugin-server.exe")
+#else
+    self.appending(components: "bin", "swift-wasm-plugin-server")
+#endif
   }
 }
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -488,6 +488,11 @@ extension Driver {
 
       commandLine.appendFlag(.pluginPath)
       commandLine.appendPath(pluginPathRoot.localPluginPath)
+
+      if isFrontendArgSupported(.wasmPluginServerPath) {
```
artemcm, line 492: I think instead this chunk of code should live in `WebAssemblyToolchain`'s implementation of `toolchain.addPlatformSpecificCommonFrontendOptions` so that other platforms' command-lines don't unnecessarily end up with this flag.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -906,3 +911,15 @@ extension ParsedOptions {
     }
   }
 }
+
+extension VirtualPath {
```
artemcm, line 915: Similarly, we could also move this to `WebAssemblyToolchain.swift`.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
kabiroberai: To clarify, wasm macros can be invoked by any toolchain and so this server can also be used by any toolchain — it's independent of the target triple.

That said we could still be smarter about this and only pass along the wasm-server iff the user has also passed in a wasm macro. I avoided this initially because I didn't want the Driver to have to parse plugin load options to determine if a wasm macro was being loaded, but if cluttering up the options is a concern then it might be worth it. What do you think?

kabiroberai: Updated with a `hasWasmPlugins` check, let me know if this looks better!

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -309,7 +309,10 @@ extension Driver {
     }
 
     // Emit user-provided plugin paths, in order.
-    if isFrontendArgSupported(.externalPluginPath) {
+    if isFrontendArgSupported(.loadPlugin) {
+      try commandLine.appendAll(.pluginPath, .externalPluginPath, .loadPluginLibrary, .loadPlugin, from: &parsedOptions)
+      try addLoadPluginExecutableArguments(commandLine: &commandLine)
```
DougGregor, line 314: Hmm. This is going change the relative ordering of `-load-plugin-executable` arguments with respect to the other plugin-like arguments, which we generally try to avoid. I think we'll need to iterate through all of the options, selecting these ones and translating the `-external-plugin-path` ones on-the-fly. I do appreciate how this approach makes the SwiftPM changes more limited, though, so SwiftPM can "just" treat wasm plugins like any other external plugin executable.

DougGregor (changes requested):
I'd like us to keep the plugin-related arguments in order, because we've had to craft the order very carefully and need to not perturb it.

kabiroberai:
@DougGregor order should now be fixed, plus a test to verify that 🫡

PR #1583, opened 2024-04-16
Resolve main executable symlink when looking up driver-adjacent subcommand binaries

artemcm:
Resolves https://github.com/apple/swift/issues/70932

### DIFF
--- a/Sources/swift-driver/main.swift
+++ b/Sources/swift-driver/main.swift
@@ -28,11 +28,13 @@ import Dispatch
 import WinSDK
 #endif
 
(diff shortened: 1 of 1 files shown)


artemcm:
@swift-ci test

nkcsgexi (approved):
Thank you!

stevapple:
I'm afraid this would break Windows support. IIRC we were using `findExecutable` for both cases because it can handle the executable suffix for us.

kkebo:
@artemcm Do you plan to cherry-pick this to 6.0?

PR #1586, opened 2024-04-24
[Caching] Remove the error for creating multiple CAS at same location

cachemeifyoucan:
Remove the CASError that prevents creating two different CAS at the same CAS path. This allows build system to create compatible CAS at the same location even the configuration is slightly different.

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/InterModuleDependencyOracle.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/InterModuleDependencyOracle.swift
@@ -189,12 +189,6 @@ public class InterModuleDependencyOracle {
       if let cas = createdCASMap[casOpt] {
         return cas
       }
-      if let path = onDiskPath {
-        guard !seenCASPath.contains(path) else {
-          throw DependencyScanningError.casError("Cannot create two different CAS at the same OnDiskPath")
-        }
-        seenCASPath.insert(path)
-      }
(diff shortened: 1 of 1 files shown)


cachemeifyoucan:
@swift-ci please test

benlangmuir (commented):
My concern with this change is that it will make it harder to diagnose real bugs where you have mismatched configurations.  It does seems reasonable for the plugin options to differ here, because we don't have any way to interpret them, but maybe we should keep the error if the plugin path differs?

cachemeifyoucan:
> My concern with this change is that it will make it harder to diagnose real bugs where you have mismatched configurations

While this is true, the benefit is quite limited as it only catches the case where it is created inside the same Oracle. You can still create with different oracle and slip through.

Also the main reason for this error is because the default OnDiskCAS implementation can race with in the process if you have multiple instances of same CAS. This error should be fixed (hopefully) and no longer a problem.

akyrtzi:
> > My concern with this change is that it will make it harder to diagnose real bugs where you have mismatched configurations
> 
> While this is true, the benefit is quite limited as it only catches the case where it is created inside the same Oracle. You can still create with different oracle and slip through.
> 
> Also the main reason for this error is because the default OnDiskCAS implementation can race with in the process if you have multiple instances of same CAS. This error should be fixed (hopefully) and no longer a problem.

I agree, I think this API is the wrong level for a "mismatched config" check, the caller should be allowed to create the CAS instances it wants and include its own sanity checks.

akyrtzi (approved):

PR #1587, opened 2024-04-25
[6.0][Caching] Remove the error for creating multiple CAS at same location

cachemeifyoucan:
* Explanation: Remove a check in the DependencyScanningOracle that prevents creating multiple CAS at the same on disk path location. This was to prevent a race condition from multiple CAS instance that has been since fixed. On the other hand, we want to allow build system to create compatible CAS at the same path (like one connected to remote but one doesn't).
Scope: Fixes an issue when pluginCAS is ued.
Original PR: https://github.com/apple/swift-driver/pull/1586
Risk: Low. Removing error diagnostics.
Testing: Unit tests.
Reviewer: @akyrtzi


cachemeifyoucan:
@swift-ci please test

nkcsgexi (approved):

PR #1590, opened 2024-05-01
Prefer toolchain's plugins to SDK plugins if the toolchain has stdlib

rintaro:
For example, when using a `swift.org` toolchain, which contains the stdlib in the resource directory, we should use toolchain's `SwiftMacros` plugin instead of the platform's one.


rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test

rintaro:
@swift-ci Please test Windows

ktoso (approved):
Thank you, Rintaro! Looks good.

PR #1591, opened 2024-05-01
[CAS] Fix task continuation misuse when CAS error happens

cachemeifyoucan:
Do not resume with returning after resume with throwing when there is an error returned from CAS service.


cachemeifyoucan:
@swift-ci please test

benlangmuir (approved):
Any of these changes testable?

cachemeifyoucan:
> Any of these changes testable?

No easily. Need a libSwiftScan C API implementation that returns error.

PR #1595, opened 2024-05-06
🍒[Swift 6.0]: Update Generic Unix linker selection

etcwilde:
Cherry-Picking https://github.com/apple/swift-driver/pull/1545 back to Swift 6.0

Not all generic-"unix" environments have the Gold linker available to them, and in some cases, the vendor of the toolchain may provide their own linker. In these cases, the driver should be internally consistent with the toolchain that it is shipped with.

Now that we have the clang-linker, we can lean on the linker selection in the clang-linker to determine a default linker. If the clang-linker, and thus, the swift compiler driver, are part of a specific toolchain, that clang-linker should be built for that platform with the appropriate linker defaults set. If someone overrides the linker with -use-ld, we should still honour that, but should otherwise be consistent with the appropriate toolchain linker.

This will enable building and linking on AmazonLinux 2023 without requiring folks to always pass additional flags. It also means we can produce a self-contained toolchain that is internally consistent using lld as the linker.

Fixes: rdar://123061492

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -16,29 +16,6 @@ import func TSCBasic.lookupExecutablePath
 import struct TSCBasic.AbsolutePath
 
 extension GenericUnixToolchain {
-  private func defaultLinker(for targetTriple: Triple) -> String? {
-    if targetTriple.os == .openbsd || targetTriple.os == .freeBSD ||
-         targetTriple.environment == .android ||
-         targetTriple.isFullyStaticLinux {
-      return "lld"
-    }
-
-    switch targetTriple.arch {
-    case .arm, .aarch64, .armeb, .thumb, .thumbeb:
-      // BFD linker has issues wrt relocation of the protocol conformance
-      // section on these targets, it also generates COPY relocations for
-      // final executables, as such, unless specified, we default to gold
-      // linker.
-      return "gold"
-    case .x86, .x86_64, .ppc64, .ppc64le, .systemz:
-      // BFD linker has issues wrt relocations against protected symbols.
-      return "gold"
-    default:
-      // Otherwise, use the default BFD linker.
-      return ""
-    }
-  }
-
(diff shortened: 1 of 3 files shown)


etcwilde:
*Explanation:* This patch has the Swift driver fallback on the clang-linker for default linker selection instead of hard-coding `gold` on Linux. The clang-linker can be configured to choose a different linker at build time, which means that we can have a toolchain that "just works" on Amazon Linux 2023, which does not have `gold`, but does have `lld`.
*Original PR:* https://github.com/apple/swift-driver/pull/1545
*Reviewed By:* @compnerd, @keith, @shahmishal, @artemcm, @finagolfin
*Resolves:* rdar://123061492
*Tests:*Several driver tests do not pass an explicit linker, those would fail if a linker weren't being chosen. This is also tested by any part of the SDK build that uses the Swift driver for linking (or clang as it is the clang-linker).

etcwilde:
@swift-ci please test

etcwilde:
We'll also need to back-port the corresponding change to the default linker in the clang-linker or the Linux builds will fail: https://github.com/apple/swift/pull/73463

etcwilde:
@swift-ci please test

nkcsgexi (approved):

Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
```diff
@@ -71,35 +48,11 @@ extension GenericUnixToolchain {
       commandLine.appendFlag("-shared")
       fallthrough
     case .executable:
-        // Select the linker to use.
-      var linker: String?
-      if let arg = parsedOptions.getLastArgument(.useLd) {
-        linker = arg.asSingle
+      // Select the linker to use.
+      if let arg = parsedOptions.getLastArgument(.useLd)?.asSingle {
+        commandLine.appendFlag("--fuse-ld=\(arg)")
```
drodriguez, line 53: It should be `-fuse-ld=\(arg)`. Only one dash.

drodriguez: Maybe cherry-pick https://github.com/swiftlang/swift-driver/pull/1624 as well?

PR #1596, opened 2024-05-06
Revert "Update Generic Unix linker selection"

etcwilde:
Reverts apple/swift-driver#1545

We're hitting issues where the driver is picking up BFD instead of Gold and mislinking:
```
/usr/bin/ld: /home/build-user/build/buildbot_linux/libdispatch-linux-x86_64/src/swift/CMakeFiles/swiftDispatch.dir/Block.swift.o: relocation R_X86_64_PC32 against protected symbol `$s8Dispatch0A13WorkItemFlagsVSYAAMc' can not be used when making a shared object
/usr/bin/ld: final link failed: bad value
```

Telling the built clang to use gold did not work so I'll need to figure out which clang the driver is trying to use here: https://github.com/apple/swift/pull/73463

### DIFF
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -176,8 +176,8 @@ extension DarwinToolchain {
       }
     }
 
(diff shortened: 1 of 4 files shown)


etcwilde:
@swift-ci please test

drexin (approved):

PR #1597, opened 2024-05-06
Do not warn on failure to in-process query target info or supported features

artemcm:
These warnings are not actionable by the user. Instead issue a remark, so as to avoid having these messages being raised as issues in build logs.


owenv (approved):

artemcm:
@swift-ci test

PR #1601, opened 2024-05-09
Fallback on `swift-frontend -scan-dependencies ...` fallback path in case `libSwiftScan` shared library cannot be loaded.

artemcm:
Resolves rdar://127768140

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
@@ -23,8 +23,14 @@ import class Foundation.JSONDecoder
 import var Foundation.EXIT_SUCCESS
 
 extension Diagnostic.Message {
-  static func warn_scanner_frontend_fallback() -> Diagnostic.Message {
-    .warning("Fallback to `swift-frontend` dependency scanner invocation")
+  static func warn_scan_dylib_not_found() -> Diagnostic.Message {
+    .warning("Unable to locate libSwiftScan. Fallback to `swift-frontend` dependency scanner invocation.")
+  }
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
```diff
@@ -171,12 +174,17 @@ public extension Driver {
...
-    try interModuleDependencyOracle.verifyOrCreateScannerInstance(fileSystem: fileSystem,
-                                                                  swiftScanLibPath: scanLibPath)
-    if isCachingEnabled {
-      self.cas = try interModuleDependencyOracle.getOrCreateCAS(pluginPath: try getCASPluginPath(),
-                                                                onDiskPath: try getOnDiskCASPath(),
-                                                                pluginOptions: try getCASPluginOptions())
+    do {
+      try interModuleDependencyOracle.verifyOrCreateScannerInstance(fileSystem: fileSystem,
+                                                                    swiftScanLibPath: scanLibPath)
+      if isCachingEnabled {
```
artemcm, line 180: @cachemeifyoucan should we error if caching is enabled but we failed to load libSwiftScan?

cachemeifyoucan: Error out sounds good to me.

cachemeifyoucan (commented):
This only fixes broken libSwiftScan. I do believe if you want driver to load a broken scanner, you deserve a hard error with no fallback. Warning doesn't do anything in this case and user will assume they didn't do anything horrible if they don't hit any error afterwards.

artemcm:
@swift-ci test

artemcm:
@swift-ci test

cachemeifyoucan (approved):
Thanks. LGTM

owenv (approved):

PR #1602, opened 2024-05-09
[Caching] Handle emit module job correctly for swift caching

cachemeifyoucan:
There are two situations when emiting modules are not correctly handled when swift caching is enabled:
* When only `-emit-module` task is requested from driver, in this case, we are contructing a `CompileJob` but only generating module output. SwiftDriver still expects all swift inputs will have a cache key even only the first file is generating outputs and has a cache key.
* When using `-experimental-emit-module-separately`, then swift-driver needs to construct a `EmitModuleJob`, which wasn't taught the concept of caching and not producing any cache key.

rdar://127768967

### DIFF
--- a/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
+++ b/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
@@ -119,6 +119,12 @@ extension Driver {
       commandLine.appendPath(abiPath.file)
       outputs.append(abiPath)
     }
+    let cacheContributingInputs = inputs.enumerated().reduce(into: [(TypedVirtualPath, Int)]()) { result, input in
+      // only the first swift input contributes cache key to an emit module job.
+      guard result.isEmpty, input.element.type == .swift else { return }
+      result.append((input.element, input.offset))
+    }
+    let cacheKeys = try computeOutputCacheKeyForJob(commandLine: commandLine, inputs: cacheContributingInputs)
     return Job(
       moduleName: moduleOutputInfo.name,
       kind: .emitModule,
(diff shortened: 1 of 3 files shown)


cachemeifyoucan:
@swift-ci please test

Sources/SwiftDriver/Jobs/EmitModuleJob.swift
```diff
@@ -119,14 +119,21 @@ extension Driver {
       commandLine.appendPath(abiPath.file)
       outputs.append(abiPath)
     }
+    let cacheContributingInputs = inputs.enumerated().reduce(into: [(TypedVirtualPath, Int)]()) { result, input in
+      // only the first swift input contributes to an emit module job.
```
xymus, line 123: What do you mean by contribute here? I assume it's specific to the cache?

xymus (approved):

Sources/SwiftDriver/Jobs/EmitModuleJob.swift
cachemeifyoucan: The way we create cache key for compilation is we have one cache key per input file, and each cache key contains all the output files for that specific input (like .o, .d, .diag). All the module level outputs are attached to the first swift input file.

This is just computing all the input file that has a cache key because there is an output produced that linked to the cache key for this input file.

artemcm: Could you please add this to the comment here?

artemcm (approved):

Sources/SwiftDriver/Jobs/EmitModuleJob.swift
cachemeifyoucan: Added to the same function in `CompileJob` which is more complicated and worth explanation.

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:

PR #1603, opened 2024-05-10
[6.0][Caching] Handle emit module job correctly for swift caching

cachemeifyoucan:
Explanation: Fix cache key mis-match between swift-frontend and swift-driver that can cause caching inconsistency when only emit module job is scheduled by swift-driver.
Scope: Fixes issue that build system can wrongly think cache miss when it is actually cache hit, or misidentify certain job as not cacheable and always run the compilation.
Issue: rdar://127768967
Original PR: https://github.com/apple/swift-driver/pull/1602
Testing: unit test
Reviewer: @artemcm @xymus

### DIFF
--- a/Sources/SwiftDriver/Jobs/CompileJob.swift
+++ b/Sources/SwiftDriver/Jobs/CompileJob.swift
@@ -397,9 +397,20 @@ extension Driver {
     } else {
       displayInputs = primaryInputs
     }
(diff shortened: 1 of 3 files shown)


cachemeifyoucan:
@swift-ci please test

nkcsgexi (approved):

PR #1607, opened 2024-05-13
[6.0🍒] Fallback on `swift-frontend -scan-dependencies ...` fallback path in case `libSwiftScan` shared library cannot be loaded.

artemcm:
Cherry-pick of https://github.com/apple/swift-driver/pull/1601
-----------------------------------
**Explanation**: On failure to load `libSwiftScan` when Explicitly Built Modules are enabled, attempt to run a dependency scan by invoking `swift-frontend -scan-dependencies ...` on the shell and parsing out the JSON output. 

**Risk**: Low. This change affects a code-path that is followed only in the event that the Swift toolchain is not configured with a correctly (as-expected) placed libSwiftScan or the event that there is a code-signing difference between the client process and the libSwiftScan it attempts to load. Previously, in these cases we encountered a hard failure. Now, we will try a best-effort fallback path attempt. Default code-path remains unaffected.

**Testing**: Test added to the automated test suite. 

**Reviewed By**: @cachemeifyoucan, @owenv 

Resolves rdar://127768140

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
@@ -23,8 +23,14 @@ import class Foundation.JSONDecoder
 import var Foundation.EXIT_SUCCESS
 
 extension Diagnostic.Message {
-  static func warn_scanner_frontend_fallback() -> Diagnostic.Message {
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci test

nkcsgexi (approved):
Thank you!

artemcm:

PR #1608, opened 2024-05-13
Revert "Revert "Update Generic Unix linker selection""

etcwilde:
Reverts apple/swift-driver#1596

Alright ready to re-land this. The issue was that, during the build, the just-built Swift driver would choose the clang from the system instead of the just-built clang that was configured to use `gold` or `lld`. The system clang was configured to use the gnu bfd linker instead of gold or lld. BFD has a known issue with our use of protected symbols and would fail during the Dispatch build with the following error:

> error: link command failed with exit code 1 (use -v to see invocation)
/usr/bin/ld: /home/build-user/build/buildbot_linux/libdispatch-linux-x86_64/src/swift/CMakeFiles/swiftDispatch.dir/Block.swift.o: relocation R_X86_64_PC32 against protected symbol `$s8Dispatch0A13WorkItemFlagsVSYAAMc' can not be used when making a shared object
> /usr/bin/ld: final link failed: bad value

Telling the just-built driver to use the just-built clang is more consistent with other parts of the build. Technically, for cross-compiling purposes, neither should be in use here, but at least this is more internally consistent.

> Not all generic-"unix" environments have the Gold linker available to them, and in some cases, the vendor of the toolchain may provide their own linker. In these cases, the driver should be internally consistent with the toolchain that it is shipped with.
> 
> Now that we have the clang-linker, we can lean on the linker selection in the clang-linker to determine a default linker. If the clang-linker, and thus, the swift compiler driver, are part of a specific toolchain, that clang-linker should be built for that platform with the appropriate linker defaults set. If someone overrides the linker with `-use-ld`, we should still honour that, but should otherwise be consistent with the appropriate toolchain linker.
> 
> This will enable building and linking on AmazonLinux 2023 without requiring folks to always pass additional flags. It also means we can produce a self-contained toolchain that is internally consistent using `lld` as the linker.
> 
> LLD removes the start/stop symbols as of https://reviews.llvm.org/D96914, which the Swift runtime used to register the runtime type metadata and other runtime metadata tables. I'm waiting to merge this until https://github.com/apple/swift/pull/72061 merges as we won't know when to pass `nostart-stop-gc` with this change.
> 
> Fixes: rdar://123061492

### DIFF
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -176,8 +176,8 @@ extension DarwinToolchain {
       }
     }
 
-    if let arg = parsedOptions.getLastArgument(.useLd) {
-      commandLine.appendFlag("-fuse-ld=\(arg.asSingle)")
+    if let arg = parsedOptions.getLastArgument(.useLd)?.asSingle {
+      commandLine.appendFlag("-fuse-ld=\(arg)")
     }
 
     if let arg = parsedOptions.getLastArgument(.ldPath)?.asSingle {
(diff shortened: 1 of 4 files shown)


artemcm (approved):

etcwilde:
@swift-ci please test

PR #1610, opened 2024-05-14
[Explicit Module Builds] Get source locations from the dependency scanner and emit diagnostics with them

artemcm:
- Adopt new libSwiftScan API to get source location for each emitted diagnostic: buffer identifier, line number, column number.
This change uses functionality added in: https://github.com/apple/swift/pull/73600

### DIFF
--- a/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
+++ b/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
@@ -21,6 +21,7 @@ import struct Foundation.Data
 import protocol TSCBasic.DiagnosticData
 import struct TSCBasic.AbsolutePath
 import struct TSCBasic.Diagnostic
+import protocol TSCBasic.DiagnosticLocation
 
 public enum DependencyScanningError: LocalizedError, DiagnosticData, Equatable {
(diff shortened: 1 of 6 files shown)


Sources/SwiftDriver/SwiftScan/SwiftScan.swift
```diff
@@ -71,8 +72,18 @@ public enum DependencyScanningError: LocalizedError, DiagnosticData, Equatable {
 }
 
 @_spi(Testing) public struct ScannerDiagnosticPayload {
+  @_spi(Testing) public struct SourceLocation : DiagnosticLocation {
```
owenv, line 75: Can we promote this type  members to API so clients can attempt to cast to it / access the fields without parsing the textual representation?

nkcsgexi (approved):

Sources/SwiftDriver/SwiftScan/SwiftScan.swift
artemcm: Of course. Done.

owenv (approved):

artemcm:
@swift-ci test

PR #1619, opened 2024-05-21
Suppress warning about missing SDKSettings.json for non-Darwin platforms

kateinoigakukun:
The SDKSettings.json file is only used for Darwin platforms, so we should not emit a warning about it being missing for non-Darwin platforms.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2805,7 +2805,7 @@ extension Driver {
 
       if !fileSystem.exists(path) {
         diagnosticsEngine.emit(.warning_no_such_sdk(sdkPath))
-      } else if !(targetTriple?.isWindows ?? (defaultToolchainType == WindowsToolchain.self)) {
+      } else if (targetTriple?.isDarwin ?? (defaultToolchainType == DarwinToolchain.self)) {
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
@swift-ci test

al45tair:
Interesting. The other option would be to put an `SDKSettings.json` file in your SDK. I _believe_ the minimal set of keys is
```
{
  "DisplayName": "<display name here>",
  "Version": "<SDK version>",
  "VersionMap": {},
  "CanonicalName": "<canonical name here>"
}
```
Everything else is optional. I don't have a strong opinion either way; @MaxDesiatov, WDYT? At least in principle, it would be nice to have Xcode support for the SDKs at some point, and including an `SDKSettings.json` file might make sense for that?

MaxDesiatov:
It feels weird to me to require all Swift SDKs to include Darwin-specific stuff in them. I'd rather see the file itself marked as optional and not trigger any warnings anymore.

MaxDesiatov (approved):

kateinoigakukun:
Placing `SDKSettings.json` in SDKs is definitely an option, but we can't always modify SDK contents (e.g. Android NDK is not under our control. `/` root directory can be used as SDK on non-darwin systems)
So I think making the file itself optional seems reasonable on non-Darwin platforms. For future IDE support, we can provide such metadata from Swift SDK's metadata file.

PR #1621, opened 2024-05-24
Fail compilation with an error diagnostic when specified `SWIFT_DRIVER_<tool>_EXEC` argument does not exist

artemcm:
Resolves rdar://128636621


artemcm:
@swift-ci test

artemcm:
@swift-ci test

PR #1624, opened 2024-05-29
Fix dash count to fuse-ld

etcwilde:
`-fuse-ld` only has one dash, while `--ld-path` has two. We should pass them down correctly.


etcwilde:
@swift-ci please test

shahmishal (approved):

etcwilde:

artemcm (approved):

PR #1625, opened 2024-05-30
[unix] Fix linker argument in Unix toolchains

drodriguez:
The spelling of the linker is `-fuse-ld=`, not `--fuse-ld=`, as it can be seen in [1] and as used in the C++ driver in [2].

The code was correct before #1545 using `-fuse-ld=`, but was changed to `--fuse-ld=`. The test was testing for the command line containing the same spelling, so it would had never failed, but running the actual command line would had failed. That code was eventually reverted and reverted back in #1608.

[1]: https://github.com/apple/llvm-project/blob/e33819dff1cbf7a90b0216a4993126cb11440d45/clang/include/clang/Driver/Options.td#L5267
[2]: https://github.com/apple/swift/blob/319f36b01baa082fd2d2640d6e320594752702d0/lib/Driver/UnixToolChains.cpp#L234

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -50,9 +50,9 @@ extension GenericUnixToolchain {
     case .executable:
       // Select the linker to use.
       if let arg = parsedOptions.getLastArgument(.useLd)?.asSingle {
-        commandLine.appendFlag("--fuse-ld=\(arg)")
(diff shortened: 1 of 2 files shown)


drodriguez:
@swift-ci please smoke test

drodriguez:
@etcwilde can you have a look and review this one? Thanks!

etcwilde:
Already fixed https://github.com/apple/swift-driver/pull/1624.

drodriguez:
Just late by 5 hours. Typical me :D

Thanks for fixing!

PR #1626, opened 2024-06-03
Improve performance of precompiled module invalidation in incremental builds

owenv:
Update the graph traversal so that we never need to visit a node more than once.


owenv:
@swift-ci please test

artemcm (approved):
This makes sense to me. Thank you, Owen!

PR #1628, opened 2024-06-06
[Explicit Module Builds][Incremental Builds] Re-compile module dependencies whose dependencies are up-to-date themselves but are themselves newer

artemcm:
For example consider the following module graph:
```mermaid
graph LR;
    test-->J;
    J-->G;
```

Where on an incremental build we detect that although `G` binary module product is *newer* than its textual source, said binary module product is also *newer* than a prior binary module product of `J`. Which means that although each of the modules is up-to-date with respect to its own textual source inputs, `J`'s module dependency's binary input has been updated elsewhere and `J` needs to be re-built. 

Resolves rdar://129225956

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
@@ -138,75 +138,6 @@ extension IncrementalCompilationState.FirstWaveComputer {
             mandatoryJobsInOrder: mandatoryJobsInOrder)
   }
 
-  /// We must determine if any of the module dependencies require re-compilation
-  /// Since we know that a prior dependency graph was not completely up-to-date,
-  /// there must be at least *some* dependencies that require being re-built.
-  ///
-  /// If a dependency is deemed as requiring a re-build, then every module
-  /// between it and the root (source module being built by this driver
-  /// instance) must also be re-built.
-  private func computeInvalidatedModuleDependencies(on moduleDependencyGraph: InterModuleDependencyGraph)
-  throws -> Set<ModuleDependencyId> {
-    let mainModuleInfo = moduleDependencyGraph.mainModule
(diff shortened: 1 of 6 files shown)


Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
```diff
@@ -175,20 +181,34 @@ extension IncrementalCompilationState.FirstWaveComputer {
...
                                                                                                                  fileSystem: fileSystem, reporter: reporter) {
-        reporter?.reportExplicitDependencyWillBeReBuilt(moduleId.moduleNameForDiagnostic, reason: "Out-of-date")
-        modulesRequiringRebuild.insert(moduleId)
+      reporter?.reportExplicitDependencyWillBeReBuilt(moduleId.moduleNameForDiagnostic, reason: "Out-of-date")
+      modulesRequiringRebuild.insert(moduleId)
+    } else if let outputModTime = try? fileSystem.lastModificationTime(for: VirtualPath.lookup(moduleInfo.modulePath.path)) {
+      // If a prior variant of this module dependnecy exists, and is older than any of its direct or transitive
+      // module dependency outputs, it must also be re-built
+      // FIXME: This needs caching for mod times, otherwise there's too much repeated `stat` traffic
+      for depId in reachabilityMap[moduleId] ?? [] {
```
owenv, line 202: I'm not sure I understand why we need to look at the transitive dependencies here, as opposed to just the direct dependencies.

artemcm: You are right. Changed the code. Thanks!

artemcm:
@swift-ci test

artemcm:
@swift-ci test

owenv (approved):

artemcm:

PR #1629, opened 2024-06-11
[6.0 🍒][Explicit Module Builds][Incremental Builds] Re-compile module dependencies whose dependencies are up-to-date themselves but are newer

artemcm:
Cherry-pick of https://github.com/apple/swift-driver/pull/1628
------------------------------------------------------

**Explanation**: For example consider the following module graph:
```mermaid
graph LR;
    test-->J;
    J-->G;
```
Where on an incremental build we detect that although `G` binary module product is *newer* than its textual source, said binary module product is also *newer* than a prior binary module product of `J`. Which means that although each of the modules is up-to-date with respect to its own textual source inputs, `J`'s module dependency's binary input has been updated elsewhere and `J` needs to be re-built. 

**Risk**: Low. This change adds to incremental explicitly-built module re-build filtering logic to also check timestamps of module dependency outputs of a given module dependency. This was not handled before and could lead to unsound incremental builds. The risk is relatively low because this is still off on the default code-path. 

**Testing**: Automated tests added to the compiler and driver suite.
**Reviewed By**: @owenv

Resolves rdar://129225956

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/CommonDependencyOperations.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/CommonDependencyOperations.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import func TSCBasic.topologicalSort
+import protocol TSCBasic.FileSystem
 
 @_spi(Testing) public extension InterModuleDependencyGraph {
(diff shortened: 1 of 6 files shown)


artemcm:
@swift-ci test

nkcsgexi (approved):

artemcm:
@swift-ci test

artemcm:

PR #1631, opened 2024-06-12
Revert "[Explicit Module Builds][Incremental Builds] Re-compile module dependencies whose dependencies are up-to-date themselves but are themselves newer"

artemcm:
Reverts apple/swift-driver#1628

Reverting because the test here fails nondeterministically when run in parallel. Need to make changes to the test harness to isolate it better.


artemcm:
@swift-ci test

adrian-prantl (approved):

artemcm:
Let's fix-forward instead, with https://github.com/apple/swift-driver/pull/1632.

PR #1632, opened 2024-06-12
[Tests][Explicit Module Builds] Fix per-test state in incremental explicitly-built module tests to avoid conflicts on parallel test runs

artemcm:


artemcm:
@swift-ci test

artemcm:
Let's fix-forward instead, with https://github.com/apple/swift-driver/pull/1632.

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
```diff
@@ -19,10 +19,7 @@ import TestUtilities
 
 // MARK: - Instance variables and initialization
 final class IncrementalCompilationTests: XCTestCase {
-
   var tempDir: AbsolutePath = try! AbsolutePath(validating: "/tmp")
```
adrian-prantl, line 22: We're not hardcoding `/tmp` to use as a temporary directory, are we? We should have something similar to LLVM's implementation:
https://github.com/llvm/llvm-project/blob/ca33796d54ce6d2c711032b269caf32851c5915a/llvm/lib/Support/Unix/Path.inc#L1452

artemcm: That's the default, but it does get overridden for each test in the `setUp`.

Tests/SwiftDriverTests/IncrementalCompilationTests.swift
```diff
@@ -90,15 +109,38 @@ final class IncrementalCompilationTests: XCTestCase {
 
   override func setUp() {
     self.tempDir = try! withTemporaryDirectory(removeTreeOnDeinit: false) {$0}
```
adrian-prantl, line 111: I see!

adrian-prantl (approved):

PR #1605, opened 2024-05-13
Allow setting the clang++ environment variable

etcwilde:
'+' isn't allowed in environment variable names so it's impossible to set the Clang++ executable at the moment. Replace `+` with `X` so that we can set `SWIFT_DRIVER_CLANGXX_EXEC` to set the clang++ executable.

Fixes: rdar://128007218

PR #1638, opened 2024-06-21
🍒[Swift 6.0] Allow overriding clang++ with an environment variable

etcwilde:
Cherry-picking: https://github.com/apple/swift-driver/pull/1605

> '+' isn't allowed in environment variable names so it's impossible to set the Clang++ executable at the moment. Replace + with X so that we can set `SWIFT_DRIVER_CLANGXX_EXEC` to set the clang++ executable.
> 
> Fixes: rdar://128007218

Explanation: Replaces `+` with `X` in the toolchain while doing tool lookup from an environment variable because `+` is not allowed in valid environment variable names. This prevents us from telling the driver where to look for the clang++ binary.

Risk: Low. Only affects projects that set the `SWIFT_DRIVER_CLANGXX_EXEC` environment variable. Since that doesn't do anything today, that would be a silly thing to do. Otherwise, projects cannot currently set `SWIFT_DRIVER_CLANG++_EXEC` because that's an invalid environment variable name.

Testing: Added test to verify that the substitution is made correctly.

Fixes: rdar://128007218

### DIFF
--- a/Sources/SwiftDriver/Toolchains/Toolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/Toolchain.swift
@@ -196,7 +196,10 @@ extension Toolchain {
 
   /// - Returns: String in the form of: `SWIFT_DRIVER_TOOLNAME_EXEC`
   private func envVarName(for toolName: String) -> String {
-    let lookupName = toolName.replacingOccurrences(of: "-", with: "_").uppercased()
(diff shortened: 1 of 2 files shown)


etcwilde:
@swift-ci please test

nkcsgexi (approved):

PR #1641, opened 2024-06-26
[Linux] Add build-id support.

al45tair:
Add a `-build-id` option to the driver to let users control the build ID setting used by the linker.  Pass `--build-id` to the linker by default, so that build IDs are enabled by default but using the linker's default type; if someone wishes to specify a particular option, they can do so using the new driver option.

Also turn on build IDs when *building* the driver.

rdar://116798309


al45tair:
@swift-ci Please smoke test

al45tair:
@swift-ci Please smoke test Linux platform

al45tair:
@swift-ci Please smoke test macOS platform

al45tair:
@swift-ci Please test

Sources/SwiftOptions/Options.swift
```diff
@@ -58,6 +58,8 @@ extension Option {
   public static let bridgingHeaderDirectoryForPrint: Option = Option("-bridging-header-directory-for-print", .separate, attributes: [.helpHidden, .frontend, .noDriver], metaVar: "<path>", helpText: "Directory for bridging header to be printed in compatibility header")
   public static let bridgingHeaderPchKey: Option = Option("-bridging-header-pch-key", .separate, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Cache Key for bridging header pch")
   public static let bsdk: Option = Option("-bsdk", .joinedOrSeparate, attributes: [.noDriver, .argumentIsPath], helpText: "path to the baseline SDK to import frameworks")
+  public static let buildIdEQ: Option = Option("-build-id=", .joined, alias: Option.buildId)
```
artemcm, line 61: LGTM. 
The usual reminder that these `Options.swift` are to be generated as follows:
https://github.com/swiftlang/swift-driver?tab=readme-ov-file#rebuilding-optionsswift
In case they were not, for this change.

artemcm (approved):

Sources/SwiftOptions/Options.swift
al45tair: Oh, interesting. They were not. I will investigate that — thank you.

al45tair:
@artemcm Hmmm. Looks like I'm not the only one that's edited `Options.swift` directly.

al45tair:
@swift-ci Please test

PR #1616, opened 2024-05-17
Add In-process plugin server path

rintaro:
For https://github.com/apple/swift/pull/73725

PR #1642, opened 2024-06-26
[6.0] Add In-process plugin server path

rintaro:
Cherry-pick #1616 into `release/6.0`

(copied from https://github.com/swiftlang/swift/pull/74740)
* **Explanation**: swift-syntax libraries in compiler has been complied with `-enable-library-evolution` because the toolchain provides them as a public interface for building plugins (i.e. `lib/swift/host/*.swiftmodule`). But that was not great for the performance in the compiler. This PR changes that by separating swift-syntax libs for the compiler from the one for the plugins. So the compiler and the plugin now can use different version/ABI of swift-syntax libraries from the compiler. Now swift-syntax for the compiler are compiled without library evolution.
* **Scope**: Macro plugins
* **Risk**: Mid. This changes the mechanism of in-process plugins significantly. Now in-process plugins are communicate with the compiler using the same messaging serialization with other plugins (i.e. `swift-plugin-server`, and other executable plugins), instead of directly calling the `Macro.Type` expansion method. Also, to separate the swift-syntax libraries, this introduces another set of shared swift-syntax libraries in the toolchain.
* **Testing**: Passes current test suite.
* **Issue**: rdar://130532628
* **Reviewer**: Hamish Knight (@hamishknight) Alex Hoppen (@ahoppen) Ben Barham (@bnbarham)

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -781,10 +781,19 @@ extension Driver {
     guard isFrontendArgSupported(.pluginPath) else {
       return
     }
+    let pluginPathRoot = VirtualPath.absolute(try toolchain.executableDir.parentDirectory)
+
+    if isFrontendArgSupported(.inProcessPluginServerPath) {
+      commandLine.appendFlag(.inProcessPluginServerPath)
+#if os(Windows)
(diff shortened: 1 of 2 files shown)


rintaro:
https://github.com/apple/swift/pull/74740
https://github.com/apple/swift-syntax/pull/2702
https://github.com/apple/llvm-project/pull/8929
https://github.com/apple/swift-installer-scripts/pull/309
@swift-ci Please test

rintaro:
Decided not to merge

PR #1649, opened 2024-07-06
Ignore prebuilt module cache when building the prebuilt module cache

DougGregor:
When building modules for an SDK, we need to ignore any existing prebuilt modules to avoid building a partial prebult module cache. Do so by passing an intentially-bad path for the prebuilt module cache path that's derived from the output path (but not the same as that path). This prohibits the frontend scanning job from adding the default prebuilt module cache path, while ensuring that we find no prebuilt modules.

Fixes rdar://131183526.


DougGregor:
@swift-ci please test

tshortli (approved):

PR #1650, opened 2024-07-06
Ignore prebuilt module cache when building the prebuilt module cache

DougGregor:
Explanation: When building modules for an SDK, we need to ignore any existing prebuilt modules to avoid building a partial prebuilt module cache. Make sure we do ignore them by passing down an intentionally-bogus path for the existing prebuilt module cache.
Scope: Prebuilt module cache generation logic only, very narrow.
Risk: Very low due to narrow scope.
Testing: Passes current test suite.
Main PR: https://github.com/swiftlang/swift-driver/pull/1649
Issue: rdar://131183526


DougGregor:
@swift-ci please test

tshortli (approved):

PR #1651, opened 2024-07-08
[PrebuiltModuleGen] Ensure the dependency scanner always ignores existing prebuilt modules

artemcm:
Since https://github.com/swiftlang/swift/pull/72291 landed, this tool has been in a state where it can accidentally pick up existing prebuilt deps instead of generating new ones from-scratch (`.swiftinterface`).

### DIFF
--- a/Sources/swift-build-sdk-interfaces/main.swift
+++ b/Sources/swift-build-sdk-interfaces/main.swift
@@ -169,6 +169,19 @@ do {
     args.append("-Xfrontend")
     args.append(outputDir.appending(component: "__nonexistent__").pathString)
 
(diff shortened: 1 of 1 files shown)


Sources/swift-build-sdk-interfaces/main.swift
```diff
@@ -169,6 +169,19 @@ do {
     args.append("-Xfrontend")
     args.append(outputDir.appending(component: "__nonexistent__").pathString)
```
artemcm, line 170: We most-likely don't need this anymore, but I'd like to remove it in a follow-up change after we cherry-pick this onto the release branch.

cachemeifyoucan (approved):

artemcm:
@swift-ci test

PR #1653, opened 2024-07-08
[Explicit Module Builds] Remove dead/incorrect code for handling header inputs of binary module dependencies

artemcm:
As of https://github.com/swiftlang/swift/pull/72067, we instead serialize `.h` inputs directly into binary `.swiftmodule` files, because their clients may not be able to use the dependnecies' corresponding `.pch` files due to a compilation context mismatch. The clients of such binary modules then consume the serialized `.h` files directly, and compile them, implicitly, using explicit module dependencies collected during scan of such header files.

Resolves rdar://131261765


cachemeifyoucan (approved):
LGTM

artemcm:
@swift-ci test

PR #1654, opened 2024-07-08
[6.0 🍒][Explicit Module Builds] Remove dead/incorrect code for handling header inputs of binary module dependencies

artemcm:
- **Explanation:** As of https://github.com/swiftlang/swift/pull/72067, we serialize (bridging) `.h` inputs directly into binary `.swiftmodule` files, because their clients may not be able to use the binary module's corresponding `.pch` files due to a compilation context mismatch. The clients of such binary modules then consume the serialized `.h` files directly, and compile them, implicitly, using explicit module dependencies collected during scan of such header files. This change removes leftover code that relied on a prior, incorrect, assumption that we pass on binary dependencies' PCH inputs to their clients. 
- **Scope:** This affects Explicit Module Builds where source client being compiled depends on a binary-only `.swiftmodule` dependency which was built with a bridging header.
- **Risk:** Low. This change only deletes a code-path that previously was extremely likely to lead to a compilation error downstream.
- **Issue/Radar:** rdar://131261765
- **Original PR:** https://github.com/swiftlang/swift-driver/pull/1653
- **Reviewed By:** @cachemeifyoucan


artemcm:
@swift-ci test

DougGregor (approved):

PR #1659, opened 2024-07-10
[6.0 🍒][Test Only] Make `CachingBuildTests.testDependencyScanning` be adaptable to scanner diagnostic revisions

artemcm:
- **Explanation:** Adapt the test to changing compiler-side infrastructure which may no longer cause the driver to throw an exception on the kind of failure this test is meant to test. Make it handle both new and existing behavior.
- **Scope:** Test-only change.
- **Risk:** None. Test-only change.
- **Original PR:** https://github.com/swiftlang/swift-driver/pull/1658
- **Reviewed By:** @cachemeifyoucan


artemcm:
@swift-ci test

bnbarham (approved):

PR #1666, opened 2024-07-15
[6.0] Allow enabling embedded Swift without WMO when not generating SIL

owenv:
This allows modes like -index-file to work the same way they do when not using embedded Swift

Explanation: Allow compiler modes which don't emit SIL to build embedded Swift without WMO, to facilitate indexing
Scope: Only affects embedded Swift
Risk: Low - makes an existing check more permissive
Original PR: https://github.com/swiftlang/swift-driver/pull/1513
Reviewed By: @artemcm


bnbarham (approved):

owenv:
@swift-ci please test

PR #1668, opened 2024-07-19
PrebuiltModuleGen: fail prebuilt module generation job when critical modules failed to build

nkcsgexi:
Generating prebuilt modules for critical modules is important because missing them has a bigger impact on the overall build time. This change adds a report of generated modules when the job finishes and fails the prebuilt module job if some selected critical modules were missing.

rdar://132094476

### DIFF
--- a/Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
+++ b/Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
@@ -213,7 +213,13 @@ fileprivate class ModuleCompileDelegate: JobExecutionDelegate {
       stderrStream.flush()
     }
     let keyModules = ["Swift", "SwiftUI", "Foundation"]
-    return keyModules.allSatisfy { compiledModules.keys.contains($0) }
+    return keyModules.allSatisfy {
+      if compiledModules.keys.contains($0) {
+        return true
+      }
+      stderrStream.send("Missing critical module: \($0)\n")
+      return false
+    }
   }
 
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
```diff
@@ -201,13 +202,34 @@ fileprivate class ModuleCompileDelegate: JobExecutionDelegate {
...
+    Driver.stdErrQueue.sync {
+      stderrStream.send("===================================================\n")
+      sortedModules.forEach {
+        stderrStream.send("\($0.key): \($0.value)\n")
+      }
+      stderrStream.send("===================================================\n")
+      stderrStream.flush()
+    }
+    let keyModules = ["Swift", "SwiftUI", "Foundation"]
+    return keyModules.allSatisfy { compiledModules.keys.contains($0) }
```
artemcm, line 216: We could also emit to stderr specifically which one is missing, to save the trouble to the person examining this output.

artemcm (approved):

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
nkcsgexi: Will add!

nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

PR #1674, opened 2024-07-31
[Incremental Builds][Explicit Module Builds] Treat Binary-only Swift module dependencies as up-to-date

artemcm:
Instead of conservatively assuming them to be new/updated in order to invalidate their dependents. This behavior is no longer necessary since the implementation of https://github.com/swiftlang/swift-driver/pull/1628 (adce1337b5dba9610401e932d1dcf28a0bf114e6) which will ensure that if a binary module dependnecy is newer than any of its dependents they will get invalidated and re-built.


artemcm:
@swift-ci test

cachemeifyoucan (approved):
LGTM

owenv (approved):

PR #1675, opened 2024-07-31
Don't plan compileJob if emitModule is all that is needed in a separate job

cachemeifyoucan:
When the requested output is `-emit-module` and also using `-experimental-emit-module-separately`, do not schedule compileJob for each individual source files since all outputs already satisfied by the separately emit module job. The compileJob will only produce unnecessary outputs like per-file modules and slow down the build.

rdar://84979778


cachemeifyoucan:
@swift-ci please test

artemcm (approved):

PR #1676, opened 2024-07-31
[Explicit Module Builds] Simplify `-explain-module-dependency` behavior and introduce `-explain-module-dependency-detailed`

artemcm:
The former will now simply print the first discovered path to the specified dependency module. While the latter will preserve prior behavior of finding *all possible paths* to the specified module.

### DIFF
--- a/Sources/SwiftOptions/Options.swift
+++ b/Sources/SwiftOptions/Options.swift
@@ -499,7 +499,8 @@ extension Option {
   public static let experimentalSpiImports: Option = Option("-experimental-spi-imports", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Enable experimental support for SPI imports")
   public static let experimentalSpiOnlyImports: Option = Option("-experimental-spi-only-imports", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Enable use of @_spiOnly imports")
   public static let enableExperimentalSwiftBasedClosureSpecialization: Option = Option("-experimental-swift-based-closure-specialization", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Use the experimental Swift based closure-specialization optimization pass instead of the existing C++ one")
-  public static let explainModuleDependency: Option = Option("-explain-module-dependency", .separate, attributes: [], helpText: "Emit remark/notes describing why compilation may depend on a module with a given name.")
+  public static let explainModuleDependencyDetailed: Option = Option("-explain-module-dependency-detailed", .separate, attributes: [], helpText: "Emit remarks describing every possible dependency path that explains why compilation may depend on a module with a given name.")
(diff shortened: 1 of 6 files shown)


artemcm:
@swift-ci test

Sources/SwiftOptions/Options.swift
```diff
@@ -500,6 +500,7 @@ extension Option {
   public static let experimentalSpiOnlyImports: Option = Option("-experimental-spi-only-imports", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Enable use of @_spiOnly imports")
   public static let enableExperimentalSwiftBasedClosureSpecialization: Option = Option("-experimental-swift-based-closure-specialization", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Use the experimental Swift based closure-specialization optimization pass instead of the existing C++ one")
   public static let explainModuleDependency: Option = Option("-explain-module-dependency", .separate, attributes: [], helpText: "Emit remark/notes describing why compilation may depend on a module with a given name.")
+  public static let explainModuleDependencyDetailed: Option = Option("-explain-module-dependency-detailed", .separate, attributes: [], helpText: "Emit remark/notes describing why compilation may depend on a module with a given name.")
```
tshortli, line 503: Is it worth clarifying in the help text what makes this option different from `-explain-module-dependency`?

Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
```diff
@@ -2248,30 +2248,61 @@ final class ExplicitModuleBuildTests: XCTestCase {
...
+          $0.message.text == "Module 'testTraceDependency' depends on 'A'"})
+
+        for diag in driver.diagnosticEngine.diagnostics {
+          print(diag.behavior)
+          print(diag.message)
+        }
+        XCTAssertEqual(driver.diagnosticEngine.diagnostics.filter { $0.behavior == .note}.count, 1)
+        XCTAssertTrue(driver.diagnosticEngine.diagnostics.contains { $0.behavior == .note &&
+          ($0.message.text == "[testTraceDependency] -> [A] -> [A](ObjC)" ||
+           $0.message.text == "[testTraceDependency] -> [C](ObjC) -> [B](ObjC) -> [A](ObjC)")})
```
tshortli, line 2304: Is the expected output here identical to `-explain-module-dependency-detailed`? If so, could the test case be adjusted such that different output is expected and the test is verifying that the behavior of the flags differ?

artemcm: The expected output is *one* of the two paths emitted in the above `-explain-module-dependency-detailed` test, and we check that only one path is emitted. I just wrote the test to not rely on which one will be detected first by the algorithm, even though it's fairly predictable currently.

tshortli: Ah I see, I missed the difference on a quick read.

tshortli (approved):

artemcm:
@swift-ci test

artemcm:

PR #1681, opened 2024-08-18
SwiftDriver: forward `-sysroot` provided by the user

compnerd:
This addresses a missing component of the initial Android support. When the user specifies `-sysroot`, that should always be given precedence as it is explicitly specified by the user. The `ANDROID_NDK_ROOT` environment variable is set by the Android NDK and is used as a _default_ value in the scenario that the user does not specify the `-sysroot` option. This makes Android both behave more like the Windows platform and also ensures that the user has full control over the behaviour of the toolchain.

Take the opportunity to refactor some of the code to extract the Android NDK specific helpers into an uninhabited enum and provide some behavioural test cases.

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -218,9 +218,12 @@ extension GenericUnixToolchain {
       }
 
       if targetTriple.environment == .android {
-        if let sysroot = try getAndroidNDKSysrootPath() {
-          commandLine.appendFlag("--sysroot")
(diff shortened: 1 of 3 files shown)


compnerd:
CC: @finagolfin

I _think_ that this might have been the misunderstanding that we had. The `ANDROID_NDK_ROOT` was never meant to be the way to configure the behaviour of the frontend, but rather a fallback. The initial implementation did not provide the control to the user that it should have, but did unblock some work. This change now fixes that limitation and "demotes" the `ANDROID_NDK_ROOT` environment variable to the fallback path where it always should have been.

compnerd:
@swift-ci please test

hyp (approved):

hyp:
thank you

Issue #1682, opened 2024-08-21
SwiftDriverTests.SwiftDriverTests testAndroidNDK failed on macOS Apple Silicon bot


shahmishal:
```
Test Case '-[SwiftDriverTests.SwiftDriverTests testAndroidNDK]' started.
/Users/ec2-user/jenkins/workspace/oss-swift-incremental-RA-macos-apple-silicon/swift-driver/Tests/SwiftDriverTests/SwiftDriverTests.swift:7983: error: -[SwiftDriverTests.SwiftDriverTests testAndroidNDK] : XCTAssertTrue failed
Test Case '-[SwiftDriverTests.SwiftDriverTests testAndroidNDK]' failed (0.012 seconds).
```

https://ci.swift.org/job/oss-swift-incremental-RA-macos-apple-silicon/6815/console

shahmishal:
Might be related to https://github.com/swiftlang/swift-driver/pull/1681

PR #1684, opened 2024-08-21
Tests: adjust tests for Darwin

compnerd:
The Darwin toolchain is incomplete, missing `swift-autolink-extract`. Add a workaround for the missing tool to allow the tests to pass. This then uncovers another issue with ASi builds where the prebuilt NDK is only available on x86_64 hosts. While we could loosen the NDK lookup, we do still occassionally use some of the tools, so for now lets limit the test to x86_64.

Fixes: #1682


compnerd:
@swift-ci please test

hyp (approved):

PR #1685, opened 2024-08-26
Android: add the right suffix for the compiler-rt profiler library

finagolfin:
A bunch of profiler tests in the compiler validation suite were recently enabled on non-Darwin platforms by swiftlang/swift#75572 and then failed natively on Android because they couldn't find this profiler library, which this pull fixed. This is [the same change I used to find the sanitizer libraries for Android a couple years ago](https://github.com/swiftlang/swift-driver/pull/720/files#diff-b2b6d2b2cf451a616f869ff28282d9fb3bbf966edeb03abfa7c47f12b6851377R109).

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -285,9 +285,10 @@ extension GenericUnixToolchain {
       }
 
       if parsedOptions.hasArgument(.profileGenerate) {
+        let environment = (targetTriple.environment == .android) ? "-android" : ""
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci test

finagolfin:
Do I still need to submit the equivalent pull for the legacy C++ driver too, or do we not have to bother as much anymore, especially since that Driver is going away and this is a niche usecase?

artemcm:
> Do I still need to submit the equivalent pull for the legacy C++ driver too, or do we not have to bother as much anymore, especially since that Driver is going away and this is a niche usecase?

I'd say that's not necessary unless you can come up with some kind if niche requirements for where the legacy driver is needed. My hope is that it isn't.

finagolfin:
Ready for merge.

PR #1698, opened 2024-09-18
[6.0.2] Only pass a versioned prebuilt-modules for Mac Catalyst if it exists.

cwakamo:
Some clients still have unversioned prebuilt modules, but SwiftDriver assumed that the versioned path would always exist for macOS (at least for the purposes of passing it when compiling for Mac Catalyst).

This change duplicates some logic from the Swift frontend to check for the existence of the versioned macOS prebuilt modules directory before passing it, optionally stripping off trailing `.0` version components.

This change also updates the driver so that it passes a prebuilt modules directory relative to the resource directory rather than relative to the compiler binary, also matching the Swift frontend's behavior.

This addresses <rdar://problem/136047054>.

(cherry picked from commit b1573ef55119629057a57591916ff4d6d7d5af9e)

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -419,12 +419,27 @@ public final class DarwinToolchain: Toolchain {
     // doesn't always match the macosx sdk version so the compiler may fail to find
     // the prebuilt module in the versioned sub-dir.
     if frontendTargetInfo.target.triple.isMacCatalyst {
+      let resourceDirPath = VirtualPath.lookup(frontendTargetInfo.runtimeResourcePath.path)
+      let basePrebuiltModulesPath = resourceDirPath.appending(components: "macosx", "prebuilt-modules")
+
+      // Ensure we pass a path that exists. This matches logic used in the Swift frontend.
+      let prebuiltModulesPath: VirtualPath = try {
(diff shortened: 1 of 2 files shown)


cwakamo:
@swift-ci please test

cwakamo:
* Explanation: This change ensures that swift-driver passes a prebuilt-modules directory for Mac Catalyst that exists instead of blindly passing a versioned prebuilt-modules directory. (Some clients still do not have a versioned prebuilt-modules directory due to the fact that the C++ driver doesn't support it correctly for Mac Catalyst.)
* Scope: Affects compilation for Mac Catalyst only.
* Issue: rdar://problem/136047054
* Original PR: https://github.com/swiftlang/swift-driver/pull/1695
* Risk: The risk is that Catalyst compilation will no longer use the prebuilt modules because it's pointed at the wrong path, but this is mitigated by the presence of tests which ensure that the driver passes the correct path for Catalyst in both the versioned and unversioned cases.
* Testing: Unit tests have been added; this has also been tested manually with a client that previously did not work due to the unconditional use of the versioned prebuilt-modules directory.
* Reviewer: @artemcm reviewed the original change.

artemcm (approved):

cwakamo:
Closing this PR without merging so as not to add undue risk to the 6.0.2 release. Will continue to pursue this for `release/6.0` for future 6.0.x releases however.

PR #1699, opened 2024-09-19
Pass down `-public-module-name` to the frontend

xymus:


xymus:
@swift-ci Please test

Sources/SwiftOptions/Options.swift
```diff
@@ -705,6 +705,7 @@ extension Option {
   public static let protocolRequirementAllowList: Option = Option("-protocol-requirement-allow-list", .separate, attributes: [.noDriver, .argumentIsPath], metaVar: "<path>", helpText: "File containing a new-line separated list of protocol names")
   public static let protocolRequirementAllowList_: Option = Option("--protocol-requirement-allow-list", .separate, alias: Option.protocolRequirementAllowList, attributes: [.noDriver, .argumentIsPath], metaVar: "<path>", helpText: "File containing a new-line separated list of protocol names")
   public static let publicAutolinkLibrary: Option = Option("-public-autolink-library", .separate, attributes: [.helpHidden, .frontend, .noDriver, .moduleInterface], helpText: "Add public dependent library")
+  public static let publicModuleName: Option = Option("-public-module-name", .separate, attributes: [.frontend], helpText: "Public facing module name to use in diagnostics and documentation")
```
artemcm, line 708: A reminder that `Options.swift` is to be generated according to the following procedure:
https://github.com/swiftlang/swift-driver?tab=readme-ov-file#rebuilding-optionsswift
In case it wasn't already.

artemcm (approved):

Sources/SwiftOptions/Options.swift
xymus: It was but I committed only the relevant lines.

PR #1703, opened 2024-09-23
Re-land: Allow for non-external lookup of libSwiftScan symbols and centralize the scanning instance use in the driver.

artemcm:
>- The driver now holds a reference to the 'swiftScanLibInstance' which is shared with the 'interModuleDependencyOracle', but is also used for target info and supported compiler feature queries. This means a single SwiftScan instance is shared across all uses in a given driver instance.
>- Allow SwiftScan to be instantiated *without* a path to an external 'libSwiftScan.dylib', which will cause it to 'dlopen' with a NULL argument, and expect the scanner symbols to be found in the image that the driver code is a part of. This is useful for when compiler-based tooling (SourceKit) links the driver directly and uses its C API.

Added a deprecation for the prior version of `verifyOrCreateScannerInstance`.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -158,6 +158,10 @@ public struct Driver {
   /// Whether we are using the driver as the integrated driver via libSwiftDriver
   public let integratedDriver: Bool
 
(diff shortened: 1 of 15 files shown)


artemcm:
@swift-ci test

cachemeifyoucan (approved):

PR #1708, opened 2024-10-08
Only make input paths absolute when -working-directory is passed

rmaz:
This is a partial revert of https://github.com/swiftlang/swift-driver/commit/a9f35bc15f68d4ae0cd3949afa6c78e1a3df8f80

When remotely building and sharing Swift artifacts it is important to maintain
relative paths to avoid serializing any paths that will not be valid on different
machines. Change the logic back to the previous model, and the one used
in the old swift driver, where paths are made absolute only when a 
`-working-directory` flag is passed.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -329,7 +329,8 @@ public struct Driver {
     case .absolute(let path):
       return path
     case .relative(let path):
-      return workingDirectory.map { AbsolutePath($0, path) }
+      let cwd = workingDirectory ?? fileSystem.currentWorkingDirectory
+      return cwd.map { AbsolutePath($0, path) }
     case .standardInput, .standardOutput, .temporary, .temporaryWithKnownContents, .fileList:
       fatalError("Frontend target information will never include a path of this type.")
     }
@@ -688,14 +689,14 @@ public struct Driver {
                                                                                         compilerMode: compilerMode)
 
     // Compute the working directory.
-    let cwd = fileSystem.currentWorkingDirectory
-    workingDirectory = try parsedOptions.getLastArgument(.workingDirectory).map { workingDirectoryArg in
+    self.workingDirectory = try parsedOptions.getLastArgument(.workingDirectory).map { workingDirectoryArg in
+      let cwd = fileSystem.currentWorkingDirectory
       return try cwd.map{ try AbsolutePath(validating: workingDirectoryArg.asSingle, relativeTo: $0) } ?? AbsolutePath(validating: workingDirectoryArg.asSingle)
-    } ?? cwd
-
-    // Apply the working directory to the parsed options.
-    if let workingDirectory = self.workingDirectory {
-      try Self.applyWorkingDirectory(workingDirectory, to: &self.parsedOptions)
+    }
+    
+    if let specifiedWorkingDir = self.workingDirectory {
+      // Apply the working directory to the parsed options if passed explicitly.
+      try Self.applyWorkingDirectory(specifiedWorkingDir, to: &self.parsedOptions)
     }
 
     let staticExecutable = parsedOptions.hasFlag(positive: .staticExecutable,
@@ -776,6 +777,8 @@ public struct Driver {
       }
 
       if let workingDirectory = self.workingDirectory {
+        // Input paths are prefixed with the working directory when specified,
+        // apply the same logic to the output file map keys.
         self.outputFileMap = outputFileMap?.resolveRelativePaths(relativeTo: workingDirectory)
       } else {
         self.outputFileMap = outputFileMap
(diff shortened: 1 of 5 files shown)


rmaz:
@swift-ci please test

cachemeifyoucan (commented):
The direction of the change works for me. Forgot distributed build is using the relative path to do the compilation.

Note there is an alternative to use `-scanner-prefix-map` to alter the path returned from the scanner but I guess that option requires knowing the path on the remote node before hand.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -688,14 +689,14 @@ public struct Driver {
...
+      let cwd = fileSystem.currentWorkingDirectory
       return try cwd.map{ try AbsolutePath(validating: workingDirectoryArg.asSingle, relativeTo: $0) } ?? AbsolutePath(validating: workingDirectoryArg.asSingle)
-    } ?? cwd
-
-    // Apply the working directory to the parsed options.
-    if let workingDirectory = self.workingDirectory {
-      try Self.applyWorkingDirectory(workingDirectory, to: &self.parsedOptions)
+    }
+    
+    if let specifiedWorkingDir = self.workingDirectory {
```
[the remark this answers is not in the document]
rmaz: I tried that initially, but then the output paths (and profile data paths) end up absolute. I didn't track down where that was happening, but there seems to be multiple places in the code that would end up prefixing paths if `workingDirectory` was set.

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -45,12 +45,15 @@ private var testInputsPath: AbsolutePath {
   }
 }
 
-func toPath(_ path: String, base: AbsolutePath = localFileSystem.currentWorkingDirectory!) throws -> VirtualPath {
-  return try VirtualPath(path: path).resolvedRelativePath(base: base)
+func toPath(_ path: String, isRelative: Bool = true) throws -> VirtualPath {
```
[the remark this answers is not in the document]
rmaz: > I am slightly prefer leaving the tests in the new state but add -working-directory to the driver command.

For the changed tests? For ` testDOTFileEmission` that would work, for the others it would require adding `isRelative: false` to all the `toPath()` calls, which would be a bigger change. 

> Also, would be nice to add a test to check the behavior of path arguments with/without -working-directory

I added `testRelativeInputs` on L7239 for this

Sources/SwiftDriver/Driver/Driver.swift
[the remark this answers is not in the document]
cachemeifyoucan: Maybe driver will think that working-directory is passed via command-line so it is passing working directory (in abs path) frontend flag plus some other flags derived from working directory?

Sounds fine with me.

cachemeifyoucan (approved):
Thanks for fixing!

rmaz:
@swift-ci please test

rmaz:
@swift-ci Please test Windows platform

Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -3039,10 +3040,10 @@ final class SwiftDriverTests: XCTestCase {
     XCTAssertEqual(emitModuleJob.outputs[0].file, try toPath("Test.swiftmodule"))
     XCTAssertEqual(emitModuleJob.outputs[1].file, try toPath("Test.swiftdoc"))
     XCTAssertEqual(emitModuleJob.outputs[2].file, try toPath("Test.swiftsourceinfo"))
-    XCTAssertEqual(emitModuleJob.outputs[3].file, try toPath("Test.swiftinterface"))
+    XCTAssertEqual(emitModuleJob.outputs[3].file, try VirtualPath(path: "./Test.swiftinterface"))
     XCTAssertEqual(emitModuleJob.outputs[4].file, try toPath("Test.private.swiftinterface"))
     XCTAssertEqual(emitModuleJob.outputs[5].file, try toPath("Test-Swift.h"))
-    XCTAssertEqual(emitModuleJob.outputs[6].file, try toPath("Test.tbd"))
+    XCTAssertEqual(emitModuleJob.outputs[6].file, try VirtualPath(path: "./Test.tbd"))
```
artemcm, line 3046: @rmaz do you recall why this change was needed for only these specific outputs?

rmaz: I don't, but changing it to `toPath()` gives us:
```
XCTAssertEqual failed: ("./Test.swiftinterface") is not equal to ("Test.swiftinterface")
```
So somewhere a CWD of `.` is being prefixed only on the `swiftinterface` path and the `tbd` path.

rmaz: After some code wandering, it is because the path is unspecified in the compiler args, so falls back to this codepath to determine it:
https://github.com/swiftlang/swift-driver/blob/eba19529226c95a602bfd1cecaef3ca920bb199f/Sources/SwiftDriver/Driver/Driver.swift#L3623-L3626

This ends up calling `dirname` on a relative path with no folder components, which ends up here:
https://github.com/swiftlang/swift-tools-support-core/blob/b464fcd8d884e599e3202d9bd1eee29a9e504069/Sources/TSCBasic/Path.swift#L598-L605

This ends up calculating the parent as `.` and generates the filename as `modulename.swiftinterface`.

artemcm: Thank you!

PR #1710, opened 2024-10-09
Fix LTO support for WebAssembly targets

kateinoigakukun:
It seems .bc input file handling and `-flto` clang linker driver option were forgotten to be ported.


kateinoigakukun:
@swift-ci test

artemcm (approved):

MaxDesiatov (approved):

PR #1711, opened 2024-10-10
Maintain relative SDK paths with no -working-directory flag

rmaz:
When providing relative SDK paths to the swift-driver, the path should be
passed directly to the frontend as a relative path. This is necessary to
support remotely building Swift objects that are shared across machines
without serializing invalid absolute paths.


rmaz:
@swift-ci please test

rmaz:
@swift-ci Please test Windows platform

artemcm (approved):
Makes sense to me.

PR #1714, opened 2024-10-17
[Tests] Remove test of static linking against musl.

al45tair:
We don't support plain musl for now, so don't bother trying to do a static linking test for it.

rdar://138021505


al45tair:
@swift-ci Please test

mikeash (approved):

PR #1716, opened 2024-10-17
Output a Windows-style response file for swift-frontend and clang

ahoppen:
https://github.com/swiftlang/swift-driver/pull/1000 switched to use POSIX-style response files but swift-frontend expects Windows-style response files on Windows.

Switch back to generating Windows-style response files on Windows and pass `--rsp-quoting=windows` to `clang` to tell it to pass the response file in Windows mode.

Alternatives considered:
- Have a separate response file generation logic for clang and Swift: This seems like a continuous source of bugs if you always need to think about which response file format you want, especially if the behavior is not intuitive (`clang.exe` expecting POSIX-style response files)
- Instead of parsing `--rsp-quoting=windows` on Windows, use `clang-cl.exe`, which also causes response files to be parsed in Windows-style: I think it’s better to be explicit here instead of relying on clang’s implicit behavior based on which executable name was invoked.
- Change swift-frontend to accept POSIX-style response files: swift-frontend and the old Swift driver share the same response file parsing logic. Changing swift-frontend without changing the old driver would require a bunch of flag parsing, which isn’t desirable. Also, using Windows-style response files on Windows seems like the better-fitting solution (CMake, for example, outputs Windows-style response files).

Reverts https://github.com/swiftlang/swift-driver/pull/1000 with some logic added on top.

### DIFF
--- a/Sources/SwiftDriver/Execution/ArgsResolver.swift
+++ b/Sources/SwiftDriver/Execution/ArgsResolver.swift
@@ -196,9 +196,6 @@ public final class ArgsResolver {
   }
 
   private func createResponseFileIfNeeded(for job: Job, resolvedArguments: inout [String], useResponseFiles: ResponseFileHandling) throws -> Bool {
-    func quote(_ string: String) -> String {
-      return "\"\(String(string.flatMap { ["\\", "\""].contains($0) ? "\\\($0)" : "\($0)" }))\""
-    }
     guard useResponseFiles != .disabled else {
       return false
     }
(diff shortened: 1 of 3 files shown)


ahoppen:
@swift-ci Please test

artemcm (approved):

ahoppen:
@swift-ci Please test macOS

ahoppen:
While I investigate a little more, do you use lld as a linker (ie. pass `--use-lld`) or any other linker? I assume the issue is that we should only pass `-rsp-quoting=windows` to `clang`.

ahoppen:
https://github.com/swiftlang/swift-driver/pull/1717 should fix the issue. I assume you were building a static library without `--use-lld`, which used `link.exe`, which doesn’t support `--rsp-quoting=windows`.

hjyamauchi:
@ahoppen  Yeah probably we were using link.exe. I will try locally test the fix and update here. Thank you for looking into this!

hjyamauchi:
@ahoppen I was able to confirm that https://github.com/swiftlang/swift-driver/pull/1717 fixes the issue. Thank you!

ahoppen:
Thanks for verifying, @hjyamauchi!

PR #1717, opened 2024-10-22
Only pass `--rsp-quoting=windows` to `clang`

ahoppen:
We were unconditionally passing `--rsp-quoting` to the linker on Windows. As far as I can tell, there are three different linkers we invoke on Windows:
- `clang.exe`: This is what this fix is intended to cover
- `lld-link.exe`: Defaults to Windows quoting style in response files and `--rsp-quoting=windows` is a no-op.
- `link.exe`: Does not accept `--rsp-quoting=windows`

Move `--rsp-quoting=windows` to the point where we decide to use `clang.exe` as the linker on Windows and only pass it to `clang.exe`.


ahoppen:
@swift-ci Please test

hjyamauchi (approved):

PR #1719, opened 2024-10-25
[SwiftOptions] Add missing options based on Options.inc from swift@main

xedin:
Sync up Options.swift with Options.td from swift's main branch by using `makeOptions` script.


xedin:
@swift-ci please test

owenv (approved):

PR #1726, opened 2024-11-06
[Darwin] Add external plugin paths into platform specified by PLATFORM_DIR

DougGregor:
The driver currently has logic to infer the location of the platform based on the conventional layout of SDKs within platforms. However, that doesn't work well when SDKs have been built up in other places.

Introduce support for a second convention, that the PLATFORM_DIR environment variable can specify the path to the platform. When present, also add external plugin paths pointing into that platform.

Fixes rdar://138370421.

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -456,29 +456,52 @@ public final class DarwinToolchain: Toolchain {
     }
 
     if driver.isFrontendArgSupported(.externalPluginPath) {
-      // Determine the platform path. For simulator platforms, look into the
-      // corresponding device platform instance.
-      let origPlatformPath = VirtualPath.lookup(sdkPath)
-        .parentDirectory
-        .parentDirectory
-        .parentDirectory
-      let platformPath: VirtualPath
-      if let simulatorRange = origPlatformPath.basename.range(of: "Simulator.platform") {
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

artemcm (approved):

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -456,29 +456,52 @@ public final class DarwinToolchain: Toolchain {
...
-      commandLine.appendFlag("\(platformPathRoot.pluginPath.name)#\(platformPathRoot.pluginServerPath.name)")
+      // If the PLATFORM_DIR environment variable is set, also add plugin
+      // paths into it.
+      if let platformDir = env["PLATFORM_DIR"],
+         let platformPath = try? VirtualPath(path: platformDir) {
+        addPluginPaths(
+          forPlatform: platformPath,
+          commandLine: &commandLine
+        )
+      }
```
rintaro, line 476: I think `env["PLATFORM_DIR"]` based paths precedes the SDK path based paths.
If there are plugins with the same name in the both directories, earlier `-external-plugin-path` option takes the precedence.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test Windows

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
DougGregor: Good point, I've moved it first

PR #1731, opened 2024-11-18
Update Yams version to fix Android build

jakepetroules:


jakepetroules:
@swift-ci please smoke test

artemcm (approved):

compnerd (approved):
This seems fine, but please cross-check update-check-config.json in the swift repository to ensure that everything is in sync.

jakepetroules:
swiftlang/swift-driver#1731
swiftlang/swift#77713

@swift-ci please test

jakepetroules:
> This seems fine, but please cross-check update-check-config.json in the swift repository to ensure that everything is in sync.

https://github.com/swiftlang/swift/pull/77713

jakepetroules:
swiftlang/swift-driver#1731
swiftlang/swift#77713

@swift-ci please test

finagolfin:
Thanks, I've updated my Android CI and Termux toolchain to [use your new swift-toolchain-sqlite package for SwiftPM](https://github.com/finagolfin/swift-android-sdk/commit/51f5139f1d97e3f3c8b986a44a3e6ec23a038965#diff-25ea11a421d152e7629a53b6deea0f96d97fc9b0e4383a86aa270f200ecf8ad2R43), will submit that upstream once I test it more. Any plans to switch over sourcekit-lsp and this repo to use it too? It would be good to get rid of that external SQLite dependency altogether.

I also have to [keep patching swift-argument-parser and a few other packages](https://github.com/finagolfin/swift-android-sdk/blob/51f5139f1d97e3f3c8b986a44a3e6ec23a038965/swift-android-ci.patch#L45) because their tags haven't advanced enough to use my upstreamed Android pulls: let me know if we can bump those too.

In particular, swift-argument-parser 1.5.0 has my similar Android overlay patch: these pulls could probably bump the toolchain to use that tag too.

jakepetroules:
> Any plans to switch over sourcekit-lsp and this repo to use it too?

I thought we switched over everything in the Swift project that was using sqlite to use the new package; did I miss something? It doesn't look like sourcekit-lsp directly uses it.

> I also have to [keep patching swift-argument-parser and a few other packages](https://github.com/finagolfin/swift-android-sdk/blob/51f5139f1d97e3f3c8b986a44a3e6ec23a038965/swift-android-ci.patch#L45) because their tags haven't advanced enough to use my upstreamed Android pulls: let me know if we can bump those too.

Are you asking for new tags to be created for these repos? I could look at doing so for llbuild and TSC at least.

finagolfin:
> I thought we switched over everything in the Swift project that was using sqlite to use the new package; did I miss something? It doesn't look like sourcekit-lsp directly uses it.

I simply did a search for which binaries link against the SQLite3 library, did not check exactly how sourcekit-lsp got that pulled in. It's possible that with your recent pulls for Android, it's no longer a dependency there. 😃

Do you plan to switch linux and other ELF platforms like BSD over to using `swift-toolchain-sqlite` next?

> Are you asking for new tags to be created for these repos?

No, I think most have had new tags after my Android overlay changes: the Swift toolchain is simply using the older tags still, so if you want to bump more tags like this in the toolchain, I'm all for it. 👍

PR #1732, opened 2024-11-19
[cmake] Set up rpath for macOS binaries

Steelskin:
The rpath was not set for macOS binaries, resulting in binaries that needed post-build modifications to be runnable. This change sets the rpath for every binary built and fixes the installation folders for Swift libraries.


Sources/SwiftDriverExecution/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ set_property(GLOBAL APPEND PROPERTY SWIFTDRIVER_EXPORTS SwiftDriverExecution)
 set_target_properties(SwiftDriverExecution PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
 
-install(TARGETS SwiftDriverExecution
-  ARCHIVE DESTINATION lib
-  LIBRARY DESTINATION lib
-  RUNTIME DESTINATION bin)
+set_target_properties(SwiftDriverExecution PROPERTIES
+  INSTALL_RPATH "$<$<PLATFORM_ID:Darwin>:@loader_path/../pm/llbuild>")
```
finagolfin, line 29: This shouldn't be needed, since there's another copy in `@executable_path/../lib/swift/macosx`. I've already submitted a pull to remove this copy, swiftlang/swift#77647.

PR #1733, opened 2024-11-19
[Explicit Module Builds][Incremental Builds] Treat binary Swift module explicit dependencies as always up-to-date

artemcm:
The driver does not have the capability to rebuild them, so there is no point to invalidating these dependency nodes. Crucially, if these dependencies are newer than some other dependency between the source module and it, it will still cause invalidation of dependencies which can and must be re-built.

This change fixes the remaining case left by the initial attempt at this in https://github.com/swiftlang/swift-driver/pull/1674.


owenv (approved):

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:

PR #1734, opened 2024-11-19
[Explicit Module Builds] Always filter out up-to-date module dependencies from execution

artemcm:
With an escape hatch of `-always-rebuild-module-dependencies` (https://github.com/swiftlang/swift/pull/77717). This change makes it so the driver is always going to produce a build plan which contains only explicit module pre-compile tasks for modules which are not already up-to-date with respect to all of their inputs.

This means that the driver will skip up-to-date module tasks even when `-incremental` is not specified. This is more in line with prior behavior when using Implicit Module builds and results in a better user experience when compiling code with `swiftc` on the command-line. For example, consequtive invocations of `swiftc test.swift -o test.out` should not have to re-build module dependencies of `test.swift` each time.

Swift's `-incremental` is a heavy-weight set of machinery for selective re-compilation of a target in single-file/batch mode, whereas checking which module dependencies are up-to-date is a much simpler mechanism which we have no reason not to run on each compile. 

This change takes the existing logic that ran during a full Incremental build first wave to determine which dependencies need to be re-built and moves it to a common non-incremental code-path.

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/CommonDependencyOperations.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/CommonDependencyOperations.swift
@@ -287,6 +287,29 @@ internal extension InterModuleDependencyGraph {
     return modulesRequiringRebuild
   }
 
+  /// From a set of provided module dependency pre-compilation jobs,
+  /// filter out those with a fully up-to-date output
+  func filterMandatoryModuleDependencyCompileJobs(_ allJobs: [Job],
+                                                  fileSystem: FileSystem,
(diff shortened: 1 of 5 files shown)


artemcm:
@swift-ci test

artemcm:
Test in this PR requires https://github.com/swiftlang/swift-driver/pull/1733 to land, first.

owenv (approved):

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:

artemcm:
@swift-ci test

artemcm:

artemcm:
@swift-ci test

artemcm:

PR #1735, opened 2024-11-19
[Dependency Scanning] Deprecate/Remove batch scanning capability

artemcm:
Batch dependency scanning was added as a mechanism to support multiple compilation contexts within a single module dependency graph. The Swift compiler and the Explicitly-built modules model has long since abandoned this approach and this code has long been stale. It is time to remove it.

Compiler-side change: https://github.com/swiftlang/swift/pull/77719


artemcm:
@swift-ci test

artemcm:

owenv (approved):

PR #1736, opened 2024-11-20
Exclude `.autolink` files from static linking on WebAssembly

kateinoigakukun:
The `.autolink` files were being included in static archives and they were silently ignored by wasm-ld linker but they started emitting warnings for archive members that are neither object files nor bitcode files[^1]. This change filters out `.autolink` files from the list of inputs to the static linker job to avoid the warnings.

[^1]: https://github.com/llvm/llvm-project/commit/bcc9b9d80c61ea6521215e9826281041a3f73b05

### DIFF
--- a/Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
@@ -198,7 +198,7 @@ extension WebAssemblyToolchain {
       commandLine.appendFlag("crs")
       commandLine.appendPath(outputFile)
 
-      commandLine.append(contentsOf: inputs.map { .path($0.file) })
+      commandLine.append(contentsOf: inputs.lazy.filter { $0.type != .autolink }.map { .path($0.file) })
       return try resolvedTool(.staticLinker(lto))
     }
   }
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
@swift-ci test

Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
```diff
@@ -198,7 +198,10 @@ extension WebAssemblyToolchain {
       commandLine.appendFlag("crs")
       commandLine.appendPath(outputFile)
 
-      commandLine.append(contentsOf: inputs.map { .path($0.file) })
+      commandLine.append(contentsOf: inputs.lazy.filter {
+                            lto == nil ? $0.type == .object
+                                       : $0.type == .object || $0.type == .llvmBitcode
```
compnerd, line 203: This logic seems complicated, it seems to be saying:

```
$0.type == .object || (lto && $0.type == .llvmBitcode)
```

You are filtering out all objects and any LLVM bitcode if LTO is enabled. That said, this seems like fixing the problem on the wrong side. Why not exclude the autolink files in the first place? I doubt that BFD will honour it.

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test Windows

MaxDesiatov (approved):

PR #1738, opened 2024-11-21
[Dependency Scanning] Deprecate/Remove inter-module dependency graph serialization SwiftScan API uses

artemcm:
This mechanism relied on serializing contents of a dependency scanner cache that included all queried scanning contexts. Instead, the scanner will move to a per-query serialization format. These API were never actually deployed and, to our knowledge, have not worked for a long time.

Compiler-side change: https://github.com/swiftlang/swift/pull/77794


artemcm:
@swift-ci test

artemcm:
@cachemeifyoucan what do you think?

cachemeifyoucan (approved):
LGTM

PR #1740, opened 2024-11-26
[FreeBSD] Fix FreeBSD build

michael-yuji:
Fix building on FreeBSD. Tested on x86_64 FreeBSD-14.1-RELEASE-p6

### DIFF
--- a/Sources/SwiftDriver/Utilities/System.swift
+++ b/Sources/SwiftDriver/Utilities/System.swift
@@ -59,7 +59,7 @@ func quoteArgument(_ argument: String) -> String {
 #endif
 }
 
-#if canImport(Darwin) || os(Linux) || os(Android) || os(OpenBSD)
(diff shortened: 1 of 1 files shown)


compnerd (approved):

compnerd:
@swift-ci please test

Sources/SwiftDriver/Utilities/System.swift
```diff
@@ -59,7 +59,7 @@ func quoteArgument(_ argument: String) -> String {
 #endif
 }
 
-#if canImport(Darwin) || os(Linux) || os(Android) || os(OpenBSD)
+#if canImport(Darwin) || os(Linux) || os(Android) || os(OpenBSD) || os(FreeBSD)
```
jakepetroules, line 62: Not a blocker, but IMO this would be more simply expressed as if-Windows-else-assume-POSIX, rather than explicitly listing every OS name.

This same code most likely works for NetBSD, QNX, Solaris, etc...

jakepetroules (approved):

PR #1743, opened 2024-12-02
SwiftDriver: adjust handling of `-resource-dir` parameter

compnerd:
When computing the resource directory for the Swift registrar, we need to add a couple of components that we had missed. Correct the path computation so that we can find the object file properly. This was detected while running the test suite with the early swift driver on Windows where `-resource-dir` is explicitly passed.


compnerd:
@swift-ci please test

artemcm (approved):

compnerd:
@swift-ci please test

compnerd:

compnerd:

PR #1750, opened 2024-12-06
swift-driver: adjust the path computation for the legacy driver

compnerd:
When running on some platforms we may not have a resolved path given to the invocation. Separately handle the invocation and the executable path to compute the desired path.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1751, opened 2024-12-09
Fix some of `ProcessEnv.vars` and `AbsolutePath` warnings

MaxDesiatov:
Fixed some of these warnings:

```
warning: 'vars' is deprecated: Use `block` instead
```
```
- warning: 'init(_:)' is deprecated: use throwing `init(validating:)` variant instead
```

Fixing the former also improves compatibility with Windows env vars handling.


MaxDesiatov:
@swift-ci test

PR #1753, opened 2024-12-10
swift-driver: Driver support for SamplePGO

kavon:
For rdar://135443278

This makes the flags available in the new driver. It also infers an extra flag for the frontend that matches how `clang` works.


kavon:
@swift-ci test

artemcm (approved):

kavon:
@swift-ci test

PR #1756, opened 2024-12-11
Fix the legacy driver path construction on non-Windows platforms

kateinoigakukun:
`swift-driver` tries to invoke `swiftc-legacy-driver.` (trailing dot) on non-Windows platforms when legacy driver is requested since https://github.com/swiftlang/swift-driver/pull/1746

This is because `URL.pathExtension` returns an empty string if the URL does not have an extension but `URL.appendPathExtension` appends a period before the extension even if given an empty string. Thus `url.deletingPathExtension().appendingPathExtension(url.pathExtension)` will not be equal to `url` if `url` does not have an extension.

### DIFF
--- a/Sources/swift-driver/main.swift
+++ b/Sources/swift-driver/main.swift
@@ -83,10 +83,26 @@ do {
     // invocation for a multi-call binary.
     let executable: URL = URL(fileURLWithPath: ProcessInfo.processInfo.arguments[0])
     let invocation: URL = URL(fileURLWithPath: CommandLine.arguments[0])
+
+    #if compiler(>=6.2)
     let legacyExecutablePath: URL =
+         executable.deletingLastPathComponent()
+                   .appendingPathComponent("\(invocation.deletingPathExtension().lastPathComponent)-legacy-driver")
+                   .appendingPathExtension(executable.pathExtension)
+    #else
+    // WORKAROUND: Check if the original path has a non-empty extension because
+    // `url.deletingPathExtension().appendingPathExtension(url.pathExtension)`
+    // will not be equal to `url` but be `\(url).` if `url` doesn't have an
+    // extension on Swift 6.0. Remove this when we drop support for Swift 6.0
+    // as a bootstrapping compiler.
+    // See https://github.com/swiftlang/swift-foundation/issues/1080
+    var legacyExecutablePath: URL =
         executable.deletingLastPathComponent()
                   .appendingPathComponent("\(invocation.deletingPathExtension().lastPathComponent)-legacy-driver")
-                  .appendingPathExtension(executable.pathExtension)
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
Wait, it seems like a swift-foundation side issue:
```swift
import Foundation
let url = URL(fileURLWithPath: "/tmp/x")
print(url.deletingPathExtension().appendingPathExtension(url.pathExtension))
// Swift 5.10: file:///tmp/x
// Swift 6.0: file:///tmp/x.
```

kateinoigakukun:
@compnerd The team said it was an unintentional change and fixed in the main branch. https://github.com/swiftlang/swift-foundation/issues/1080

kateinoigakukun:
@swift-ci smoke test

Sources/swift-driver/main.swift
```diff
@@ -83,10 +83,18 @@ do {
...
     let invocation: URL = URL(fileURLWithPath: CommandLine.arguments[0])
-    let legacyExecutablePath: URL =
+    var legacyExecutablePath: URL =
         executable.deletingLastPathComponent()
                   .appendingPathComponent("\(invocation.deletingPathExtension().lastPathComponent)-legacy-driver")
-                  .appendingPathExtension(executable.pathExtension)
+    // WORKAROUND: Check if the original path has a non-empty extension because
+    // `url.deletingPathExtension().appendingPathExtension(url.pathExtension)`
+    // will not be equal to `url` but be `\(url).` if `url` doesn't have an
+    // extension.on Swift 6.0. Remove this when we drop support for Swift 6.0
```
kkebo, line 92: A typo?

```suggestion
    // extension on Swift 6.0. Remove this when we drop support for Swift 6.0
```

Sources/swift-driver/main.swift
```diff
@@ -83,10 +83,18 @@ do {
     // invocation for a multi-call binary.
     let executable: URL = URL(fileURLWithPath: ProcessInfo.processInfo.arguments[0])
     let invocation: URL = URL(fileURLWithPath: CommandLine.arguments[0])
-    let legacyExecutablePath: URL =
+    var legacyExecutablePath: URL =
```
compnerd, line 86: Can you duplicate this code instead and guard it with a compiler version check please?

compnerd (changes requested):

Sources/swift-driver/main.swift
kateinoigakukun: Ok, gated behind 6.2 until we cherry-pick the fix to 6.1 branch.

compnerd:

compnerd:

PR #1757, opened 2024-12-11
Tests: remove some debugging code

compnerd:
Remove a number of cases where we would print the command line or using `-v` on commands which are being executed which results in a significant amount of output in the test run.


compnerd:
@swift-ci please test

cachemeifyoucan (approved):

compnerd:

cachemeifyoucan:
I think the macOS failure is a real one. There is something wrong with the job planning. It seems it is not planning build the clang `SwiftShim` module.

compnerd:

PR #1763, opened 2024-12-19
Tests: various changes to clean up the swift-driver test suite (NFC)

compnerd:
Various changes to clean up the test suite implementation.
- Adopt `XCTAssertJobInvocationMatches` in more test cases
- Replace use of `XCTAssert(... == ...)` in favour of `XCTAssertEqual`
- Replace use of `XCTAssertTrue(!...)` in favour of `XCTAssertFalse`
- Some stylistic compression of `... ; return`
- Coddle braces as per Swift style
- Adopt `XCTUnwrap` in a number of test cases

The majority of this change is intended to reduce the LoC and ease debugging by allowing the test failure to indicate what the expectation vs computed values are. Additionally, it makes the test suite run to completion in more cases by removing forced unwraps.


compnerd:
@swift-ci please test

cachemeifyoucan (approved):
We need a linter to implement all those changes :)

compnerd:
Yeah, but since I'm trying to figure out how to clean up the test failures on Windows, might as well as take the opportunity to clean up the ones that are around the areas that I'm looking at due to the failures.

PR #1766, opened 2024-12-20
Jobs: adjust swift module wrapping job (NFC)

compnerd:
With `-emit-module`, the output type computed is the swiftmodule. This output is subsequently wrapped by the module wrap job which will embed the swift module into an object file. This would trigger the assertion previously. Clean up some of the planning code associated with this in tandem.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1767, opened 2024-12-20
SwiftDriver: honour `SWIFT_DRIVER_SWIFTSCAN_LIB` on Windows

compnerd:
We would previously not honour this flag on Windows as the layout of the toolchain is different on Windows from the other platforms. The computed path needing to be different bifurcated the implementation. Hoist a path for override handling into the shared path.


compnerd:
@swift-ci please test

MaxDesiatov (commented):
Is there an existing test that would be impacted by the change of behavior, or is a new test needed?

compnerd:
Yeah, the existing test cases are impacted by this which is how I found this in the first place.

MaxDesiatov:
We don't run this test suite on Windows CI nodes, same as SwiftPM I assume?

compnerd:
Correct, this is why I've been making all these changes for the test suite - been trying to clean up the test suite so that I can get it passing on Windows again.

MaxDesiatov (approved):

PR #1769, opened 2024-12-20
Tests: avoid forming an in-place driver job

compnerd:
We would previously form an in-place job on Windows which would then terminate the tests. Explicitly use an output file map which generates a build record which prevents the in-place job.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1771, opened 2024-12-24
Driver: simplify `Driver.computeSDKPath(_:compilerMode:toolchain:targ…

compnerd:
…etTriple:fileSystem:iagnosticsEngine:env:)`

Simplify the logic in this method to help reduce some of the dependency on the `FileSystem` type from tools-support-core. The incidental benefit here is to reduce the dependency on `AbsolutePath` and `RelativePath`.

The semantic changes incurred by this change is that the `-sdk` parameter for the frontend is now always converted to an absolute path from the relative path.


compnerd:
@swift-ci please test

PR #1773, opened 2025-01-01
libclang_rt.profile*.a search path should use osNameUnversioned

michael-yuji:
On platforms where the triple includes the os version, in this case `x86_64-unknown-freebsd14.2`,
the swift-driver is trying to link against various clang rt libraries in `clang/lib/freebsd14.2`, but the actual
clang convention is to have the libraries installed in `clang/lib/freebsd` instead, see [clang/lib/Driver/ToolChain.cpp#L686](https://github.com/swiftlang/llvm-project/blob/985636ed916ef54e2df74f9d72fc20656b11c094/clang/lib/Driver/ToolChain.cpp#L686).

Although the right solution is probably to implement `getOSLibName` in `swift-driver`; we can just use
`osNameUnversioned` for now as the only special case where `getOSLibName != osNameUnversioned`
is for `Solaris`, which is not yet supported.


michael-yuji:
@swift-ci please test

jakepetroules (approved):

PR #1775, opened 2025-01-08
Remove error for invalid swift caching configuration

cachemeifyoucan:
Currently, swift-driver emits an error when trying to importing a swift binary module that has a bridging header when swift caching is enabled. This turns out to be a very common configuration that we need to support.

It turns out most of the time, allowing the compilation to continue will result in a successful compilation. Even though in rare occasions where multiple bridging headers from the dependencies are involved with overlapping content, the compilation is going to fail with confusing error message, we will address that in swift compiler instead.

Thus the current error message is removed to allow more supported projects.


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:

artemcm (approved):

PR #1781, opened 2025-01-21
[Options] Update options.

al45tair:
We need to do this to remove references to implicit backtracing import, but it also looks like there are some other options that haven't been picked up yet.

rdar://143310410


al45tair:
I should say that this was done with `makeOptions`, following the instructions in the README.

al45tair:
@swift-ci Please test

al45tair:
@swift-ci Please test

owenv (approved):

PR #1782, opened 2025-01-22
build: add additional target exports

compnerd:
When building swift-driver with static libraries, we need to add this additional export. This is meant to support building swift-driver with static libraries for use as the early swift-driver on Windows.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1784, opened 2025-01-28
build: tweak `FetchContent` usage to follow best practices

compnerd:
We should do a singular call to `FetchContent_MakeAvailable` as per the documentation. Heed this advice which also allows us to reduce the scope of the overridden variables. Once we migrate to a new enough CMake minimum (3.25), we should be able to replace that with a `block` scope to avoid the push/pop manual adjustment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,11 +49,7 @@ find_package(Foundation QUIET)
 
 include(FetchContent)
 
-set(_SD_SAVED_BUILD_TESTING ${BUILD_TESTING})
-set(_SD_SAVED_BUILD_EXAMPLES ${BUILD_EXAMPLES})
-
-set(BUILD_TESTING NO)
-set(BUILD_EXAMPLES NO)
+set(VendoredDependencies)
 
 find_package(ArgumentParser CONFIG)
 if(NOT ArgumentParser_FOUND)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

CMakeLists.txt
```diff
@@ -49,19 +49,15 @@ find_package(Foundation QUIET)
 
 include(FetchContent)
 
-set(_SD_SAVED_BUILD_TESTING ${BUILD_TESTING})
-set(_SD_SAVED_BUILD_EXAMPLES ${BUILD_EXAMPLES})
-
-set(BUILD_TESTING NO)
-set(BUILD_EXAMPLES NO)
+set(VendoredDependencies)
```
etcwilde, line 52: Do we need to forward declare this?

CMakeLists.txt
```diff
@@ -84,9 +80,17 @@ if(NOT TSC_FOUND)
   FetchContent_Declare(ToolsSupportCore
     GIT_REPOSITORY https://github.com/apple/swift-tools-support-core
     GIT_TAG main)
-  FetchContent_MakeAvailable(ToolsSupportCore)
+  list(APPEND VendoredDependencies ToolsSupportCore)
 endif()
 
+set(_SD_SAVED_BUILD_TESTING ${BUILD_TESTING})
```
etcwilde, line 86: General rambling... but this seems like a good use of `block()`/`endblock()`, but that's only available in CMake 3.25+.

etcwilde (approved):

CMakeLists.txt
compnerd: I suppose that we don't need to forward declare it. But it does make it easier to read the logic. I can remove it if you prefer.

CMakeLists.txt
compnerd: Agreed. See the commit message :)

PR #1786, opened 2025-01-28
[Incremental Builds][Explicit Module Builds] Switch to using incremental dependency scanning

artemcm:
For Explicit Module Builds, have the driver configure the dependency scanner invocation to serialize its internal scanner cache state after a scan, and attempt to deserialize a prior build's scanner cache state, and validate its contents for an incremental re-scan.

Using functionality implemented in: https://github.com/swiftlang/swift/pull/78962

### DIFF
--- a/Sources/SwiftDriver/Execution/DriverExecutor.swift
+++ b/Sources/SwiftDriver/Execution/DriverExecutor.swift
@@ -61,22 +61,27 @@ public struct DriverExecutorWorkload {
   }
 
   public let kind: Kind
-  public let interModuleDependencyGraph: InterModuleDependencyGraph?
+
+  @available(*, deprecated, message: "use of 'interModuleDependencyGraph' on 'DriverExecutorWorkload' is deprecated")
+  public let interModuleDependencyGraph: InterModuleDependencyGraph? = nil
 
   public init(_ allJobs: [Job],
               _ incrementalCompilationState: IncrementalCompilationState?,
-              _ interModuleDependencyGraph: InterModuleDependencyGraph?,
               continueBuildingAfterErrors: Bool) {
     self.continueBuildingAfterErrors = continueBuildingAfterErrors
     self.kind = incrementalCompilationState
       .map {.incremental($0)}
       ?? .all(allJobs)
-    self.interModuleDependencyGraph = interModuleDependencyGraph
   }
 
+  static public func all(_ jobs: [Job]) -> Self {
+    .init(jobs, nil, continueBuildingAfterErrors: false)
+  }
+
+  @available(*, deprecated, message: "use all(_ jobs: [Job]) instead")
   static public func all(_ jobs: [Job],
-                         _ interModuleDependencyGraph: InterModuleDependencyGraph? = nil) -> Self {
-    .init(jobs, nil, interModuleDependencyGraph, continueBuildingAfterErrors: false)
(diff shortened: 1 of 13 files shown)


nkcsgexi:
@swift-ci please test

artemcm:
I think this PR will need the compiler change to land in a nightly snapshot first, in order to pass the tests.

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
```diff
@@ -131,7 +128,9 @@ extension IncrementalCompilationState.FirstWaveComputer {
     // In the case where there are no compilation jobs to run on this build (no source-files were changed),
     // we can skip running `beforeCompiles` jobs if we also ensure that none of the `afterCompiles` jobs
     // have any dependencies on them.
+    // ACTODO: If the beforeCompiles contains modules, invalidate all sources
```
owenv, line 131: I think this TODO, needs to be addressed before merging, right?

cachemeifyoucan:
> I think this PR will need the compiler change to land in a nightly snapshot first, in order to pass the tests.

Shouldn't swift-driver tests work with/without swift compiler changes? It needs to handle both cases otherwise it is going to be a trouble for standalone development.

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
artemcm: This comment was stale. I am now convinced that in a scenario where `beforeCompiles` contains modules, the existing cross-module incremental machinery is guaranteed to have also, based on stale module dependencies, invalidated he appropriate source-files, and we will have required source compiles as well. Which means we do not need a further mechanism here.

artemcm:
@swift-ci test

artemcm:
> Shouldn't swift-driver tests work with/without swift compiler changes? It needs to handle both cases otherwise it is going to be a trouble for standalone development.

We often evolve the drive in-sync with the compiler and make inter-dependent changes between the two. The driver is tested against the current development snapshot compiler, and maintaining **test** compatibility with much older toolchains would be a large burden here that is not worthwhile, in my opinion. Standalone development works quite well with the same snapshot toolchains as are used in CI here.

artemcm:
@swift-ci test

artemcm:

Sources/SwiftDriver/Execution/DriverExecutor.swift
```diff
@@ -61,22 +61,18 @@ public struct DriverExecutorWorkload {
   }
 
   public let kind: Kind
-  public let interModuleDependencyGraph: InterModuleDependencyGraph?
 
   public init(_ allJobs: [Job],
```
cachemeifyoucan, line 65: Do we need to deprecate when we change public APIs? I don't see `swift-build` is affected by this change but I am not sure if there are other uses.
Same for `.all()`.

cachemeifyoucan (approved):
LGTM function wise.

artemcm:
@swift-ci test

Sources/SwiftDriver/Execution/DriverExecutor.swift
artemcm: Thank you for catching this. Brought back the deleted interface with deprecation warnings.

artemcm:

PR #1800, opened 2025-02-12
[Test] Allow for diagnostic verifier to expect an alternative diagnostic

artemcm:
And fix 'testTraceDependency' where we can have two possibilities for the emitted note.


artemcm:
@swift-ci test

etcwilde (approved):

PR #1801, opened 2025-02-12
Exclude `swift-autolink-extractor` no-op job for embedded Wasm

MaxDesiatov:
This driver job has no effect (all autolinking sections are empty when building in this mode) and is actually prone to failure when Wasm reference types are used:

```
error: autolink-extract command failed with exit code 1 (use -v to see invocation)
.build/wasm32-unknown-none-wasm/release/externref.build/externref.c.o' (invalid data relocation: table)
```

### DIFF
--- a/Sources/SwiftDriver/Jobs/AutolinkExtractJob.swift
+++ b/Sources/SwiftDriver/Jobs/AutolinkExtractJob.swift
@@ -19,7 +19,7 @@ import struct TSCBasic.RelativePath
 extension Driver {
   /*@_spi(Testing)*/ public var isAutolinkExtractJobNeeded: Bool {
       switch targetTriple.objectFormat {
-      case .wasm where targetTriple.os != .noneOS:
+      case .wasm where !parsedOptions.isEmbeddedEnabled:
           fallthrough
 
       case .elf:
--- a/Sources/SwiftDriver/Jobs/AutolinkExtractJob.swift
+++ b/Sources/SwiftDriver/Jobs/AutolinkExtractJob.swift
@@ -18,16 +18,18 @@ import struct TSCBasic.RelativePath
 // FIXME: Also handle Cygwin and MinGW
 extension Driver {
   /*@_spi(Testing)*/ public var isAutolinkExtractJobNeeded: Bool {
+    mutating get {
       switch targetTriple.objectFormat {
       case .wasm where !parsedOptions.isEmbeddedEnabled:
-          fallthrough
+        fallthrough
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

artemcm (approved):

kateinoigakukun (dismissed):
Is this issue only for Embedded mode? I guess it could happen with WASI targets as well, isn't it?

MaxDesiatov:
@swift-ci test windows

MaxDesiatov:
> Is this issue only for Embedded mode? I guess it could happen with WASI targets as well, isn't it?

Right, WASI is not impacted by this PR though. The issue might happen, but I suggest we fix it in `swift-autolink-extractor` itself, since WASI does need that utility. I assume it uses autolinking mechanisms when linking core libraries?

This PR can unblock experiments with reference types in embedded Wasm at least. If we ever need libraries support in embedded Wasm, we can revert this PR later.

kateinoigakukun:
Okay, it makes sense to skip the autolink-extract step for Embedded mode anyway. Could you use the Embedded feature flag instead of target OS to make the decision?
I'm working with Emscripten recently and OS == .wasi doesn't work for the case.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/SwiftDriver/Jobs/AutolinkExtractJob.swift
```diff
@@ -18,8 +18,16 @@ import struct TSCBasic.RelativePath
 // FIXME: Also handle Cygwin and MinGW
 extension Driver {
   /*@_spi(Testing)*/ public var isAutolinkExtractJobNeeded: Bool {
-    [.elf, .wasm].contains(targetTriple.objectFormat) && lto == nil &&
-    linkerOutputType != nil
+      switch targetTriple.objectFormat {
+      case .wasm where targetTriple.os != .noneOS:
```
kateinoigakukun, line 22:
```suggestion
      case .wasm where !parsedOptions.isEmbeddedEnabled:
```

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

kateinoigakukun (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #1804, opened 2025-02-14
SwiftDriver: enable `-sanitize=` on Windows selectively

compnerd:
ASAN, UBSAN are available on Windows while TSAN is not. The toolchain has started building the runtimes and will be distributing them with the toolchain builds. This change enables the support in the driver for the santizers on Windows.


compnerd:
@swift-ci please test

artemcm (approved):

artemcm:
The "new" style seems like the way to go, especially since even `clang` logic prefers it. Let's go with it here, and then also consider if we should move other platforms to this layout.

PR #1806, opened 2025-02-14
Fix newly-introduced tests for non-Darwin platforms to account for additional autolink-extract job

artemcm:


artemcm:
@swift-ci test

tshortli (approved):

PR #1809, opened 2025-02-14
[Test] Fix a typo in the JobExecutorTests

cachemeifyoucan:
There is a typo in "testInputModifiedDuringMultiJobBuild" that calls `swift` instead of `swiftc`. This drops swift into interactive mode and can cause the entire test suite to return 0 on linux even there are other errors.

### DIFF
--- a/Tests/SwiftDriverTests/JobExecutorTests.swift
+++ b/Tests/SwiftDriverTests/JobExecutorTests.swift
@@ -351,7 +351,7 @@ final class JobExecutorTests: XCTestCase {
       let main = path.appending(component: "main.swift")
       try localFileSystem.writeFileContents(main, bytes: "let foo = 1")
 
-      var driver = try Driver(args: ["swift", main.pathString])
+      var driver = try Driver(args: ["swiftc", main.pathString])
       let jobs = try driver.planBuild()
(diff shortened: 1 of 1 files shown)


cachemeifyoucan:

artemcm (approved):
Thank you.

cachemeifyoucan:
Testing linux only for now since I expect this to fail before https://github.com/swiftlang/swift-driver/pull/1808 is merged.

Tests/SwiftDriverTests/JobExecutorTests.swift
```diff
@@ -351,7 +351,7 @@ final class JobExecutorTests: XCTestCase {
       let main = path.appending(component: "main.swift")
       try localFileSystem.writeFileContents(main, bytes: "let foo = 1")
 
-      var driver = try Driver(args: ["swift", main.pathString])
+      var driver = try Driver(args: ["swiftc", main.pathString])
       let jobs = try driver.planBuild()
       XCTAssertEqual(jobs.count, 1)
```
artemcm, line 356: This needs to change as well as the line below. Which sort of then goes against the point of this test.

PR #1811, opened 2025-02-16
[Unix] Use the new `-sysroot` flag for all non-Darwin Unix platforms, not just Android

finagolfin:
This makes two changes:

1. Use the new flag on all Unix platforms, as in the original frontend pull where it was introduced, swiftlang/swift#72352.
2. When compiling for Android and `ANDROID_NDK_ROOT` is unset, fall back to the `-sdk` as the sysroot when linking.

I'm primarily concerned with 2., as it got `swift-crypto` cross-compiling again with my trunk Android SDK bundle, but I went ahead and updated 1. also, when I looked into how that new `-sysroot` flag was introduced in that linked compiler frontend pull.

I tested this pull on linux x86_64 locally and saw no test regressions.

@compnerd, I'd like to expand this pull to get rid of `AndroidNDK.getDefaultSysrootPath(in:)` altogether, if you aren't using it anymore. Please review, since you added all this `-sysroot` code for [your new cross-compilation vision](https://github.com/swiftlang/swift/blob/main/docs/CrossCompilationModel.md).

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -217,14 +217,14 @@ extension GenericUnixToolchain {
         commandLine.appendPath(try VirtualPath(path: opt.argument.asSingle))
       }
 
-      if targetTriple.environment == .android {
-        if let sysroot = parsedOptions.getLastArgument(.sysroot)?.asSingle {
-          commandLine.appendFlag("--sysroot")
-          try commandLine.appendPath(VirtualPath(path: sysroot))
-        } else if let sysroot = AndroidNDK.getDefaultSysrootPath(in: self.env) {
-          commandLine.appendFlag("--sysroot")
-          try commandLine.appendPath(VirtualPath(path: sysroot.pathString))
-        }
+      if let sysroot = parsedOptions.getLastArgument(.sysroot)?.asSingle {
+        commandLine.appendFlag("--sysroot")
+        try commandLine.appendPath(VirtualPath(path: sysroot))
+      } else if targetTriple.environment == .android,
+        let sysroot = AndroidNDK.getDefaultSysrootPath(in: self.env)
+      {
+        commandLine.appendFlag("--sysroot")
+        try commandLine.appendPath(VirtualPath(path: sysroot.pathString))
       } else if let path = targetInfo.sdkPath?.path {
         commandLine.appendFlag("--sysroot")
         commandLine.appendPath(VirtualPath.lookup(path))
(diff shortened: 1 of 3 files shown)


finagolfin:
@swift-ci please test

finagolfin:
@swift-ci please smoke test linux

Sources/SwiftDriver/Toolchains/GenericUnixToolchain.swift
```diff
@@ -152,17 +152,16 @@ public final class GenericUnixToolchain: Toolchain {
...
-      } else if let sysroot = AndroidNDK.getDefaultSysrootPath(in: self.env) {
-        commandLine.appendFlag("-sysroot")
-        try commandLine.appendPath(VirtualPath(path: sysroot.pathString))
-      }
+    if let sysroot = driver.parsedOptions.getLastArgument(.sysroot)?.asSingle {
+      commandLine.appendFlag("-sysroot")
+      try commandLine.appendPath(VirtualPath(path: sysroot))
+    } else if driver.targetTriple.environment == .android,
+      let sysroot = AndroidNDK.getDefaultSysrootPath(in: self.env)
+    {
```
compnerd, line 160:
```suggestion
      let sysroot = AndroidNDK.getDefaultSysrootPath(in: self.env) {
```

Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
```diff
@@ -217,14 +217,14 @@ extension GenericUnixToolchain {
...
-        } else if let sysroot = AndroidNDK.getDefaultSysrootPath(in: self.env) {
-          commandLine.appendFlag("--sysroot")
-          try commandLine.appendPath(VirtualPath(path: sysroot.pathString))
-        }
+      if let sysroot = parsedOptions.getLastArgument(.sysroot)?.asSingle {
+        commandLine.appendFlag("--sysroot")
+        try commandLine.appendPath(VirtualPath(path: sysroot))
+      } else if targetTriple.environment == .android,
+        let sysroot = AndroidNDK.getDefaultSysrootPath(in: self.env)
+      {
```
compnerd, line 225:
```suggestion
        let sysroot = AndroidNDK.getDefaultSysrootPath(in: self.env) {
```

compnerd (approved):
This change seems reasonable. We are still using the fallback path, so please do not remove that.

Sources/SwiftDriver/Toolchains/GenericUnixToolchain.swift
finagolfin: Thanks for the review, but do you mind if I leave the formatting this way? I have no preference and merely ran this through `swift format` to get this layout.

compnerd: If that is from swift-format, that is good.

finagolfin: Yep, ran it through again to double check.

PR #1812, opened 2025-02-18
[Tests] Fix another early termination in tests

cachemeifyoucan:
When running swift-driver test suite with an older compiler that doesn't support bridging header chaining, the second part of the bridging header caching test will just be a single job, from which the default executor will run the swift-frontend in-place. This will cause the test suite to early exit regardless of the state of the test suite.


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
ping

nkcsgexi (approved):

PR #1813, opened 2025-02-18
[Caching] Change SDK prefix mapping to parent directory

cachemeifyoucan:
Change the prefix mapping of the SDK to the parent path of the actualy SDK path. This preserves the SDK name in the path so tools can make assumption with the name to infer informations if needed.

rdar://144886542


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:

artemcm (approved):

benlangmuir:
Could we instead change the code that is looking at the directory name to instead look at SDKSettings? It looks like the driver already knows how to read that file. Being able to canonicalize across different SDK paths seems better to me.

PR #1814, opened 2025-02-19
Tests: reduce enabled tests on Linux

compnerd:
These seem to be causing a regression on Linux somehow. Restore the original `#if os(macOS)` check around the region, but retain some of the enabled tests which are passing.


compnerd:
@swift-ci please test

shahmishal:
Merging to unblock PR testing.

PR #1819, opened 2025-02-21
[Unix] Use the new `-sysroot` flag for all non-Darwin Unix platforms, not just Android (#1811)

finagolfin:
__Explanation:__ This makes two changes:

1. Use the new flag on all Unix platforms, as in the original frontend pull where it was introduced, swiftlang/swift#72352.
2. When compiling for Android and `ANDROID_NDK_ROOT` is unset, fall back to the `-sdk` as the sysroot when linking.

I'm primarily concerned with 2., as it got a few packages cross-compiling again with both my trunk and 6.1 Android SDK bundles, but I went ahead and updated 1. also, when I looked into how that new `-sysroot` flag was introduced in that linked compiler frontend pull.

__Scope:__ Only affects Unix platforms and Android

__Issue:__ None

__Original PR:__ #1811

__Risk:__ Very low

__Testing:__ Passed all CI on trunk and appears to fix getting my Android SDK bundle working with this new swift-driver again, because of 2.

__Reviewer:__ @compnerd

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -217,14 +217,14 @@ extension GenericUnixToolchain {
         commandLine.appendPath(try VirtualPath(path: opt.argument.asSingle))
       }
 
-      if targetTriple.environment == .android {
-        if let sysroot = parsedOptions.getLastArgument(.sysroot)?.asSingle {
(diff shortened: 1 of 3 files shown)


finagolfin:
@swift-ci please test

finagolfin:
Ping @nkcsgexi, small pull for 6.1 that primarily only affects cross-compiling for Android.

nkcsgexi (approved):

finagolfin:
Thanks for the quick review. 😃

PR #1821, opened 2025-02-25
[SDK Support] Make sure to look in `swift_static` rather than just `swift`.

al45tair:
If we're building a statically linked executable, we should look in `usr/lib/swift_static` for `swiftrt.o`, rather than looking in `usr/lib/swift`.  Not doing this breaks the Static SDK for Linux.

rdar://145549541


al45tair:
**Explanation**: Without this change, when `-static-stdlib` or `-static-executable` are used, we look in the wrong place for `swiftrt.o`. With the Static SDK for Linux, this won't work because there is no `swiftrt.o` in `usr/lib/swift`, since we only support static linking.
**Risk**: Low. Only affects static linking with cross-compilation.
**Original PR**: #1820 
**Reviewed by**: TBC
**Resolves**: rdar://145549541
**Tests**: None at present. I'm going to add a test to the Static SDK for Linux build process that will make sure that we don't have a repeat of this kind of problem.

al45tair:
@swift-ci Please test

nkcsgexi (approved):

PR #1823, opened 2025-02-27
[SE-0458] Pass -strict-memory-safety through to the frontend

DougGregor:


DougGregor:
@swift-ci please test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -301,6 +301,7 @@ extension Driver {
                               .enableUpcomingFeature,
                               .disableUpcomingFeature,
                               from: &parsedOptions)
+    try commandLine.appendLast(.strictMemorySafety, from: &parsedOptions)
```
tshortli, line 304: Don't we need to check if it's supported, or has it been implemented in the frontend long enough?

tshortli: I guess it's opt-in so there isn't a big risk of incompatibility for this one.

tshortli (approved):

DougGregor:
Since its opt in, I didn't bother to check

PR #1824, opened 2025-02-27
Rename env var with more accurate name

bkhouri:
The SWIFTPM_SWBUILD_FRAMEWORK environment variable used in the Package.swift is not indicative of the actual behaviour.

Rename the environment variable to SWIFT_DRIVER_USE_STSC_DYLIB to more clearly indicate the behavioural change when it is set.


bkhouri:
@swift-ci please test

artemcm (approved):

PR #1826, opened 2025-03-03
Revert "[Planning] Avoid batching compile job twice"

owenv:
This reverts commit 7ba9d485ab44cca5c61cf974679d6fba91feb1ca.

This is breaking swift-build CI, because SwiftDriver now fails to return a build plan to clients when there are 0 mandatory jobs

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -90,17 +90,16 @@ extension Driver {
       incrementalCompilationState = nil
     }
 
-    let batchedJobs: [Job]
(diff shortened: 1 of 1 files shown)


owenv:
@swift-ci test

owenv:
This reverts one commit from https://github.com/swiftlang/swift-driver/pull/1825, but leaves the other 2

artemcm (approved):

cachemeifyoucan:
@owenv I will take a look. Maybe we should get swift-driver bot to run swift-build test suite as well just to increase coverage? The swift-driver bot is currently very fast so I won't mind it test more (for Mac and Linux for now, before windows is switched to the new system).

Issue #1562, opened 2024-03-13
This driver no longer looks in a passed-in `-sdk` for the Swift modules and libraries


finagolfin:
@compnerd and some Windows devs have been trying out Android cross-compilation from Windows, and in response to some issues they've been seeing, I suggested [symlinking the Swift files into the C sysroot in the Android NDK](https://github.com/apple/swift/pull/72161#discussion_r1523247310). I then tried it myself on linux, using my Android SDK for Swift for convenience, and found that it works with the legacy C++ Driver, but not this one:
```
> wget https://github.com/finagolfin/swift-android-sdk/releases/download/5.9.2/swift-5.9.2-android-24-sdk.tar.xz
> tar xf swift-5.9.2-android-24-sdk.tar.xz
> ln -s ~/swift-5.9.2-android-24-sdk/usr/lib/swift android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/swift
> ./swift-5.9.2-RELEASE-ubi9/usr/bin/swiftc -tools-directory ~/android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/ -target aarch64-unknown-linux-android24 -sdk android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/sysroot swift/test/Interpreter/hello_toplevel.swift  -v -L swift-5.9.2-android-24-sdk/usr/lib/aarch64-linux-android/ -disallow-use-new-driver
<unknown>:0: warning: legacy driver is now deprecated; consider avoiding specifying '-disallow-use-new-driver'
Swift version 5.9.2 (swift-5.9.2-RELEASE)
Target: aarch64-unknown-linux-android24
[...]

PR #1622, opened 2024-05-24
[Explicit Module Builds] Adopt dependency scanner Link Libraries support

artemcm:
- Adopt new libSwiftScan API to query link libraries for each module dependency
- Introduce '-explicit-auto-linking' to turn explicit dependency link library information into flags passed directly to the linker invocation

PR #1827, opened 2025-03-04
Stop passing `-resource-dir` to the Frontend if the flag wasn't passed in on the command-line

finagolfin:
As noted in #1562, this swift-driver always passes a `-resource-dir` to the `swift-frontend`, whereas the C++ Driver doesn't, unless one was explicitly passed in by the user. Given [the recent push to deprecate this flag](https://github.com/swiftlang/swift/blob/main/docs/CrossCompilationModel.md#flags), try to only use it when needed.

I tested this on linux locally and saw no test regressions, so push it through the full CI now.


Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -392,11 +392,13 @@ extension Driver {
       try addPathArgument(.absolute(workingDirectory), to: &commandLine, remap: jobNeedPathRemap)
     }
 
+    if parsedOptions.hasArgument(.resourceDir) || kind == .scanDependencies {
```
finagolfin, line 395: I didn't pass it for `scanDependencies` jobs either initially, then a single test `ExplicitModuleBuildTests.testExplicitLinkLibraries` failed. It appears that a `swift-frontend -scan-dependencies` job alone does not find the resource-dir properly, so I set this flag for that kind of job for now.

finagolfin:
@swift-ci please test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
artemcm: What was the failure? It's possible this job relies on being able to look up linked libraries relative to resource directory.

finagolfin: > What was the failure?

It complains that it didn't find a single library:
```
/home/finagolfin/swift-driver/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift:549: error: ExplicitModuleBuildTests.testExplicitLinkLibraries : XCTUnwrap failed: expected non-nil value of type "LinkLibraryInfo" -
```
The test, which you just added in #1622 last summer, looks in the Swift resource directory next to whatever compiler was used, as these swift-driver tests require a Swift toolchain to be in the `PATH` when they're run, and makes sure various parts of the stdlib are available. This currently works because this swift-driver queries the frontend with `-print-target-info` then passes that Swift `-resource-dir` to all subsequent jobs, like the `-scan-dependencies` frontend job that this test invokes.

However, this seems like a frontend bug that this swift-driver is covering up, as it is the frontend that finds the Swift resource-dir for all other jobs anyway. I will look into why it does not do so only for `-scan-dependencies` in the frontend and fix that first, just ran this through the CI to see if any other tests broke on other platforms and they didn't.

> It's possible this job relies on being able to look up linked libraries relative to resource directory.

Yep, the only question is who should supply the path to that Swift resource directory? We cannot easily do it in the test because it is just picking up the Swift frontend and resource-dir from the `PATH`, ie it could be anywhere.

I see no reason why the `swift-frontend -scan-dependencies` job can't find it, whereas the frontend always easily finds it for all other jobs, so I will look into fixing this in the frontend before coming back to this pull. If I'm missing anything, let me know.

finagolfin:
@swift-ci please test linux

finagolfin:
@swift-ci please test linux

finagolfin:
@swift-ci please test macos

finagolfin:
@swift-ci please test linux

finagolfin:
@swift-ci please test macos

finagolfin:
@swift-ci please test linux and macos

finagolfin:
@swift-ci please test linux

finagolfin:
@swift-ci please test macos

finagolfin:
@swift-ci please test linux

finagolfin:
@swift-ci please test macos

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
artemcm: Thank you for the investigation! It makes sense that this would be due to a difference with the in-process scanner. And your theory on the different `PATH` seems plausible. I am okay with passing `resource-dir` here to the scanner invocation only in the meantime, to make up for that difference. And I will see if I can discern making the in-process scanner potentially not need that.

finagolfin:
@swift-ci please test

finagolfin:
@artemcm, ready for review, passed CI and I ran it through a toolchain build and source compatibility run in the linked compiler pull above too.

The linux toolchain built fine while the mac  toolchain build failed when installing some unrelated lldb file. The source compatibility run shows as failed, but examining the results shows that everything passed, likely some CI misconfiguration. The Windows toolchain build failed because that compiler pull found a misconfiguration in the Windows CI, but since [the Windows CI run on this repo takes hours and built much of the toolchain fine](https://ci-external.swift.org/job/swift-driver-PR-windows/1343/) anyway, that's irrelevant.

artemcm (approved):

artemcm:

PR #1829, opened 2025-03-04
Re-land [Planning] Avoid batching compile job twice

cachemeifyoucan:
This re-land the change in 7ba9d485ab44cca5c61cf974679d6fba91feb1ca with compatibility fixes and additional tests.

The fix makes sure the skipped jobs are returned in the planning as well for compatibility reasons. Not after this change:

For the build systems that use swift-driver that doesn't support incremental build (swiftpm), asking for an incremental planning will still get the full list of jobs, but the skipped jobs might not be batched. The planned jobs are still complete, but might not be optimally batched.

For the build systems that supports incremental build (swift-build), the build planning is extracted from incremental state directly. But note swift-driver lazily wrapped the compile jobs in Executor so swift-driver can't return an empty list of jobs. Add a test to make sure that is the case.

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -90,16 +90,18 @@ extension Driver {
       incrementalCompilationState = nil
     }
 
-    return try (
-      // For compatibility with swiftpm, the driver produces batched jobs
-      // for every job, even when run in incremental mode, so that all jobs
-      // can be returned from `planBuild`.
-      // But in that case, don't emit lifecycle messages.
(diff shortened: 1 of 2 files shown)


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
Alternatives are:
* We are also batching the skipped job and return
* We keep the original logics but be very careful when to compute hash key. Assume the build system that supports caching also do the incremental build correctly by reading from incremental states, if there is an incremental state, don't compute hash for returned job list.

cachemeifyoucan:
@artemcm @owenv any opinion on this change?

artemcm (approved):

PR #1832, opened 2025-03-06
[PCH][OutputFileMap] GeneratePCH output should use single output entry

cachemeifyoucan:
The build system has been emitting the output path of generating PCH job in the module level entry in the output file map, while swift-driver is expecting the entry in an entry that accosiated with imported bridging header.

This patch unifies the behavior by picking the implementation of the swift-build, where it is a module level entry. This works better for future when bridging header can be chained, thus creating PCH job even there isn't a imported objc header explicitly provided.

rdar://146395316


cachemeifyoucan:
@swift-ci please test

artemcm (approved):

cachemeifyoucan:
@swift-ci please test

owenv (approved):

cachemeifyoucan:

PR #1836, opened 2025-03-12
Allow module aliases to be raw identifiers.

allevato:
This is the companion to the frontend changes for SE-0451 in https://github.com/swiftlang/swift/pull/76636. The first component of a `-module-alias` flag is allowed to be a raw identifier (without the backticks); for example, `-module-alias //some/module:name=ActualModuleName` allows users to write this:

```
import `//some/module:name`
```

The *target* of the alias (the actual module name) must still be a valid Swift identifier, since that corresponds to file system artifacts that need stricter naming conventions.


allevato:
@swift-ci please test

rauhul:
Oh this will be so much nicer for bazel projects! ❤️

allevato:

allevato:
@swift-ci please test

PR #1837, opened 2025-03-12
[Explicit Module Builds] Add support for '-clang-target-variant' flag

artemcm:
Adds equivalent functionality to that specified by '-clang-target' but for the zippered library variant on Darwin platforms.

Using new compiler flags added in https://github.com/swiftlang/swift/pull/79930.


artemcm:
@swift-ci test

tshortli (approved):

xymus (approved):

PR #1838, opened 2025-03-12
Added a getSingleFrontendInvocationFromDriverArgumentsV3 function whi…

cwakamo:
…ch takes `env` and `executor` parameters (in addition to what V2 takes).

This allows clients of swift-driver which need a custom environment (e.g. to specify the path to libSwiftScan or swift-frontend) or which need a custom executor to get a single frontend invocation given a set of driver arguments. This commit also promotes SimpleExecutor to `@_spi(Testing) public` from `internal` so it can be used in tests of this new function.

This addresses <rdar://problem/146875054>.


artemcm (approved):
Thank you.

cwakamo:
@swift-ci please test

PR #1841, opened 2025-03-14
Fix incorrect link job options for embedded WASI

MaxDesiatov:
`WebAssemblyToolchain` didn't check for `parsedOptions.isEmbeddedEnabled` when forming `swiftrt.o` and `static-executable-args.lnk` options, and neither of these files exist when building with Embedded Swift for WASI.


MaxDesiatov:
@swift-ci test

kateinoigakukun (approved):

PR #1846, opened 2025-03-20
Ensure we don't skip precompile jobs if a planned interface verification depends on recompiling modules

owenv:
The test has a longer comment explaining the scenario this addresses, where a change in flags or environment triggers updated PCM hashes when it should have been canonicalized away as the sources don't need to recompile.

rdar://146668157


owenv:
@swift-ci test

artemcm (approved):
Thank you.

PR #1847, opened 2025-03-20
Fix a hypothetical corner case for incremental planning

cachemeifyoucan:
The jobsBeforeCompile that is emitModuleSparately job should also run if the jobsAfterCompile depends on it.


cachemeifyoucan:
@swift-ci please test

owenv:
The git history on 04ca5cc seems to imply this could reintroduce an old issue:

```
[Incremental Builds] Skip `jobsBeforeCompiles` if no compilation jobs are to be run.

Otherwise, for example, when using the `emit-module-separately` mode, the `emit-module` job is *always* mandatory, even on a null build.
```
Any way we can test this to make sure that's not the case?

cachemeifyoucan:
Hmm, I guess we currently not planning it if we don't plan any explicit module dependencies. Ideally, I actually want to remove the verifyEmittedInterface job if nothing is planned. What do you all think?

PR #1848, opened 2025-03-20
Expose a new public property on the driver to retrieve an InterModuleDependencyGraph

owenv:
Swift Build will use this to access header dependencies for clang modules


owenv:
@swift-ci test

artemcm (approved):
Thank you!

PR #1849, opened 2025-03-21
Plan swiftinterface job and verify interface job together

cachemeifyoucan:
Always plan the job producing swiftinterface and verifying
    swiftinterface together. This make sure the swiftinterface exists on
    disk or in CAS when verification job is running.

    rdar://147501517

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -77,7 +77,7 @@ public final class IncrementalCompilationState {
       jobsInPhases.allJobs.first(where: {$0.kind == .generatePCH}),
       &driver)
     self.mandatoryJobsInOrder = firstWave.mandatoryJobsInOrder
-    self.jobsAfterCompiles = jobsInPhases.afterCompiles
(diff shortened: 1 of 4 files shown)


Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -77,7 +77,7 @@ public final class IncrementalCompilationState {
       jobsInPhases.allJobs.first(where: {$0.kind == .generatePCH}),
       &driver)
     self.mandatoryJobsInOrder = firstWave.mandatoryJobsInOrder
-    self.jobsAfterCompiles = jobsInPhases.afterCompiles
+    self.jobsAfterCompiles = firstWave.jobAfterCompile
```
artemcm, line 80: The name is singular `job` but the property is an array. I think it should be `jobs` even though it is usually just going to be one verification task.

artemcm (commented):
I think this broadly makes sense. I can't think of a scenario where we need to wait for the second wave to determine if any kind of post-compile task is needed, but if we find one, we could make that work too.

cachemeifyoucan:
@swift-ci please test

owenv (approved):

artemcm (approved):

PR #1851, opened 2025-03-22
SwiftDriver: handle `-libc` on Windows

compnerd:
Windows has a special flag `-libc` that is responsible for determining the libc variant (ABI flavour) to link against. This flag defaults to `MD` if unspecified. We did not replicate the behaviour from the C++ implementation of the Swift driver to the Swift implementation of the driver which prevented the use of the early swift-driver.


compnerd:
@swift-ci please test

artemcm (approved):

PR #1852, opened 2025-03-23
Added a new `getSingleFrontendInvocationFromDriverArgumentsV4` function.

cwakamo:
This variant adds a `compilerExecutableDir` parameter which gets passed along to the `Driver` during initialization. This allows callers of `getSingleFrontendInvocationFromDriverArgumentsV4` to set a custom `compilerExecutableDir`, which influences paths passed to the frontend for things like the in-process plugin library.

This addresses <rdar://problem/147692019>.


cwakamo:
@swift-ci please test

artemcm (approved):

PR #1853, opened 2025-03-26
[Incremental] Track and return skipped non-compile job

cachemeifyoucan:
This fixes a fallout from #1829 that skipped non-compile jobs are not returned from `planBuild()` during incremental build if there are no compile jobs are being schedule from the first place. This can result into an empty list of job being returned, and `swift-build` will treat that as an error.

Now the skipped non-compile jobs are properly tracked and returned.

rdar://147874145


cachemeifyoucan:
@swift-ci please test

owenv (approved):

artemcm (approved):

PR #1857, opened 2025-03-29
Detect OpenBSD opt-out of BTCFI.

3405691582:
On the swift side, we added a new build flavor in swiftlang/swift#80389 to opt-out of BTCFI as a way of working around swiftlang/swift#80059. We communicate this to swift-driver via the frontend with FrontendTargetInfo.


3405691582:
@swift-ci please test.

DougGregor (approved):

PR #1859, opened 2025-03-31
[SwiftCaching] Tweak warning for disable pch when caching is enabled

cachemeifyoucan:
Tweak the warning message when using -disable-bridging-pch when using together with -cache-compile-job since it can be confusing. The new message should be clear that the automatic resolution from the driver is to ignore -disable-bridging-pch and enable compilation caching.

rdar://148284738


cachemeifyoucan:
@swift-ci please test

artemcm (approved):

PR #1860, opened 2025-04-01
Android: remove build flags that are no longer needed

finagolfin:
Removing the definition of `_GNU_SOURCE` is no longer needed since swiftlang/swift-tools-support-core#500 made sure the right `strerror_r()` is used.


finagolfin:
@swift-ci test

finagolfin:
@artemcm, Android-only change that I built and tested locally, please review.

PR #1863, opened 2025-04-03
Allow linking `_Concurrency` for WASI with Embedded Swift

MaxDesiatov:
Currently, when building packages for WASI with Embedded Swift, libraries such as `libswift_Concurrency.a` and `libswift_ConcurrencyDefaultExecutor.a` are not discoverable and require passing `-Xlinker <swift-sdk-path>/usr/lib/swift/embedded/wasm32-unknown-wasip1` option manually. This path can be inferred by the driver, which simplifies build invocations for users significantly, while the rest of linkage options (`-lswift_Concurrency` etc) can be specified in toolset files.

### DIFF
--- a/Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
@@ -125,10 +125,17 @@ extension WebAssemblyToolchain {
         commandLine.appendPath(path)
       }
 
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@kateinoigakukun friendly ping

kateinoigakukun (approved):
We will need to add the same logic to generic unix toolchain too later but the change itself seems good to me 👍

PR #1866, opened 2025-04-08
[SwiftScanCAS] Make sure that CAS size limitation can take effect

akyrtzi:
Previously the last CAS instance that was closing the database was not the one that was getting the CAS size limitation setting, resulting in the limit not being in effect.

rdar://148836977


akyrtzi:
@swift-ci Please test

cachemeifyoucan (approved):
LGTM as a temporary workaround.

PR #1867, opened 2025-04-08
Remove outdated Windows instructions

jakepetroules:
Custom build instructions for Windows have no longer been necessary since the introduction of swift-toolchain-sqlite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,21 +21,11 @@ On most platforms you can build using:
 $ swift build
 ```
 
-However, on Windows, some additional work must be done by the developer.
-
-Due to the default version of swift-tools-support-core that `Package.resolved` references, we must first update the package dependencies.
-
-```
-swift package update
-```
-
-Then, we can build the package using:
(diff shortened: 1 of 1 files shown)


artemcm (approved):
Thanks.

jakepetroules:
@swift-ci test

compnerd (changes requested):
The sqlite options are no longer needed, however, I do believe that the additional option for linking in `swiftCore` is required as SPM will link with `swiftc` instead of `clang` of `clang++`.

jakepetroules:
> The sqlite options are no longer needed, however, I do believe that the additional option for linking in `swiftCore` is required as SPM will link with `swiftc` instead of `clang` of `clang++`.

Fixed, how does that look now?

Also, I think this linker issue is something we can fix as part of the transition to Swift Build.

compnerd:
@jakepetroules I think you forgot to push the changes?

jakepetroules:
I did, but GitHub was stuck in "processing changes" and now it appears to have lost them... will try again.

jakepetroules:
@swift-ci test

jakepetroules:
@compnerd Does this look good to you now?

compnerd (approved):
Thanks!

PR #1874, opened 2025-04-14
[Explicit Module Builds] Move generation of explicit auto-link flags to a toolchain-specific method

artemcm:
- This way we can adopt a feature of a specific platform's linker separately from the others.
- Adopt `-possible_l*` and `-possible_framework *` flags on Darwin platforms.

### DIFF
--- a/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
+++ b/Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
@@ -114,6 +114,20 @@ public final class DarwinToolchain: Toolchain {
     }
   }
 
+  public func addAutoLinkFlags(for linkLibraries: [LinkLibraryInfo], to commandLine: inout [Job.ArgTemplate]) {
+    for linkLibrary in linkLibraries {
+      if !linkLibrary.isFramework {
+        commandLine.appendFlag(.Xlinker)
+        commandLine.appendFlag("-possible-l\(linkLibrary.linkName)")
+      } else {
+        commandLine.appendFlag(.Xlinker)
+        commandLine.appendFlag("-possible_framework")
+        commandLine.appendFlag(.Xlinker)
+        commandLine.appendFlag(linkLibrary.linkName)
+      }
+    }
+  }
+
   public func defaultSDKPath(_ target: Triple?) throws -> AbsolutePath? {
     let hostIsMacOS: Bool
     #if os(macOS)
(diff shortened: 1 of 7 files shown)


artemcm:
@swift-ci test Linux platform

artemcm:
@swift-ci test

artemcm:

artemcm:

artemcm:
@swift-ci test

artemcm:

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
```diff
@@ -114,6 +114,20 @@ public final class DarwinToolchain: Toolchain {
     }
   }
 
+  public func addAutoLinkFlags(for linkLibraries: [LinkLibraryInfo], to commandLine: inout [Job.ArgTemplate]) {
+    for linkLibrary in linkLibraries {
+      if !linkLibrary.isFramework {
+        commandLine.appendFlag(.Xlinker)
+        commandLine.appendFlag("-possible-l\(linkLibrary.linkName)")
```
cachemeifyoucan, line 121: Do we have a rule for what is the linker compatibility from swift-driver? This is quite a new option that doesn't work with older linker and doesn't work with `ld-classic`.

artemcm: We do not, but since this entire code-path is currently behind an opt-in flag (`-explicit-auto-linking`), my thinking is to defer to the client to figure that out, at the moment. We have some precedent in `swift-build` of enabling compiler features based on the current system linker version.

On Darwin, with older linkers which do not support this option, this feature does not really make much sense anyway since using the non-`-possible` flags changes the semantics too much compared to the current auto-linking mechanism (load directives in object files).

owenv (approved):

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
cachemeifyoucan: I am fine with defer to swift-build to figure out but if you use the option from command-line, there is no link that the error message of unknown linker flag has anything to do with `-explicit-auto-linking`?

artemcm: No, there's not really a link there. I'm not sure I see an obvious way to establish such a link, especially given that we use `clang` as the linker driver. Maybe we could emit a warning when the user enables this feature, but that's got its own obvious downsides. And I'm not sure it's worth it, given that this is a fully opt-in behavior.

cachemeifyoucan (approved):

Sources/SwiftDriver/Toolchains/DarwinToolchain.swift
cachemeifyoucan: I don't have better idea either.

PR #1875, opened 2025-04-15
[Jobs] Switch `EmitSupportedFeaturesJob` to use `-emit-supported-argu…

xedin:
…ments` option

This is a driver side change for https://github.com/swiftlang/swift/pull/80833


xedin:
https://github.com/swiftlang/swift/pull/80833
@swift-ci please test

Sources/SwiftDriver/Jobs/EmitSupportedFeaturesJob.swift
```diff
@@ -31,7 +30,7 @@ extension Toolchain {
     var commandLine: [Job.ArgTemplate] = swiftCompilerPrefixArgs.map { Job.ArgTemplate.flag($0) }
     var inputs: [TypedVirtualPath] = []
     commandLine.append(contentsOf: [.flag("-frontend"),
-                                    .flag("-emit-supported-features")])
+                                    .flag("-emit-supported-arguments")])
```
artemcm, line 33: We are in a place where, unfortunately, this code needs to be compatible with compilers that do not yet have https://github.com/swiftlang/swift/pull/80833, so we must maintain compatibility with both. So we need something like:
```
if isFrontendArgSupported(.emitSupportedArguments) {
  commandLine.append(.flag("-emit-supported-arguments"))
} else {
  commandLine.append(.flag("-emit-supported-features"))
}

artemcm:
> [swiftlang/swift#80833](https://github.com/swiftlang/swift/pull/80833) @swift-ci please test

There is no cross-repository testing in this direction, so in order to test this in CI, https://github.com/swiftlang/swift/pull/80833 needs to get merged and then make it into a nightly snapshot (which is what swift-driver CI uses).

Sources/SwiftDriver/Jobs/EmitSupportedFeaturesJob.swift
xedin: Should I leave it as `-features` for now then?

xedin: Btw, I don't think that we can actually do that since this call is the source of "supported arguments" :)

xedin:
@swift-ci please test

xedin:

xedin:
Closing this for now since we need maintain backward compatibility.

PR #1879, opened 2025-04-22
Do not add -sysroot flag when ANDROID_NDK_ROOT environment is set

marcprux:
When the environment `ANDROID_NDK_ROOT` is set, the driver manually adds it as a `--sysroot` flag. This breaks building with an Android SDK when this variable is set, with confusing errors. I first saw this at https://github.com/finagolfin/swift-android-sdk/issues/207#issuecomment-2718259494, and more recently ran into it again (https://github.com/swift-android-sdk/swift-docker/actions/runs/14584558227/job/40907674425) while working on the official Android SDK build script, where it fails a build with:

```
<unknown>:0: error: missing required module 'SwiftAndroid'
```

In both cases, manually un-setting `ANDROID_NDK_ROOT` (which is set by default in GitHub workflows) resolved the issue.

This flag has some history (https://github.com/swiftlang/swift-driver/pull/1560, https://github.com/swiftlang/swift-driver/pull/1681, and https://github.com/swiftlang/swift-driver/pull/1811), but I feel like it should just be excised altogether since the Android SDK can specify its own `sdkRootPath` in the `swift-sdk.json` file. If Windows needs it due to lack of Swift SDK support, then it could be gated inside an `#if os(Windows)` check.

And BTW, the `#if arch(x86_64)` check is incorrect: the Android NDK works on both Intel and ARM macOS machines, despite being stored under "darwin-x86_64", as the binaries are universal:

```
zap ~ % file $ANDROID_HOME/ndk/27.0.12077973/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang
~/Library/Android/sdk/ndk/27.0.12077973/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64]
~/Library/Android/sdk/ndk/27.0.12077973/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang (for architecture x86_64):	Mach-O 64-bit executable x86_64
~/Library/Android/sdk/ndk/27.0.12077973/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang (for architecture arm64):	Mach-O 64-bit executable arm64
```

### DIFF
--- a/Sources/SwiftDriver/CMakeLists.txt
+++ b/Sources/SwiftDriver/CMakeLists.txt
@@ -117,6 +117,7 @@ add_library(SwiftDriver
   Utilities/FileMetadata.swift
   Utilities/FileType.swift
   Utilities/PredictableRandomNumberGenerator.swift
+  Utilities/PythonArchitecture.swift
   Utilities/RelativePathAdditions.swift
   Utilities/Sanitizer.swift
   Utilities/StringAdditions.swift
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -353,6 +353,9 @@ public struct Driver {
   /// Original ObjC Header passed from command-line
   let originalObjCHeaderFile: VirtualPath.Handle?
 
+  /// Whether to import the bridging header as internal (vs public).
+  let importBridgingHeaderAsInternal: Bool
+
   /// Enable bridging header chaining.
   let bridgingHeaderChaining: Bool
 
@@ -374,6 +377,9 @@ public struct Driver {
   /// Path to the serialized diagnostics file of the emit-module task.
   let emitModuleSerializedDiagnosticsFilePath: VirtualPath.Handle?
 
+  /// Path to the serialized diagnostics file of the dependency scanning task.
+  let dependencyScanSerializedDiagnosticsPath: VirtualPath.Handle?
+
   /// Path to the discovered dependencies file of the emit-module task.
   let emitModuleDependenciesFilePath: VirtualPath.Handle?
 
@@ -382,6 +388,12 @@ public struct Driver {
 
   /// Path to the Objective-C generated header.
   let objcGeneratedHeaderPath: VirtualPath.Handle?
+
+  /// Path to the SIL output file.
+  let silOutputPath: VirtualPath.Handle?
+
+  /// Path to the LLVM IR output file.
+  let llvmIROutputPath: VirtualPath.Handle?
 
   /// Path to the loaded module trace file.
   let loadedModuleTracePath: VirtualPath.Handle?
@@ -875,9 +887,23 @@ public struct Driver {
 
     self.executor = executor
 
+    if args.count > 1 && args[1] == "-repl" {
+      #if arch(arm64)
+        checkIfMatchingPythonArch(
+          cwd: ProcessEnv.cwd, envBlock: envBlock, toolchainArchitecture: .arm64, diagnosticsEngine: diagnosticsEngine)
+      #elseif arch(x86_64)
+        checkIfMatchingPythonArch(
+          cwd: ProcessEnv.cwd, envBlock: envBlock, toolchainArchitecture: .x64, diagnosticsEngine: diagnosticsEngine)
+      #elseif arch(x86)
+        checkIfMatchingPythonArch(
+          cwd: ProcessEnv.cwd, envBlock: envBlock, toolchainArchitecture: .x86, diagnosticsEngine: diagnosticsEngine)
+      #endif
+    }
+
     if case .subcommand = try Self.invocationRunMode(forArgs: args).mode {
       throw Error.subcommandPassedToDriver
     }
+
     var args = args
     if let additional = env["ADDITIONAL_SWIFT_DRIVER_FLAGS"] {
       args.append(contentsOf: additional.components(separatedBy: " "))
@@ -1148,10 +1174,12 @@ public struct Driver {
     }
     self.producePCHJob = maybeNeedPCH
 
-    if let objcHeaderPathArg = parsedOptions.getLastArgument(.importObjcHeader) {
-      self.originalObjCHeaderFile = try? VirtualPath.intern(path: objcHeaderPathArg.asSingle)
+    if let importBridgingHeaderOption = parsedOptions.last(for: .importBridgingHeader, .internalImportBridgingHeader) {
+      self.originalObjCHeaderFile = try? VirtualPath.intern(path: importBridgingHeaderOption.argument.asSingle)
+      self.importBridgingHeaderAsInternal = importBridgingHeaderOption.option == .internalImportBridgingHeader
     } else {
       self.originalObjCHeaderFile = nil
+      self.importBridgingHeaderAsInternal = false
     }
 
     if parsedOptions.hasFlag(positive: .autoBridgingHeaderChaining,
@@ -1241,6 +1269,14 @@ public struct Driver {
         emitModuleSeparately: emitModuleSeparately,
         outputFileMap: self.outputFileMap,
         moduleName: moduleOutputInfo.name)
+    self.dependencyScanSerializedDiagnosticsPath = try Self.computeSupplementaryOutputPath(
+        &parsedOptions, type: .dependencyScanDiagnostics, isOutputOptions: [.serializeDiagnostics],
+        outputPath: .dependencyScanSerializeDiagnosticsPath,
+        compilerOutputType: compilerOutputType,
+        compilerMode: compilerMode,
+        emitModuleSeparately: emitModuleSeparately,
+        outputFileMap: self.outputFileMap,
+        moduleName: moduleOutputInfo.name)
     self.emitModuleDependenciesFilePath = try Self.computeSupplementaryOutputPath(
         &parsedOptions, type: .emitModuleDependencies, isOutputOptions: [.emitDependencies],
         outputPath: .emitModuleDependenciesPath,
@@ -1266,6 +1302,11 @@ public struct Driver {
         emitModuleSeparately: emitModuleSeparately,
         outputFileMap: self.outputFileMap,
         moduleName: moduleOutputInfo.name)
+    // SIL and IR outputs are handled through directory options and file maps
+    // rather than single output paths, so we set these to nil to enable
+    // the supplementary output path logic in FrontendJobHelpers
+    self.silOutputPath = nil
+    self.llvmIROutputPath = nil
 
     if let loadedModuleTraceEnvVar = env["SWIFT_LOADED_MODULE_TRACE_FILE"] {
       self.loadedModuleTracePath = try VirtualPath.intern(path: loadedModuleTraceEnvVar)
@@ -2034,8 +2075,8 @@ extension Driver {
 
     // Put bridging header as first input if we have it
     let allInputs: [TypedVirtualPath]
-    if let objcHeader = originalObjCHeaderFile {
-      allInputs = [TypedVirtualPath(file: objcHeader, type: .objcHeader)] + inputFiles
+    if let bridgingHeader = originalObjCHeaderFile {
+      allInputs = [TypedVirtualPath(file: bridgingHeader, type: .objcHeader)] + inputFiles
     } else {
       allInputs = inputFiles
     }
@@ -2751,7 +2792,9 @@ extension Driver {
     }
 
     // Check that we're one of the known supported targets for sanitizers.
-    if !(targetTriple.isWindows || targetTriple.isDarwin || targetTriple.os == .linux || targetTriple.os == .wasi) {
+    if !(targetTriple.isWindows ||
+      targetTriple.isDarwin ||
+      [.freeBSD, .linux, .wasi].contains(targetTriple.os)) {
       diagnosticEngine.emit(
         .error_unsupported_opt_for_target(
           arg: "-sanitize=",
@@ -3837,6 +3880,18 @@ extension Driver {
                   .intern()
     }
     return try VirtualPath.intern(path: moduleName.appendingFileTypeExtension(type))
+  }
+
+  static func hasFileMapEntry(outputFileMap: OutputFileMap?, fileType: FileType) -> Bool {
+    guard let outputFileMap = outputFileMap else { return false }
+
+    // Check if any input has this file type in the output file map
+    for inputFile in outputFileMap.entries.keys {
+      if outputFileMap.entries[inputFile]?[fileType] != nil {
+        return true
+      }
+    }
+    return false
   }
 
   /// Determine if the build system has created a Project/ directory for auxiliary outputs.
--- a/Sources/SwiftDriver/Driver/OutputFileMap.swift
+++ b/Sources/SwiftDriver/Driver/OutputFileMap.swift
@@ -172,6 +172,11 @@ public struct OutputFileMap: Hashable, Codable {
     }
     return result
   }
+
+  /// Check if the output file map has any entries for the given file type
+  public func hasEntries(for fileType: FileType) -> Bool {
+    return entries.values.contains { $0[fileType] != nil }
+  }
 }
 
 /// Struct for loading the JSON file from disk.
--- a/Sources/SwiftDriver/Execution/ArgsResolver.swift
+++ b/Sources/SwiftDriver/Execution/ArgsResolver.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import class Foundation.NSLock
+import struct Foundation.UUID
 
 import func TSCBasic.withTemporaryDirectory
 import protocol TSCBasic.FileSystem
@@ -23,6 +24,11 @@ public enum ResponseFileHandling {
   case forced
   case disabled
   case heuristic
+}
+
+public enum ResolvedCommandLine {
+  case plain([String])
+  case usingResponseFile(resolved: [String], responseFileContents: [String])
 }
 
 /// Resolver for a job's argument template.
@@ -73,6 +79,18 @@ public final class ArgsResolver {
     let usingResponseFile = try createResponseFileIfNeeded(for: job, resolvedArguments: &arguments,
                                                            useResponseFiles: useResponseFiles)
     return (arguments, usingResponseFile)
+  }
+
+  public func resolveArgumentList(for job: Job, useResponseFiles: ResponseFileHandling = .heuristic)
+  throws -> ResolvedCommandLine {
+    let tool = try resolve(.path(job.tool))
+    let resolvedArguments = [tool] + (try resolveArgumentList(for: job.commandLine))
+    var actualArguments = resolvedArguments
+    let usingResponseFile = try createResponseFileIfNeeded(for: job, resolvedArguments: &actualArguments,
+                                                           useResponseFiles: useResponseFiles)
+    return usingResponseFile ? .usingResponseFile(resolved: actualArguments,
+                                                  responseFileContents: resolvedArguments)
+                             : .plain(actualArguments)
   }
 
   public func resolveArgumentList(for commandLine: [Job.ArgTemplate]) throws -> [String] {
@@ -189,14 +207,13 @@ public final class ArgsResolver {
       (job.supportsResponseFiles && !commandLineFitsWithinSystemLimits(path: resolvedArguments[0], args: resolvedArguments)) {
       assert(!forceResponseFiles || job.supportsResponseFiles,
              "Platform does not support response files for job: \(job)")
-      // Match the integrated driver's behavior, which uses response file names of the form "arguments-[0-9a-zA-Z].resp".
-      let hash = SHA256().hash(resolvedArguments.joined(separator: " ")).hexadecimalRepresentation
-      let responseFilePath = temporaryDirectory.appending(component: "arguments-\(hash).resp")
+      let uuid = UUID().uuidString
+      let responseFilePath = temporaryDirectory.appending(component: "arguments-\(uuid).resp")
 
       // FIXME: Need a way to support this for distributed build systems...
       if let absPath = responseFilePath.absolutePath {
         let argumentBytes = ByteString(resolvedArguments[2...].map { $0.spm_shellEscaped() }.joined(separator: "\n").utf8)
-        try fileSystem.writeFileContents(absPath, bytes: argumentBytes, atomically: true)
+        try fileSystem.writeFileContents(absPath, bytes: argumentBytes)
         resolvedArguments = [resolvedArguments[0], resolvedArguments[1], "@\(absPath.pathString)"]
       }
 
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
@@ -193,6 +193,11 @@ public extension Driver {
       commandLine.appendFlag(.resolvedPluginVerification)
     }
 
+    if let depScanSerializedDiagnosticsPath = dependencyScanSerializedDiagnosticsPath {
+      commandLine.appendFlag("-serialize-diagnostics-path")
+      commandLine.appendPath(VirtualPath.lookup(depScanSerializedDiagnosticsPath))
+    }
+
     // Pass on the input files
     commandLine.append(contentsOf: inputFiles.filter { $0.type == .swift }.map { .path($0.file) })
     return (inputs, commandLine)
--- a/Sources/SwiftDriver/Jobs/CompileJob.swift
+++ b/Sources/SwiftDriver/Jobs/CompileJob.swift
@@ -97,7 +97,7 @@ extension Driver {
          .moduleTrace, .yamlOptimizationRecord, .bitstreamOptimizationRecord, .pcm, .pch,
          .clangModuleMap, .jsonCompilerFeatures, .jsonTargetInfo, .jsonSwiftArtifacts,
          .indexUnitOutputPath, .modDepCache, .jsonAPIBaseline, .jsonABIBaseline,
-         .swiftConstValues, .jsonAPIDescriptor, .moduleSummary, .moduleSemanticInfo,
+         .swiftConstValues, .jsonAPIDescriptor, .moduleSummary, .moduleSemanticInfo, .dependencyScanDiagnostics,
          .cachedDiagnostics, .jsonSupportedFeatures, nil:
       return false
     }
@@ -107,7 +107,7 @@ extension Driver {
   /// corresponding primary set of outputs and, if not identical, the output
   /// paths to record in the index data (empty otherwise).
   mutating func addCompileInputs(primaryInputs: [TypedVirtualPath],
-                                 indexFilePath: TypedVirtualPath?,
+                                 indexFilePaths: [TypedVirtualPath],
                                  inputs: inout [TypedVirtualPath],
                                  inputOutputMap: inout [TypedVirtualPath: [TypedVirtualPath]],
                                  outputType: FileType?,
@@ -149,11 +149,11 @@ extension Driver {
       assert(!primaryInputs.isEmpty)
       usesPrimaryFileInputs = true
       primaryInputFiles = primaryInputs
-    } else if let path = indexFilePath {
+    } else if !indexFilePaths.isEmpty {
       // If -index-file is used, we perform a single compile but pass the
       // -index-file-path as a primary input file.
       usesPrimaryFileInputs = true
-      primaryInputFiles = [path]
+      primaryInputFiles = indexFilePaths
     } else {
       usesPrimaryFileInputs = false
       primaryInputFiles = []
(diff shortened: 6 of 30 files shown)


marcprux:
I've marked this ready for review, @jakepetroules, @finagolfin, and @compnerd.

See also the discussion at https://github.com/swiftlang/swift-build/issues/495#issuecomment-2878480135

finagolfin:
I agree with getting rid of this. As much as the idea is to automatically set things up and make things automatically work, I think it's better to have the user configure this themselves. For example, this env variable is automatically set on the github CI, so even if the user explicitly configures the `sdkRootPath` to some other NDK, that will be overridden by this environment variable.

Better to not have such magic and make the setup more explicit, as it would be after this pull.

compnerd:
This is going to break all of our builds, please do not commit this as is. We need the ability to be able to pass the `-sysroot` through. I don't think that this is the right approach.

marcprux:
> This is going to break all of our builds

Then how about gating it within an `#if os(Windows)` check? This issue a blocker for Android builds on macOS and Linux, so we need some solution.

marcprux:
@compnerd Any thoughts on either @jakepetroules's questions or my suggestion? I'd really like to reach some solution so we can un-block the Android SDK work.

finagolfin:
@compnerd or @weliveindetail, could you apply this change to your CI and let us know if it causes any error and what the full output is? Because that may turn up an underlying bug that is better fixed rather than papering over it with this Android default.

marcprux:
This hasn't made it into the swift-6.2 brach, has it? I'm still seeing the build failures when I have `ANDROID_NDK_ROOT` set, which is especially problematic because the developer _needs_ to set `ANDROID_NDK_HOME` to run the [SDK setup script](https://github.com/swiftlang/swift-docker/pull/467#issuecomment-2864149387) initially, but needs to un-set `ANDROID_NDK_ROOT`. These are both set to the same value in GitHub actions.

@finagolfin Can you add this to the [project board](https://github.com/orgs/swiftlang/projects/17)?

jakepetroules:
Can we get this updated and landed? @compnerd do you still have concerns?

compnerd:
I think that we should wait until @finagolfin's frontend/driver changes are done and then re-test before merging this. The `--sysroot` flag is needed to allow us to pull content from the NDK without copying it into the Swift SDK, so I would prefer that we do not merge this change.

finagolfin:
My frontend change in swiftlang/swift#79621 only affects `-sdk` usage, does not affect this `-sysroot` usage at all.

It is unclear why this env variable detection and `-sysroot` flag insertion is so important to your builds, as you could just detect it manually in your build scripts and add it in your config yourself. This is really just a convenience method for swift-driver to do that for you, but the problem is that it breaks other build environments that have multiple NDKs and don't use the env variable but manually pass in the NDK path.

Is there some ordering issue or something, on the TBC CI builds with this pull? If you guys would simply try this patch and replace it with manually passing in this flag to your build config, we could determine if this was really necessary to you or not, from the resulting error output, if any.

finagolfin:
@marcprux, mind rebasing?

finagolfin:
@swift-ci test

finagolfin:
@swift-ci test windows

finagolfin:
@Steelskin, can you apply this pull to your Windows CI, along with passing in this `-sysroot ANDROID_NDK_ROOT` flag manually through wherever your Android build config is, similar to how [the Powershell script `build.ps1` is already doing this](https://github.com/swiftlang/swift/blob/e1d7c70545e3539b668dcd5cb13fc6ca4b0c62d4/utils/build.ps1#L1602), and report any build errors that you see?

The problem with this default is that it forces everyone to use the NDK at `ANDROID_NDK_ROOT` or have to unset that variable each time, when many have alternate NDKs they want to use through `swift sdk configure` instead, swiftlang/swift-package-manager#8856.

Steelskin:
> @Steelskin, can you apply this pull to your Windows CI, along with passing in this `-sysroot ANDROID_NDK_ROOT` flag manually through wherever your Android build config is, similar to how [the Powershell script `build.ps1` is already doing this](https://github.com/swiftlang/swift/blob/e1d7c70545e3539b668dcd5cb13fc6ca4b0c62d4/utils/build.ps1#L1602), and report any build errors that you see?
> 
> The problem with this default is that it forces everyone to use the NDK at `ANDROID_NDK_ROOT` or have to unset that variable each time, when many have alternate NDKs they want to use through `swift sdk configure` instead, [swiftlang/swift-package-manager#8856](https://github.com/swiftlang/swift-package-manager/pull/8856).

I am in the process of changing our build to more closely match what is being done in `buid.ps1`, but this is still a work in progress. In the meantime, it is going to be a bit challenging to get this to run so bear with me.
We run our smoke android test with the following:

```
          swift build `
            --package-path ${{ github.workspace }}/SourceCache/cassowary `
            --triple ${{ matrix.arch }}-unknown-linux-android${{ inputs.ANDROID_API_LEVEL }} `
            --sdk "${{ steps.android-swift-env.outputs.sysroot }}" `
            -Xswiftc -sdk -Xswiftc "${{ steps.android-swift-env.outputs.sdkroot }}" `
            -Xswiftc -sysroot -Xswiftc "${{ steps.android-swift-env.outputs.sysroot }}" `
            -Xswiftc -I -Xswiftc "${{ steps.android-swift-env.outputs.sdkroot }}\usr\include" `
            -Xswiftc -Xclang-linker -Xswiftc -resource-dir -Xswiftc -Xclang-linker -Xswiftc ${{ steps.android-swift-env.outputs.clang-resource-dir }}
```

Does that sound about correct to you or would you like me to edit some of the flags and/or the environment here?

finagolfin:
Are you sure you need both `--sdk sdk-path` and `-Xswiftc -sdk -Xswiftc sdk-path`? The latter seems redundant, but perhaps there is a bug I'm unaware of.

Other than that, I'd like you to run these flags you've pasted but using a swift-driver with this pull of Marc's applied: see if removing this `ANDROID_NDK_ROOT` detection causes any problem on your CI when you're passing it in with `-Xswiftc -sysroot -Xswiftc ndk-path` instead.

Steelskin:
> Are you sure you need both `--sdk sdk-path` and `-Xswiftc -sdk -Xswiftc sdk-path`? The latter seems redundant, but perhaps there is a bug I'm unaware of.

Not sure, this is how the build is configured. I can toy with what arguments we actually need at a later time.

> Other than that, I'd like you to run these flags you've pasted but using a swift-driver with this pull of Marc's applied: see if removing this `ANDROID_NDK_ROOT` detection causes any problem on your CI when you're passing it in with `-Xswiftc -sysroot -Xswiftc ndk-path` instead.

Alright, got it. All I did was change the `swift-driver` repo to point to `swift-everywhere:main`. The build is ongoing, I'll let you know how it goes.

Steelskin:
@finagolfin Our CI succeeded with the current `swift-everywhere:main` branch.

finagolfin:
@compnerd, if you would approve or let us know if you need to test anything else, I'd like to get this in.

finagolfin:
Pinging @compnerd again, would like to get this into 6.2 next, as [I have had to work around this override on my Android CI](https://github.com/finagolfin/swift-android-sdk/commit/419014771d1ad90fd671d7eb98901d0f9cfc0081#diff-76992bc322b65ab94f299c05ba3cbd8b3ab2e32ebfdde6db5adba70fa9f5a26aR298) and so have others. If you have no further concerns, please approve and we'll get this in.

compnerd:
I still have concerns around this - the user needs to pass in `-sysroot` manually (which is what @Steelskin pointed out). As long as we that I can restore this change if the user needs to specify the path, I am okay with merging. The point is that the user shouldn't have to provide the paths. `ANDROID_NDK_PATH` is a variable that the NDK sets. Using that to locate the NDK unless the user overrides it explicitly seems reasonable.

finagolfin:
> Using that to locate the NDK unless the user overrides it explicitly seems reasonable.

The problem is that in general the `ANDROID_NDK_ROOT` variable is not set, so most of the time this adds nothing for the user. However, it is set in github CI and some other environments, but this pull passed the new `-sysroot` flag that almost nobody is using yet, plus `--sysroot` for the linker, so on such CI it depends if the user wants the default NDK using the new `-sysroot` flag. Since SwiftPM doesn't really support using such external C/C++ SDKs yet with the bundle feature we're pushing now, swiftlang/swift-package-manager#8958, almost nobody is using such a setup.

I understand that you guys at TBC got it to work, but because most people won't, I think it's best if you pass in the `-sysroot` manually for now. As we get more experience with this flag in the coming year, we can revisit this. What do you think?

marcprux:
> The problem is that in general the `ANDROID_NDK_ROOT` variable is not set, so most of the time this adds nothing for the user.

This is especially true since `ANDROID_NDK_ROOT` is not the currently recommended environment variable to point to the NDK installation. It should be `ANDROID_NDK_HOME` (see https://github.com/android/ndk-samples/wiki and https://developer.android.com/ndk/guides/graphics/getting-started#creating).

This is the default variable that will be set by environments like [Homebrew](https://github.com/Homebrew/homebrew-cask/blob/main/Casks/a/android-ndk.rb#L45), while others (like [Github runners](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#environment-variables-2)) set both for good measure.

compnerd:
I would be okay with changing the behaviour to refer to `ANDROID_NDK_HOME` rather than `ANDROID_NDK_ROOT`. I wonder how that got mixed up.

BTW, I don't think that we should be passing in the other flags - that sounds like something we should clean up. The idea is that with `ANDROID_NDK_ROOT` (well, `ANDROID_NDK_HOME`) set, all the appropriate flags should be computed in the driver.

finagolfin:
> I don't think that we should be passing in the other flags - that sounds like something we should clean up.

Which other flags are you referring to here?

> The idea is that with ANDROID_NDK_ROOT (well, ANDROID_NDK_HOME) set, all the appropriate flags should be computed in the driver.

If by that you mean `-sysroot`, which this pull removes, that flag does not work well yet, so setting it as the default is more likely to break builds.

finagolfin:
@compnerd, now that this is not needed for your TBC CI, can we go ahead and get it in? We all agree that we should simplify these Android flags much more, but we are still a fair ways away from that:

1. Even with this env variable checked, your Windows cross-compilation for Android currently requires a ton of flags:
```
swift build `
            --package-path ${{ github.workspace }}/SourceCache/cassowary `
            --triple ${{ matrix.arch }}-unknown-linux-android${{ inputs.ANDROID_API_LEVEL }} `
            --sdk "${{ steps.android-swift-env.outputs.sysroot }}" `
            -Xswiftc -sdk -Xswiftc "${{ steps.android-swift-env.outputs.sdkroot }}" `
            -Xswiftc -sysroot -Xswiftc "${{ steps.android-swift-env.outputs.sysroot }}" `
            -Xswiftc -I -Xswiftc "${{ steps.android-swift-env.outputs.sdkroot }}\usr\include" `
            -Xswiftc -Xclang-linker -Xswiftc -resource-dir -Xswiftc -Xclang-linker -Xswiftc ${{ steps.android-swift-env.outputs.clang-resource-dir }}
```
2. This env variable checking, that Marc reverts in this pull, currently breaks cross-compilation to Android on all non-TBC CI.

The good news is that SDK bundles already abstract away most of these flags for users, though I know those don't work on Windows or for user-installed platform C/C++ SDKs, like the Android NDK, yet.

Let's just fix this breakage for now, and we can whittle away those extra Android flags over time.

jakepetroules:
Reading through this in more detail to try to ensure I understand correctly...

There are two flags relevant for the Swift compiler to build for Android:
- `-sdk`: path to the Swift SDK (Swift-specific bits)
- `--sysroot`: path to the Android NDK sysroot

Are those flags working in the driver today and is SwiftPM passing them both?

> For example, this env variable is automatically set on the github CI, so even if the user explicitly configures the sdkRootPath to some other NDK, that will be overridden by this environment variable.

I am confused about this part. Why would that be the case? `sdkRootPath` maps to `--sysroot`, which when passed will override any environment that is set, right?

finagolfin:
> There are two flags relevant for the Swift compiler to build for Android:
-sdk: path to the Swift SDK (Swift-specific bits)
--sysroot: path to the Android NDK sysroot
Are those flags working in the driver today and is SwiftPM passing them both?

Close, historically, the two flags used to cross-compile are `-sdk`, for the platform C/C++ sysroot, and `-resource-dir`, for the Swift runtime modules and libraries. Recently, the `-sysroot` flag was added and [switching to `-sdk`/`-sysroot` over time was proposed](https://github.com/swiftlang/swift/blob/main/docs/CrossCompilationModel.md), but that largely has not been implemented yet.

This pull reverts the env variable checking because it inserts that new `-sysroot` flag regardless of which pair of cross-compilation flags above are being used, breaking most current Android cross-compilation with the `-sdk`/`-resource-dir` flags.

You'll notice that the long TBC CI command above is an exception that does not pass the `-resource-dir` flag to the Swift compiler, though it does pass in the different `-Xclang-linker -resource-dir -Xclang-linker ${{ steps.android-swift-env.outputs.clang-resource-dir }}` flag to clang. They apparently get away with this because they don't use Swift SDK bundles on Windows, where bundles don't work yet, and by manually passing in all those extra flags.

> I am confused about this part. Why would that be the case? sdkRootPath maps to --sysroot, which when passed will override any environment that is set, right?

`sdkRootPath` in a Swift bundle is used to set the `-sdk` flag when compiling. It is also separately passed with `--sysroot` to the Swift clangImporter, but the bigger issue is that you are now passing in two different platform C/C++ sysroots, one with `-sdk` and another with `-sysroot` in the code this pull changes, which causes conflicts.

It is not worth getting into all the details, but the basic point is the meaning of `-sdk` changes, depending on whether it is used with the current standard of `-resource-dir` or the proposed new `-sysroot`. In the former, `-sdk` is the platform C/C++ SDK, whereas in the latter, it is the path to the Swift runtime libraries. You cannot simply pass in `-sysroot` no matter what, which is why this revert is needed.

jakepetroules:
Thanks for the explanation, this is making things a little clearer. Can you also clarify whether you're building via SwiftPM, or building with a different build system?

If `sdkRootPath` is supposed to correspond to the Swift SDK bits and not the platform sysroot, is SwiftPM doing the right thing today? It looks like `-sdk` is being passed for Swift, but this one for the C compiler needs correction.

```swift
if let sdkDir = swiftSDK.pathsConfiguration.sdkRootPath {
    let sysrootFlags = [triple.isDarwin() ? "-isysroot" : "--sysroot", sdkDir.pathString]
    self.extraFlags.cCompilerFlags.insert(contentsOf: sysrootFlags, at: 0)
}
```

> but that largely has not been implemented yet.

What's still missing?

> This pull reverts the env variable checking because it inserts that new -sysroot flag regardless of which pair of cross-compilation flags above are being used, breaking most current Android cross-compilation with the -sdk/-resource-dir flags.

Could this have some smarter detection? e.g. don't add the default `--sysroot` flag if `-sdk` is set and points to something that looks like an NDK sysroot?

finagolfin:
> Can you also clarify whether you're building via SwiftPM, or building with a different build system?

The build system is irrelevant, as these are Swift compiler flags that any build system has to use. I merely focused on SwiftPM because it is what most use, but these Swift compiler flags used for cross-compilation are the same regardless of build system, whether CMake or Bazel.

> If sdkRootPath is supposed to correspond to the Swift SDK bits and not the platform sysroot, is SwiftPM doing the right thing today?

OK, I can see you are confused, so let me get into the slightly longer expanation. Until recently, there were supposed to be two options to cross-compile using the `-sdk` flag. You could specify "`-sdk`, for the platform C/C++ sysroot" alone, contrary to what you just wrote, and "`-resource-dir`, for the Swift runtime modules and libraries." If you happened to have a non-Darwin C/C++ sysroot that _also_ contained the Swift runtime in `<sdk>/usr/lib/swift/`, you could specify `-sdk` alone, but [only with the legacy C++ Driver](https://github.com/swiftlang/swift/blob/b7b5a2a19dd4e5ca9a4c16c5684f2eb4b7f76725/lib/Driver/ToolChains.cpp#L1584): that was never brought over to this new `swift-driver` (I'm working on a pull to bring that capability over to the `swift-frontend`, which this `swift-driver` invokes to find such flags, swiftlang/swift#79621).

Well, when the `-sdk`/`-sysroot` switch was proposed in the cross-compilation document I linked above, my guess is that its authors thought they could reuse that capability to look in `-sdk` for the Swift runtime libraries instead, and split off the platform C/C++ sysroot into the new `-sysroot` flag. However, they were likely unaware that this new `swift-driver` does not look in `-sdk` for the Swift runtime libraries, and their effort to switch to `-sdk`/`-sysroot` appears to have collapsed.

The SwiftPM flag you pasted is correct under the `-sdk`/`-resource-dir` cross-compilation model almost everybody is now using, but you are right that it may cause problems if we ever made the switch to the `-sdk`/`-sysroot` model that almost nobody is now using and is mostly unimplemented.

The only reason the TBC CI is able to use `-sdk`/`-sysroot` above is because either Swift cross-compilation flag pair is then used to automatically set a bunch of `-I`/`-L`/`--sysroot` flags internal to the Swift compiler, so they replace that with a bunch of manually passed-in flags like those shown above. That is brittle and tends to break though, better to choose a pair of front-facing cross-compilation flags and keep it simple.

> What's still missing?

The main one is actually having this `swift-driver` look in `-sdk` for the Swift runtime libraries! After a year of my pointing out that that basic capability doesn't work to those who wanted to switch to `-sdk`/`-sysroot` and not getting a response until recently, I finally stepped up to implement it myself. Another is that `-sysroot` still has bugs, since almost nobody is using it.

> Could this have some smarter detection? e.g. don't add the default --sysroot flag if -sdk is set and points to something that looks like an NDK sysroot?

Given all the `-sdk`/`-sysroot` problems detailed above, best to just take it out for now, as Marc does in this pull. Another issue is that there's no guarantee `ANDROID_NDK_ROOT` is the NDK that users want, so we'd need to have a story in place of how to override it.

A final aspect that I don't like is that we have an explicit flag for this in SwiftPM, better to use that than these magic environment variables. You think you're making life easier for the user, but you're really just introducing hidden variables that can break things in hard to debug ways.

Let me finally note that I'm not against the `-sdk`/`-sysroot` switch, but I do have doubts that it can be made to work and most importantly, we shouldn't be pushing in this `-sysroot` flag for Android alone until the underlying `-sdk`/`-sysroot` support is fully implemented.

finagolfin:
Ping @compnerd or @jakepetroules, can we go ahead and get this in? The behavior this pull reverts is clearly not needed on Windows, where it can always be passed in manually, and as detailed above, it breaks most Android builds when used with SDK bundles, which only support `-sdk`/`-resource-dir` and not the `-sysroot` flag it adds.

This pull gets Android building again on CI environments which set this environment variable, where Marc and I have had to unset this variable to avoid this broken behavior and warn others to do the same.

Since `-sdk`/`-sysroot` cross-compilation currently doesn't work well, with even the Windows build above for Android having to add several flags to make up for those problems, it is best to not jam in the `-sysroot` flag like this. Once we get it working, and I'm still working on that `swift-frontend` pull to help get it there, we can revisit adding something like this later on down the line.

artemcm:
I chatted with @jakepetroules, @owenv, @compnerd and @etcwilde and have a proposal. 

One key thing we’d like to do is to better delineate is the use of the various discussed flags here. The ones we’re generally concerned with are:
* `-sysroot`: Native system root on platforms that have one. 
* `-sdk`: Swift SDK root
* `-resource-dir`: Compiler toolchain-specific contents and headers. In the case of Swift, these are header files used by the ClangImporter, SwiftShims, etc. It is our expectation that this should generally never be specified by the user. 

One some platforms, sys root and SDK root are the same (read Apple platforms), in which case no explicit `-sysroot` is necessary, only `-sdk` is needed. On Windows there is not really a concept of a sysroot, so similarly only `-sdk` is required. All other platforms should specify both `-sdk` and `-sysroot`.

It’s also critical that we do not conflate the required Swift bits (resource dir and SDK) and the native system sysroot and keep those concepts separate.

Regarding `ANDROID_NDK_HOME`, there is value in continuing to use this environment variable to match other toolchains, and to minimize the option set needed from a user when using an Android triple - either by just building with `swiftc` or via a build system like SwiftPM.
 For the matter of this PR, we would like to establish the above semantics for each of these options, but it would also be clearly valuable to ease with the transition here and one thing we can do is change this PR to disable `ANDROID_NDK_HOME` inference only when `-resource-dir` is specified to accommodate the current workflows. With a long term goal of removing this change.

finagolfin:
> Compiler toolchain-specific contents and headers. In the case of Swift, these are header files used by the ClangImporter, SwiftShims, etc. It is our expectation that this should generally never be specified by the user.

> All other platforms should specify both -sdk and -sysroot.

This is currently aspirational and doesn't really work on any platform, hence [all the flags listed for Windows above too](https://github.com/swiftlang/swift-driver/pull/1879#issuecomment-3168820504).

> Regarding ANDROID_NDK_HOME, there is value in continuing to use this environment variable to match other toolchains, and to minimize the option set needed from a user when using an Android triple - either by just building with swiftc or via a build system like SwiftPM.

There _would_ be value if `-sdk`/`-sysroot` currently worked on any platform, it currently does not. To take just one example, [the Android doc has listed the `-sdk`/`-resource-dir` flags to build a simple executable for the last five years](https://github.com/swiftlang/swift/commit/e6f591377278b1b4ad7211f370f15ab9d20d2202#diff-b761356ee260fb5b722f31ac680efa41044fa64eb1302a089d9e65f577ef0cdbR108), since I switched it over along with simplifying for the NDK's unified sysroot. There has never been a `-sdk`/`-sysroot` equivalent that worked, without adding a bunch more flags to make up for the fact it doesn't work.

> For the matter of this PR, we would like to establish the above semantics for each of these options, but it would also be clearly valuable to ease with the transition here and one thing we can do is change this PR to disable ANDROID_NDK_HOME inference only when -resource-dir is specified to accommodate the current workflows. With a long term goal of removing this change.

`-sdk`/`-sysroot` will require significantly more work for it to be viable to target non-Darwin platforms, better to just take this out for now as Marc does here. If you doubt that, take a look at swiftlang/swift#79657 and swiftlang/swift#81842 to dive into the issues that remain. Until those issues are meaningfully fixed, foisting this broken `-sdk`/`-sysroot` approach on users is actually breaking builds: it is _not_ helping.

If and when `-sdk`/`-sysroot` works well, and I'm trying to help that with my frontend pull swiftlang/swift#79621, it would make sense to discuss inserting it like this then, but not now when it doesn't work.

marcprux:
I'd like to raise attention to this issue again, since it is still the case that trying to build for Android with `ANDROID_NDK_HOME` set will cause the build to fail on every platform except Windows, and requires workarounds like [this](https://github.com/skiptools/swift-android-action/blob/cf6903e8d98c73b4d5198cae5fd84fbc80ff5c59/action.yml#L248C25-L248C33).

If improving this PR to special-case Windows is not satisfactory, what _would_ be an acceptable solution? Is there any way we can get some solution — even a stopgap one — in place before we start publishing the SDK?

finagolfin:
Their compromise will work to elide the broken code for most SDK bundles, but I think we should remove the broken `-sysroot` flag insertion altogether, as it mostly doesn't work right now, in case someone mistakenly doesn't set the `-resource-dir` flag and hits that mostly broken env variable code path.

Up to you what you want to do...

ephemer:
Am I understanding correctly that having `ANDROID_NDK_HOME` set will set flags in the Swift compiler that would be unset otherwise? But the flags it sets are actually redundant: the settings they refer to can be – and in many cases in practice indeed are – configured by other flags?

If so, the way I understand this, we have an environment variable that is intended to be used as a fallback actually overwriting things it shouldn't.

While I understand the motivation for using a platform-standard environment variable (ANDROID_NDK_HOME) for configuring certain settings, the issue is that we have multiple redundant flags for kind of doing the same thing. This makes using the environment variable as a fallback "only" is not effective. The redundancy in question cannot be trivially removed though, and doing so would be a much larger fix.

Is there an alternative of ensuring only one of the two sets of flags has an effect?

finagolfin:
> Am I understanding correctly that having ANDROID_NDK_HOME set will set flags in the Swift compiler that would be unset otherwise?

Yes

> But the flags it sets are actually redundant: the settings they refer to can be – and in many cases in practice indeed are – configured by other flags?

Not merely redundant, they actually break the build in the vast majority of cases if that environment variable is set, as all SDK bundles currently set the `-resource-dir` flag instead, which leads to conflicts with the `-sysroot` flag set here.

> we have an environment variable that is intended to be used as a fallback actually overwriting things it shouldn't.

The way I would phrase it is that this is a convenience method to find the Android sysroot for you. However, since almost everybody uses `-sdk`/`-resource-dir` right now, that convenience method is instead breaking almost all CI builds! To me, that is a no-brainer, breaking builds should always take priority over convenience methods.

Artem and the others above have put forth a compromise, ie only invoke this convenience method if `-resource-dir` is not set. I would rather we just took out this mostly broken convenience method, as this pull does now, but it is up to Marc, who submitted this pull.

marcprux:
I'll plead once more for attention to this. With the imminent release of the Swift SDK for Android, anyone who happens to have `ANDROID_NDK_ROOT` set in their environment is going to encounter a mystifying error when they try to build their packages (including GitHub actions):

```
<unknown>:0: error: missing required module 'SwiftAndroid'
```

I just don't see why we are consulting this environment at all, at least on non-Windows platforms.

finagolfin:
> One key thing we’d like to do is to better delineate is the use of the various discussed flags here. The ones we’re generally concerned with are:

> -sysroot: Native system root on platforms that have one.
-sdk: Swift SDK root
-resource-dir: Compiler toolchain-specific contents and headers. In the case of Swift, these are header files used by the ClangImporter, SwiftShims, etc. It is our expectation that this should generally never be specified by the user.
One some platforms, sys root and SDK root are the same (read Apple platforms), in which case no explicit -sysroot is necessary, only -sdk is needed. On Windows there is not really a concept of a sysroot, so similarly only -sdk is required. All other platforms should specify both -sdk and -sysroot.

@artemcm, let me demonstrate a core way in which this understanding you have laid out is blatantly wrong on _all_ platforms. There is currently a test in the compiler validation suite that ensures that the _resource directory_, whether implicit or explicitly passed in with the `-resource-dir` flag, is [looked in _first_ for runtime swiftmodules for the target platform](https://github.com/swiftlang/swift/blob/c1fcdc41344ca6140b063e318049d2084e82d61e/test/Serialization/runtime-import-from-sdk.swift#L34):
```
// If the compiler's resource directory *does* contain a runtime swiftmodule, we
// should *not* use the one in the SDK. (We assume that the resource directory
// built with this compiler does contain a standard library.)

// RUN: %empty-directory(%t/mcp)
// RUN: %target-swift-frontend(mock-sdk: -sdk %t/bad-sdk) -module-cache-path %t/mcp -typecheck -verify -disable-implicit-concurrency-module-import -disable-implicit-string-processing-module-import %s
```
This test is currently only run on Darwin CI, but the underlying logic is the same on _all_ platforms, which is why I was able to [enable it on linux and Windows also with a couple minor  tweaks to the test format alone](https://github.com/swiftlang/swift/pull/81842/files#diff-1cab99569bce1c27672e2ed71e850f80cc44d0a75cba48077d6950ded4240a20), ie no changes to the Swift compiler.

This directly contradicts your claim that the `-sdk` is where Swift SDK files like the platform runtime swiftmodules are always looked for. In fact, the Swift resource directory is always where the compiler looks for Swift runtime modules first, only falling back to the `-sdk` if it doesn't find them there. On top of that, there are other gaping holes in your desired approach of looking for Swift SDK files like the runtime libraries in the passed-in `-sdk`, #1562, which makes the `-sdk`/`-sysroot` approach you espouse unworkable right now, unless you [manually make up for it with a dozen more flags like Fabrice is above](https://github.com/swiftlang/swift-driver/pull/1879#issuecomment-3168820504).

All SDK bundles currently use the long-time `-sdk`/`-resource-dir` approach to cross-compilation, while almost nobody is using the new `-sdk`/`-sysroot` alternative you are pushing. That is because of significant flaws in this new approach.

I am not saying let's not move to the new approach: I'm simply saying let's not break the old approach by pushing the new approach in when it doesn't really work yet. This pull by Marc would give us more time to fix the new `-sysroot` approach before we [come back to it later, as I proposed months ago](https://github.com/swiftlang/swift-driver/pull/1879#issuecomment-3197733910).

jakepetroules:
@swift-ci test

jakepetroules:
@compnerd I think we should merge this at this point, since SwiftPM is already passing the necessary flags to activate the NDK without relying on the env var, and also: the code here is behind `#if arch(x86_64)`. So it's basically not even taking effect in the first place on macOS, nor for a number of users on other platforms as well.

I think most of the other discussion on this PR is perhaps better had in swiftlang/swift#79657

jakepetroules:
@swift-ci test windows

compnerd:
Discussed with @jakepetroules offline. The variable here is **correct**. This is the *older* variable. `ANDROID_NDK_ROOT` predates `ANDROID_NDK_HOME` and is used by many projects (including OpenSSL and godot).

I think that we can merge this for now, but be okay with the idea of a quick revert if we find that things are broken. We are still fighting a Debug Info generation issue that was introduced in main and so cannot test this right away.

I will want to see this restored in the future, as this avoids the need to pass `-sysroot` in favour of an environment variable that can be set once.

finagolfin:
The issue is not the name of the environment variable: the issue is that you only use this new `-sysroot` flag for Android cross-compilation on Windows hosts, whereas [SDK bundles use `-sdk`/`-resource-dir` instead](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md#swift-sdkjson-files), which passing in this incompatible `-sysroot` flag breaks.

As such, it will not make sense to ever bring these inserted flags back until there is some change in that SDK bundle JSON format that allows it. I realize you don't use SDK bundles on Windows yet, but since you can always pass this `-sysroot` flag in manually or in a local config file, better to take out this magic method that is currently breaking Android CI, as we had to [unset this env var in the official swiftlang workflows](https://github.com/swiftlang/github-workflows/blob/d7f8c780c9994917d6088e165e0fee6fb0bd06a1/.github/workflows/scripts/install-and-build-with-sdk.sh#L800) to work around this `-sysroot` bug.

Regarding what to do in the future about helping Android devs configure this easier, we can discuss that more with Jake and decide, since he is baking some NDK support into the swift-build package.

Merging now since we are in agreement and we can remove those linked CI workarounds once this fix is merged, then let's discuss more elsewhere and come up with a better solution if needed.

marcprux:
Can we cherry-pick this for 6.3?

PR #1880, opened 2025-04-22
[Jobs] Add a job for `-print-supported-features`

xedin:
This makes it possible to use `-print-supported-features` directly
on `swift{c}` and without any extra flags just like `-print-target-info`.

### DIFF
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -713,6 +713,15 @@ extension Driver {
                                               swiftCompilerPrefixArgs: swiftCompilerPrefixArgs)
     }
 
+    if parsedOptions.hasArgument(.printSupportedFeatures) {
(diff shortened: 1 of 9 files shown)


xedin:
@swift-ci please test

xedin:

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -713,6 +713,13 @@ extension Driver {
                                               swiftCompilerPrefixArgs: swiftCompilerPrefixArgs)
     }
 
+    if parsedOptions.hasArgument(.printSupportedFeatures) {
+      return try toolchain.printSupportedFeaturesJob(
```
artemcm, line 717: We should probably check if the frontend supports this flag and gracefully error exit if it does not.

xedin: ✅

xedin:
@swift-ci please test

artemcm (approved):

xedin:

PR #1883, opened 2025-04-28
[SwiftCaching] Don't leak output path in the explicit module map JSON

cachemeifyoucan:
Using an abstract name for the module path for swift caching build because caching build loads modules directly from CAS and the path is insignificant. The abstract module path allows cache hits across different output locations.

rdar://149960868


cachemeifyoucan:
@swift-ci please test

artemcm (approved):

PR #1891, opened 2025-05-05
Revert "[SwiftCaching] Don't leak output path in the explicit modulemap JSON"

cachemeifyoucan:
The absolute path for the output modules are used by LLDB to look up modules on disk. Revert before LLDB can understand CAS.

This reverts commit 90c6ece2360730a19baed7d6c5797310945bea32.

rdar://150334469


cachemeifyoucan:
@swift-ci please test

PR #1895, opened 2025-05-13
Align Clang resource directory lookup with Clang driver logic

kateinoigakukun:
Previously, we used `Triple.platformName(conflatingDarwin: true)` to derive the OS directory under <ResourceDir>/lib/, which was documented as “the name clang uses”. However, this was in fact closer to what Swift itself uses for resource directories, and it diverged from Clang’s behavior in some cases.

For example, for wasm32-unknown-wasip1-threads, Swift uses "wasi" as the OS name, but Clang uses "wasip1", matching the result of Triple::getOS, in their resource directory lookup.

This patch aligns the behavior with Clang’s logic for looking up the compiler resource directory libraries.


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):
Thank you!

PR #1900, opened 2025-05-14
Pass '-fpie' to the clang linker driver on Linux, instead of 'pie'

artemcm:
Resolves rdar://143793051

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -101,9 +101,9 @@ extension GenericUnixToolchain {
         commandLine.appendPath(toolsDir)
       }
 
-      // Executables on Linux get -pie
+      // Executables on Linux get -fpie
       if targetTriple.os == .linux && linkerOutputType == .executable {
-        commandLine.appendFlag("-pie")
+        commandLine.appendFlag("-fpie")
       }
 
       // On some platforms we want to enable --build-id
--- a/Tests/SwiftDriverTests/SwiftDriverTests.swift
+++ b/Tests/SwiftDriverTests/SwiftDriverTests.swift
@@ -2070,6 +2070,16 @@ final class SwiftDriverTests: XCTestCase {
     }
 
     do {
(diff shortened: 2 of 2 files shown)


artemcm:
@swift-ci test

nkcsgexi (approved):
Thank you!

etcwilde (approved):
I think the fix is fine. How haven't we hit this before though?
Also, maybe for embedded applications, do we have any support for emitting programs that aren't position independent?

etcwilde:
CC @compnerd

Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
```diff
@@ -101,9 +101,9 @@ extension GenericUnixToolchain {
         commandLine.appendPath(toolsDir)
       }
 
-      // Executables on Linux get -pie
+      // Executables on Linux get -fpie
       if targetTriple.os == .linux && linkerOutputType == .executable {
-        commandLine.appendFlag("-pie")
+        commandLine.appendFlag("-fpie")
```
compnerd, line 106: I'm confused by this change. The spelling is `-pie` for the linker driver invocation, but `-fpie` for the compiler invocation (unless you are on certain platforms, where the spelling is `-fPIE`) under certain circumstances.

`-pie` is the equivalent of `-shared` but for executables.

compnerd (changes requested):

Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
artemcm: Thanks for clearing that up, I had confused the two. @cachemeifyoucan pointed out that the warning we get appears to be due to the Clang driver emitting it for platforms where the toolchain default for executables is PIE.

cachemeifyoucan: I am actually unsure where the warning coming from before seeing the build log. I suggest it is better if you just pass `-Xlinker -pie` if you worried about unused argument warning.

PR #1905, opened 2025-05-19
[Incremental Builds] Separately check whether we can skip 'emit-module' on an incremental module-only build

artemcm:
When the incremental machinery is in a state that no source files require to be re-compiled, it may also determine that all the corresponding before-compile tasks may also be skipped, such as `emit-module`. On emit-module-only builds however, this logic is invalid because the individual file compilation is always skipped by not being included in the build plan. 

This PR adds logic for specifically cases where no source files must be compiled on an incremental build because the build plan does not require source compilation, adding logic to ensure that the `emit-module` tasks are skipped only when *all* of the `.swift` inputs to the tasks are older than all of the existing outputs of such tasks.

Resolves rdar://151626629


artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:

artemcm:

artemcm:
@swift-ci test

artemcm:
@swift-ci test

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
```diff
@@ -170,6 +182,21 @@ extension IncrementalCompilationState.FirstWaveComputer {
     }
   }
 
+  /// Figure out if the emit-module tasks are *not* mandatory. This functionality only runs if there are not actual
+  /// compilation tasks to be run in this build, for example on an emit-module-only build.
+  private func computeCanSkipEmitModuleTasks(_ buildRecord: BuildRecord) throws -> Bool {
+    guard let emitModuleJob = jobsInPhases.beforeCompiles.first(where: { $0.kind == .emitModule }) else {
+      return true
```
cachemeifyoucan, line 189: Maybe it is easier to think that if there isn't an emitModule job, there is nothing to skip, so return false here since not skipping logic is simpler and should return the same result if this job doesn't exist.

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
```diff
@@ -170,6 +182,21 @@ extension IncrementalCompilationState.FirstWaveComputer {
...
+    }
+    // If any of the outputs do not exist, they must be re-computed
+    guard try emitModuleJob.outputs.allSatisfy({ try fileSystem.exists($0.file) }) else {
+      return false
+    }
+    // Ensure that no output is older than any of the inputs
```
cachemeifyoucan, line 195: I also feel like if we should only skip if emitModule is the only beforeCompile task (or there isn't any other explicit module tasks) because that might suggest the dependency is changed, so emitModule needs to be re-run.

artemcm:
@swift-ci test

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
artemcm: Good point, changed to `false`.

Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
artemcm: Also a good point, thanks for pointing that out. Added another check to ensure that if non-emit-module tasks are present in `beforeCompiles` then we do not skip `emit-module`.

artemcm:

owenv (approved):

cachemeifyoucan (approved):
Thanks!

artemcm:
@swift-ci test

PR #1909, opened 2025-05-30
Color output on screen256 color terminals

etcwilde:
The Swift driver only emitted color on xterm-256color terminals. screen-256color terminals also have the ability to output colors so we can colorize the output there too.

Also fixing the build deprecation warning on `ProcessEnv.vars`.


etcwilde:
@swift-ci please test

artemcm (approved):

PR #1914, opened 2025-06-02
[6.2 🍒][Incremental Builds] Separately check whether we can skip 'emit-module' on an incremental module-only build

artemcm:
Cherry-pick of https://github.com/swiftlang/swift-driver/pull/1905
- **Explanation**: When the incremental machinery is in a state that no source files require to be re-compiled, it may also determine that all the corresponding before-compile tasks may also be skipped, such as emit-module. On emit-module-only builds however, this logic is invalid because the individual file compilation is always skipped by not being included in the build plan.
This change adds logic for specifically cases where no source files must be compiled on an incremental build because the build plan does not require source compilation, adding logic to ensure that the emit-module tasks are skipped only when all of the .swift inputs to the tasks are older than all of the existing outputs of such tasks.

- **Scope**: Invocations of the driver which only emit a module (no object files or executable) and use Incremental Builds (`-incremental`). 

- **Risk**: Low, the new logic adds checks for whether an emitted module is up-to-date in order to ensure it gets re-compiled when needed. It should only ever result in more compilation tasks. 

• **Problem**: rdar://151626629

- **Reviewed By**: @cachemeifyoucan, @owenv  

- **Original PR**: https://github.com/swiftlang/swift-driver/pull/1905

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/FirstWaveComputer.swift
@@ -134,11 +134,23 @@ extension IncrementalCompilationState.FirstWaveComputer {
       jobCreatingPch: jobCreatingPch)
 
     // In the case where there are no compilation jobs to run on this build (no source-files were changed),
-    // we can skip running `beforeCompiles` jobs if we also ensure that none of the `afterCompiles` jobs
-    // have any dependencies on them.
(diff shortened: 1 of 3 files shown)


artemcm:
@swift-ci test

artemcm:
@swift-ci test

artemcm:

nkcsgexi (approved):

PR #1915, opened 2025-06-02
FreeBSD: Fix rpath to swift runtimes

etcwilde:
The build_os in build-script includes the platform version number from the target triple. The runtimes are installed to a location that does not include the version number, resulting in the swift-driver failing to launch due to not finding libswiftCore. Fixing the rpath to avoid requiring folks to manually specify the correct location through environment variables.


etcwilde:
@swift-ci please test

artemcm:
@swift-ci please test Linux platform

PR #1916, opened 2025-06-02
Revert "[Incremental Builds] Separately check whether we can skip 'emit-module' on an incremental module-only build"

artemcm:
Reverts swiftlang/swift-driver#1905

Looks like it may have caused an intermittent test failure.


artemcm:
@swift-ci test

artemcm:
There is a Linux test failure after this change that hits *some* CI runs.

owenv (approved):

PR #1918, opened 2025-06-04
[Options] Switch -Isystem to -I if the front end doesn't support -Isystem

ian-twilightcoder:
If the front end doesn't support -Isystem, pass it as -I instead.

rdar://152602098


ian-twilightcoder:
@swift-ci test

ian-twilightcoder:
@swift-ci test macOS

ian-twilightcoder:
@swift-ci test Windows

ian-twilightcoder:
Upon further reflection and discussion, I don't think we want to do this.

PR #1919, opened 2025-06-05
Re-land [Incremental Builds] Separately check whether we can skip 'emit-module' on an incremental module-only build

artemcm:
This change was originally landed in https://github.com/swiftlang/swift-driver/pull/1905 and reverted in https://github.com/swiftlang/swift-driver/pull/1916 due to an intermittently-failing test it caused. This is a re-application of the original change with an added test fix. The `testExplicitModuleBuildDoesNotSkipPrecompiledModulesWhenOnlyVerificationIsInvalidated` was a bit too fast with the new module up-to-date checking after the inputs got generated, for the granularity of `stat` calls on Linux, so this change adds a wait to this test. 

Resolves rdar://151626629


artemcm:
@swift-ci test

cachemeifyoucan (approved):

owenv (approved):

artemcm:

PR #1922, opened 2025-06-10
Expose `BuildRecordInfo.computeHash`

plemarquand:
SwiftPM will compute a hash of the arguments passed to `swift build`, `swift test`, etc to determine if a full rebuild is necessary. Currently, it doesn't take in to account whether `.doesNotAffectIncrementalBuild` is set on swift compiler options and so any time any compiler option changes it will perform a full rebuild.

Expose the function that swift-driver uses to compute the argument hash which takes in to account `.doesNotAffectIncrementalBuild` so that SwiftPM can leverage it to reduce unnecessary rebuilds.

In order to keep `BuildRecordInfo` hidden behind `_spi` the function was moved out to the standalone `BuildRecordArguments` struct.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/BuildRecordInfo.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/BuildRecordInfo.swift
@@ -97,7 +97,7 @@ import class Dispatch.DispatchQueue
     else {
       return nil
     }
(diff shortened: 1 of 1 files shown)


plemarquand:
@swift-ci please test

artemcm (approved):

plemarquand:
@swift-ci please test windows

PR #1923, opened 2025-06-11
File Hashing for Incremental Builds

benb:
This PR:
* Adds a command line option `--enable-incremental-file-hashing`, and a corresponding `--disable-incremental-file-hashing`, defaulting to disabled
* When file-hashing is enabled, both the source files and external dependency files are hashed with SHA256
* Those hashes are used to verify that a file is unchanged even if the timestamp is updated
* They are serialized with the rest of the ModuleDependencyGraph info in the blob field
* The serialized version is given a minor version bump to account for this change
* Tests added

RFC discussion here: https://forums.swift.org/t/rfc-file-hashes-for-swift-incremental-compilation/

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
@@ -666,7 +666,8 @@ extension ModuleDependencyGraph {
   /// - Minor number 2: Use `.swift` files instead of `.swiftdeps` in ``DependencySource``
   /// - Minor number 3: Use interned strings, including for fingerprints and use empty dependency source file for no DependencySource
   /// - Minor number 4: Absorb the data in the ``BuildRecord`` into the module dependency graph.
-  @_spi(Testing) public static let serializedGraphVersion = Version(1, 4, 0)
+  /// - Minor number 5: SHA256 hashes for files in externalDepNode and inputInfo blobs. 
+  @_spi(Testing) public static let serializedGraphVersion = Version(1, 5, 0)
 
   /// The IDs of the records used by the module dependency graph.
   fileprivate enum RecordID: UInt64 {


drodriguez:
@swift-ci please smoke test

Broken tree state because swiftlang/swift#82153 had not landed yet.

drodriguez:
@swift-ci please smoke test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -44,6 +45,14 @@ extension Driver.ErrorDiagnostics: CustomStringConvertible {
   }
 }
 
+public struct FileMetadata {
```
owenv: Could this be marked `@_spi(Testing)`? I don't think it needs to be API

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -960,11 +969,20 @@ public struct Driver {
...
-          .lastModificationTime(for: $0.file) else { return nil }
-        return ($0, modTime)
+
+    let incrementalFileHashes = parsedOptions.hasFlag(positive: .enableIncrementalFileHashing,
+                                                      negative: .disableIncrementalFileHashing,
+                                                      default: false)
+    self.recordedInputMetadata = .init(uniqueKeysWithValues:
+      Set(inputFiles).compactMap { inputFile -> (TypedVirtualPath, FileMetadata)? in 
+        guard let modTime = try? fileSystem.lastModificationTime(for: inputFile.file) else { return nil }
+        guard let data = try? fileSystem.readFileContents(inputFile.file)  else { return nil }
```
owenv: Could you move this inside the `if` below so that we avoid the read if the flag is off?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -44,6 +45,14 @@ extension Driver.ErrorDiagnostics: CustomStringConvertible {
   }
 }
 
+public struct FileMetadata {
+    public let mTime: TimePoint
+    public let hash: String
+    init(mTime: TimePoint, hash: String = "") {
```
owenv: nit: IMO making hash optional would be preferable to using "" to represent no hash

owenv (commented):
LGTM, just had a few minor comments

Sources/SwiftDriver/Driver/Driver.swift
benb: I'm afraid this needs to be exposed in the DriverExecutor protocol, so I think it needs to be `public`.

Sources/SwiftDriver/Driver/Driver.swift
benb: Nice one, done.

Sources/SwiftDriver/Driver/Driver.swift
benb: Done.

benb:
I had to fix the added tests - the old logic stopped working after rebasing this branch past https://github.com/swiftlang/swift-driver/commit/7c709c552ff15606a70ed8327b2d34f487886e77

Honestly, I'm not sure my tests are the right approach, however the majority of tests in IncrementalCompilationTests are failing for me locally on main, so I don't want to copy those approaches directly. Maybe I am doing something wrong locally? Most other tests are passing for me, though.

Sources/SwiftDriver/Driver/Driver.swift
owenv: Ah, I forgot this showed up in that API. We may need a strategy to stage in the changes to the DriverExecutor protocol so they don't break https://github.com/swiftlang/swift-build, or make synchronized changes to the repos. Maybe you could keep the old `execute` requirements around temporarily for default implementations of the new ones to call?

artemcm:
> however the majority of tests in IncrementalCompilationTests are failing for me locally on main, so I don't want to copy those approaches directly.

What kinds of failures are you seeing? I recommend building and running tests against most-recent `main` snapshot from `https://www.swift.org/install/macos/` or a locally-built ToT `main` of the compiler.

Sources/SwiftDriver/Driver/Driver.swift
benb: Ah, I didn't realise there was an API here exposed to swift-build. 

I've added a commit that restores the original methods to `DriverExecutor`, with an extension that gives default implementations of the new methods, and added in definitions with `fatalError()` if these legacy methods are called on the implementations defined in swift-driver itself.

https://github.com/swiftlang/swift-driver/pull/1923/commits/3ecc73491f17bde92242e17490652b0456496abf

benb:
> > however the majority of tests in IncrementalCompilationTests are failing for me locally on main, so I don't want to copy those approaches directly.
> 
> What kinds of failures are you seeing? I recommend building and running tests against most-recent `main` snapshot from `https://www.swift.org/install/macos/` or a locally-built ToT `main` of the compiler.

`Test Suite 'IncrementalCompilationTests' failed at 2025-06-13 17:30:25.064.
         Executed 37 tests, with 31 failures (0 unexpected) in 49.195 (49.197) seconds`

I'll try with a snapshot.

benb:
> I'll try with a snapshot.
Using swiftly main-snapshot on a commit from yesterday:
```
$ ~/.swiftly/bin/swift --version
Apple Swift version 6.2-dev (LLVM d266fd9f80a7945, Swift 665515c781999a8)
Target: arm64-apple-macosx15.0
Build config: +assertions
```

I'm still getting lots of failures in IncrementalCompilationTests on current main (8c130e3dc2b6):
```
$ ~/.swiftly/bin/swift test
…
Test Suite 'IncrementalCompilationTests' failed at 2025-06-13 19:33:51.708.
         Executed 34 tests, with 31 failures (0 unexpected) in 132.160 (132.161) seconds
…
```
Although I see strictly speaking there are '0 unexpected' failures, but I'm not sure how or why these are defined as being 'expected'.

benb:
@owenv Is there anything I'm missing to unblock this PR?

owenv:
No, sorry, I lost track of this. Let's kick off the CI tests and see how those results look.

@swift-ci test

owenv:
@swift-ci test Windows

owenv (approved):

benb:
Rebased and squashed fixup commits.

@swift-ci test
@swift-ci test Windows

owenv:
@swift-ci test

Sources/SwiftDriver/Driver/Driver.swift
cachemeifyoucan: I prefer this to move to a file in `SwiftDriver/Utilities/`. Remember to update the CMakeFile if you add a new file.

artemcm:

artemcm (approved):

benb:
> There are some incremental state reporter messages need to be adjusted, or report a different message. Like `reportExplicitDependencyOutOfDate` says `older than input file`.

This message is still accurate, alas. I haven't added hash logic here, and I don't think this PR can help in this case without adding a lot more work. 

I may have understood this wrong, but I think in this case we are comparing swiftmodule files that are dependencies or transitive dependencies of a file under compilation in the current module ("Module A"). And if the timestamp of a dependency-of-a-dependency ("Module C") is newer than that of the dependency ("Module B"), this is taken as meaning we have to recompile Module B before we can proceed. Of course, if the hash of Module C is unchanged, we could skip it.

However, I don't think there's an easy place to store the hash data for these (and we are not storing timestamps for them). Furthermore, I think you would want to compare the hash of Module C with that recorded during the compilation of Module B, not during the previous compilation of the current module (Module A).

I also think this logic is invoked during WMO compilation, and not just during Incremental mode.

Does this sound right?

cachemeifyoucan:
> This message is still accurate, alas. I haven't added hash logic here, and I don't think this PR can help in this case without adding a lot more work.

Ah, this is the report for module dependency only and I guess we are not extending hash check to module dependencies yet (and that is not computed in swift driver, but in scanner). The remark for incremental build only says `skipped` without stating if the file changed or it is older, so I guess it is fine.

Thanks for looking and address the feedback. I think the newly added `FileMetadata.swift` should be in `Sources/SwiftDriver/Utilities` not in its current location where there are no other swift source files. LGTM otherwise.

benb:
> I think the newly added `FileMetadata.swift` should be in `Sources/SwiftDriver/Utilities` not in its current location where there are no other swift source files.

Whoops, fixed. Thanks for your review.

benb:
Can someone with the power run the required swift-ci checks again, please?

artemcm:
@swift-ci test

artemcm:

cachemeifyoucan:

benb:
Rebased past b0a01d3a283f6aa63ac64d0aa5af3d0cc8b64fc7 to fix merge conflicts, now we need to run the CI tests again…

artemcm:
@swift-ci test

artemcm:

Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
```diff
@@ -536,7 +536,7 @@ final class ExplicitModuleBuildTests: XCTestCase {
                   main.nativePathString(escaped: true)] + sdkArgumentsForTesting
       var driver = try Driver(args: args)
       let _ = try driver.planBuild()
-      let dependencyGraph = try XCTUnwrap(driver.explicitDependencyBuildPlanner?.dependencyGraph)
+      let dependencyGraph = try XCTUnwrap(driver.intermoduleDependencyGraph)
```
artemcm, line 539: Thank you.

benb:
I can't see the merge button - does a committer need to merge on my behalf?

PR #1924, opened 2025-06-11
Bump `swift-tools-version` to 5.10 in `Package.swift`

MaxDesiatov:
As Swift CI nodes use Swift 5.10 or later, we no longer need to maintain compatibility with Swift 5.7. This allows adopting features introduced in that span of versions, e.g. `package` access control.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.10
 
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #1927, opened 2025-06-11
Add swift crash producer support

cachemeifyoucan:
When swift compiler crashed with swift caching enabled, create a crash reproducer automatically.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -23,6 +23,7 @@ import struct TSCBasic.AbsolutePath
 import struct TSCBasic.ByteString
 import struct TSCBasic.Diagnostic
 import struct TSCBasic.FileInfo
+import struct TSCBasic.ProcessResult
 import struct TSCBasic.RelativePath
 import struct TSCBasic.SHA256
 import var TSCBasic.localFileSystem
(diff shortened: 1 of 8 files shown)


nkcsgexi:
Super excited to see this is happening! Why does it have to be guarded behind CAS enablement?

cachemeifyoucan:
> Super excited to see this is happening! Why does it have to be guarded behind CAS enablement?

For now, it is. When CAS is enabled, it is easy to capture the full dependency just from the failing swift-frontend command. It is not that trivial from a non-CAS build since all the clang dependencies are hidden. clang is adding a way to create reproducer that requires a full scan, but that needs to orchestrated by driver/build system, not that easy to do from a single crashed swift-frontend command.

cachemeifyoucan:
Rebase and ping. This is the driver support for https://github.com/swiftlang/swift/pull/82128

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -4019,3 +4024,36 @@ extension Driver {
...
+    }
+#else
+    guard case .signalled = status else {
+      return;
+    }
+#endif
+    guard isFrontendArgSupported(.genReproducer), enableCaching, job.kind.supportCaching else {
+      return
+    }
+    // TODO: check flag is supported.
```
artemcm: Which flag? The code above is already checking `-gen-reproducer`.

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@artemcm @nkcsgexi Can you review the new patch again? The previous implementation registered a callback that extends the lifetime of `Driver` to be hold on by `Executor` and its delegations. That has unintended consequences for the resources like CAS lifetime. 

The new approach adds a new function to `JobExecutionDelegate` to avoid let driver doing all the work. This will also make swift-build implementation easier.

cachemeifyoucan:

PR #1934, opened 2025-06-17
[ABIChecking] swift-api-digester doesn't support -isystem

ian-twilightcoder:
swift-api-digester doesn't support the -isystem clang flag like it does the -iframework one. Use -I for now.

rdar://153664788


ian-twilightcoder:
@swift-ci test

PR #1935, opened 2025-06-18
[Subcommands] Ensure an adjacent binary exists before using it

doinkythederp:
This pull request updates the handling of subcommands in the Swift Driver to ensure correct behavior when the subcommand executable is not adjacent to swift-driver itself. Now, if the adjacent binary doesn't exist, the driver will fall back to a $PATH-based search rather than attempting to run the missing executable (this seems to be the intended behavior based on the comment above this code).

I also expanded the error message for the failure case here.

This change is important because allows users to use subcommands that didn't come bundled with Swift.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -77,8 +77,11 @@ public struct Driver {
 
     public var description: String {
       switch self {
-      case .unknownOrMissingSubcommand(let subcommand):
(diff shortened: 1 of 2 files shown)


artemcm (approved):

artemcm:
@swift-ci test

lewisfm:
The error message for Windows seems to be about Predicate's string formatting, but I didn't change code related to that. Is there anything I need to do here to resolve the CI error? This is my first time contributing to Swift, so I don't really know the procedure.

PR #1939, opened 2025-06-25
Perform response file write atomically

artemcm:
Switch over to using `FileSystem.writeFileContents(_ path:, bytes:, atomically:)`. Otherwise we may encounter a race when multiple driver invocations attempt to write a response file for an identical task, such as a shared module dependency pre-compilation job.


artemcm:
@swift-ci test

cachemeifyoucan (approved):
LGTM

PR #1942, opened 2025-06-27
add a check for a matching python architecture

charles-zablit:
Fixes https://github.com/swiftlang/swift/issues/79981.

On Windows, if there is a mismatch between the architecture of the installed Python and the architecture of the toolchain, `swift repl` will crash immediately with `0xC000007B`. This patch adds a check to warn the user of the architecture mismatch right before invoking the REPL, which will crash. We also implement the same check for Darwin.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -875,17 +875,22 @@ public struct Driver {
 
     self.executor = executor
 
-    let invocationMode = try Self.invocationRunMode(forArgs: args).mode
-    if case .subcommand = invocationMode {
-      throw Error.subcommandPassedToDriver
+    if args.count > 1 && args[1] == "-repl" {
+      #if arch(arm64)
+        checkIfMatchingPythonArch(
+          cwd: ProcessEnv.cwd, envBlock: envBlock, toolchainArchitecture: .arm64, diagnosticsEngine: diagnosticsEngine)
+      #elseif arch(x86_64)
+        checkIfMatchingPythonArch(
+          cwd: ProcessEnv.cwd, envBlock: envBlock, toolchainArchitecture: .x64, diagnosticsEngine: diagnosticsEngine)
+      #elseif arch(x86)
+        checkIfMatchingPythonArch(
+          cwd: ProcessEnv.cwd, envBlock: envBlock, toolchainArchitecture: .x86, diagnosticsEngine: diagnosticsEngine)
+      #endif
     }
 
-    #if os(Windows)
-      if case .normal(isRepl: true) = invocationMode {
-        checkIfMatchingPythonArch(
-          cwd: ProcessEnv.cwd, envBlock: envBlock, diagnosticsEngine: diagnosticsEngine)
-      }
-    #endif
+    if case .subcommand = try Self.invocationRunMode(forArgs: args).mode {
+      throw Error.subcommandPassedToDriver
+    }
 
     var args = args
     if let additional = env["ADDITIONAL_SWIFT_DRIVER_FLAGS"] {
--- a/Sources/SwiftDriver/Utilities/PythonArchitecture.swift
+++ b/Sources/SwiftDriver/Utilities/PythonArchitecture.swift
@@ -1,3 +1,14 @@
+//===--------------- Driver.swift - Swift Driver --------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
 import Foundation
 
 import struct TSCBasic.AbsolutePath
@@ -11,114 +22,187 @@ import typealias TSCBasic.ProcessEnvironmentBlock
 import func TSCBasic.getEnvSearchPaths
 import func TSCBasic.lookupExecutablePath
 
+#if os(Windows)
+  import WinSDK
+#endif
+
 /// Check that the architecture of the toolchain matches the architecture
 /// of the Python installation.
 ///
 /// When installing the x86 toolchain on ARM64 Windows, if the user does not
-/// install an x86 version of Python, they will get acryptic error message
+/// install an x86 version of Python, they will get a cryptic error message
 /// when running lldb (`0xC000007B`). Calling this function before invoking
-/// lldb gives them a warning to help troublshoot the issue.
+/// lldb gives them a warning to help troubleshoot the issue.
 ///
 /// - Parameters:
 ///   - cwd: The current working directory.
-///   - env: A dictionary of the environment variables and their values. Usually of the parent shell.
+///   - env: The parent shell's ProcessEnvironmentBlock.
 ///   - diagnosticsEngine: DiagnosticsEngine instance to use for printing the warning.
 public func checkIfMatchingPythonArch(
-  cwd: AbsolutePath?, envBlock: ProcessEnvironmentBlock, diagnosticsEngine: DiagnosticsEngine
+  cwd: AbsolutePath?, envBlock: ProcessEnvironmentBlock,
+  toolchainArchitecture: ExecutableArchitecture, diagnosticsEngine: DiagnosticsEngine
 ) {
-  #if arch(arm64)
-    let toolchainArchitecture = COFFBinaryExecutableArchitecture.arm64
-  #elseif arch(x86_64)
-    let toolchainArchitecture = COFFBinaryExecutableArchitecture.x64
-  #elseif arch(x86)
-    let toolchainArchitecture = COFFBinaryExecutableArchitecture.x86
-  #else
-    return
-  #endif
-  let pythonArchitecture = COFFBinaryExecutableArchitecture.readWindowsExecutableArchitecture(
-    cwd: cwd, envBlock: envBlock, filename: "python.exe")
+  #if os(Windows) || os(macOS)
+    #if os(Windows)
+      let pythonArchitecture = Process.readWindowsExecutableArchitecture(
+        cwd: cwd, envBlock: envBlock, filename: "python.exe")
+    #elseif os(macOS)
+      let pythonArchitecture = Process.readDarwinExecutableArchitecture(
+        cwd: cwd, envBlock: envBlock, filename: "python3")
+      if pythonArchitecture == .universal {
+        return
+      }
+    #endif
 
-  if toolchainArchitecture != pythonArchitecture {
-    diagnosticsEngine.emit(
-      .warning(
-        """
-        There is an architecture mismatch between the installed toolchain and the resolved Python's architecture:
-        Toolchain: \(toolchainArchitecture)
-        Python: \(pythonArchitecture)
-        """))
-  }
+    guard toolchainArchitecture == pythonArchitecture else {
+      diagnosticsEngine.emit(
+        .warning(
+          """
+          There is an architecture mismatch between the installed toolchain and the resolved Python's architecture:
+          Toolchain: \(toolchainArchitecture)
+          Python: \(pythonArchitecture)
+          """))
+      return
+    }
+  #endif
 }
 
 /// Some of the architectures that can be stored in a COFF header.
-enum COFFBinaryExecutableArchitecture: String {
+public enum ExecutableArchitecture: String {
   case x86 = "X86"
   case x64 = "X64"
   case arm64 = "ARM64"
+  case universal = "Universal"
   case unknown = "Unknown"
 
-  static func fromPEMachineByte(machine: UInt16) -> Self {
-    // https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#machine-types
-    switch machine {
-    case 0x014c: return .x86
-    case 0x8664: return .x64
-    case 0xAA64: return .arm64
-    default: return .unknown
+  #if os(Windows)
+    static func fromPEMachineByte(machine: Int32) -> Self {
+      // https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#machine-types
+      switch machine {
+      case IMAGE_FILE_MACHINE_I386: return .x86
+      case IMAGE_FILE_MACHINE_AMD64: return .x64
+      case IMAGE_FILE_MACHINE_ARM64: return .arm64
+      default: return .unknown
+      }
     }
-  }
+  #endif
 
-  /// Resolves the filename from the `Path` environment variable and read its COFF header to determine the architecture
-  /// of the binary.
-  ///
-  /// - Parameters:
-  ///   - cwd: The current working directory.
-  ///   - env: A dictionary of the environment variables and their values. Usually of the parent shell.
-  ///   - filename: The name of the file we are resolving the architecture of.
-  /// - Returns: The architecture of the file which was found in the `Path`.
-  static func readWindowsExecutableArchitecture(
-    cwd: AbsolutePath?, envBlock: ProcessEnvironmentBlock, filename: String
-  ) -> Self {
-    let searchPaths = getEnvSearchPaths(pathString: envBlock["Path"], currentWorkingDirectory: cwd)
-    guard
-      let filePath = lookupExecutablePath(
-        filename: filename, currentWorkingDirectory: cwd, searchPaths: searchPaths)
-    else {
-      return .unknown
+  #if os(macOS)
+    static func fromMachoCPUType(cpuType: Int32) -> Self {
+      // https://en.wikipedia.org/wiki/Mach-O
+      switch cpuType {
+      case 0x0100_0007: return .x86
+      case 0x0100_000c: return .arm64
+      default: return .unknown
+      }
     }
-    guard let fileHandle = FileHandle(forReadingAtPath: filePath.pathString) else {
-      return .unknown
+  #endif
+}
+
+extension Process {
+  #if os(Windows)
+    /// Resolves the filename from the `Path` environment variable and read its COFF header to determine the architecture
+    /// of the binary.
+    ///
+    /// - Parameters:
+    ///   - cwd: The current working directory.
+    ///   - env: A dictionary of the environment variables and their values. Usually of the parent shell.
+    ///   - filename: The name of the file we are resolving the architecture of.
+    /// - Returns: The architecture of the file which was found in the `Path`.
+    static func readWindowsExecutableArchitecture(
+      cwd: AbsolutePath?, envBlock: ProcessEnvironmentBlock, filename: String
+    ) -> ExecutableArchitecture {
+      let searchPaths = getEnvSearchPaths(
+        pathString: envBlock["Path"], currentWorkingDirectory: cwd)
+      guard
+        let filePath = lookupExecutablePath(
+          filename: filename, currentWorkingDirectory: cwd, searchPaths: searchPaths)
+      else {
+        return .unknown
+      }
+      guard let fileHandle = FileHandle(forReadingAtPath: filePath.pathString) else {
+        return .unknown
+      }
+
+      defer { fileHandle.closeFile() }
+
+      // Infering the architecture of a Windows executable from its COFF header involves the following:
+      // 1. Get the COFF header offset from the pointer located at the 0x3C offset (4 bytes long).
+      // 2. Jump to that offset and read the next 6 bytes.
+      // 3. The first 4 are the signature which should be equal to 0x50450000.
+      // 4. The last 2 are the machine architecture which can be infered from the value we get.
+      //
+      // The link below provides a visualization of the COFF header and the process to get to it.
+      // https://upload.wikimedia.org/wikipedia/commons/1/1b/Portable_Executable_32_bit_Structure_in_SVG_fixed.svg
+      guard (try? fileHandle.seek(toOffset: 0x3C)) != nil else {
+        return .unknown
+      }
+      guard let offsetPointer = try? fileHandle.read(upToCount: 4),
+        offsetPointer.count == 4
+      else {
+        return .unknown
+      }
+
+      let peHeaderOffset = offsetPointer.withUnsafeBytes { $0.load(as: UInt32.self) }
+
+      guard (try? fileHandle.seek(toOffset: UInt64(peHeaderOffset))) != nil else {
+        return .unknown
+      }
+      guard let coffHeader = try? fileHandle.read(upToCount: 6), coffHeader.count == 6 else {
+        return .unknown
+      }
+
+      let signature = coffHeader.prefix(4)
+      let machineBytes = coffHeader.suffix(2)
+
+      guard signature == Data([0x50, 0x45, 0x00, 0x00]) else {
+        return .unknown
+      }
+
+      let machine = machineBytes.withUnsafeBytes { $0.load(as: UInt16.self) }
+      return .fromPEMachineByte(machine: Int32(machine))
     }
+  #endif
 
-    defer { fileHandle.closeFile() }
+  #if os(macOS)
+    static func readDarwinExecutableArchitecture(
+      cwd: AbsolutePath?, envBlock: ProcessEnvironmentBlock, filename: String
+    ) -> ExecutableArchitecture {
+      let magicNumber: UInt32 = 0xcafe_babe
 
-    // Infering the architecture of a Windows executable from its COFF header involves the following:
-    // 1. Get the COFF header offset from the pointer located at the 0x3C offset (4 bytes long).
-    // 2. Jump to that offset and read the next 6 bytes.
-    // 3. The first 4 are the signature which should be equal to 0x50450000.
-    // 4. The last 2 are the machine architecture which can be infered from the value we get.
-    //
-    // The link below provides a visualization of the COFF header and the process to get to it.
-    // https://upload.wikimedia.org/wikipedia/commons/1/1b/Portable_Executable_32_bit_Structure_in_SVG_fixed.svg
-    fileHandle.seek(toFileOffset: 0x3C)
-    guard let offsetPointer = try? fileHandle.read(upToCount: 4),
-      offsetPointer.count == 4
-    else {
-      return .unknown
+      let searchPaths = getEnvSearchPaths(
+        pathString: envBlock["PATH"], currentWorkingDirectory: cwd)
+      guard
+        let filePath = lookupExecutablePath(
+          filename: filename, currentWorkingDirectory: cwd, searchPaths: searchPaths)
+      else {
+        return .unknown
+      }
+      guard let fileHandle = FileHandle(forReadingAtPath: filePath.pathString) else {
+        return .unknown
+      }
+
+      defer {
+        try? fileHandle.close()
+      }
+
+      // The first 4 bytes of a Mach-O header contain the magic number. We use it to determine if the binary is
+      // universal.
+      // https://github.com/apple/darwin-xnu/blob/main/EXTERNAL_HEADERS/mach-o/loader.h
+      let magicData = fileHandle.readData(ofLength: 4)
+      let magic = magicData.withUnsafeBytes { $0.load(as: UInt32.self).bigEndian }
+
+      if magic == magicNumber {
+        return .universal
+      }
+
+      // If the binary is not universal, the next 4 bytes contain the CPU type.
+      guard (try? fileHandle.seek(toOffset: 4)) != nil else {
+        return .unknown
+      }
+      let cpuTypeData = fileHandle.readData(ofLength: 4)
+      let cpuType = cpuTypeData.withUnsafeBytes { $0.load(as: Int32.self) }
+      return .fromMachoCPUType(cpuType: cpuType)
     }
-
-    let peHeaderOffset = offsetPointer.withUnsafeBytes { $0.load(as: UInt32.self) }
-
-    fileHandle.seek(toFileOffset: UInt64(peHeaderOffset))
-    guard let coffHeader = try? fileHandle.read(upToCount: 6), coffHeader.count == 6 else {
-      return .unknown
-    }
-
-    let signature = coffHeader.prefix(4)
-    let machineBytes = coffHeader.suffix(2)
-
-    guard signature == Data([0x50, 0x45, 0x00, 0x00]) else {
-      return .unknown
-    }
-
-    return .fromPEMachineByte(machine: machineBytes.withUnsafeBytes { $0.load(as: UInt16.self) })
-  }
+  #endif
 }


charles-zablit:
@swift-ci please test

charles-zablit:
@swift-ci please test

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1438,42 @@ extension Driver {
 
     return (.subcommand(subcommand), updatedArgs)
   }
+
+  private static func checkIfMatchingPythonArch() {
```
adrian-prantl: Can you add a comment here explaining that this exists because users who install the x86 toolchain on arm64 windows will get an extremely cryptic error message otherwise?

adrian-prantl:
This is going to be huge usability improvement. I'd like @artemcm to sign off though due the unusual nature of the patch.

charles-zablit:
@swift-ci please test

Sources/SwiftDriver/Driver/Driver.swift
charles-zablit: Completely forgot to add it after our discussion. Fixed 👍

artemcm:

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1438,49 @@ extension Driver {
 
     return (.subcommand(subcommand), updatedArgs)
   }
+
+  /// Ensure that the architecture of the toolchain matches the architecture
+  /// of the Python installation.
+  ///
+  /// When installing the x86 toolchain on ARM64 Windows, if the user does not
+  /// install an x86 version of Python, they will get acryptic error message
```
artemcm:
```suggestion
  /// install an x86 version of Python, they will get a cryptic error message
```

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1438,49 @@ extension Driver {
...
+    do {
+        try process.run()
+        process.waitUntilExit()
+
+        let data = pipe.fileHandleForReading.readDataToEndOfFile()
+        guard let output = String(data: data, encoding: .utf8) else {
+          return;
+        }
+        if !output.lowercased().contains(arch) {
+          stderrStream.send("There is an architecture mismatch between the installed toolchain and the resolved Python's architecture:\n")
```
artemcm: Could you please use the driver's `diagnosticEngine` and define a new `.warning` for this message instead of using the stderr stream directly?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1438,49 @@ extension Driver {
...
+    #if arch(arm64)
+    let arch = "arm64"
+    #elseif arch(x86_64)
+    let arch = "amd64"
+    #elseif arch(x86)
+    let arch = "x86"
+    #else
+    return;
+    #endif
+    let process = Process()
```
artemcm: By the time we are in `invocationRunMode` we have an instance of the `DriverExecutor`, so we should plumb that here and use it to run this subprocess instead of using `Process` APIs directly. You can see an example of this being done in `xcrunFind`.

artemcm (changes requested):
Is it possible to find a way to do this without executing Python code? e.g. can we locate the `python.exe` executable and do something equivalent to what `file` does on Darwin?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1425,6 +1426,9 @@ extension Driver {
     if firstArg == "repl" {
         updatedArgs.remove(at: 1)
         updatedArgs.append("-repl")
+        #if os(Windows)
+        checkIfMatchingPythonArch()
+        #endif
```
compnerd: Does this problem not exist on Darwin as well? If you run `arch -m x86_64 swift repl` that should fail similarly.

Sources/SwiftDriver/Driver/Driver.swift
compnerd: It is quite non-cryptic IMO. 0xC000007B is a `NTSTATUS`; 0xC => Kernel; 0x7B = 123 = ERROR_INVALID_NAME. This is the same behaviour as Linux when the loader is not found.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1438,49 @@ extension Driver {
...
+    #elseif arch(x86_64)
+    let arch = "amd64"
+    #elseif arch(x86)
+    let arch = "x86"
+    #else
+    return;
+    #endif
+    let process = Process()
+    process.executableURL = URL(fileURLWithPath: "C:\\Windows\\System32\\cmd.exe")
+    process.arguments = ["/c", "python.exe", "-c", "import platform; print(platform.machine())"]
```
compnerd: This seems like a pretty bad way to check if the DLL matches. Are we checking if *python* matches the architecture or if the current runtime as indicated by `PYTHONHOME` matchess?

charles-zablit: We are checking that the Python that is resolved from the `Path` matches the architecture. I have tried some combinations of configurations:

Everything below was run on an ARM64 Windows VM with the x86 toolchain installed.

| Python Path pointing to | PYTHONHOME pointing to | Output                               |
|-------------------------|------------------------|--------------------------------------|
| ARM64                   | Nothing                | Crash with warning                   |
| ARM64                   | ARM64                  | Crash with warning                   |
| ARM64                   | AMD64                  | Crash with warning                   |
| AMD64                   | Nothing                | Launches fine                        |
| AMD64                   | ARM64                  | Fatal Python error: init_fs_encoding |
| AMD64                   | AMD64                  | Launches fine                        |


I think that checking if the Python that is in the `Path` is the correct one is enough. If there is a mismatch between the Python that is in the Path and the one that is in `PYTHONHOME`, that is another problem and there are enough information printed to the user to help them figure it out.

Sources/SwiftDriver/Driver/Driver.swift
charles-zablit: I meant cryptic from the point of view of the user. They only get this error message with no indication of what is failing and how to fix it (see the screenshot below). I will change the wording in the comment 👍 

- https://github.com/swiftlang/swift/issues/79981

lldb_crash

charles-zablit:
> Is it possible to find a way to do this without executing Python code? e.g. can we locate the `python.exe` executable and do something equivalent to what `file` does on UNIX?

There is no non third party alternative to `file` on Windows. There is a way to check programmatically that `python.exe` is the correct architecture according to [this SO thread](https://stackoverflow.com/questions/54834984/how-do-i-determine-the-architecture-of-an-executable-binary-on-windows-10#:~:text=The%20architecture%20of%20the%20executable,the%20beginning%20of%20the%20file). I'm happy to try to implement this in Swift, but I think it will add some complexity to the code compared to just calling Python?

Sources/SwiftDriver/Driver/Driver.swift
charles-zablit: Switched to using `DriverExecutor`.

Sources/SwiftDriver/Driver/Driver.swift
charles-zablit: Switched to using the diagnosticsEngine.

Sources/SwiftDriver/Driver/Driver.swift
charles-zablit: I was not able to get the problem to happen on Darwin. `arch -x86_64 swift repl` starts the REPL correctly.

Maybe we should bring this check to other platforms as well. I focused on Windows in this patch because all the reports I received were from Windows users on ARM64 machines.

artemcm:
> > Is it possible to find a way to do this without executing Python code? e.g. can we locate the `python.exe` executable and do something equivalent to what `file` does on UNIX?
> 
> There is no non third party alternative to `file` on Windows. There is a way to check programmatically that `python.exe` is the correct architecture according to [this SO thread](https://stackoverflow.com/questions/54834984/how-do-i-determine-the-architecture-of-an-executable-binary-on-windows-10#:~:text=The%20architecture%20of%20the%20executable,the%20beginning%20of%20the%20file). I'm happy to try to implement this in Swift, but I think it will add some complexity to the code compared to just calling Python?

If we can reduce the set of dependencies the driver needs in order to perform all the tasks it needs to then we should. Even though `swift repl` does depend on having Python on the system, it would be really nice to avoid the driver having to - especially by firing up the interpreter and executing code. 

Furthermore, as-is, when executing this check in the driver on a system where Python is missing, the user will still get this warning with some surprising `pythonArchitecture` contents printed. 

I know there is `Dumpbin.exe` that is present on the system when Windows SDK is installed, which can act similarly to `file`, but it also doesn't seem all that bad to try and read the file machine type from the COFF header - that way we can ensure the binary exists in the first place and emit an appropriate error in that case too.

artemcm:
To add to the above comment, the driver today on most code paths plans the build without launching any subprocesses at all, which is crucial in some environments where the driver is used as a library in an environment where it is not allowed to launch processes itself.

While I'm not aware of this being a hard requirement on Windows today, it's still worthwhile to avoid adding new process launch sites when we can.

charles-zablit:
@swift-ci please test

charles-zablit:
The COFF header approach seems to work fine 👍 

I am unsure what's the best way to resolve the path to `python`. My current approach manually searches for it with `lookupExecutablePath` to avoid executing a subprocess.
However, we could use `where.exe` to achieve the same thing and use the actual resolution mechanism from Windows.

compnerd:
> The COFF header approach seems to work fine 👍 
> 
> 
> 
> I am unsure what's the best way to resolve the path to `python`. My current approach manually searches for it with `lookupExecutablePath` to avoid executing a subprocess.
> 
> However, we could use `where.exe` to achieve the same thing and use the actual resolution mechanism from Windows.

Looking up in Path is the right thing to do - if it is not in Path, it won't be found anyway.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1444,100 @@ extension Driver {
...
+        .warning(
+          """
+          There is an architecture mismatch between the installed toolchain and the resolved Python's architecture:
+          Toolchain: \(toolchainArchitecture)
+          Python: \(pythonArchitecture)
+          """))
+    }
+  }
+
+  enum ExecutableArchitecture: String {
```
artemcm: The name of this type should indicate that it is referring to a binary file, perhaps even specifically to a COFF file.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1444,100 @@ extension Driver {
...
+    }
+  }
+
+  static func readWindowsExecutableArchitecture(cwd: AbsolutePath?, env: [String: String])
+    -> ExecutableArchitecture
+  {
+    guard let pathEnv = env["Path"]?.split(separator: ";") else {
+      return .unknown
+    }
+    let searchPaths = pathEnv.compactMap { path in
```
artemcm: This can be replaced with `TSCBasic.getEnvSearchPaths`:
```
let searchPaths = getEnvSearchPaths(pathString: env["PATH"], currentWorkingDirectory: cwd)
```

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1444,100 @@ extension Driver {
...
+      switch machine {
+      case 0x014c: return .x86
+      case 0x8664: return .x64
+      case 0xAA64: return .arm64
+      default: return .unknown
+      }
+    }
+  }
+
+  static func readWindowsExecutableArchitecture(cwd: AbsolutePath?, env: [String: String])
```
artemcm: We should either make the name of this method reflect that it is Python-specific or parameterize in on the executable name and pass in `python.exe`.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1444,100 @@ extension Driver {
...
   }
+
+  /// Ensure that the architecture of the toolchain matches the architecture
+  /// of the Python installation.
+  ///
+  /// When installing the x86 toolchain on ARM64 Windows, if the user does not
+  /// install an x86 version of Python, they will get acryptic error message
+  /// when running lldb (`0xC000007B`). Calling this function before invoking
+  /// lldb gives them a warning to help troublshoot the issue.
+  public static func checkIfMatchingPythonArch(
```
artemcm: Could we please move the new checking code to a file in `Utilities` from `Driver.swift`?

Package.resolved
```diff
@@ -1,4 +1,5 @@
 {
+  "originHash" : "a4d231bdb2709a1f4d2fa06d5a165bad5e3f77e6173f659dfcdc931dd3acdaf3",
```
artemcm: Could you please remove this file from this diff? We try to update `Package.resolved` in a standalone change, when needed.

artemcm (approved):
I have a couple of small nits, but otherwise this looks good, thank you!

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1434,6 +1444,100 @@ extension Driver {
...
+    else {
+      return .unknown
+    }
+    guard let fileHandle = FileHandle(forReadingAtPath: filePath.pathString) else {
+      return .unknown
+    }
+
+    defer { fileHandle.closeFile() }
+
+    // https://upload.wikimedia.org/wikipedia/commons/1/1b/Portable_Executable_32_bit_Structure_in_SVG_fixed.svg
```
artemcm: Just on the off chance this link dies, it's worth a sentence here to elaborate on the fact that we're reading in the specific machine kind offset.

charles-zablit:
@swift-ci please check

charles-zablit:
> I have a couple of small nits, but otherwise this looks good, thank you!

Thanks! Fixed the issues and added documentation to the new methods.

artemcm (approved):

artemcm:
@swift-ci test

artemcm:

artemcm:

charles-zablit:
Odd, the build error happens only on Windows:

```
C:\Users\swift-ci\jenkins\workspace\swift-driver-PR-windows\swift-driver\Sources\SwiftDriver\Driver\Driver.swift:870:13: error: cannot find 'checkIfMatchingPythonArch' in scope
 868 |     #if os(Windows)
 869 |       if args[1] == "-repl" {  // a `-` gets automatically added to `swift repl`.
 870 |         try checkIfMatchingPythonArch(
     |             `- error: cannot find 'checkIfMatchingPythonArch' in scope
 871 |           cwd: ProcessEnv.cwd, env: env, diagnosticsEngine: diagnosticsEngine)
 872 |       }
```

Do I need to manually import the method?

EDIT: Added the file to `CMakeList.txt`.

charles-zablit:

charles-zablit:
@swift-ci please test

artemcm:

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,122 @@
...
+  ///
+  /// - Parameters:
+  ///   - cwd: The current working directory.
+  ///   - env: A dictionary of the environment variables and their values. Usually of the parent shell.
+  ///   - filename: The name of the file we are resolving the architecture of.
+  /// - Returns: The architecture of the file which was found in the `Path`.
+  static func readWindowsExecutableArchitecture(
+    cwd: AbsolutePath?, env: [String: String], filename: String
+  ) -> Self {
+    let searchPaths = getEnvSearchPaths(pathString: env["Path"], currentWorkingDirectory: cwd)
```
jakepetroules: Environment on Windows is case-insensitive and this WILL lead to very subtle bugs in cases where `Path` happens to be cased as `PATH`.

Do we have an `Environment` abstraction available to SwiftDriver like we have in SwiftPM that we can use instead of a raw `[String: String]` dictionary?

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,122 @@
...
+import func TSCBasic.getEnvSearchPaths
+import func TSCBasic.lookupExecutablePath
+
+/// Check that the architecture of the toolchain matches the architecture
+/// of the Python installation.
+///
+/// When installing the x86 toolchain on ARM64 Windows, if the user does not
+/// install an x86 version of Python, they will get acryptic error message
+/// when running lldb (`0xC000007B`). Calling this function before invoking
+/// lldb gives them a warning to help troublshoot the issue.
```
jakepetroules: nit: "troublshoot" mis-spelled

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,122 @@
...
+import protocol TSCBasic.FileSystem
+import protocol TSCBasic.OutputByteStream
+import func TSCBasic.getEnvSearchPaths
+import func TSCBasic.lookupExecutablePath
+
+/// Check that the architecture of the toolchain matches the architecture
+/// of the Python installation.
+///
+/// When installing the x86 toolchain on ARM64 Windows, if the user does not
+/// install an x86 version of Python, they will get acryptic error message
```
jakepetroules: nit: "acryptic" => "a cryptic"

jakepetroules (changes requested):

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
artemcm: We have `TSCBasic.ProcessEnvironmentBlock` which I just transitioned most uses in the driver to: https://github.com/swiftlang/swift-driver/commit/fe4a66c541aea0bddbb996a27af83a9484c06dce. 

Though this is missing a bunch of good stuff that SPM's `Environment` has, it does use `ProcessEnvironmentKey` which appears to differentiate case sensitivity for Windows in particular. @charles-zablit we should switch to that here.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Fixed, thanks!

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Fixed, thanks!

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Switched to using `TSCBasic.ProcessEnvironmentBlock` 👍

charles-zablit:

charles-zablit:
@swift-ci please test

charles-zablit:

charles-zablit:
@swift-ci please test

charles-zablit:

charles-zablit:
@swift-ci please test

charles-zablit:

charles-zablit:
@swift-ci please test

charles-zablit:

charles-zablit:
I was able to track down the build issue on Windows to the `args` not being accessible. Reusing the same logic as `invocationRunMode` fixes the issue.

charles-zablit:
@swift-ci please test

charles-zablit:

charles-zablit:
Removed an unnecessary `throw`, now the CI passes.

charles-zablit:
> I think that I would strongly like to see the same change for macOS as well if this goes in. This is not specific to Windows - the other platforms have the same issue. Linux/FreeBSD do not as they don't have X64 emulation layers for ARM64 yet.

Added the same check on macOS by reading the Mach-O header.

I was only able to get the warning to show with my brew install of Python, as the `/usr/local/bin/python3` python is a universal binary.

```
> arch -x86_64 .build/x86_64-apple-macosx/debug/swift repl
warning: There is an architecture mismatch between the installed toolchain and the resolved Python's architecture:
Toolchain: x64
Python: arm64
Welcome to Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3).
Type :help for assistance.
```

charles-zablit:
@swift-ci please test

charles-zablit:

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
...
+    let pythonArchitecture = ExecutableArchitecture.readWindowsExecutableArchitecture(
+      cwd: cwd, envBlock: envBlock, filename: "python.exe")
+  #elseif os(macOS)
+    let pythonArchitecture = ExecutableArchitecture.readDarwinExecutableArchitecture(
+      cwd: cwd, envBlock: envBlock, filename: "python3")
+  #else
+    return
+  #endif
+
+  if pythonArchitecture == .universal {
```
artemcm: `#if` are evaluated early in the compilation pipeline so this code won't compile on a platform that's not Windows or macOS.

charles-zablit: Fixed by adding a `  #if os(Windows) || os(macOS)` in the function. Calling `checkIfMatchingPythonArch` on Linux is a no-op. Thanks!

charles-zablit:
@swift-ci please test

charles-zablit:

charles-zablit:

adrian-prantl (approved):

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -868,9 +868,16 @@ public struct Driver {
...
-    if case .subcommand = try Self.invocationRunMode(forArgs: args).mode {
+    let invocationMode = try Self.invocationRunMode(forArgs: args).mode
+    if case .subcommand = invocationMode {
       throw Error.subcommandPassedToDriver
     }
+
+    if case .normal(isRepl: true) = invocationMode {
+      checkIfMatchingPythonArch(
+        cwd: ProcessEnv.cwd, envBlock: envBlock, diagnosticsEngine: diagnosticsEngine)
+    }
```
compnerd, line 889: Why not use `switch` for this?

```swift
try switch Self.invocationRunMode(forArgs: args).mode {
case .subcommand:
  throw Error.subcommandPassedToDriver
case let .normal(isRepl: true):
  checkIfMatchingPythonArch(cwd: ProcessEnv.cwd, envBlock: envBlock,
                            diagnosticsEngine: diagnosticsEngine)
}
```

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
+import Foundation
```
compnerd, line 12: Missing copyright header

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
...
+/// When installing the x86 toolchain on ARM64 Windows, if the user does not
+/// install an x86 version of Python, they will get a cryptic error message
+/// when running lldb (`0xC000007B`). Calling this function before invoking
+/// lldb gives them a warning to help troubleshoot the issue.
+///
+/// - Parameters:
+///   - cwd: The current working directory.
+///   - env: The parent shell's ProcessEnvironmentBlock.
+///   - diagnosticsEngine: DiagnosticsEngine instance to use for printing the warning.
+public func checkIfMatchingPythonArch(
```
compnerd, line 41: I think that hoisting the `toolchainArchitecture` computation to the call site and passing that down would be better. `Process.ensurePython(at: AbsolutePath?, in: ProcessEnvironmentBlock, matches: ExecutableArchitecture, diagnostics: DiagnosticsEngine)` as a signature feels more natural.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
...
+
+/// Some of the architectures that can be stored in a COFF header.
+enum ExecutableArchitecture: String {
+  case x86 = "X86"
+  case x64 = "X64"
+  case arm64 = "ARM64"
+  case universal = "Universal"
+  case unknown = "Unknown"
+
+  static func fromPEMachineByte(machine: UInt16) -> Self {
```
compnerd: This is likely better to guard to `os(Windows)`.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
...
+    // https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#machine-types
+    switch machine {
+    case 0x014c: return .x86
+    case 0x8664: return .x64
+    case 0xAA64: return .arm64
+    default: return .unknown
+    }
+  }
+
+  static func fromMachoCPUType(cpuType: Int32) -> Self {
```
compnerd: This is probably better to guard under `os(macOS)`.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
...
+  case x86 = "X86"
+  case x64 = "X64"
+  case arm64 = "ARM64"
+  case universal = "Universal"
+  case unknown = "Unknown"
+
+  static func fromPEMachineByte(machine: UInt16) -> Self {
+    // https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#machine-types
+    switch machine {
+    case 0x014c: return .x86
```
compnerd: I would rather pull these constants from `WinSDK` (they have named constants for these).

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
...
+  #if os(Windows)
+    /// Resolves the filename from the `Path` environment variable and read its COFF header to determine the architecture
+    /// of the binary.
+    ///
+    /// - Parameters:
+    ///   - cwd: The current working directory.
+    ///   - env: A dictionary of the environment variables and their values. Usually of the parent shell.
+    ///   - filename: The name of the file we are resolving the architecture of.
+    /// - Returns: The architecture of the file which was found in the `Path`.
+    static func readWindowsExecutableArchitecture(
```
compnerd, line 112: This might be better suited as an extension on `Process`?

```swift
Process.executableArchitecture(atPath: String) -> ExecutableArchitecture
```

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
...
+      cwd: AbsolutePath?, envBlock: ProcessEnvironmentBlock, filename: String
+    ) -> Self {
+      let searchPaths = getEnvSearchPaths(
+        pathString: envBlock["Path"], currentWorkingDirectory: cwd)
+      guard
+        let filePath = lookupExecutablePath(
+          filename: filename, currentWorkingDirectory: cwd, searchPaths: searchPaths)
+      else {
+        return .unknown
+      }
```
compnerd, line 122: For Windows, this seems like a lot of complexity where you can just invoke `PathFindOnPathW` which allows you to augment the system paths as well. If you only want to search system paths, you can use the simpler `SearchPathW`.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
...
+
+      // Infering the architecture of a Windows executable from its COFF header involves the following:
+      // 1. Get the COFF header offset from the pointer located at the 0x3C offset (4 bytes long).
+      // 2. Jump to that offset and read the next 6 bytes.
+      // 3. The first 4 are the signature which should be equal to 0x50450000.
+      // 4. The last 2 are the machine architecture which can be infered from the value we get.
+      //
+      // The link below provides a visualization of the COFF header and the process to get to it.
+      // https://upload.wikimedia.org/wikipedia/commons/1/1b/Portable_Executable_32_bit_Structure_in_SVG_fixed.svg
+      fileHandle.seek(toFileOffset: 0x3C)
```
compnerd: Why the `fileHandle` approach? Why not use `Data(contentsOf:)` and access the raw data?

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,189 @@
...
+      guard let offsetPointer = try? fileHandle.read(upToCount: 4),
+        offsetPointer.count == 4
+      else {
+        return .unknown
+      }
+
+      let peHeaderOffset = offsetPointer.withUnsafeBytes { $0.load(as: UInt32.self) }
+
+      fileHandle.seek(toFileOffset: UInt64(peHeaderOffset))
```
compnerd: Can this not fail?

compnerd (commented):
This feels like it is going in the right direction. There are some things which can be done cleaner though.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,207 @@
...
+      let pythonArchitecture = Process.readWindowsExecutableArchitecture(
+        cwd: cwd, envBlock: envBlock, filename: "python.exe")
+    #elseif os(macOS)
+      let pythonArchitecture = Process.readDarwinExecutableArchitecture(
+        cwd: cwd, envBlock: envBlock, filename: "python3")
+    #endif
+
+    if pythonArchitecture == .universal {
+      return
+    }
```
compnerd: This is macOS specific, please hoist this into the `#if os(macOS)` condition.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,207 @@
...
+
+    if toolchainArchitecture != pythonArchitecture {
```
compnerd: I'd personally use `guard pythonArchitecture == toolchainArchitecture else {`.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
```diff
@@ -0,0 +1,207 @@
...
+  case universal = "Universal"
+  case unknown = "Unknown"
+
+  #if os(Windows)
+    static func fromPEMachineByte(machine: UInt16) -> Self {
+      // https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#machine-types
+      switch machine {
+      case 0x014c: return .x86
+      case 0x8664: return .x64
+      case 0xAA64: return .arm64
```
compnerd:
```suggestion
      case IMAGE_FILE_MACHINE_I386: return .x86
      case IMAGE_FILE_MACHINE_AMD64: return .x64
      case IMAGE_FILE_MACHINE_ARM64: return .arm64
```

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Fixed, thanks!

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Fixed, thanks!

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Fixed, thanks!

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Does `swift-driver` add other entries to the `PATH` implicitly? I wanted to reuse the same logic to be sure we would "see" the same entries as the driver.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: My only motivation to use `fileHandle` instead of `Data` is to avoid loading the whole file in memory. The python binaries are quite small but I still wanted to avoid loading multiple MBs in memory just to read the first bytes.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: `seek(toFileOffset:` cannot, but `seek(toOffset:` can. I replaced it because the former is now deprecated anyways.

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Switched to the WinSDK values, thanks!

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Fixed, thanks!

Sources/SwiftDriver/Utilities/PythonArchitecture.swift
charles-zablit: Fixed, thanks!

charles-zablit:
@swift-ci please test

charles-zablit:
@swift-ci please test

charles-zablit:
@swift-ci please test windows

jakepetroules:
@charles-zablit This change broke the iOS build:

PythonArchitecture.swift:102:11: error: cannot find type 'Process' in scope
extension Process {

charles-zablit:
> @charles-zablit This change broke the iOS build:
> 
> PythonArchitecture.swift:102:11: error: cannot find type 'Process' in scope extension Process {

I have opened a fix PR here: https://github.com/swiftlang/swift-driver/pull/1984

PR #1948, opened 2025-07-01
[6.2][Macro] Do not drop `in-process-plugin-server-path` option

cachemeifyoucan:
- **Explanation**: Fix a regression that `in-process-plugin-server-path` is accidentally dropped in explicit module build.
  
- **Scope**: Library macro plugins doesn't work when turn on explicit module and cause build failures.
  
- **Issues**: rdar://154780122
  
- **Original PRs**: https://github.com/swiftlang/swift-driver/pull/1947
  
- **Risk**: Low. Preserve an additional flag
  
- **Testing**: Unit Test
  
- **Reviewers**: @artemcm


cachemeifyoucan:
@swift-ci please test

PR #1949, opened 2025-07-02
Add `-scanner-prefix-map-paths` option and fix multiArg option parsing

sina-mahdavi:
The current command-line option for specifying prefix mappings for the dependency scanner does not handle paths containing equal signs. This PR adds a new option `-scanner-prefix-map-paths` that has a multiArg format instead of being `=`-separated and fixes this issue. The old `-scanner-prefix-map` option is still preserved, but is made driver-only and gets translated to the new option format by the driver.
This also fixes the parsing of `.multiArg` options in `OptionParsing.swift` so that their corresponding arguments are not added to the general inputs list and are attached to their option correctly.

Corresponding PR on Swift: https://github.com/swiftlang/swift/pull/82745

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
@@ -62,6 +62,9 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
   /// Does this compile support `.explicitInterfaceModuleBuild`
   private let supportsExplicitInterfaceBuild: Bool
 
+  /// status of the scanner prefix mapping option supported by the frontend
+  private let supportsScannerPrefixMapPaths: Bool
+
   /// Cached command-line additions for all main module compile jobs
   private struct ResolvedModuleDependenciesCommandLineComponents {
     let inputs: [TypedVirtualPath]
@@ -88,7 +91,8 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
               supportsExplicitInterfaceBuild: Bool = false,
               cas: SwiftScanCAS? = nil,
               prefixMap:  [(AbsolutePath, AbsolutePath)] = [],
-              supportsBridgingHeaderPCHCommand: Bool = false) throws {
+              supportsBridgingHeaderPCHCommand: Bool = false,
+              supportsScannerPrefixMapPaths: Bool = false) throws {
     self.dependencyGraph = dependencyGraph
     self.toolchain = toolchain
     self.integratedDriver = integratedDriver
@@ -96,6 +100,7 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
     self.reachabilityMap = try dependencyGraph.computeTransitiveClosure()
     self.supportsExplicitInterfaceBuild = supportsExplicitInterfaceBuild
     self.supportsBridgingHeaderPCHCommand = supportsBridgingHeaderPCHCommand
+    self.supportsScannerPrefixMapPaths = supportsScannerPrefixMapPaths
     self.cas = cas
     self.prefixMap = prefixMap
     let mainModuleId: ModuleDependencyId = .swift(dependencyGraph.mainModuleName)
@@ -216,7 +221,12 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
       // Add prefix mapping. The option is cache invariant so it can be added without affecting cache key.
       for (key, value) in prefixMap {
         commandLine.appendFlag("-cache-replay-prefix-map")
-        commandLine.appendFlag(value.pathString + "=" + key.pathString)
+        if supportsScannerPrefixMapPaths {
+          commandLine.appendFlag(value.pathString)
+          commandLine.appendFlag(key.pathString)
+        } else {
+          commandLine.appendFlag(value.pathString + "=" + key.pathString)
+        }
       }
 
       jobs.append(Job(
@@ -277,7 +287,12 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
       // Add prefix mapping. The option is cache invariant so it can be added without affecting cache key.
       for (key, value) in prefixMap {
         commandLine.appendFlag("-cache-replay-prefix-map")
-        commandLine.appendFlag(value.pathString + "=" + key.pathString)
+        if supportsScannerPrefixMapPaths {
+          commandLine.appendFlag(value.pathString)
+          commandLine.appendFlag(key.pathString)
+        } else {
+          commandLine.appendFlag(value.pathString + "=" + key.pathString)
+        }
       }
 
       jobs.append(Job(
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -1031,10 +1031,15 @@ extension Driver {
   }
 
   public mutating func addCacheReplayMapping(to commandLine: inout [Job.ArgTemplate]) {
-    if isCachingEnabled && isFrontendArgSupported(.scannerPrefixMap) {
+    if isCachingEnabled && isFrontendArgSupported(.cacheReplayPrefixMap) {
       for (key, value) in prefixMapping {
         commandLine.appendFlag("-cache-replay-prefix-map")
-        commandLine.appendFlag(value.pathString + "=" + key.pathString)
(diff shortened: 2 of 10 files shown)


Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -216,7 +216,8 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
       // Add prefix mapping. The option is cache invariant so it can be added without affecting cache key.
       for (key, value) in prefixMap {
         commandLine.appendFlag("-cache-replay-prefix-map")
-        commandLine.appendFlag(value.pathString + "=" + key.pathString)
+        commandLine.appendFlag(value.pathString)
```
cachemeifyoucan: swift-driver needs to be compatible with a range of swift compiler. I suggest
```
isFrontendArgSupported(.scannerPrefixMapPaths)
```
then do the new option, otherwise, join them with `=`

Sources/SwiftOptions/OptionParsing.swift
```diff
@@ -182,10 +182,8 @@ extension OptionTable {
           throw OptionParseError.missingArgument(
             index: index - 1, argument: argument)
         }
-        parsedOptions.addOption(option, argument: .multiple(Array()))
-        arguments[index..<endIdx].map { String($0) }.forEach { parsedOptions.addInput($0) }
+        parsedOptions.addOption(option, argument: .multiple(arguments[index..<endIdx].map { String($0) }))
```
cachemeifyoucan, line 185: Let's add a test for this fix.

cachemeifyoucan (commented):
Thanks for handling this.

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
sina-mahdavi: Done, I also did the same with the other places where I was only using `isFrontendArgSupported(.scannerPrefixMapPaths)`. Now I'm instead checking if the new option is supported, and if it's not checking if the old one is supported and using the old format instead of completely skipping that part.

Sources/SwiftOptions/OptionParsing.swift
sina-mahdavi: I found a test `testParsingMultiArg` that was already doing some basic tests on the old multiArg parsing, so I added some new assertions to that to test the new logic and make sure it's working correctly.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -1031,10 +1031,15 @@ extension Driver {
   }
 
   public mutating func addCacheReplayMapping(to commandLine: inout [Job.ArgTemplate]) {
-    if isCachingEnabled && isFrontendArgSupported(.scannerPrefixMap) {
+    if isCachingEnabled && (isFrontendArgSupported(.scannerPrefixMapPaths) || isFrontendArgSupported(.scannerPrefixMap)) {
```
cachemeifyoucan: I don't think this is needed. The check just needs to be functionally correct, and doesn't need to match the flag used. `scannerPrefixMap` predates `scannerPrefixMapPaths` so if it passed the check, the other check is not useful.

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -146,23 +147,23 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
   /// module's target version (PCM Args).
   /// 2. Generate jobs for all Clang modules, now that we have each module's set of PCM versions it must be
   /// built against.
-  public mutating func generateExplicitModuleDependenciesBuildJobs() throws -> [Job] {
+  public mutating func generateExplicitModuleDependenciesBuildJobs(driver: Driver) throws -> [Job] {
```
cachemeifyoucan: Didn't realize in the first review. I don't like passing `Driver` here. We can pre-compute all the flag queries when we initialize ExplicitDependencyBuildPlanner. I think you can do something like:
* if no prefix map option is supported, just don't set prefixMap
* set a `Bool` in the planner to suggest which option to use if prefix map is not empty.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
sina-mahdavi: I'm not sure if I understand this correctly. So you mean only `isFrontendArgSupported(.scannerPrefixMap)` would be enough here? These two options are mutually exclusive on the frontend because in the versions that support `-scanner-prefix-map-paths` I have made the old `-scanner-prefix-map` option driver-only. So if we just check for `.scannerPrefixMap` in the frontend, the check won't pass on the new versions on the frontend after the corresponding Swift PR for this.

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
sina-mahdavi: Sure! If none of the options are supported, `prefixMap` will already be empty so the for loops that check for these options will not get executed any way. Otherwise if at least one of the options is supported, I changed it to pass a boolean to the constructor in the new commit instead of passing the whole driver object for keeping the supported option type.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
cachemeifyoucan: Good call. How about checking `.cacheReplayPrefixMap` instead.?

sina-mahdavi: Right! I think that would work. Did `.scannerPrefixMap` and `.cacheReplayPrefixMap` start being supported at the same time? If so we can replace the two conditions with just that.

sina-mahdavi: I changed this (and another similar case in `Driver.swift`) to use `isFrontendArgSupported(.cacheReplayPrefixMap)` instead of `isFrontendArgSupported(.scannerPrefixMap) || isFrontendArgSupported(.scannerPrefixMapPaths)`.

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -9,6 +9,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
+import SwiftOptions
```
cachemeifyoucan: This is no longer needed.

cachemeifyoucan (approved):
LGTM with one small comment. 

We can't merge this till landing the swift compiler change.

cachemeifyoucan:
@swift-ci please test

sina-mahdavi:
> LGTM with one small comment.
> 
> We can't merge this till landing the swift compiler change.

Thank you! Also, should I squash my commits into one before merging?

sina-mahdavi:
@swift-ci please test

cachemeifyoucan:
> Thank you! Also, should I squash my commits into one before merging?

Yes. Please never push individual review feedback fixes to any repositories. You can either use `squash and merge` or manually update the PR with a merged commit.

sina-mahdavi:
@swift-ci please test

sina-mahdavi:

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@sina-mahdavi  Running the test again with the compiler after your swift change landed results in test failure.

I think it is a purely a test problem, where we check for all paths are prefixed mapped (so there should be no arguments start with mapped path) but since you split the prefix map option, now there are arguments (I think it is just prefix map option) that still contains path with the prefix T_T

sina-mahdavi:
> @sina-mahdavi Running the test again with the compiler after your swift change landed results in test failure.
> 
> I think it is a purely a test problem, where we check for all paths are prefixed mapped (so there should be no arguments start with mapped path) but since you split the prefix map option, now there are arguments (I think it is just prefix map option) that still contains path with the prefix T_T

I see, I think this last change should fix that, I'm skipping the arguments to `-cache-replay-prefix-map` when searching for the unmapped paths.

sina-mahdavi:
@swift-ci please test

sina-mahdavi:

sina-mahdavi:
@cachemeifyoucan I think the failing tests are fixed now 😄

cachemeifyoucan:
LGTM. I will merge for you.

PR #1954, opened 2025-07-07
[Tests] Do not share underlying Oracle for caching tests

cachemeifyoucan:
After rewriting the scanning service, oracle can get a race condition when CAS configuration is changed in between builds. It causes clang dependency scanning using a wrong CAS when scanning after some other thread updated the CAS location.

Fix the occasional race condition by not sharing the orcale between all the tests sicne this is not a valid real world configuration. In the future, we should teach clang scanner not to hold a global CAS instance so service can be shared when switching CAS locations in a racy way.

rdar://154592458

### DIFF
--- a/Tests/SwiftDriverTests/CachingBuildTests.swift
+++ b/Tests/SwiftDriverTests/CachingBuildTests.swift
@@ -193,8 +193,6 @@ private func checkCachingBuildJobDependencies(job: Job,
 
 
 final class CachingBuildTests: XCTestCase {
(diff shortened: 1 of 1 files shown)


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:

cachemeifyoucan:

cachemeifyoucan:

cachemeifyoucan:

PR #1955, opened 2025-07-08
Change Triple.isiOS to no longer match tvOS

jakepetroules:
This behavior is confusing and leads to wrong answers in a number of places. All usage sites of isiOS have been audited to ensure they're now correct.

- unsupportedTargetVariant was showing an error message including -target-variant, which should have only matched iOS and not tvOS
- getDefaultDwarfVersion checked isiOS in two places, but just so happened to always get the right answer anyways (tvOS's first version was 9, so checking for tvOS < 9 was nonsensical and didn't matter, and iOS and tvOS version numbers have always been aligned, so a < 18 check was right for both)

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple.swift
+++ b/Sources/SwiftDriver/Utilities/Triple.swift
@@ -1473,19 +1473,16 @@ extension Triple.OS {
     self == .aix
   }
 
-  /// isMacOSX - Is this a Mac OS X triple. For legacy reasons, we support both
-  /// "darwin" and "osx" as OS X triples.
+  /// Is this an Apple macOS triple.
+  /// - note: For legacy reasons, we support both "darwin" and "macosx" as macOS triples.
   public var isMacOSX: Bool {
     self == .darwin || self == .macosx
   }
 
-  /// Is this an iOS triple.
-  /// Note: This identifies tvOS as a variant of iOS. If that ever
-  /// changes, i.e., if the two operating systems diverge or their version
-  /// numbers get out of sync, that will need to be changed.
-  /// watchOS has completely different version numbers so it is not included.
+  /// Is this an Apple iOS triple.
+  /// - note: Contrary to historical behavior with regard to LLVM's Triple type, this does NOT match tvOS in order to avoid confusion moving forward.
   public var isiOS: Bool {
-    self == .ios || isTvOS
+    self == .ios
   }
 
   /// Is this an Apple tvOS triple.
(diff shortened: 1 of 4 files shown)


Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -1656,6 +1656,32 @@ extension Triple {
     }
   }
 
+  /// Parse the version number as with getOSVersion.  This should
+  /// only be called with tvOS or generic triples.
+  ///
+  /// This accessor is semi-private; it's typically better to use `version(for:)` or
+  /// `Triple.FeatureAvailability`.
+  public var _tvOSVersion: Version {
```
jakepetroules, line 1664: This part (and changes to _iOSVersion) could use extra scrutiny, I'm not super clear what the true intent is here in terms of what's supposed to be able to map to what.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test windows

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -1640,13 +1638,15 @@ extension Triple {
...
-    case .ios, .tvos:
+    case .ios:
       var version = self.osVersion
       // Default to 5.0 (or 7.0 for arm64).
       if version.major == 0 {
         version.major = arch == .aarch64 ? 7 : 5
       }
       return version
+    case .tvos:
+      return osVersion
```
artemcm, line 1649: Should this become `fatalError("conflicting triple info")` like `.watchos`?

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -1656,6 +1656,32 @@ extension Triple {
...
+    case .darwin, .macosx:
+      // Ignore the version from the triple.  This is only handled because the
+      // the clang driver combines OS X and iOS support into a common Darwin
+      // toolchain that wants to know the iOS version number even when targeting
+      // OS X.
+      return Version(9, 0, 0)
+    case .ios, .tvos:
+      var version = self.osVersion
+      if version.major == 0 {
+        version.major = 9
```
artemcm, line 1676: Could you add a comment for why we default to `9`?

artemcm:
I believe we must also update `getDefaultDwarfVersion`.

Sources/SwiftDriver/Utilities/Triple.swift
jakepetroules: That fails some existing tests; would rather leave it for now.

Sources/SwiftDriver/Utilities/Triple.swift
jakepetroules: Done!

jakepetroules:
> I believe we must also update `getDefaultDwarfVersion`.

From PR description:

getDefaultDwarfVersion checked isiOS in two places, but just so happened to always get the right answer anyways (tvOS's first version was 9, so checking for tvOS < 9 was nonsensical and didn't matter, and iOS and tvOS version numbers have always been aligned, so a < 18 check was right for both)

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test windows

PR #1959, opened 2025-07-22
[ToolingInterface] Fix `V2` API should forward `forceNoOutputs` argument

xedin:
While working on `swift-ide-test` integration I noticed that the `swift_getSingleFrontendInvocationFromDriverArgumentsV2` API doesn't forward `forceNoOutputs` argument.


xedin:
@swift-ci please test

PR #1960, opened 2025-07-22
[ToolingInterface] Fix `getSingleFrontendInvocationFromDriverArgumentsV3`

xedin:
... to avoid adding driver path to arguments

The path to the driver is passed down already and it would be added to the arguments by `getSingleFrontendInvocationFromDriverArgumentsV5`.


xedin:
@swift-ci please test

artemcm (approved):

artemcm:
Thanks!

PR #1961, opened 2025-07-29
CrashReproducer generation improvements

cachemeifyoucan:
Add some improvements to crash reproducer:
* Allow environmental variable `SWIFT_CRASH_DIAGNOSTICS_DIR` to specify where the reproducer is written to.
* Add a testcase for generation reproducer since we can direct reproducer into a directory that can be cleared after testcase
* Do use parseable output when producing reproducer (to avoid polluting the stdout where build system might except build commands)


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
ping

artemcm (approved):

PR #1962, opened 2025-07-30
Swift Driver: Unknown Subcommand Location

etcwilde:
Include the path that the Swift driver is looking for a subcommand in the error message. This makes it easier to see where the driver is looking, which isn't always next to the driver.

This is only applied to the swift-legacy-driver and swiftc-legacy-driver symlinks.


etcwilde:
@swift-ci please test

artemcm (approved):

PR #1963, opened 2025-08-01
[ArgsResolver] Create specified temporary directory, if necessary

artemcm:
Resolves rdar://157302464

### DIFF
--- a/Sources/SwiftDriver/Execution/ArgsResolver.swift
+++ b/Sources/SwiftDriver/Execution/ArgsResolver.swift
@@ -45,6 +45,10 @@ public final class ArgsResolver {
 
     if let temporaryDirectory = temporaryDirectory {
       self.temporaryDirectory = temporaryDirectory
+      let temporaryDirExists = try fileSystem.exists(self.temporaryDirectory)
+      if !temporaryDirExists, let temporaryDirAbsPath = self.temporaryDirectory.absolutePath {
+        try fileSystem.createDirectory(temporaryDirAbsPath, recursive: true)
+      }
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci test

Sources/SwiftDriver/Execution/ArgsResolver.swift
```diff
@@ -45,6 +45,10 @@ public final class ArgsResolver {
 
     if let temporaryDirectory = temporaryDirectory {
       self.temporaryDirectory = temporaryDirectory
+      let temporaryDirExists = try fileSystem.exists(self.temporaryDirectory)
+      if !temporaryDirExists, let temporaryDirAbsPath = self.temporaryDirectory.absolutePath {
+        try fileSystem.createDirectory(temporaryDirAbsPath, recursive: true)
```
owenv, line 50: Should this be try? in case the directory was created after the check on the line above?

artemcm: The `createDirectory` implementation accounts for this and will not throw if it encounters the failure based on the fact that a racing thread may have created this directory already:
https://github.com/swiftlang/swift-tools-support-core/blob/c4309ae736691e0baf7d47268cd1912e44f1db30/Sources/TSCBasic/FileSystem.swift#L500

So it will only throw if creation failed for some unrelated reason, which I think we should continue letting it do.

owenv: 👍

PR #1965, opened 2025-08-07
[Build] Fix linking of llbuild

drexin:
In the cmake build, the lilbrary to link was specified as `libllbuild`, which caused the build to pass `-llibllbuild` to the linker, instead of `-lllbuild`.

### DIFF
--- a/Sources/SwiftDriverExecution/CMakeLists.txt
+++ b/Sources/SwiftDriverExecution/CMakeLists.txt
@@ -17,7 +17,7 @@ target_link_libraries(SwiftDriverExecution PUBLIC
   SwiftOptions
   SwiftDriver)
 target_link_libraries(SwiftDriverExecution PRIVATE
-  libllbuild
+  llbuild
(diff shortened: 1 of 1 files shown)


drexin:
@swift-ci test

artemcm (approved):

compnerd:
The original spelling is *correct*. The problem is that `llbuild` is the executable target and the library target is `libllbuild`.

See https://github.com/swiftlang/swift-llbuild/blob/main/products/llbuild/CMakeLists.txt#L4
https://github.com/swiftlang/swift-llbuild/blob/main/products/libllbuild/CMakeLists.txt#L10

The fact that `-llibllbuild` is correct given the name of the target (`liblibllbuild`).  However, the output name is tweaked to be renamed to `libllbuild` at https://github.com/swiftlang/swift-llbuild/blob/main/products/libllbuild/CMakeLists.txt#L13.

This change broke the early swift-driver on windows.

PR #1966, opened 2025-08-07
Revert "[Build] Fix linking of llbuild"

compnerd:
Reverts swiftlang/swift-driver#1965

This fixes the build on Windows

### DIFF
--- a/Sources/SwiftDriverExecution/CMakeLists.txt
+++ b/Sources/SwiftDriverExecution/CMakeLists.txt
@@ -17,7 +17,7 @@ target_link_libraries(SwiftDriverExecution PUBLIC
   SwiftOptions
   SwiftDriver)
 target_link_libraries(SwiftDriverExecution PRIVATE
-  llbuild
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

drexin:
Can we maybe have a version that works on all systems? Because reverting this breaks local builds for me.

compnerd:
Hmm, how are you wiring up the dependency for llbuild to swift-driver? That should work properly because `libllbuild` should point to the CMake target, which will set up the linking correctly. Are you trying to build without building the dependencies?

compnerd:
As an aside - we should rename the llbuild targets so that this can be avoided.

compnerd:
https://github.com/swiftlang/swift-llbuild/pull/1005 should fix the naming issue

compnerd:
Okay - we should not merge this if we merge https://github.com/swiftlang/swift-llbuild/pull/1005 as that renames the library target.

PR #1972, opened 2025-08-19
PrebuiltModuleGen: dedupe prebuilt module jobs

nkcsgexi:
This prevents issues that we can find same module in multiple different locations.

rdar://158615938

### DIFF
--- a/Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
+++ b/Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
@@ -812,13 +812,13 @@ extension Driver {
       return (key, outputPaths)
     })
 
-    func collectSwiftModuleNames(_ ids: [ModuleDependencyId]) -> [String] {
-      return ids.compactMap { id in
+    func collectUniqueSwiftModuleNames(_ ids: [ModuleDependencyId]) -> [String] {
+      return Array(Set(ids.compactMap { id in
         if case .swift(let module) = id {
           return module
         }
         return nil
-      }
+      }))
     }
 
     func getSwiftDependencies(for module: String) -> [String] {
@@ -827,7 +827,7 @@ extension Driver {
       guard !dependencies.isEmpty else {
         return []
       }
-      return collectSwiftModuleNames(dependencies)
+      return collectUniqueSwiftModuleNames(dependencies)
     }
 
     func getOutputPaths(withName modules: [String], loadableFor arch: Triple.Arch) throws -> [TypedVirtualPath] {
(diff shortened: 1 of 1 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
```diff
@@ -772,6 +772,21 @@ extension Driver {
     return allJobs
   }
 
+  /// Dedupe jobs by output paths
+  func dedupeJobs(_ jobs: [Job]) -> [Job] {
+    var seen: Set<String> = []
+    var result: [Job] = []
+    for job in jobs {
+      assert(job.outputs.count == 1)
```
artemcm: Would be worthwhile to also assert on job kind being an emit module from interface here.

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
```diff
@@ -772,6 +772,21 @@ extension Driver {
...
+      let key = job.outputs.first!.file.absolutePath!.pathString
```
artemcm: Instead of `file`, it should be sufficient to have the hash set operate on `fileHandle`, as those will map to the same underlying absolute path string, and it won't be necessary to do `.absolutePath!.pathString`.

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
```diff
@@ -772,6 +772,21 @@ extension Driver {
     return allJobs
   }
 
+  /// Dedupe jobs by output paths
+  func dedupeJobs(_ jobs: [Job]) -> [Job] {
```
artemcm: Nit: something like `dedupeCompileSwiftModuleFromInterfaceJobs` would be a more-descriptive name

artemcm (approved):

Sources/SwiftDriver/Jobs/PrebuiltModulesJob.swift
nkcsgexi: I actually changed the fix of deduping to an earlier step. Mind taking another look?

nkcsgexi:
@swift-ci please smoke test

nkcsgexi:
@swift-ci please smoke test

nkcsgexi:
@swift-ci please smoke test

nkcsgexi:
@swift-ci please smoke test

nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test Windows

nkcsgexi:
@swift-ci please test Windows

nkcsgexi:
@swift-ci please clean test Windows

PR #1978, opened 2025-08-28
Always diagnose failures in job execution

bnbarham:
Execution of the `MultiJobExecutor` throws a
`Driver.ErrorDiagnostics.emitted` if the job was not successful, ie. it assume any errors were already emitted. Make sure to emit an error in all paths of `ExecuteJobRule.executeJob` so that failures aren't hidden.

### DIFF
--- a/Sources/SwiftDriverExecution/MultiJobExecutor.swift
+++ b/Sources/SwiftDriverExecution/MultiJobExecutor.swift
@@ -655,9 +655,8 @@ class ExecuteJobRule: LLBuildRule {
       context.cancelBuildIfNeeded(result)
       value = .jobExecution(success: success)
     } catch {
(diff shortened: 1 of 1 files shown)


bnbarham:
@swift-ci please test

bnbarham:
Interested in opinions for what this message should be - right now this becomes an `AnyDiagnostic` which then just prints the underlying error. But in the case that it isn't a `DiagnosticData`, we could also add something extra to it, eg. `Internal issue: ''. Please file an issue at ...` or something similar.

(This came up in investigating the fatalError we were seeing sometimes on Windows: https://github.com/swiftlang/swift/issues/83606#issuecomment-3229837312)

owenv (approved):

PR #1979, opened 2025-09-02
Allow multiple files to be indexed in a single invocation using the `-index-file` driver mode

ahoppen:
This allows us to index multiple files in a single frontend invocation during background indexing, allowing us to share common work like module loading.

One thing that I’m not sure about is that in the previously existing single file indexing mode, we were adding `-o <module-name>.indexData` to the frontend invocation and I couldn’t figure out what that’s needed for. With the multi-file indexing mode we add that argument for every single file to index, eg. in my test case we have `-o main.indexData -o main.indexData`. As far as I can tell this does no harm but it’s worth considering during review.

### DIFF
--- a/Sources/SwiftDriver/Jobs/CompileJob.swift
+++ b/Sources/SwiftDriver/Jobs/CompileJob.swift
@@ -107,7 +107,7 @@ extension Driver {
   /// corresponding primary set of outputs and, if not identical, the output
   /// paths to record in the index data (empty otherwise).
   mutating func addCompileInputs(primaryInputs: [TypedVirtualPath],
-                                 indexFilePath: TypedVirtualPath?,
+                                 indexFilePaths: [TypedVirtualPath],
                                  inputs: inout [TypedVirtualPath],
                                  inputOutputMap: inout [TypedVirtualPath: [TypedVirtualPath]],
                                  outputType: FileType?,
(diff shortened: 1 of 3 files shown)


artemcm (approved):

ahoppen:
@swift-ci Please test

Sources/SwiftDriver/Jobs/CompileJob.swift
```diff
@@ -149,11 +149,11 @@ extension Driver {
       assert(!primaryInputs.isEmpty)
       usesPrimaryFileInputs = true
       primaryInputFiles = primaryInputs
-    } else if let path = indexFilePath {
+    } else if !indexFilePaths.isEmpty {
       // If -index-file is used, we perform a single compile but pass the
       // -index-file-path as a primary input file.
```
bnbarham, line 154: Comments a little out of date now, but the gist is still the same (just a single -> plural really)

bnbarham:
> we were adding -o <module-name>.indexData to the frontend invocation

I wonder if we needed a concrete output at some point and this was some workaround? If you're not seeing errors without it, I'd be fairly tempted to remove?

ahoppen:
> I wonder if we needed a concrete output at some point and this was some workaround? If you're not seeing errors without it, I'd be fairly tempted to remove?

I just tried to remove this and don’t feel confident in doing so. Too much seems to hinge on the fact that we have an output for each primary input file and I don’t know the driver well enough to fully understand what the consequences of removing the output might be. So, I would propose taking this change as-is. Removing the output path could be a separate PR but I’m currently not planning to tackle this.

PR #1981, opened 2025-09-09
Update `testExecutableFallbackPath` to pass on all platforms

bnbarham:
Amazon Linux 2023 includes a dsymutil in its LLVM package, which is installed on our builders (checking separately if that is actually required). The intent of this test seems to be mostly to make sure we are able to find all tools when allowing fallbacks, so update it to just check for that.


bnbarham:
@swift-ci please test

artemcm (approved):

PR #1984, opened 2025-09-12
gate Process extension behind preprocessor check

charles-zablit:
This patch fixes a build issue on iOS introduced by https://github.com/swiftlang/swift-driver/pull/1942.

The extension is only present on macOS and Windows and will be absent on iOS.


charles-zablit:
@swift-ci please test

PR #1987, opened 2025-09-15
[wasm] Pass `--table-base` to linker to reserve low function addresses

kateinoigakukun:
WebAssembly does not have a reserved address space by default, so we need to explicitly reserve low addresses for extra inhabitants for enum types with pointer payloads. https://github.com/swiftlang/swift/pull/39300 added `--global-base` to reserve low data addresses, but we also need to reserve low function addresses with `--table-base` for function pointers because WebAssembly uses a separate address space for function pointers.

Corresponding legacy-driver change: https://github.com/swiftlang/swift/pull/84285


kateinoigakukun:
@swift-ci test

PR #1988, opened 2025-09-15
[6.2][wasm] Pass `--table-base` to linker to reserve low function addresses

kateinoigakukun:
Explanation: Fix memory corruption on WebAssembly by properly reserving low-function addresses
Original PRs: https://github.com/swiftlang/swift-driver/pull/1987
Risk: Low, this only removes a spurious error for library evolution.
Testing: Ensured at desk that the minimum problematic code won't perform bad memory access with ASan instrumentation
Reviewers: TBD

### DIFF
--- a/Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
@@ -160,8 +160,11 @@ extension WebAssemblyToolchain {
       // The value of lowest valid address, called "global base", must be always
       // synchronized with `SWIFT_ABI_WASM32_LEAST_VALID_POINTER` defined in
       // apple/swift's runtime library.
+      let SWIFT_ABI_WASM32_LEAST_VALID_POINTER = 4096
       commandLine.appendFlag(.Xlinker)
-      commandLine.appendFlag("--global-base=4096")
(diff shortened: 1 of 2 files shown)


Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
```diff
@@ -160,8 +160,11 @@ extension WebAssemblyToolchain {
       // The value of lowest valid address, called "global base", must be always
       // synchronized with `SWIFT_ABI_WASM32_LEAST_VALID_POINTER` defined in
       // apple/swift's runtime library.
+      let SWIFT_ABI_WASM32_LEAST_VALID_POINTER = 4096
```
DougGregor, line 163: Doesn't need to be this PR specifically, but should we export this value as part of the target info that the frontend prints? That way, we don't need to have the hardcoded value duplicated across code bases.

kateinoigakukun: Sounds like a good idea, let me make a follow up change on the main branch!

DougGregor: Awesome, thank you! I don't think we need to cherry-pick the change to `release/6.2` or hold up this PR for it, but it would be excellent future-proofing and will help with other platforms where we might need to propagate this information.

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test Windows

PR #1989, opened 2025-09-15
Use UUID for response file names and add 'resolveArgumentList' API which contains original command-line if a response file is used

artemcm:
Using the content hash in the response file name caused us to require using the atomic FS write API, due to a possibility that concurrently-running driver processes may encounter a need to build a common, shared module dependency with an identical command-line recipe. In such a scenario there was a possibility of a race in writing out the response file for this command. 

We have since seen challenges with the current atomic FS write APIs and this PR is an alternative approach to tackling this problem. Instead of using the content hash, we will use UUID to ensure that we do not have races when multiple drivers are creating a response file for an identical command. In addition, we add an API for build system clients so that they can directly examine the pre-response-file-creation command line arguments directly from the `ArgsResolver`, without needing to read the response file from the filesystem.

### DIFF
--- a/Sources/SwiftDriver/Execution/ArgsResolver.swift
+++ b/Sources/SwiftDriver/Execution/ArgsResolver.swift
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import class Foundation.NSLock
+import struct Foundation.UUID
 
 import func TSCBasic.withTemporaryDirectory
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci test

owenv (approved):

artemcm:
@swift-ci test

artemcm:

PR #1993, opened 2025-09-22
Handle "-internal-import-bridging-header"

DougGregor:
The bridging header is effectively imported as a "public" import. This new option makes it an "internal" import, so that entities accessible via the bridging header cannot be part of public API.

Driver part of rdar://74011750.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -353,6 +353,9 @@ public struct Driver {
   /// Original ObjC Header passed from command-line
   let originalObjCHeaderFile: VirtualPath.Handle?
 
+  /// Whether to import the bridging header as internal (vs public).
+  let importBridgingHeaderAsInternal: Bool
+
   /// Enable bridging header chaining.
   let bridgingHeaderChaining: Bool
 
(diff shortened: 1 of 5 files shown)


DougGregor:
https://github.com/swiftlang/swift/pull/84410

@swift-ci please test

tshortli (approved):

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -1165,10 +1168,12 @@ public struct Driver {
...
     self.producePCHJob = maybeNeedPCH
 
-    if let objcHeaderPathArg = parsedOptions.getLastArgument(.importObjcHeader) {
-      self.originalObjCHeaderFile = try? VirtualPath.intern(path: objcHeaderPathArg.asSingle)
+    if let importBridgingHeaderOption = parsedOptions.last(for: .importBridgingHeader, .internalImportBridgingHeader) {
+      self.originalObjCHeaderFile = try? VirtualPath.intern(path: importBridgingHeaderOption.argument.asSingle)
+      self.importBridgingHeaderAsInternal = importBridgingHeaderOption.option == .internalImportBridgingHeader
     } else {
       self.originalObjCHeaderFile = nil
+      self.importBridgingHeaderAsInternal = false
```
cachemeifyoucan, line 1176: I think this is not quite correct. I think it maybe need to be internal if library evolution is enabled.

The special case that falls here is that the current module doesn't have a bridging header but there is dependency swift binary module that has bridging header, thus scanner can create a chained bridging for current module. In this case, there isn't a bridging header import flag that indicates if current module does internal import or not.

DougGregor:
@swift-ci please test Windows

Sources/SwiftDriver/Driver/Driver.swift
DougGregor: I don't think I understand how to construct this special case. In the place where the driver used to always generate `-import-pch` for the chained header, it now checks whether the bridging header is internal and will pass the new `-internal-import-pch` instead (which treats the PCH as internal-only). Is that not sufficient?

cachemeifyoucan: A swift testcase for this is: https://github.com/swiftlang/swift/blob/main/test/ScanDependencies/bridging-header-autochaining.swift#L51-L55

Requirement is:
* Module has no bridging header
* Binary module dependency has bridging header
* Enable auto chaining. 

In this case, driver invocation will not have `-*import-bridging-header`, swift-driver needs to figure out which `-import-pch` flag flavor to pass the PCH flag to swift-frontend.

DougGregor: When the binary module dependency has an internal bridging header (the new thing), it's not recorded that it has a bridging header at all, so there is no chaining. No?

cachemeifyoucan: In my example, internal bridging header is not using internal bridging header, but the legacy bridging header.

PR #1994, opened 2025-09-26
[Incremental Builds]skip startingPoint of changed .swiftmodule

WeijunDeng:
If a change in the fingerprint of an interface causes a change in the swiftmodule, but other swiftmodules that depend on this swiftmodule do not depend on this interface, then those other swiftmodules should not be recompiled.

This means that recompilation is only triggered when the fingerprint of a dependent interface changes.

For example, adding a new public class will not cause the dependent swiftmodules to recompile.


WeijunDeng:
close and see another PR #2001

PR #1995, opened 2025-09-29
Add -sil-output-dir and -ir-output-dir driver options

rjmansfield:
This enables generating SIL and LLVM IR files during normal compilation using the frontend -sil-output-path and -ir-output-path flags.

Supports both single-file and WMO modes with ile map integration.

rdar://160297898

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -667,8 +667,32 @@ extension Driver {
                                                         moduleOutputInfo: ModuleOutputInfo,
                                                         moduleOutputPaths: SupplementalModuleTargetOutputPaths,
                                                         includeModuleTracePath: Bool,
-                                                        indexFilePaths: [TypedVirtualPath]) throws -> [TypedVirtualPath] {
+                                                        indexFilePaths: [TypedVirtualPath],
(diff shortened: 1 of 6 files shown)


rjmansfield:
https://github.com/swiftlang/swift/pull/84392 is the PR for the swift-frontend flags

rjmansfield:
@swift-ci please test

rjmansfield:
@swift-ci please test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -763,6 +810,30 @@ extension Driver {
         finalOutputPath: serializedDiagnosticsFilePath,
         input: input,
         flag: "-serialize-diagnostics-path")
+
+      // Add SIL and IR outputs when explicitly requested via directory options, file maps, or -save-temps
+      let saveTempsWithoutFileMap = parsedOptions.hasArgument(.saveTemps) && outputFileMap == nil
```
DougGregor, line 815: SwiftDriver is meant to provide some forward/backward compatibility with different swift-frontend versions. In this case where `-save-temps` has been provided, can this check whether the frontend supports the `-sil-output-dir` option before it it concludes that it should pass the option down? Otherwise, new driver with old swift-frontend will fail for existing users of `-save-temps`.

rjmansfield:
@swift-ci please test

rjmansfield:
@swift-ci please test

PR #1997, opened 2025-09-30
Fix optimization record path handling in primary file compilation mode

rjmansfield:
When using -save-optimization-record-path in primary file mode, the user
 provided path was being ignored and used a derived path instead.
-save-optimization-record-path was working correctly in WMO mode due to taking a different code path.


rjmansfield:
@swift-ci please test

rjmansfield:
@swift-ci please test

rjmansfield:
@swift-ci please test

PR #1998, opened 2025-09-30
Fix swift driver hanging on SIGINT by adding timeout forced exit

rjmansfield:
When users interrupt swift driver with ctrl+c, the llbuild engine.build() call can hang indefinitely even after child processes are terminated. Add a two second timeout to force exit if graceful termination fails.

Resolves https://github.com/swiftlang/swift-driver/issues/1798


rjmansfield:
@swift-ci please test

PR #2001, opened 2025-10-14
[Incremental Builds]When a file changed, do not affect nodes with a fingerprint that have not changed.

WeijunDeng:
When a file changed, do not affect nodes with a fingerprint that have not changed.

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraphParts/Tracer.swift
@@ -86,6 +86,15 @@ extension ModuleDependencyGraph.Tracer {
 
     // If this use also provides something, follow it
     for use in graph.nodeFinder.uses(of: definition) {
+      if (definition.key.aspect == .interface) {
+        // .swift or .swiftmodule
+        if case .sourceFileProvide(name: _) = definition.key.designator {
+          if use.keyAndFingerprint.fingerprint != nil {
+            // When a file changed, do not affect nodes that have a fingerprint and have not changed.
+            continue
+          }
+        }
+      }
       collectNextPreviouslyUntracedDependent(of: use)
     }
     traceDeparture(pathLengthAfterArrival);


WeijunDeng:
@artemcm

artemcm:
Could you please add a test of a scenario this affects?

WeijunDeng:
### Case 1
**Before:**
```
// A.swift
public class A {
  public let b = B()
}
```
```
// B.swift
public class B {
}
```
**After:**
```
// B.swift
public class B {
}
public class B_2 {
}
```
**Now:**
remark: Incremental compilation: Traced: interface of source file from B.swift.swiftdeps in B.swift -> interface of top-level name 'B' in B.swift -> implementation of source file from A.swift.swiftdeps in A.swift
**Expected:**
Do not recompile A.swift 🤔 （This PR can fix this case.）

### Case 2
**Before：**
```
// A.swift
foo(123)
func foo(_: Any) { print("Hello Any") }
```
```
// B.swift
// func foo(_: Int) { print("Hello Int") }
```
**After：**
```
// B.swift
func foo(_: Int) { print("Hello Int") }
```
**Now(Expected):**
remark: Incremental compilation: Traced: interface of source file from B.swift.swiftdeps in B.swift -> interface of top-level name 'foo' in B.swift -> implementation of source file from A.swift.swiftdeps in A.swift

### Case 3
**Before:**
```
// A.swift
func a() {
    if (1==1) {
    }
}
```
```
// B.swift
public class B {
//    static func == (lhs: B, rhs: B) -> Bool {
//        return true
//    }
}
```
**After:**
```
// B.swift
public class B {
    static func == (lhs: B, rhs: B) -> Bool {
        return true
    }
}
```
**Now:**
remark: Incremental compilation: Traced: interface of top-level name '==' in B.swift -> implementation of source file from A.swift.swiftdeps in A.swift
**Expected:**
Do not recompile A.swift 🤔 （This PR can not fix this case. I am debugging.）

WeijunDeng:
> Could you please add a test of a scenario this affects?

please see above. @artemcm

PR #2013, opened 2025-11-12
Added functionality to -register-module-dependency flag

sepy97:
rdar://164596852
This flag tells the driver to include the module in the dependency scanning, but not load it. Added test for the flag.

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
@@ -145,6 +145,14 @@ public extension Driver {
       commandLine.appendFlag(.dependencyScanCacheRemarks)
     }
 
+    // command-line flag registers a dependency with the dependency scanner
+    // without causing it to be implicitly imported in compilation tasks
+    for moduleNameArg in parsedOptions.arguments(for: .registerModuleDependency) {
+      let moduleName = moduleNameArg.argument.asSingle
+      commandLine.appendFlag(.importModule)
+      commandLine.appendFlag(moduleName)
+    }
+
     if shouldAttemptIncrementalCompilation &&
        parsedOptions.contains(.incrementalDependencyScan) {
       if let serializationPath = buildRecordInfo?.dependencyScanSerializedResultPath {
--- a/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
+++ b/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
@@ -803,6 +803,82 @@ final class ExplicitModuleBuildTests: XCTestCase {
     }
   }
 
+  func testRegisterModuleDependencyFlag() throws {
+    let (stdlibPath, shimsPath, _, _) = try getDriverArtifactsForScanning()
+    try withTemporaryDirectory { path in
+      try localFileSystem.changeCurrentWorkingDirectory(to: path)
+      let moduleCachePath = path.appending(component: "ModuleCache")
+      try localFileSystem.createDirectory(moduleCachePath)
+      let main = path.appending(component: "testRegisterModuleDependency.swift")
+      // Note: We're NOT importing module E in the source code
+      try localFileSystem.writeFileContents(main, bytes:
+        """
+        import A;
+        """
+      )
+      let cHeadersPath: AbsolutePath =
+      try testInputsPath.appending(component: "ExplicitModuleBuilds")
+        .appending(component: "CHeaders")
+      let swiftModuleInterfacesPath: AbsolutePath =
+      try testInputsPath.appending(component: "ExplicitModuleBuilds")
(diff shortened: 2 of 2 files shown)


Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
```diff
@@ -803,6 +803,73 @@ final class ExplicitModuleBuildTests: XCTestCase {
...
+                                     "-module-cache-path", moduleCachePath.nativePathString(escaped: true),
+                                     "-working-directory", path.nativePathString(escaped: true),
+                                     "-disable-implicit-concurrency-module-import",
+                                     "-disable-implicit-string-processing-module-import",
+                                     "-register-module-dependency", "E",
+                                     "-emit-loaded-module-trace",
+                                     main.nativePathString(escaped: true)] + sdkArgumentsForTesting)
+      //guard driver.isFrontendArgSupported(.registerModuleDependency) else {
+      //  throw XCTSkip("Compiler does not support -register-module-dependency")
+      //}
```
nkcsgexi: Could you remove this commented-out guard? It's no longer necessary.

sepy97: Is this guard not needed at all?

artemcm: The guard is not needed here at all because we are not relying on the compiler having this flag, rather we simply need the compiler to support `-import-module` which we know it does.

Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
```diff
@@ -145,6 +145,15 @@ public extension Driver {
       commandLine.appendFlag(.dependencyScanCacheRemarks)
     }
 
+    // enabling scanning dependencies and building the module without including it in the compilation
```
artemcm: Nitpick: I'd reword this to be more specific to the functionality of this flag. Something like "Register a command-line-specified dependency with the dependency scanner, without implicitly importing it in compilation tasks".

Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
```diff
@@ -145,6 +145,15 @@ public extension Driver {
...
+    if (parsedOptions.contains(.registerModuleDependency) ){
+      if let moduleNameArg = parsedOptions.getLastArgument(.registerModuleDependency) {
+        let moduleName = moduleNameArg.asSingle
+        commandLine.appendFlag(.importModule)
+        commandLine.appendFlag(moduleName)
+      }
+    }
```
artemcm: Conditionals in Swift do not require parentheses. We can also fold this into a single conditional with:
```swift
    if let registerDependencyName = parsedOptions.getLastArgument(.registerModuleDependency)?.asSingle {
      commandLine.appendFlag(.importModule)
      commandLine.appendFlag(registerDependencyName)
    }
```
This way, if `parsedOptions.getLastArgument(.registerModuleDependency)` is `nil`, then the RHS overall is `nil`. 

But on top of that, we should also ensure that we can handle multiple `-register-module-dependency` flags, and this logic will only handle the last one appearing on the command line.

Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
```diff
@@ -145,6 +145,17 @@ public extension Driver {
       commandLine.appendFlag(.dependencyScanCacheRemarks)
     }
 
+    // command-line flag registers a dependency with the dependency scanner without including it in compilation tasks
```
artemcm: Sorry, just a small tweak still, the dependency is technically "included" as an input to the compilation.
```suggestion
    // command-line flag registers a dependency with the dependency scanner without causing it to be implicitly imported it in compilation tasks
```

Sources/SwiftDriver/ExplicitModuleBuilds/ModuleDependencyScanning.swift
```diff
@@ -145,6 +145,17 @@ public extension Driver {
...
+    if parsedOptions.contains(.registerModuleDependency) {
+      if let moduleNamesArg = parsedOptions.getLastArgument(.registerModuleDependency) {
```
artemcm: This isn't quite right still, I think simply deleting these two lines of code should get us the desired behavior.

artemcm:
@swift-ci test

nkcsgexi (approved):

nkcsgexi:
@swift-ci test Windows

sepy97:
@swift-ci test windows platform

sepy97:
@swift-ci test windows platform

sepy97:
@swift-ci test windows platform

PR #2018, opened 2025-11-18
SwiftDriver: explicitly select RSP handling on Windows

compnerd:
When cross-compiling from Windows to Android, if we spill past the command line limit, we would need to emit a response file. The response file is emitted into the style of the build rather than the host. This means that on Unix platforms, we need to indicate that the response file is Windows not Unix.

We rely on the internal behaviour of the second flag being special cased and emit the `--rsp-quoting=windows` parameter in the second argument position to ensure that it is passed on the command line (as if the argument is pushed into the response file, it will not be possible to honour it for processing the file).


compnerd:
@swift-ci please test

compnerd:
This should help repair the nightlies on Windows

compnerd:
I think that unblocking the nightlies is important enough. We can re-evaluate this once things are green.

PR #2022, opened 2025-11-19
Skip testEmitClangHeaderMinAccess when frontend doesn't support it.

rjmansfield:


rjmansfield:
@swift-ci please test

PR #2026, opened 2025-11-20
Wasm: bump default stack size to 128K

MaxDesiatov:
The default of 64K is too low, especially as tail calls are not available by default on Wasm. It also means Swift Testing is unusable in `main` development snapshots due to stack overflows.

rdar://160218251


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #2035, opened 2025-12-02
Skip default CMO when emitting module separately

rjmansfield:
Cross-module optimization is incompatible with emitting the module separately. Previously, an assertion caught this condition, but the assertion would not fire in release builds.

This changes the logic to simply skip enabling default CMO when emitModuleSeparately is true, rather than asserting or erroring.


rjmansfield:
@swift-ci please test

PR #2045, opened 2025-12-12
Tests: repair output file map writing in tests

compnerd:
This adjusts the OFM emission on Windows to ensure that the arc separator does not cause issues. `\` is treated as the escape character in JSON and this prevents the loading of the OFM.


compnerd:
Seems that these changes were reverted.

PR #2050, opened 2025-12-18
Revert "[Sanitizers] Add support for -sanitize=memtag-stack"

nkcsgexi:
Reverts swiftlang/swift-driver#2016

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -2754,17 +2754,12 @@ extension Driver {
       // Support is determined by existence of the sanitizer library.
       // FIXME: Should we do this? This prevents cross-compiling with sanitizers
       //        enabled.
-      var sanitizerSupported = true
-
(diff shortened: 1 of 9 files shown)


nkcsgexi:
@swift-ci please smoke test

DanBlackwell:
Hi @nkcsgexi has this broken something?

nkcsgexi:
hey @DanBlackwell , I'm testing whether this can fix a failure we observed in the build log as such:

> error: unsupported option '-sanitize=address' for target 'x86_64-unknown-windows-msvc'. 

However, it seems we still fail after taking the revert.

DanBlackwell:
> hey @DanBlackwell , I'm testing whether this can fix a failure we observed in the build log as such:
> 
> > error: unsupported option '-sanitize=address' for target 'x86_64-unknown-windows-msvc'. 
> 
> However, it seems we still fail after taking the revert.

I'm not sure if I understand that target triple correctly, but I wouldn't be surprised if clang and msvc address sanitizer implementations are incompatible - is that triple new?

PR #2053, opened 2025-12-26
Wasm: Use `-sysroot` value as the clang linker sysroot if provided

kateinoigakukun:
For other Unix-like platforms, the driver already respects `-sysroot` over `-sdk` as the native sysroot (not Swift "SDK"). See the following PRs for reference:
* https://github.com/swiftlang/swift-driver/pull/1811
* https://github.com/swiftlang/swift/pull/72352

For WebAssembly, this change makes the driver behave consistently with other platforms. This consistency is important for supporting swiftc as a swift-build's underlying linker driver for Wasm targets, as UnixLd.xcspec now uses `-sysroot` by default:
https://github.com/swiftlang/swift-build/blob/02e9f6778b621375b5a5a2fe40f9a4b55bb44ba0/Sources/SWBGenericUnixPlatform/Specs/UnixLd.xcspec#L61-L68

### DIFF
--- a/Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
@@ -114,7 +114,11 @@ extension WebAssemblyToolchain {
       }
       commandLine.append(contentsOf: inputFiles)
 
-      if let path = targetInfo.sdkPath?.path {
+      // Prefer -sysroot as the native sysroot path if provided.
+      if let sysroot = parsedOptions.getLastArgument(.sysroot)?.asSingle {
+        commandLine.appendFlag("--sysroot")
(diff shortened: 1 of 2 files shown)


finagolfin:
This new `-sysroot` approach largely doesn't work, swiftlang/swift#79657, and nobody is using it on Unix, eg [in the compiler validation suite](https://github.com/swiftlang/swift/blob/285c943374cbeb0c48abbef1f6f32a19949bf279/test/lit.cfg#L419). We tried putting it in for Unix only because some Windows people were using this new flag there, and they wanted to try cross-compiling to Android from Windows, but [this long command they use](https://github.com/swiftlang/swift-driver/pull/1822#issuecomment-3120342816) shows you it still requires more work.

It is possible that your Wasm work will do fine with it- I don't know- as your SDK bundle may be more self-contained, but it definitely still has problems on Unix.

jakepetroules (approved):

jakepetroules:
@swift-ci test

PR #2054, opened 2026-01-05
Fix optimization record path support for multi-threaded WMO

rjmansfield:
Handles multiple producers error by ensuring each input file gets a unique optimization record output path across all compilation modes.

### DIFF
--- a/Sources/SwiftDriver/Driver/OutputFileMap.swift
+++ b/Sources/SwiftDriver/Driver/OutputFileMap.swift
@@ -134,7 +134,63 @@ public struct OutputFileMap: Hashable, Codable {
     var outputFileMap = OutputFileMap()
     outputFileMap.entries = try result.toVirtualOutputFileMap()
 
+    outputFileMap.fixDuplicateOptimizationRecordPaths(diagnosticEngine: diagnosticEngine)
+
     return outputFileMap
+  }
+
+  /// Detect and fix duplicate optimization record paths in the output file map.
+  /// When multiple inputs point to the same optimization record file, synthesize unique paths.
+  private mutating func fixDuplicateOptimizationRecordPaths(diagnosticEngine: DiagnosticsEngine) {
+    var optRecordPaths: [VirtualPath.Handle: [VirtualPath.Handle]] = [:]
+
+    for (inputFile, outputs) in entries {
+      for outputType in [FileType.yamlOptimizationRecord, FileType.bitstreamOptimizationRecord] {
+        if let optRecordPath = outputs[outputType] {
+          optRecordPaths[optRecordPath, default: []].append(inputFile)
+        }
+      }
+    }
+
+    // Fix duplicates by synthesizing unique paths
+    for (optRecordPath, inputFiles) in optRecordPaths where inputFiles.count > 1 {
+      let pathName = VirtualPath.lookup(optRecordPath).basename
+      diagnosticEngine.emit(.warning(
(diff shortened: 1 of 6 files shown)


rjmansfield:
@swift-ci please test

nkcsgexi:
hey @rjmansfield , the code change is kind of large for review. Could you split it into smaller commits, or ideally, separate PRs?

Sources/SwiftDriver/Driver/OutputFileMap.swift
```diff
@@ -134,9 +134,65 @@ public struct OutputFileMap: Hashable, Codable {
...
     outputFileMap.entries = try result.toVirtualOutputFileMap()
 
+    outputFileMap.fixDuplicateOptimizationRecordPaths(diagnosticEngine: diagnosticEngine)
+
     return outputFileMap
   }
 
+  /// Detect and fix duplicate optimization record paths in the output file map.
+  /// When multiple inputs point to the same optimization record file, synthesize unique paths.
+  private mutating func fixDuplicateOptimizationRecordPaths(diagnosticEngine: DiagnosticsEngine) {
```
artemcm, line 144: What can cause this to happen? Shouldn't fixing/avoiding this be the responsibility of the output file map generator client of the driver? 

If the driver goes ahead and changes output paths for some inputs here, then the client won't be able to get at them, because they won't match the client's specification, even if incorrect. 

I'm curious to learn more about how we came about this problem.

rjmansfield:
> hey @rjmansfield , the code change is kind of large for review. Could you split it into smaller commits, or ideally, separate PRs?

OK, I'll try my best to break it up.

Sources/SwiftDriver/Driver/OutputFileMap.swift
rjmansfield: This was what was happening when #1997 got merged. We hit output maps with multiple file entries for the same opt record when building multithreaded. e.g.

`SwiftDriverExecution/MultiJobExecutor.swift:209: Fatal error: multiple producers for output CMakeFiles/swiftSynchronization.dir/RelWithDebInfo/Synchronization.opt.bitstream`

The testOptimizationRecordMultipleProducersError test reproduces this case.

swiftpm can generate the module-level opt-record entries in the file map. However, now with multithreaded WMO , the frontend will write to the same file with a module level file entry for opt records or IR output from different threads. I don't think the synthesis of filenames is ideal by any means but it was added to keep compatibility from swiftpm and output maps

Do you think it more sense to change swiftpm to only generate per file entries, and then phase an error about using module level entries for opt records/IR? Or what path forward makes sense to you?  Sorry, in retrospect I should've discussed this issue prior to implementing.

artemcm: Yeah, I think it would make the most sense to have the clients (swiftpm/swiftbuild) always generate per-file entries **and** a per-module entry.

The right way to model this from the build system probably perspective should match how these outputs are produced by the compiler:
1) per-file outputs for multi-threaded WMO
2) per-file outputs for non-WMO builds
3) single per-module output for single-threaded WMO

The build system client can either implement the logic to do this, or it can always specify both kinds of output entries and avoid any ambiguity there. It would then still probably need logic to know which outputs actually got generated though.

rjmansfield: Okay, makes sense. I'll abandon this PR.

PR #2058, opened 2026-01-09
tests: avoid unnecessary escaping of paths

compnerd:
The extra escaping of paths results in the lookup for the dependencies in the CAS to be quoted further than desired. As a result, the `VirtualPath` fails to resolve and the test fails. Avoid the unnecessary escaping to allow the test to pass.

### DIFF
--- a/Tests/SwiftDriverTests/CachingBuildTests.swift
+++ b/Tests/SwiftDriverTests/CachingBuildTests.swift
@@ -228,15 +228,14 @@ final class CachingBuildTests: XCTestCase {
           cHeadersPath.appending(component: "Bridging.h")
       let sdkArgumentsForTesting = (try? Driver.sdkArgumentsForTesting()) ?? []
       var driver = try Driver(args: ["swiftc",
-                                     "-I", cHeadersPath.nativePathString(escaped: true),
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

compnerd:
Yeah, local testing indicates that this works. The Windows failure in GHA is expected - we need an updated snapshot that has the changes for the VFS layering issue.

Separating the two concerns is good - this at least gets us into a good state for the command line invocation (after this change, we only escape the paths when emitting a JSON file in the tests or the RSP emission). I do agree that avoiding the re-tokenisation is definitely better, I do not know why we do that.

compnerd:
```
commit cfde1fb693bb71f107ec09dfe92bb87ca15e990a
Author: Artem Chikin <achikin@apple.com>
Date:   Wed Apr 19 11:14:17 2023 -0700

    [Dependency Scanning] Always tokenize scanning query command-line strings

    Other parts of the scanner lib (e.g. target-info query) already do this. We must always make sure to process the incoming command-line strings and run them through 'llvm::cl::TokenizeGNUCommandLine' in order to process escaped paths.

    Part of rdar://106712169
```

@artemcm might remember more context around this

PR #2060, opened 2026-01-13
Fix optimization record handling

rjmansfield:
This PR fixes several issues with the optimization record handling and is a continuation of https://github.com/swiftlang/swift-driver/pull/2054 with multiple commits.

Handles multiple producers error by ensuring each input file gets a unique optimization record output path across all compilation modes.
    
All the various compilation modes make path selection somewhat complicated so the priority is output file map > explicit flags > derived from input filename
    
Within output file map entries:
 
For single-threaded WMO: module-level entry > per-file entry > derived
For multi-threaded WMO: per-file entry > derived (but never module-level)
For non-WMO: per-file entry > derived
    
If both output file map entries and explicit -save-optimization-record-path flags are specified, the output file map wins and a warning is emitted.
    
When no output file map entry exists, explicit -save-optimization-record-path flags are used. Otherwise, paths are derived from input filenames.
    
Several tests were added to cover the various compilation modes and scenarios.

rdar://164884975
rdar://164494085

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -708,6 +708,15 @@ extension Driver {
 
     /// Process inputs for supplementary output generation (SIL/IR/opt-records)
     func processInputsForSupplementaryOutput(inputs: [TypedVirtualPath], outputType: FileType, flag: String, directory: String?) throws {
+      // For single-threaded WMO with optimization records, prefer module-level entry if it exists
+      if isSingleThreadedWMO && outputType.isOptimizationRecord {
+        let emptyPathHandle = try VirtualPath.intern(path: "")
+        if let moduleLevelPath = outputFileMap?.entries[emptyPathHandle]?[outputType] {
+          flaggedInputOutputPairs.append((flag: flag, input: nil, output: TypedVirtualPath(file: moduleLevelPath, type: outputType)))
+          return
+        }
+      }
+
       for inputFile in inputs {
         // Check output file map first for per-file entry
         if let outputFileMapPath = try outputFileMap?.existingOutput(inputFile: inputFile.fileHandle, outputType: outputType) {
@@ -717,6 +726,11 @@ extension Driver {
           flaggedInputOutputPairs.append((flag: flag, input: inputFile, output: outputPath))
         } else if parsedOptions.hasArgument(.saveTemps) {
           // When using -save-temps without explicit directories, output to current directory
+          let outputPath = try generateSupplementaryOutputPath(for: inputFile, outputType: outputType, directory: ".")
+          flaggedInputOutputPairs.append((flag: flag, input: inputFile, output: outputPath))
+        } else if outputType.isOptimizationRecord && (parsedOptions.hasArgument(.saveOptimizationRecord) || parsedOptions.hasArgument(.saveOptimizationRecordEQ)) {
+          // Multi-threaded WMO: when -save-optimization-record is used without explicit
+          // -save-optimization-record-path and without file map entries, derive per-file paths
           let outputPath = try generateSupplementaryOutputPath(for: inputFile, outputType: outputType, directory: ".")
           flaggedInputOutputPairs.append((flag: flag, input: inputFile, output: outputPath))
         }
@@ -746,7 +760,19 @@ extension Driver {
         }
 
         let directory = directoryOption.flatMap { parsedOptions.getLastArgument($0)?.asSingle }
-        let hasFileMapEntries = outputFileMap?.hasEntries(for: outputType) ?? false
+
+        // For multi-threaded WMO opt-records, check for per-file entries (not module-level)
+        // For other cases, check for any entries
+        let hasFileMapEntries: Bool
+        if outputType.isOptimizationRecord && !compilerMode.usesPrimaryFileInputs && numThreads > 1 {
+          // Multi-threaded WMO: only consider per-file entries
+          hasFileMapEntries = allInputs.contains { input in
+            (try? outputFileMap?.existingOutput(inputFile: input.fileHandle, outputType: outputType)) != nil
+          }
+        } else {
+          hasFileMapEntries = outputFileMap?.hasEntries(for: outputType) ?? false
+        }
+
         let hasOptRecordFlag = outputType.isOptimizationRecord &&
                                 (parsedOptions.hasArgument(.saveOptimizationRecord) ||
                                  parsedOptions.hasArgument(.saveOptimizationRecordEQ))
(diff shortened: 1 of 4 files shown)


rjmansfield:
@swift-ci please test

rjmansfield:
@swift-ci please test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -688,17 +688,29 @@ extension Driver {
     /// Generate directory-based output path for supplementary outputs
     func generateSupplementaryOutputPath(for input: TypedVirtualPath, outputType: FileType, directory: String) throws -> TypedVirtualPath {
       let inputBasename = input.file.basenameWithoutExt
-      let fileExtension = outputType == .sil ? "sil" : "ll"
+      let fileExtension: String
```
compnerd: I wonder if this is better handled as an extension to `FileType`:

```swift
extension FileType {
  var `extension`: String {
    return switch self {
    case .sil: "sil"
    case .llvmIR: "ll"
    case .yamlOptimizationRecord: "opt.yaml"
    case .bitstreamOptimizationRecord: "opt.bitstream"
    default: self.rawValue
    }
  }
}
```

rjmansfield: Sounds good, done.

rjmansfield:
@swift-ci please test

Sources/SwiftDriver/Jobs/CompileJob.swift
```diff
@@ -264,13 +264,43 @@ extension Driver {
...
+
+    // If we have N explicit optimization record paths for N files, collect them
+    var explicitOptRecordPaths: [VirtualPath.Handle]? = nil
+    if optRecordPathCount == inputs.count && !hasOptRecordFileMapEntries {
+      let allPaths = parsedOptions.arguments(for: .saveOptimizationRecordPath)
+
+      // In multi-threaded WMO, all paths go to the single job
+      if !compilerMode.usesPrimaryFileInputs && numThreads > 1 {
+        for optPath in allPaths {
+          commandLine.appendFlag("-save-optimization-record-path")
```
artemcm:
```suggestion
          commandLine.appendFlag(.saveOptimizationRecordPath)
```
We're not quite consistent about doing this but we should try to be.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -708,6 +727,10 @@ extension Driver {
           // When using -save-temps without explicit directories, output to current directory
           let outputPath = try generateSupplementaryOutputPath(for: inputFile, outputType: outputType, directory: ".")
           flaggedInputOutputPairs.append((flag: flag, input: inputFile, output: outputPath))
+        } else if outputType.isOptimizationRecord && (parsedOptions.hasArgument(.saveOptimizationRecord) || parsedOptions.hasArgument(.saveOptimizationRecordEQ)) {
+          // For opt-records with explicit flag but no file map entry, derive path from input filename
```
artemcm: I'm not sure I follow this comment, if there is an explicit flag, shouldn't we use `explicitOptRecordPathMap`? Or do you mean there is a *single* command-line flag and we are doing multi-threaded WMO or single-file compilation here?

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -719,21 +742,49 @@ extension Driver {
...
+        if directory != nil || hasFileMapEntries || (parsedOptions.hasArgument(.saveTemps) && !hasFileMapEntries && !outputType.isOptimizationRecord) || hasOptRecordFlag {
           let inputsToProcess: [TypedVirtualPath]
           if compilerMode.usesPrimaryFileInputs {
             inputsToProcess = input.map { [$0] } ?? []
           } else {
             inputsToProcess = allInputs
           }
 
-          try processInputsForSupplementaryOutput(inputs: inputsToProcess, outputType: outputType, flag: flag, directory: directory)
+          let isSingleThreadedWMO = !compilerMode.usesPrimaryFileInputs && numThreads < 2
```
artemcm: We end up computing this so often that we may as well make this a computed property on the `Driver` and reference that throughout, since I think it's valid to think of this as the only real edge case and all other scenarios share the per-source file behavior.

e.g. 
```
var isSingleThreadedWMO: Bool {
  return compilerMode.usesPrimaryFileInputs && numThreads < 2
}
```

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -859,6 +913,63 @@ extension Driver {
...
+              (try? outputFileMap?.existingOutput(inputFile: input.fileHandle, outputType: optRecordType)) != nil
+            }
+          } else {
+            inputHasOptRecordEntry = outputFileMap?.hasEntries(for: optRecordType) ?? false
+          }
+        }
+
+        // In multi-threaded WMO with -save-optimization-record but no explicit paths or file map entries,
+        // pass nil to trigger per-file path generation
+        let isMultiThreadedWMOWithAutoGenPaths = !compilerMode.usesPrimaryFileInputs && numThreads > 1 &&
```
artemcm, line 939: This scenario adds a decent amount of complexity overall. What do you think of not supporting this flow? That is, rely on the build system to ensure it either specifies a single top-level entry for single-threaded WMO or specifies per-file entries otherwise. 

Would there be an existing use case or build system we break if we emit an error on this kind of configuration (single output path flag but per-input opt records)?

rjmansfield: I don't know enough about all the use cases to say how much we'd break if we made this an error. I suspect not a lot, but I was considering cases like: `swiftc -wmo -num-threads 2 -save-optimization-record file1.swift file2.swift`, or makefiles in embedded projects that use -wmo that might also use -num_threads > 1 with -save-optimization-record. I suspect erroring would mainly break the command line usage, and partial output file map entries, but probably both are edge cases.

The other option would be to add this as a temporary measure with a deprecation warning for multithreaded WMO w/o per file paths, to flush out any misuses? Let me know which you prefer.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
rjmansfield: I meant that this is for multi-threaded WMO when using -save-optimization-record without explicit -save-optimization-record-path arguments and without file map entries. I will fix the comment up.

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
artemcm: Ah, I see, I hadn't thought through the workflow of just the `-save-optimization-record` without any explicit path arguments at all. 

I agree these are edge cases but it seems more reasonable to keep this around as a shorthand, thanks for explaining.

artemcm (approved):
Just the above comment about lifting up the various checks for single-treaded WMO into a computed property, and otherwise LGTM, thanks.

rjmansfield:
@swift-ci please test

rjmansfield:
@swift-ci please test

PR #2061, opened 2026-01-14
Scanner prefix map options should only be available with caching

cachemeifyoucan:
Scanner prefix map options requires the scanner to create a CAS file system that has all the input path remapped, thus it is not compatible when caching is not enabled. Issue a warning in the swift-driver and ignore such options when caching is not enabled.

rdar://164649047

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -299,28 +299,28 @@ public struct Driver {
   let scannerPrefixMapSDK: AbsolutePath?
   let scannerPrefixMapToolchain: AbsolutePath?
   lazy var prefixMapping: [(AbsolutePath, AbsolutePath)] = {
+    guard isFrontendArgSupported(.cacheReplayPrefixMap) else {
+      return []
+    }
     var mapping: [(AbsolutePath, AbsolutePath)] = scannerPrefixMap.map {
       return ($0.key, $0.value)
     }
-    do {
-      guard isFrontendArgSupported(.cacheReplayPrefixMap) else {
-        return []
-      }
-      if let sdkMapping = scannerPrefixMapSDK,
-         let sdkPath = absoluteSDKPath {
-        mapping.append((sdkPath, sdkMapping))
-      }
-      if let toolchainMapping = scannerPrefixMapToolchain {
-        let toolchainPath = try toolchain.executableDir.parentDirectory // usr
-                                                       .parentDirectory // toolchain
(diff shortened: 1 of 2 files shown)


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

artemcm (approved):

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -299,28 +299,32 @@ public struct Driver {
...
+    guard isFrontendArgSupported(.cacheReplayPrefixMap) else {
+      return []
+    }
     var mapping: [(AbsolutePath, AbsolutePath)] = scannerPrefixMap.map {
       return ($0.key, $0.value)
     }
+    if let sdkMapping = scannerPrefixMapSDK,
+       let sdkPath = absoluteSDKPath {
+      mapping.append((sdkPath, sdkMapping))
+    }
```
compnerd, line 311: Nit: would be nice to unwrap the values in the same order as usage.

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -299,28 +299,32 @@ public struct Driver {
...
                                                        .parentDirectory // toolchain
         mapping.append((toolchainPath, toolchainMapping))
       }
-      // The mapping needs to be sorted so the mapping is determinisitic.
-      // The sorting order is reversed so /tmp/tmp is preferred over /tmp in remapping.
-      return mapping.sorted { $0.0 > $1.0 }
     } catch {
-      return mapping.sorted { $0.0 > $1.0 }
+      // ignore error about failure to get toolchain mapping.
+    }
```
compnerd, line 316: Why not use the optional unwrapping and avoid the `do { } catch { }`?

```swift
if let toolchainPath = try? toolchain.executableDir.parentDirectory.parentDirectory {
  mapping.append((toolchainPath, toolchainMapping))
}
```

compnerd (approved):
Looks fine except for the small clean up recommendation

Sources/SwiftDriver/Driver/Driver.swift
cachemeifyoucan: The order needs to be sorted in the end. This is to that we get maximum path remapping when possible regardless of ordering.

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:

PR #2062, opened 2026-01-16
Implement -debug-module-path for implicit modules in swift-driver

adrian-prantl:
There is no fundamental technical reason to restrict this feature to explict modules only, so this patch adds support for passing the correct -debug-module-path also to implicit builds.

rdar://168256846

### DIFF
--- a/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
+++ b/Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
@@ -109,12 +109,15 @@ extension Driver {
       jobNeedPathRemap = false
     }
 
-    if isPlanJobForExplicitModule && forObject && isFrontendArgSupported(.debugModulePath),
-       let explicitModulePlanner {
-      let mainModule = explicitModulePlanner.dependencyGraph.mainModule
-      let pathHandle = moduleOutputInfo.output?.outputPath ?? mainModule.modulePath.path
-      let path = VirtualPath.lookup(pathHandle)
-      try addPathOption(option: .debugModulePath, path: path, to: &commandLine, remap: jobNeedPathRemap)
+    // Add the -debug-module-path option to the compile job.
+    // If modulePathHandle is nil, this build doesn't produce a Swift module.
+    if forObject && isFrontendArgSupported(.debugModulePath),
+       var modulePathHandle = moduleOutputInfo.output?.outputPath {
+      // Recompute the module path based on the module name, because this is effectively passing the output
+      // of the module merge action which depends on the compile action.
+      let moduleBase : VirtualPath = VirtualPath.lookup(modulePathHandle).parentDirectory.appending(component: moduleOutputInfo.name)
+      modulePathHandle = try moduleBase.replacingExtension(with: .swiftModule).intern()
+      try addPathOption(option: .debugModulePath, path: VirtualPath.lookup(modulePathHandle), to: &commandLine, remap: jobNeedPathRemap)
     }
 
     // Check if dependency scanner has put the job into direct clang cc1 mode.
--- a/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
+++ b/Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
@@ -716,10 +716,6 @@ final class ExplicitModuleBuildTests: XCTestCase {
       for job in jobs {
         XCTAssertEqual(job.outputs.count, 1)
         let outputFilePath = job.outputs[0].file
-        if job.kind == .compile && driver.isFrontendArgSupported(.debugModulePath) {
-          XCTAssertTrue(job.commandLine.contains(subsequence: ["-debug-module-path", try toPathOption("testExplicitModuleBuildJobs.swiftmodule")]))
-
-        }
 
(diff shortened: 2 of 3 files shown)


adrian-prantl:
@swift-ci test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -109,12 +109,23 @@ extension Driver {
...
+    } else if !isPlanJobForExplicitModule && isFrontendArgSupported(.debugModulePath),
+      let pathHandle = moduleOutputInfo.output?.outputPath {
+      // Recompute the module path based on the module name, because
+      // this is effectively passing the ouput of the module merge
+      // action which depends on the compile action.
+      let path = try VirtualPath.lookup(pathHandle)
+        .parentDirectory.appending(component: moduleOutputInfo.name)
+        .replacingExtension(with: .swiftModule)
+        .intern()
+      try addPathOption(option: .debugModulePath, path: VirtualPath.lookup(path), to: &commandLine, remap: jobNeedPathRemap)
```
compnerd: Why the second lookup?

adrian-prantl: Thanks, that does look confusing.
It's a `VirtualPath.Handle`. I renamed the variable to make that more obvious to the reader!

cachemeifyoucan: There is no more module merge option anymore (new driver doesn't support it).

Is it possible to pass in the swiftmodule path to the function so it is not derived again?

adrian-prantl:
@swift-ci test

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
adrian-prantl: > is it possible to pass in the swiftmodule path to the function so it is not derived again?

I would love to hear some advice for how to better implement this from someone more familiar with swift-driver :-)

cachemeifyoucan: I am not too familiar with how new swift-driver construct implicit module job but I don't think it will construct merge job anymore. `forObject` is probably still valid for when this flag is needed for implicit build, but you want to double check.

I also think the correct way to get module output path is to directly access `self.moduleOuputInfo.output.outputPath`, maybe even for explicit module build. But that seems to have a variant module output path for catalyst output. I would like to see a test case for correctly writing `-debug-module-path` when planning zippered module. (retroactively added for explicit module, for implicit module if that is supported in new driver)

Sources/SwiftDriver/Jobs/FrontendJobHelpers.swift
```diff
@@ -109,12 +109,25 @@ extension Driver {
...
-      try addPathOption(option: .debugModulePath, path: path, to: &commandLine, remap: jobNeedPathRemap)
+    // Add the -debug-module-path option to the compile job.
+    if forObject && isFrontendArgSupported(.debugModulePath) {
+      let modulePathHandle : VirtualPath.Handle
+      if isPlanJobForExplicitModule,
+         let explicitModulePlanner {
+        let mainModule = explicitModulePlanner.dependencyGraph.mainModule
+        modulePathHandle = moduleOutputInfo.output?.outputPath ?? mainModule.modulePath.path
+      } else {
+        // Recompute the module path based on the module name, because this is effectively passing the output
```
artemcm: Similarly to what @cachemeifyoucan is saying, I do not think we should be computing this here at all. I think this code should rely on `moduleOutputInfo.output?.outputPath` for both implicit and explicit module code paths, without trying to figure out what it would be if it is `nil`. 

Also, as already mentioned, it would be great to have a test for this which includes `-experimental-emit-variant-module -emit-variant-module-path <...>`.

adrian-prantl: Let me see if I can figure out how to get the path into `moduleOutputInfo.output?.outputPath` then.

adrian-prantl: I uploaded a much simplified version of the patch that unifies the explicit and implicit code paths.
One thing I couldn't resolve is switching to use `moduleOutputInfo.output?.outputPath`: it contains the unmerged module path (i.e., "Foo-1.swiftmodule") instead of the final merged module name. If you have any ideas for how to fix that, I'm happy to rework this further. IIUC, the main problem is that the compile job doesn't actually know the name of final merged module...

artemcm: I'm confused a bit, there is no unmerged or merged module name anymore in the driver. So the comment:
```
// Recompute the module path based on the module name, because this is effectively passing the output
// of the module merge action which depends on the compile action.
```
Doesn't really apply because we intend to be passing the output of the `-emit-module` (`EmitModuleJob`) action, which does not depend on the compile action. 

How are you seeing `*-1.swiftmodule` names? Taking a look at the `EmitModuleJob` construction, it seems to just use `moduleOutputInfo.output!.outputPath` directly itself.

PR #2063, opened 2026-01-20
Reapply "[SwiftCaching] Don't leak output path in the explicit module map JSON"

cachemeifyoucan:
This re-enables the change that abstract mdoule path in the explicit module map for caching build after LLDB fully support the abstract module path and loads module directly from CAS.

rdar://149960868


cachemeifyoucan:
@swift-ci please test

PR #2064, opened 2026-01-21
JobExecutor: consistently avoid canceling job when failure is encountered

nkcsgexi:
When a job is signalled to abort, such as compiler crashers, etc, we should still avoid conceling other jobs if the executor has been configured as such.


nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

PR #2066, opened 2026-01-21
SwiftDriver: adjust `SwiftScan` to properly report caching

compnerd:
CAS on Windows is now supported. Adjust the feature support reporting from SwiftScan to ensure that we can use the CAS on Windows.

### DIFF
--- a/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
+++ b/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
@@ -259,10 +259,6 @@ private extension String {
   }
 
   @_spi(Testing) public var supportsCaching : Bool {
-#if os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

cachemeifyoucan (commented):
Do we have caching tests passing in a reasonable state?

I think we should remove the `os(Windows)` guard in `CachingBuildTests.swift` in the same commit so that caching at least get some test coverage as we declare it is supported in windows.

compnerd:
With your changes (i.e. updated swift toolchain), my pending change (https://github.com/swiftlang/swift/pull/86503), and the changes that I am putting up now on swift-driver in local testing, I have all the caching and incremental build tests passing.

PR #2068, opened 2026-01-21
SwiftDriverTests: style change for validating job outputs

compnerd:
Use a bit more clever syntax to compress a long chain of `if`-`else if` statements that replicate the check.

This code can be further compressed as well but the pay off there is smaller (we could add the extension check as part of the `where` clause).

I opted to split the "runtime" components and the user specified modules for clarity not for any technical reason.

### DIFF
--- a/Tests/SwiftDriverTests/CachingBuildTests.swift
+++ b/Tests/SwiftDriverTests/CachingBuildTests.swift
@@ -205,11 +205,6 @@ final class CachingBuildTests: XCTestCase {
 #endif
   }
 
-  private func pathMatchesSwiftModule(path: VirtualPath, _ name: String) -> Bool {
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

compnerd:
Oh, interesting, didn't know that. I went to change this for the Windows tests and I just felt overwhelmed by the long chain and thought that a bit of Swift would help.

cachemeifyoucan:
I know because I did it...

Some day we will use SwiftTest and we can combine caching test with explicit module test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

Issue #2039, opened 2025-12-08
Windows: enable incremental tests


cachemeifyoucan:
After windows image updated, incremental tests no longer passes due to the added module dependencies. Currently, the incremental tests are disable on windows but need to fix and re-enable.

Issue #2040, opened 2025-12-08
Windows: enable compilation caching tests


cachemeifyoucan:
After windows support for compilation caching is added, we should update and enable compilation caching tests in swift-driver.

compnerd:
I believe that this has been done, closing the issue for now.

PR #2069, opened 2026-01-22
Enable CAS and associated tests on Windows

compnerd:
CAS on Windows is now supported. Adjust the feature support reporting from SwiftScan to ensure that we can use the CAS on Windows.

Enable the incremental compilation tests on Windows. Now that CAS is supported on Windows and the Swift scanner issues resulted, we can enable the incremental module tests.

Enable the caching build tests on Windows. Now that CAS is supported on Windows, we can enable the caching build tests.

Fixes: #2039, #2040

### DIFF
--- a/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
+++ b/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
@@ -259,10 +259,6 @@ private extension String {
   }
 
   @_spi(Testing) public var supportsCaching : Bool {
-#if os(Windows)
-    // Caching is currently not supported on Windows hosts.
-    return false
-#else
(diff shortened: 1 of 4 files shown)


compnerd:
@swift-ci please test

cachemeifyoucan (approved):
YES. LGTM

compnerd:
@swift-ci please test

compnerd:

compnerd:
Discussed this offline with @cachemeifyoucan. Basically, the current failures are a regression in the toolchain, but the windows tests already require a newer snapshot. Since we don't have that, lets go ahead and merge this. This brings us in a better state and does not block further development.

PR #2073, opened 2026-01-31
[Triple] Make a target triple "os" for firmware

ian-twilightcoder:
Make a Triple.OS to represent a generic Apple firmware platform that isn't bare metal, but isn't tied to a specific hardware platform like macOS or iOS.

rdar://165381213

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple+Platforms.swift
+++ b/Sources/SwiftDriver/Utilities/Triple+Platforms.swift
@@ -35,6 +35,9 @@ public enum DarwinPlatform: Hashable {
   /// visionOS, corresponding to the `visionos` OS name.
   case visionOS(EnvironmentWithoutCatalyst)
 
+  /// Firmware, corresponding to the `firmware` OS name.
+  case Firmware
+
   /// The most general form of environment information attached to a
   /// `DarwinPlatform`.
   ///
@@ -82,6 +85,9 @@ public enum DarwinPlatform: Hashable {
     case .visionOS:
       guard let withoutCatalyst = environment.withoutCatalyst else { return nil }
       return .visionOS(withoutCatalyst)
+    case .Firmware:
+      guard environment == .device else { return nil }
+      return .Firmware
     }
   }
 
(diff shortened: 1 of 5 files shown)


Sources/SwiftDriver/Utilities/Triple+Platforms.swift
```diff
@@ -35,6 +35,9 @@ public enum DarwinPlatform: Hashable {
   /// visionOS, corresponding to the `visionos` OS name.
   case visionOS(EnvironmentWithoutCatalyst)
 
+  /// Firmware, corresponding to the `firmware` OS name.
+  case Firmware
```
artemcm, line 39: Other enum cases here begin with a lowercase letter, other enum kinds here also do, other, non-Darwin, OS enums are also lowercase, it's worth sticking with the convention.

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -1500,13 +1505,39 @@ extension Triple.OS {
...
   }
 
+  /// Is this a Firmware triple.
+  public var isFirmware: Bool {
+    self == .firmware
+  }
+
 
   /// isOSDarwin - Is this a "Darwin" OS (macOS, iOS, tvOS, watchOS, or visionOS).
   public var isDarwin: Bool {
```
artemcm: Should we get rid of this computed property? Or rename it to indicate that it is inly checking the OS? Or rename the new getter method to indicate that it is about more than the OS? 

Otherwise I think it will confuse people as to why there are two same-named different methods to answer what _looks_ like the same question.

ian-twilightcoder: I would vote for getting rid of this one, it doesn't look like SwiftBuild uses it, are there any other clients that would have to change in lock step?

ian-twilightcoder:
@swift-ci test

Sources/SwiftDriver/Utilities/Triple+Platforms.swift
ian-twilightcoder: They do, but I think they're supposed to be the "pretty" name of the platform, i.e. "macOS" and not "macos"

Sources/SwiftDriver/Utilities/Triple.swift
artemcm: I was only thinking of just other clients internal to the driver.

artemcm (approved):

PR #2075, opened 2026-02-04
Fix module-level SIL entries in output file maps for WMO

rjmansfield:


rjmansfield:
@swift-ci please test

PR #2080, opened 2026-02-11
Switch to lossy UTF8 decoding for dependency scanner diagnostic result strings

artemcm:
The `toSwiftString` code currently force-unwraps a `String` initializer which attempts to decode the input buffer as valid UTF-8. We have seen a crash on this force-unwrap.

This change switches `toSwiftString` to use the Swift standard library's `String(decoding:,as:)` instead of Foundation's `String(data:,encoding:)`. The former returns a non-optional value which replaces non-UTF-8 characters with a Unicode replacement character, instead of failing and returning `nil`.

This will not only resolve the crash, but also result in us emitting these potentially-corrupted strings for the user to see.

Resolves rdar://157486211


artemcm:
@swift-ci test

PR #2081, opened 2026-02-11
[6.3 🍒] Switch to lossy UTF8 decoding for dependency scanner diagnostic result strings

artemcm:
Cherry-pick of https://github.com/swiftlang/swift-driver/pull/2080
---------------------------------------
- **Explanation**
The toSwiftString code currently force-unwraps a String initializer which attempts to decode the input buffer as valid UTF-8. We have seen a crash on this force-unwrap.
This change switches toSwiftString to use the Swift standard library's String(decoding:,as:) instead of Foundation's String(data:,encoding:). The former returns a non-optional value which replaces non-UTF-8 characters with a Unicode replacement character, instead of failing and returning nil.
This will not only resolve the crash, but also result in us emitting these potentially-corrupted strings for the user to see.

- **Resolves**:  rdar://157486211
- **Main branch PR**: https://github.com/swiftlang/swift-driver/pull/2080
- **Risk**: Low, this change removes a force-unwrap which was previously hit when encountering non-UTF8 strings in the dependency scanner output. It replaces it with a lossy conversion method which will not crash when encountering such invalid strings.
- **Reviewed By**: TBD
- **Testing**: Tests added to the driver test suite to validate the new logic results in the driver no longer crashing on invalid UTF8 dependency scanner output.

### DIFF
--- a/Sources/SwiftDriver/SwiftScan/DependencyGraphBuilder.swift
+++ b/Sources/SwiftDriver/SwiftScan/DependencyGraphBuilder.swift
@@ -361,9 +361,9 @@ internal extension SwiftScan {
     guard let dataPtr = string_ref.data else {
       throw DependencyScanningError.invalidStringPtr
     }
-    return String(bytesNoCopy: UnsafeMutableRawPointer(mutating: dataPtr),
-                  length: string_ref.length,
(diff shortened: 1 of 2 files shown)


artemcm:
@swift-ci test

artemcm:

artemcm:

PR #2082, opened 2026-02-11
Driver support for linking relocatable object files

owenv:
Currently, Swift Build falls back to linking relocatable objects with clang, even when a target contains Swift. Begin adding support to swiftc for the equivalent functionality so we can avoid this and ensure Swift-specific behavior is applied consistently to all linker output types.

### DIFF
--- a/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
@@ -14,6 +14,7 @@ import SwiftOptions
 
 import func TSCBasic.lookupExecutablePath
 import struct TSCBasic.AbsolutePath
+import class TSCBasic.DiagnosticsEngine
 
 extension GenericUnixToolchain {
   private func majorArchitectureName(for triple: Triple) -> String {
@@ -42,7 +43,8 @@ extension GenericUnixToolchain {
     shouldUseInputFileList: Bool,
     lto: LTOKind?,
     sanitizers: Set<Sanitizer>,
-    targetInfo: FrontendTargetInfo
+    targetInfo: FrontendTargetInfo,
+    diagnosticsEngine: DiagnosticsEngine
   ) throws -> ResolvedTool {
 #if os(Windows)
     commandLine.appendFlag("--rsp-quoting=windows")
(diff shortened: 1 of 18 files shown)


owenv:
@swift-ci test

Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
```diff
@@ -51,315 +53,361 @@ extension GenericUnixToolchain {
...
+        sanitizers: sanitizers,
+        linkerOutputType: linkerOutputType,
+        lto: lto
+      )
+      // This should be the last option, for convenience in checking output.
+      commandLine.appendFlag(.o)
+      commandLine.appendPath(outputFile)
+      return resolvedTool
+    case .relocatableObject:
+      commandLine.appendFlag("-r")
```
compnerd, line 85: I'm worried about this assumption. Why can we not continue to use `clang` to drive the linker? It has a good deal of knowledge of how to properly invoke the linker.

owenv: In GenericUnixToolchain we always end up delegating the link to either clang or clang++ in the end depending on if C++ interop is enabled or not. However, I think it still makes sense to allow swiftc to initiate the link so that build systems can select the linker driver they invoke independently of the linker output type. We're in a somewhat awkward spot on macOS/Linux right now where we're using swiftc to link executables/shared libraries and clang to link relocatable objects, and it's fairly easy to introduce bugs since the linker drivers accept slightly different sets of flags and the driver selection logic is IMO overcomplicated

owenv:
@swift-ci test

owenv:
@swift-ci test windows

Sources/SwiftDriver/Jobs/GenericUnixToolchain+LinkerSupport.swift
compnerd: I think that having swiftc initiate that link is going to introduce even more complexity. There are subtle things about ordering, lookup, etc. that need to be honoured and are hard learnt over the many years of clang's history. Reproducing that and then ensuring that we are synchronised for all time seems like a good deal of overhead to avoid the dispatch.

owenv: Alright, I think you've convinced me this might not be worth pursuing right now. I'll close this for now and run it by you again if I decide to pursue it later.

PR #2102, opened 2026-03-11
Incremental: Don't rebuild everything when only emitting the module

benb:
Currently, if we want to emit a swiftmodule / header only in incremental mode, swift-driver will schedule the work for rebuilding the object outputs regardless (although it won't actually produce that output).

This PR makes these changes:

1) Planning.swift: Changes logic when deciding when we can skip object file jobs. Explicit module builds can be scheduled without passing this `.driverExplicitModuleBuild` option (i.e. `explicitSwiftModuleMap`), and as far as I can tell checking for explicit module builds is overly conservative here anyway. 
2) EmitModuleJob.swift: makes a check stricter so .d files can still be generated even though we won't now do the rest of the work
3) Updates tests to reflect the reduced work scheduled. 

I might be barking up the wrong tree here, if so I'd appreciate any guidance on how else we can resolve this.

### DIFF
--- a/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
+++ b/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
@@ -46,21 +46,6 @@ extension Driver {
       return
     }
 
-    // Add outputs that can't be merged
-    // Workaround for rdar://85253406
-    // Ensure that the separate emit-module job does not emit `.d.` outputs.
-    // If we have both individual source files and the emit-module file emit .d files, we
-    // are risking collisions in output filenames.
-    //
-    // In cases where other compile jobs exist, they will produce dependency outputs already.
-    // There are currently no cases where this is the only job because even an `-emit-module`
-    // driver invocation currently still involves partial compilation jobs.
-    // When partial compilation jobs are removed for the `compilerOutputType == .swiftModule`
-    // case, this will need to be changed here.
-    //
-    if emitModuleSeparately {
-      return
-    }
     if let dependenciesFilePath = dependenciesFilePath {
       var path = dependenciesFilePath
       // FIXME: Hack to workaround the fact that SwiftPM/Xcode don't pass this path right now.
--- a/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
+++ b/Sources/SwiftDriver/Jobs/EmitModuleJob.swift
@@ -48,22 +48,13 @@ extension Driver {
       return
     }
 
-    // Add outputs that can't be merged
-    // Workaround for rdar://85253406
-    // Ensure that the separate emit-module job does not emit `.d.` outputs.
-    // If we have both individual source files and the emit-module file emit .d files, we
-    // are risking collisions in output filenames.
-    //
-    // In cases where other compile jobs exist, they will produce dependency outputs already.
-    // There are currently no cases where this is the only job because even an `-emit-module`
-    // driver invocation currently still involves partial compilation jobs.
-    // When partial compilation jobs are removed for the `compilerOutputType == .swiftModule`
-    // case, this will need to be changed here.
-    //
-    if emitModuleSeparately {
-      return
-    }
-    if let dependenciesFilePath = dependenciesFilePath {
+    // `emitModuleDependenciesFilePath` (above) already supplies `-emit-dependencies-path`
+    // for this job whenever `-emit-dependencies` is requested. Only fall back to the general
+    // dependencies path when it did not, otherwise the frontend receives the flag twice and
+    // rejects it ("wrong number of '-emit-dependencies-path' arguments"). This matters now
+    // that the emit-module-only build (compilerOutputType == .swiftModule) skips the compile
+    // jobs and runs this block.
+    if emitModuleDependenciesFilePath == nil, let dependenciesFilePath = dependenciesFilePath {
       var path = dependenciesFilePath
       // FIXME: Hack to workaround the fact that SwiftPM/Xcode don't pass this path right now.
       if parsedOptions.getLastArgument(.emitDependenciesPath) == nil {
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -434,7 +434,8 @@ extension Driver {
 
       assert(input.type.isPartOfSwiftCompilation)
       // We can skip the compile jobs if all we want is a module when it's
-      // built separately.
+      // built separately, unless we need per-file outputs like const values
+      // that only compile jobs can produce.
       let canSkipIfOnlyModule = compilerOutputType == .swiftModule && emitModuleSeparately
       try createAndAddCompileJob(primaryInput: input,
                                  emitModuleTrace: emitModuleTrace,
@@ -478,7 +479,7 @@ extension Driver {
   )  throws {
     // We can skip the compile jobs if all we want is a module when it's
     // built separately.
-    if parsedOptions.hasArgument(.driverExplicitModuleBuild), canSkipIfOnlyModule { return }
+    if canSkipIfOnlyModule { return }
     // If we are in the batch mode, the constructed jobs here will be batched
     // later. There is no need to produce cache key for the job.
     let compile = try compileJob(primaryInputs: [primaryInput],
--- a/Sources/SwiftDriver/Jobs/Planning.swift
+++ b/Sources/SwiftDriver/Jobs/Planning.swift
@@ -436,7 +436,8 @@ extension Driver {
 
       assert(input.type.isPartOfSwiftCompilation)
       // We can skip the compile jobs if all we want is a module when it's
-      // built separately.
+      // built separately, unless we need per-file outputs like const values
+      // that only compile jobs can produce.
       let canSkipIfOnlyModule = compilerOutputType == .swiftModule && emitModuleSeparately
       try createAndAddCompileJob(primaryInput: input,
                                  emitModuleTrace: emitModuleTrace,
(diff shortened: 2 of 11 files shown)


Tests/SwiftDriverTests/SwiftDriverTests.swift
```diff
@@ -4636,8 +4636,8 @@ final class SwiftDriverTests: XCTestCase {
       // -experimental-emit-module-separately.
       var driver = try Driver(args: ["swiftc", "foo.swift", "bar.swift", "-module-name", "Test", "-emit-module-path", rebase("Test.swiftmodule", at: root), "-experimental-emit-module-separately"])
       let plannedJobs = try driver.planBuild().removingAutolinkExtractJobs()
-      XCTAssertEqual(plannedJobs.count, 3)
-      XCTAssertEqual(Set(plannedJobs.map { $0.kind }), Set([.emitModule, .compile]))
+      XCTAssertEqual(plannedJobs.count, 1)
+      XCTAssertEqual(Set(plannedJobs.map { $0.kind }), Set([.emitModule]))
```
compnerd: Why not simplify this?


```suggestion
      XCTAssertEqual(plannedJobs[0].kind, .emitModule)
```

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -434,8 +434,10 @@ extension Driver {
 
       assert(input.type.isPartOfSwiftCompilation)
       // We can skip the compile jobs if all we want is a module when it's
-      // built separately.
+      // built separately, unless we need per-file outputs like const values
+      // that only compile jobs can produce.
       let canSkipIfOnlyModule = compilerOutputType == .swiftModule && emitModuleSeparately
+                                && !parsedOptions.hasArgument(.emitConstValues)
```
compnerd: Is there some classification that we can use? "like const values" ... are there other features would also require compile jobs?

Sources/SwiftDriver/Jobs/EmitModuleJob.swift
```diff
@@ -53,12 +53,11 @@ extension Driver {
...
-    // When partial compilation jobs are removed for the `compilerOutputType == .swiftModule`
-    // case, this will need to be changed here.
+    // When compilerOutputType == .swiftModule, there are no per-file compile jobs (they are
+    // skipped since emitModuleSeparately handles the module), so the emit-module job must
+    // produce .dependencies output itself.
     //
-    if emitModuleSeparately {
+    if emitModuleSeparately && compilerOutputType != .swiftModule {
       return
     }
```
compnerd: I don't understand this - how is this different from L45-47?

artemcm: I guess this is meant to relax the check on this line, but it does make it fully redundant in the process.

Sources/SwiftDriver/Jobs/Planning.swift
artemcm: +1 to this, this doesn't seem like a general approach, there are many other kinds of outputs which may be requested on a driver invocation which are only produced by object compilation (not emit-module) tasks.
 
I think we should trust the `compilerOutputType` and for the case you are trying to catch here perhaps we instead have the driver diagnose when the invocation specifies a module output compiler "mode" but also specifies requiring some outputs which require compilation tasks which produce object files. 

We select the compiler output type here:
https://github.com/swiftlang/swift-driver/blob/4e94366e034aca47214fc2a50ef05e61ae0e5566/Sources/SwiftDriver/Driver/Driver.swift#L2442
Based on which compilation "mode" flag is used. Admittedly this part of the driver logic seems quite messy because `-emit-module` seems to be considered a mode only if no other `.modes` flag was specified. 

But, if we are in a `compilerOutputType == .swiftModule` scenario, `-emit-module` doesn't mean "_among other things, emit a module_", but rather "_this `swiftc` invocation's intent is to emit a module_". It's reasonable that some outputs will not be compatible with this mode. 

An alternative would be to extend `determinePrimaryOutputs` to select a different `compilerOutputType` when things like `.emitConstValues` are specified.

Sources/SwiftDriver/Jobs/Planning.swift
```diff
@@ -478,7 +480,7 @@ extension Driver {
   )  throws {
     // We can skip the compile jobs if all we want is a module when it's
     // built separately.
-    if parsedOptions.hasArgument(.driverExplicitModuleBuild), canSkipIfOnlyModule { return }
+    if canSkipIfOnlyModule { return }
```
artemcm, line 484: This is a change I've wanted to make for a while since it is rather superfluous and leads to an odd difference between implicit and explicit builds. 

While I think we should do it, I am having a hard time remembering if this is still an assumption some part of `swift-build` relies on. @owenv do you recall?

artemcm (changes requested):

Sources/SwiftDriver/Jobs/Planning.swift
benb: Let me paraphrase you, and you can tell me if I have this right:

Given the current implementation of `determinePrimaryOutputs`, if we have `compilerOutputType == .swiftModule` then we must be in an de facto emit module 'mode' (albeit this is not an actual `Option.Group.Modes` mode like `.emitExecutable`).

Therefore one way of looking at is: When only specifying emit-module you are not really expecting to receive other outputs here, so we don't need to worry about them here.

Option two would be to catch other outputs (`supplementaryOutput`s?) in `determinePrimaryOutputs` and change the `compilerOutputType` to reflect this.

My opinion here is that it seems weird to make a change so that Swiftmodule is not be the 'primary output' in these cases - it really is the primary output! So I would be minded to 'trust' the compilerOutputType and remove this change.

Does this make sense?

Sources/SwiftDriver/Jobs/EmitModuleJob.swift
benb: Ach, I forgot about this weirdness. Yes, it's redundant this way. Kicking myself.

Sources/SwiftDriver/Jobs/Planning.swift
owenv: This _might_ break some assumptions but now that we've open sourced swift-build it's a lot easier to test. I'll try to take a closer look soon but I might not get to it until next week

Sources/SwiftDriver/Jobs/Planning.swift
artemcm: Yeah that's a good summary and I agree with your conclusion. 👍🏼

Sources/SwiftDriver/Jobs/Planning.swift
benb: @owenv Any more thoughts?

owenv: sorry, I haven't had a chance to look yet. I'll try to investigate asap

owenv: I ran this through some basic testing and didn't encounter any issues, so I don't think there are any concerns on my end. However, I think I'd recommend running cross-repo testing from the main swiftlang/swift repo before merging this

benb: @owenv Automated testing, or manual? Can you point me at some docs or give me more info?

owenv: It's automated testing, I kicked it off in https://github.com/swiftlang/swift/pull/79684. That'll run the full compiler test suite on macOS/Linux/Windows which might catch issues the CI in this repo would not

owenv: https://ci-external.swift.org/job/swift-PR-windows/53449/console
https://ci.swift.org/job/swift-PR-Linux-smoke-test/26883/console

The sourcekit-lsp failures here _might_ be related, I didn't take a close look yet

benb: What can we do to push this forward - is there anything I can do? 
Afraid I didn't understand the output you posted at the time, and now it has apparently expired.

owenv:
@swift-ci test

benb:
@owenv @artemcm Can someone please re-run the tests to see if they're passing now?

artemcm:
@swift-ci test

artemcm:

benb:
Thanks @artemcm, but it looks like the Windows tests are stuck permanently pending for some reason?

benb:
@owenv Can you help me try and get this moving? This got blocked by the Windows tests not being scheduled, but hopefully they should be okay. (Do I need to manually rebase first? It's been a while.)

artemcm:

benb:
The windows integration tests were failing with `wrong number of '-emit-dependencies-path' arguments (expected 1, got 2`.

Rebased and updated to add a check to prevent `-emit-dependencies-path` from being generated twice:

https://github.com/swiftlang/swift-driver/commit/0f5fd9ce590dc2aec1ed9edd69c043ed17734eaf#diff-0ea29f287e3721b48f2ceba2f6cd08a604a90a9111d10e0a087f1e0ac499eb6aR51-R57

benb:
@artemcm Can you re-approve the workflows, please?

owenv:
@swift-ci test

benb:
@owenv Windows test has sadly not materialized.

owenv:
sometimes it fails after a force push

@swift-ci test windows

benb:
Everything seems to pass now.
@owenv can I get a re-review? Test changes aside, this PR is now quite small.

owenv (approved):
lgtm. I'd like to give other folks a _little_ bit of time to chime in if they have more feedback. Assigning to myself so we get this landed in a timely manner

benb:
@owenv Can we ping the relevent people?

artemcm (approved):

benb:
@owenv Okay to merge now?

owenv:
yeah, I'm ok landing this

PR #2105, opened 2026-03-16
Add `wasm32-unknown-emscripten` triple support

MaxDesiatov:
One of the necessary steps for implementing the Emscripten pitch: https://forums.swift.org/t/pitch-emscripten-target-support-for-swift/85310.

Depends on https://github.com/swiftlang/swift/pull/88732


MaxDesiatov:
@swift-ci test

jakepetroules:
Can you also add support for this platform in Swift Build's SWBWebAssembly plugin?

MaxDesiatov:
> Can you also add support for this platform in Swift Build's SWBWebAssembly plugin?

Done in https://github.com/swiftlang/swift-build/pull/1335

MaxDesiatov:
https://github.com/swiftlang/swift/pull/87797
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

Sources/SwiftDriver/Toolchains/WebAssemblyToolchain.swift
```diff
@@ -62,6 +64,8 @@ public final class WebAssemblyToolchain: Toolchain {
 
   public let dummyForTestingObjectFormat = Triple.ObjectFormat.wasm
 
+  public var targetTriple: Triple?
```
artemcm: Could we parameterize `makeLinkerOutputFilename` instead? Even though it means updating unrelated toolchains, I'd rather we thread the triple to the decision point where we need it, instead of this kind of mutable state that's post-hoc initialized.

Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
```diff
@@ -137,7 +161,9 @@ extension WebAssemblyToolchain {
         let embeddedLibrariesPath: VirtualPath = runtimeResourcePath.appending(
           components: "embedded", targetTriple.triple
         )
-        commandLine.append(.flag("-Xlinker"))
+        if !isEmscripten {
```
artemcm: I don't know if I have a better suggestion, but it feels a bit unwieldy to special case things like this. Would there be too much duplication if we factor out the `emcc` code path out of here?

MaxDesiatov:
@swift-ci test

Sources/SwiftDriver/Jobs/WebAssemblyToolchain+LinkerSupport.swift
MaxDesiatov: Addressed by creating separate `EmscriptenToolchain`.

Sources/SwiftDriver/Toolchains/WebAssemblyToolchain.swift
MaxDesiatov: Creating a separate `EmscriptenToolchain` should address this as well.

MaxDesiatov:
@swift-ci test windows

Sources/SwiftDriver/Jobs/EmscriptenToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,148 @@
...
+      )
+
+      // `emcc` supports `-Xlinker` for passing flags to `wasm-ld`, but
+      // `-Xclang-linker` is `clang`-specific and must not be forwarded.
+      for linkerOpt in parsedOptions.arguments(for: .Xlinker) {
+        commandLine.appendFlag(.Xlinker)
+        commandLine.appendFlag(linkerOpt.argument.asSingle)
+      }
+      // `-Xemcc-linker` passes arguments directly to `emcc`.
+      try commandLine.appendAllArguments(.XemccLinker, from: &parsedOptions)
```
kateinoigakukun, line 133: If we pass `-Xemcc-linker` values to emcc, then I'm wondering if we really need a separate option name for emcc. Can't we just use `-Xclang-linker`? We can treat it just like "X flag for linker driver, not the linker itself" rather than strictly treating it only for "clang linker driver". emcc is meant to be used as drop-in-replacement as clang command, so this model should work I believe. Otherwise the build invoker has to switch the option based on if a target is emscripten or not, which is something I'm not a big fan of.

MaxDesiatov: It's not a drop-in replacement unfortunately, flags and options it supports differ substantially. I think keeping `-Xclang-linker` for it would be very confusing, as very few of the existing options work and require explicit updates. I'm totally fine with introducing `-Xlinker-driver` separately though.

kateinoigakukun: > It's not a drop-in replacement unfortunately, flags and options it supports differ substantially.

Oh, really. I know many options specific to emcc but don't know much about clang options not working for emcc. Do you have examples?

MaxDesiatov: I know of at least 3 cases that break or warn:
* `-resource-dir PATH` and `--sysroot PATH`: `emcc` has a silent flag-drop: it swallows the flag word but lets the path leak as a positional input leading to `wasm-ld: error: cannot open PATH: Is a directory`. Same bug hits `-B PATH`. Joined forms work.
* `-nostdlib`: accepted but semantically incompatible (emcc's runtime needs symbols `-nostdlib` drops)
* `-Wl,-z,max-page-size=16384`: `wasm-ld` warns and no-ops

As for the asymmetry: `emcc`-only flags like `-sENVIRONMENT=shell`, `--shell-file=` are hard errors under  `clang` (unknown argument).

Sources/SwiftOptions/Options.swift
```diff
@@ -1036,6 +1036,7 @@ extension Option {
   public static let Wwarning: Option = Option("-Wwarning", .separate, attributes: [.helpHidden, .frontend], metaVar: "<diagnostic_group>", helpText: "Treat this warning group as warning", group: .warningTreating)
   public static let Xcc: Option = Option("-Xcc", .separate, attributes: [.frontend, .synthesizeInterface], metaVar: "<arg>", helpText: "Pass <arg> to the C/C++/Objective-C compiler")
   public static let XclangLinker: Option = Option("-Xclang-linker", .separate, attributes: [.helpHidden], metaVar: "<arg>", helpText: "Pass <arg> to Clang when it is used for linking.")
+  public static let XemccLinker: Option = Option("-Xemcc-linker", .separate, attributes: [.helpHidden], metaVar: "<arg>", helpText: "Pass <arg> to emcc when it is used for linking.")
```
artemcm, line 1039: A reminder to follow the process here:
https://github.com/swiftlang/swift-driver#rebuilding-optionsswift

Sources/SwiftDriver/Jobs/EmscriptenToolchain+LinkerSupport.swift
```diff
@@ -0,0 +1,148 @@
...
+        to: &commandLine,
+        parsedOptions: &parsedOptions,
+        sanitizers: sanitizers,
+        lto: lto,
+        targetInfo: targetInfo,
+        linkerOutputType: linkerOutputType
+      )
+
+      // `emcc` supports `-Xlinker` for passing flags to `wasm-ld`, but
+      // `-Xclang-linker` is `clang`-specific and must not be forwarded.
```
artemcm, line 127: Should we warn when `-Xclang-linker` is used and we are on the `emcc` code path, instead of dropping them silently?

Sources/SwiftDriver/Jobs/WebAssemblyToolchainProtocol+LinkerSupport.swift
```diff
@@ -0,0 +1,118 @@
...
+    sanitizers: Set<Sanitizer>,
+    lto: LTOKind?,
+    targetInfo: FrontendTargetInfo,
+    linkerOutputType: LinkOutputType
+  ) throws {
+    let targetTriple = targetInfo.target.triple
+
+    // Delegate to Clang for sanitizers. It will figure out the correct linker
+    // options.
+    if linkerOutputType == .executable && !sanitizers.isEmpty {
```
artemcm, line 82: From what I can tell, in the code as-is, `linkerOutputType == .executable` is always going to be true. Do we think that can change in the future? If not, do we still need `linkerOutputType` here?

artemcm (approved):
Much cleaner, thanks!

MaxDesiatov:
Ah, sorry about the auto-merge that I forgot to disable, I'll submit a follow up PR!

Sources/SwiftDriver/Jobs/WebAssemblyToolchainProtocol+LinkerSupport.swift
MaxDesiatov: Addressed in https://github.com/swiftlang/swift-driver/pull/2136

Sources/SwiftDriver/Jobs/EmscriptenToolchain+LinkerSupport.swift
MaxDesiatov: It does emit a working in `EmscriptenToolchain` validation function, I've updated this comment to mention that in https://github.com/swiftlang/swift-driver/pull/2136

PR #2110, opened 2026-04-01
Trace events: add `-time-trace` to `swiftc`

MaxDesiatov:
Depends on https://github.com/swiftlang/swift/pull/88251.

This behaves in the same way as Clang's [`-ftime-trace`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-ftime-trace):
> Turn on time profiler. Generates JSON file based on output filename.

Generated JSON can be previewed with tools like https://ui.perfetto.dev.


Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -231,6 +241,24 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
         }
       }
 
+      // Add time trace flags so dependency module builds produce profiling data.
+      // Include the main module name in the trace file name to avoid collisions
```
artemcm: As you mention, there's a chance that in a client build system (e.g. `swift-build`) multiple targets have the same dependency module, because the dependency command-line recipe is the same. This allows the build system to dedup these dependency builds to avoid building the same dependency more than once.

If we add a module-name-unique command-line flag here, we may break this deduplication because the command-line recipes of otherwise identical dependencies will now differ due to this flag, so I think we should avoid doing this. *Or*, figure out how to take this into account during deduplication in `swift-build` though I doubt that's worth it.

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
```diff
@@ -231,6 +241,24 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
...
+      // when multiple targets compile the same dependency module.
+      if enableTimeTrace {
+        commandLine.appendFlag("-time-trace")
```
artemcm: I am not 100% sure how this option interacts with compiler caching, but additional flags/outputs will need to be taken into account. I think we should instead investigate having the dependency scanner specify these flags to module dependency command-line recipes.

artemcm (commented):
Is there a chance other tools will benefit from a `TimeTrace`-like thing in the future and we would need to hoist it out of the driver?

Sources/SwiftDriver/Driver/Driver.swift
```diff
@@ -411,6 +411,12 @@ public struct Driver {
   /// Path to the loaded module trace file.
   let loadedModuleTracePath: VirtualPath.Handle?
 
+  /// Path to the time trace profiling output file.
+  let timeTraceOutputPath: VirtualPath.Handle?
+
+  /// Time trace profiler, active when `-time-trace` is passed.
+  public var timeTrace = TimeTrace()
```
artemcm: Does this need to be `public` API?

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
MaxDesiatov: I removed it from supplementary outputs and removed suffixes. Do you still think dependency scanner investigation is needed if `-time-trace-path` is also `cacheInvariant`?

Sources/SwiftDriver/Driver/Driver.swift
MaxDesiatov: Thanks, addressed

Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
MaxDesiatov: I cleaned it up, it should no longer break deduplication.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test windows

PR #2111, opened 2026-04-06
Keep paths for prebuilt external swift modules for caching builds

hjyamauchi:
Incremental builds don't currently work on Windows because incremental builds are always invalidated due to an external dependency on the prebuilt Swift module, which fails the currency test as its path is just `Swift.swiftmodule` with no directory. We don't need to use abstract paths for prebuilt external swift modules for caching builds as they aren't the output of the current builds. This fixes incremental builds for Windows.

Issue https://github.com/swiftlang/swift/issues/88113

### DIFF
--- a/Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
+++ b/Sources/SwiftDriver/ExplicitModuleBuilds/ExplicitDependencyBuildPlanner.swift
@@ -389,7 +389,8 @@ public typealias ExternalTargetModuleDetailsMap = [ModuleDependencyId: ExternalT
                                   modulePath: TextualVirtualPath(path: swiftModulePath.fileHandle),
                                   isFramework: isFramework,
                                   moduleCacheKey: swiftModuleDetails.moduleCacheKey,
-                                  libraryLevel: dependencyInfo.libraryLevel))
+                                  libraryLevel: dependencyInfo.libraryLevel,
+                                  isPrebuiltExternal: false))
       case .clang:
         let dependencyInfo = try dependencyGraph.moduleInfo(of: dependencyId)
         let dependencyClangModuleDetails =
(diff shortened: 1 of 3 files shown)


hjyamauchi:
@swift-ci please test

hjyamauchi:
@swift-ci please test

hjyamauchi:
@swift-ci please test

hjyamauchi:
@swift-ci please test

hjyamauchi:

cachemeifyoucan (commented):
Can you please detailed what problem it fix by exactly how the abstract path cause problem?

The incremental has nothing to do with this path. This is just the path in JSON file, not the path incremental build use to figure out the change.

hjyamauchi:
Incremental builds are currently all invalidated (cause all source files to be rebuilt) with CAS enabled on Windows because the Swift module external dependency (`Swift.swiftmodule`) fails the incremental currency test [here](https://github.com/swiftlang/swift-driver/blob/7bcca29f0a1cdf2ea3a593cb4911a9c68e775a1a/Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift#L508). It fails because its module path is abstract (just the `Swift.swiftmodule` file name without a directory). The Swift module is the only *prebuilt* swift module dependency on Windows and has this issue.  The other (non-prebuilt) swift module dependencies like `_Concurrency` and `_StringProcessing` are okay because they are built from swiftinterface files which are the files to check for incrementality (despite having abstract paths).  I think that since the prebuilt one isn't built and isn't compiler output of normal swift application build, it doesn't need to have an abstract path and retain an actual module path as in this change, which solves this issue. I followed where the directory-less `Swift.swiftmodule` path comes from and it seems to come from there (the json file, then a swfitdeps file). Perhaps a prebuilt one is special?

cachemeifyoucan:
I don't think this is the correct fix. I didn't realize we encode external module dependencies in the `.swiftdeps` file (that is what I assumed by reading the code) and we need to fix the logic the check `old` vs `new`. I haven't looked at that code to see if we can check without the file path. @artemcm might have better idea.

cachemeifyoucan:
This is probably not windows only. A fast temporary workaround maybe just to keep all the paths if no prefix mapping is used.

cachemeifyoucan:
Let's try this: https://github.com/swiftlang/swift-driver/pull/2112

PR #2124, opened 2026-04-18
add -Rhelp and -R <group>

hnrklssn:
When passed to the driver, these options are remapped to -Rhelp-swiftc and -Rswiftc, respectively, before invoking the compiler. This lets the compiler gate remark availability on whether the options were passed to the driver or to the compiler directly. the -R*-swiftc versions are not to be used directly.

Paired with https://github.com/swiftlang/swift/pull/88516


hnrklssn:
https://github.com/swiftlang/swift/pull/88516
@swift-ci please test

PR #2126, opened 2026-04-23
[Caching] Abstract clang module map path with prefix mapping

cachemeifyoucan:
When using prefix mapping, don't serialize clang module path to swift-frontend. This removes the absolute path dependency when building with compilation caching and prefix mapping.

Clang module path is not actually used by the swift compiler when loading modules. It is only serialized in swift binary module and allows lldb to some fallback paths when the PCM cannot be loaded. This is not used with prefix mapping.

 
- **Explanation**: Drop clang module map path when using caching + prefix mapping.
  
- **Scope**: Affect cache hit rate for caching + prefix mapping.
  
- **Issues**: rdar://175017181
  
- **Risk**: Low. Only affect caching + prefix mapping.
  
- **Testing**: Local testing. Once everything landed, tests can be enabled in swift-build.


cachemeifyoucan:
@swift-ci please test

artemcm (approved):

nkcsgexi (approved):

PR #2127, opened 2026-04-27
Adjust incrementalCompilationCachingBasic to reproduce issue 88113

hjyamauchi:
Need to not use load module interface to reproduce issue 88113 and add a note that it needs to be fixed with prefix mapping enabled.

### DIFF
--- a/Tests/SwiftDriverTests/Helpers/IncrementalTestHarness.swift
+++ b/Tests/SwiftDriverTests/Helpers/IncrementalTestHarness.swift
@@ -157,7 +157,7 @@ final class IncrementalTestHarness {
     }
   }
 
-  init(config: TestBuildConfig = .implicitModule) throws {
+  init(config: TestBuildConfig = .implicitModule, preferInterfaceModuleLoadMode: Bool = true) throws {
     self.config = config
 
     guard let sdkArgs = try Driver.sdkArgumentsForTesting() else {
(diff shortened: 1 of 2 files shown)


hjyamauchi:
@swift-ci please test

cachemeifyoucan:
Prefix mapping incremental build is not working great right now. I have a fix but I was waiting for rewriting the tests to land it so I don't need to write test again. Let me post that.

hjyamauchi:
ok. I thought it'd be good to keep the regression test coverage for the incremental build issue. I can hold this off (or maybe ask you to incorporate this PR into your change) as you are reworking the tests? What do you prefer?

cachemeifyoucan:
https://github.com/swiftlang/swift-driver/pull/2128

I rewrite the incremental tests to swift testing so it is easy to spot any tests have different behavior.

hjyamauchi:
I confirmed that https://github.com/swiftlang/swift-driver/pull/2128 fixes this test for the `cachingPrefixMapped` config. Thank you. I updated this PR just to change the test to not to use `-module-load-mode prefer-interface` to reproduce the issue 88113. Are you open to that?

cachemeifyoucan:
We need to make sure we have coverage for both.  In general, we shouldn't use this flag to control the coverage, but using inputs that are interface-only or binary module only to ensure the coverage.

PR #2128, opened 2026-04-27
[IncrementalCompilation] Support prefix-mapped paths and external module resolution

cachemeifyoucan:
- Add prefixMapping to IncrementalDependencyAndInputSetup, threaded from Driver. Provide reverseMapPath() to convert prefix-mapped paths back to real paths during .swiftdeps and priors deserialization.
- Add pathReverser to SourceFileDependencyGraph deserialization so all identifier strings in .swiftdeps files are reverse-mapped at read time.
- Apply pathReverser during priors graph deserialization as well.
- Add externalModulePathMap to ExternalDependencyCurrencyCache for resolving abstract module paths (e.g. "B.swiftmodule") to absolute paths during filesystem currency checks. Populated from ExplicitDependencyBuildPlanner scanning results.
- Serialize/deserialize externalModulePathMap in priors file via new externalModulePathNode record (version bumped to 1.6.0).
- Add integration tests for caching and prefix-mapped caching incremental builds, including upstream dependency invalidation with diagnostics.

Assisted-By: Claude

 
- **Explanation**: Fixes incremental build for caching prefix mapping configuration to correctly identify unchanged dependency to avoid rebuilds.
  
- **Scope**: Caching incremental build only.
  
- **Issues**: rdar://174330691
  
- **Risk**: Low. Limited scope and supports a configuration that wasn't previously well supported.
  
- **Testing**: Unit-tests

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -72,6 +72,25 @@ public final class IncrementalCompilationState {
       reporter: reporter)
       .compute(batchJobFormer: &driver)
 
+    // Re-populate the external module path map from fresh scanning results.
+    // The deserialized map from priors was already used by FirstWaveComputer
+    // above for currency checks; this overwrites it so that priors serialized
+    // at the end of this build reflect up-to-date paths.
+    if let planner = explicitModulePlanner {
+      initialState.graph.blockingConcurrentAccessOrMutation {
+        let map = Dictionary(uniqueKeysWithValues:
+          planner.dependencyGraph.modules.compactMap {
+            moduleId, moduleInfo -> (String, VirtualPath.Handle)? in
+            guard case .swiftPrebuiltExternal = moduleId,
+                  case .swiftPrebuiltExternal(let details) = moduleInfo.details
+            else { return nil }
+            let abstractPath = moduleId.moduleName + "." + FileType.swiftModule.rawValue
+            return (abstractPath, details.compiledModulePath.path)
+          })
+        initialState.graph.setExternalModulePathMap(map)
+      }
+    }
+
     self.info = initialState.graph.info
 
     self.protectedState = ProtectedState(
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
@@ -17,6 +17,7 @@ import class Dispatch.DispatchQueue
 
 import class TSCBasic.DiagnosticsEngine
 import protocol TSCBasic.FileSystem
+import struct TSCBasic.AbsolutePath
 
 // Initial incremental state computation
 extension IncrementalCompilationState {
@@ -56,7 +57,8 @@ extension IncrementalCompilationState {
           buildRecordInfo,
           reporter, driver.inputFiles,
           driver.fileSystem,
-          driver.diagnosticEngine
+          driver.diagnosticEngine,
+          prefixMapping: driver.prefixMapping
         ).computeInitialStateForPlanning(driver: &driver)
     else {
       Self.removeDependencyGraphFile(driver)
@@ -114,6 +116,10 @@ extension IncrementalCompilationState {
     @_spi(Testing) public let inputFiles: [TypedVirtualPath]
     @_spi(Testing) public let fileSystem: FileSystem
 
+    /// Prefix mapping from the driver, used to reverse-map paths in .swiftdeps
+    /// files that were written with prefix mapping active.
+    @_spi(Testing) public let prefixMapping: [(AbsolutePath, AbsolutePath)]
+
     /// The state managing incremental compilation gets mutated every time a compilation job completes.
     /// This queue ensures that the access and mutation of that state is thread-safe.
     @_spi(Testing) public let incrementalCompilationQueue: DispatchQueue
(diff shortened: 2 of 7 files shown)


hjyamauchi (approved):
Does this cover the cases for the abstract paths for binary modules as in https://github.com/swiftlang/swift/issues/88113#issuecomment-4200474527 ?

cachemeifyoucan:
I don't know why issue 88113 is related here. That problem should be resolved and the example is not using prefix mapping.

Prefix mapping is to abstract the path away to allow cache hits regardless of the source/module path and in such case, swift-driver will abstract the path away so that can happen. For regular caching build, there is no path abstraction anymore, so it should just work.

If you want to try path prefix mapping, I think this patch should fix the incremental build. If you try it and let me know if you still see problem.

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -575,6 +582,8 @@ extension ModuleDependencyGraph {
     private var currencyCache = [ExternalDependency: Bool]()
     public internal(set) var externalDependencyFileHashes: [ExternalDependency: String]
     private let useFileHashes: Bool
+    /// Maps relative module paths to absolute paths for filesystem checks.
+    var externalModulePathMap: [String: String] = [:]
```
artemcm: Any reason to not have `AbsolutePath` or `VirtualPath.Handle` values here?

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -1242,6 +1280,12 @@ extension ModuleDependencyGraph {
         _ = input.name.intern(in: self.internedStringTable)
       }
 
+      // Ensure external module path map strings are interned
+      for (abstractPath, resolvedPath) in graph.currencyCache.externalModulePathMap {
```
artemcm: The code above ensures sorted iteration, I gather to ensure reproducible `.priors`. We should probably do the same here.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -583,11 +592,21 @@ extension ModuleDependencyGraph {
       self.useFileHashes = useFileHashes
     }
 
+    /// Resolve an external dependency to a filesystem path, using the module
+    /// path map for relative/abstract paths.
+    private func resolvedPath(for externalDependency: ExternalDependency) -> VirtualPath? {
+      if let resolvedPathString = externalModulePathMap[externalDependency.fileNameString],
+         let resolvedPath = try? VirtualPath(path: resolvedPathString) {
```
artemcm: Do we need to flag the case where we have a `resolvedPathString` but can't construct a `VirtualPath` out of it?

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
```diff
@@ -72,6 +72,25 @@ public final class IncrementalCompilationState {
...
+          planner.dependencyGraph.modules.compactMap {
+            moduleId, moduleInfo -> (String, String)? in
+            guard case .swiftPrebuiltExternal = moduleId,
+                  case .swiftPrebuiltExternal(let details) = moduleInfo.details
+            else { return nil }
+            let abstractPath = moduleId.moduleName + "." + FileType.swiftModule.rawValue
+            let resolvedPath = VirtualPath.lookup(details.compiledModulePath.path).name
+            return (abstractPath, resolvedPath)
+          })
+        initialState.graph.setExternalModulePathMap(map)
```
artemcm, line 90: I'm not sure I follow: when will the deserialized external module map path from a prior build get used if we re-compute it from scan results every time?

Sources/SwiftDriver/IncrementalCompilation/IncrementalDependencyAndInputSetup.swift
```diff
@@ -181,6 +189,27 @@ extension IncrementalCompilationState {
     func isPartOfBuild(_ sourceFile: SwiftSourceFile) -> Bool {
       return self.inputFiles.contains(sourceFile.typedFile)
     }
+
+    /// Reverse prefix-map a path string that was written with prefix mapping
+    /// active. Converts e.g. "/^tmp/foo.swift" back to "/real/path/foo.swift".
+    @_spi(Testing) public func reverseMapPath(_ mappedPath: String) -> String {
```
artemcm: I don't think `@_spi(Testing) public` is currently needed.

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
cachemeifyoucan: Done. The key is an abstract path that will never be converted to path so keep the key to be String.

Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
cachemeifyoucan: Sometimes you don't scan when doing incremental build, and you need to map the path to figure out if you need to scan or not. This is for that purpose.

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:

cachemeifyoucan:
ping

artemcm (approved):

PR #2132, opened 2026-05-06
[Caching] Prefix-map supplementary output file map keys

cachemeifyoucan:
With caching + prefix mapping, a few insertions into the supplementary output file map used the original (unmapped) source path as the key, producing a phantom entry alongside the correctly remapped one. Route every key through the caching prefix map so the file map matches the paths the frontend sees.

rdar://175888979


cachemeifyoucan:
@swift-ci please test

artemcm (approved):

PR #2144, opened 2026-05-21
[6.4][IncrementalCompilation] Support prefix-mapped paths and external module resolution

cachemeifyoucan:
- **Explanation**: Support proper incremental caching when all the paths are prefixed mapped and module path abstracted out. This change adds mapping between mapped file path to the actual physical file path so the incremental build computer can compute the state correctly.
  
- **Scope**: Makes incremental build for caching + prefix mapping build have the behavior as regular incremental build.
  
- **Issues**: rdar://174330691
  
- **Original PRs**: https://github.com/swiftlang/swift-driver/pull/2128
  
- **Risk**: Low. The incremental build for caching + prefix mapping is almost always clean build before and the fix is improving that situation. No behavior change for other kind of builds. 
  
- **Testing**: Unit testing added
  
- **Reviewers**: @artemcm @hjyamauchi

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState.swift
@@ -72,6 +72,25 @@ public final class IncrementalCompilationState {
       reporter: reporter)
       .compute(batchJobFormer: &driver)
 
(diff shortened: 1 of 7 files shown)


cachemeifyoucan:
@swift-ci please test

DougGregor (approved):

PR #2150, opened 2026-06-09
[Driver] Recognize embedded stdlib layout when picking plugin paths

hnrklssn:
For embedded `noneOS` targets the stdlib lives at
`<resource-dir>/embedded/<triple>/Swift.swiftmodule`, but the existing `hasToolchainStdlib` probe only checked `<resource-dir>/<platform>/`, which is `<resource-dir>/Swift.swiftmodule` (no platform subdir) for `noneOS` because `Triple.platformName()` returns nil. The probe always returned false, so the conditional swap in
`addCommonFrontendOptions` placed the SDK's `-external-plugin-path` ahead of the toolchain's `-plugin-path`. The first-emplace race in `PluginLoader` then registered the bootstrap Xcode SDK's (potentially stale) `libSwiftMacros.dylib` as `SwiftMacros`, shadowing the freshly-built one and producing
"type 'SwiftMacros.SwiftifyImportMacro' could not be found" errors when the embedded headers used macros that post-date the bootstrap platform (and some CI machines run a pretty old host OS, so they lack `_SwiftifyImport`). This should mostly be an issue when bootstrapping.

Branch the probe on `triple.os == .noneOS` so the embedded layout is checked directly, with a single `fileSystem.exists` call. Covers Apple `*-apple-none-macho` (the visible failure), plus Wasm `*-unknown-none-wasm`, ARM/RISC-V/x86 `*-none-*-eabi|elf`, and AVR.

Add 8 regression tests in `ToolchainTests` (4 Apple macho + 4 non-Apple `noneOS`), each gated with `requireFrontendSupportsTarget` so they self-skip on toolchains lacking the relevant LLVM backend.

rdar://178814859

Assisted-by: Claude Opus 4.7

### DIFF
--- a/Tests/SwiftDriverTests/ToolchainTests.swift
+++ b/Tests/SwiftDriverTests/ToolchainTests.swift
@@ -698,6 +698,81 @@ import CRT
     #endif
   }
 
+  func assertEmbeddedPluginPathOrder(
+    triple: String
+  ) async throws {
+    let sdkRoot = try testInputsPath.appending(
+      components: ["Platform Checks", "iPhoneOS.platform", "Developer", "SDKs", "iPhoneOS13.0.sdk"])
+
+    try await withTemporaryDirectory { resourceDir in
+      try localFileSystem.createDirectory(
+        resourceDir.appending(components: "embedded", triple, "Swift.swiftmodule"),
+        recursive: true)
+
+      var driver = try TestDriver(args: [
+        "swiftc", "-typecheck", "foo.swift",
+        "-target", triple,
+        "-enable-experimental-feature", "Embedded",
+        "-sdk", VirtualPath.absolute(sdkRoot).name,
+        "-resource-dir", resourceDir.pathString,
+      ])
+      guard driver.isFrontendArgSupported(.pluginPath) else { return }
+
+      let jobs = try await driver.planBuild().removingAutolinkExtractJobs()
+      let job = try #require(jobs.first)
+
+      let toolchainPluginIdx = try #require(
+        job.commandLine.firstIndex(of: .flag("-plugin-path")),
+        "Expected toolchain -plugin-path to be emitted for \(triple)")
+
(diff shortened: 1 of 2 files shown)


DougGregor (approved):
This looks reasonable, thank you!

hnrklssn:
@swift-ci please test

Tests/SwiftDriverTests/ToolchainTests.swift
```diff
@@ -698,6 +698,81 @@ import CRT
...
+      guard let externalPluginIdx =
+        job.commandLine.firstIndex(of: .flag("-external-plugin-path")) else { return }
+      #expect(toolchainPluginIdx < externalPluginIdx,
+              "-plugin-path must precede -external-plugin-path so the toolchain libSwiftMacros wins (\(triple))")
+    }
+  }
+
+  // Check that we pass the toolchain plugin path before the external plugin
+  // path for embedded targets.
+  @Test(.requireFrontendSupportsTarget("armv6-apple-none-macho"))
```
cachemeifyoucan: This is a perfect case to use parameterized tests?

hnrklssn: I was not familiar with that feature so I looked it up, but it doesn't seem like that would be compatible with `.requireFrontendSupportsTarget` since that's only evaluated once (unless I'm missing something). I could add a guard inside the test function if you want, but then it would show up as passing instead of unsupported. Maybe that's still preferable?

cachemeifyoucan: You don't need to use `requireFrontendSupportsTarget` in your Test trait. You just need to construct Collection to be passed as `argument:` and you can construct this collection with supported target.

hnrklssn: Ahh yeah that works, thanks!

hnrklssn:
@swift-ci please test

hnrklssn:

PR #2153, opened 2026-06-11
Triple: bring OS version mapping up to date

nkcsgexi:
- Map darwin25-26 → macOS 26-27 and align darwin27+
  - llvm side PR: https://github.com/llvm/llvm-project/pull/202690 
- Fix OS triple version mapping for the 26-aligned releases

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple.swift
+++ b/Sources/SwiftDriver/Utilities/Triple.swift
@@ -1601,11 +1601,16 @@ extension Triple {
         version.micro = 0
         version.minor = version.major - 4
         version.major = 10
-      } else {
+      } else if version.major < 25 {
         version.micro = 0
         version.minor = 0
-        // darwin20+ corresponds to macOS 11+.
+        // darwin20-24 corresponds to macOS 11-15.
         version.major = version.major - 9
+      } else if version.major == 25 || version.major == 26 {
+        version.micro = 0
+        version.minor = 0
+        // darwin25-26 corresponds to macOS 26-27.
+        version.major = version.major + 1
       }
 
     case .macosx:
--- a/Sources/SwiftDriver/Utilities/Triple.swift
+++ b/Sources/SwiftDriver/Utilities/Triple.swift
@@ -1573,6 +1573,32 @@ extension Triple {
 // MARK: - Darwin Versions
 
 extension Triple {
+  // Version 26 alignment helper. Each Apple OS jumped to version 26, so the
+  // last pre-jump release canonicalizes to 26; versions in the unshipped gap
+  // below 26 are left unchanged. Only `_iOSVersion` consumes this.
+
+  /// Canonicalize the last pre-26 release to the year-aligned version 26.
+  /// Gap versions (between the pre-jump release and 26) pass through unchanged.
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/Utilities/Triple.swift
```diff
@@ -1573,6 +1573,60 @@ extension Triple {
 // MARK: - Darwin Versions
 
 extension Triple {
+  // Year-26 alignment helpers, ported from LLVM's `isValidVersionForOS` /
+  // `getCanonicalVersionForOS`. Each Apple OS jumped to the year (26), leaving
```
jakepetroules: nit: it jumped to the year+1, not the year :) I would replace "year" with "version" in these sentences

nkcsgexi: Thanks! Fixed.

nkcsgexi:
@swift-ci please test

nkcsgexi:
@swift-ci please test

Tests/SwiftDriverTests/TripleTests.swift
```diff
@@ -1140,6 +1180,29 @@ import class TSCBasic.DiagnosticsEngine
...
+  /// Mirrors LLVM `getiOSVersion` / `getCanonicalVersionForOS` version-26 alignment.
+  @Test func iOSVersionAlignment() {
+    func iOSVersion(_ triple: String) -> Triple.Version { Triple(triple)._iOSVersion }
+
+    // iOS / tvOS 19 -> iOS 26; already-aligned and pre-jump values pass through;
+    // a gap version (20-25) bumps by 7.
+    #expect(iOSVersion("arm64-apple-ios19.0") == Triple.Version(26, 0, 0))
+    #expect(iOSVersion("arm64-apple-ios26.0") == Triple.Version(26, 0, 0))
+    #expect(iOSVersion("arm64-apple-ios18.0") == Triple.Version(18, 0, 0))
+    #expect(iOSVersion("arm64-apple-ios21.0") == Triple.Version(28, 0, 0))
```
jakepetroules: Is this really the right behavior? If I pass arm64-apple-ios20.0 to clang, it errors out completely, rather than transforming it to an iOS 27 triple.

nkcsgexi: I believe it's better to translate these versions mechanically at the driver layer because the compiler will eventually diagnose them anyway.

nkcsgexi: Additionally, there doesn't seem to be a great diagnostic engine at this layer in the driver that allows us to emit good diagnostics.

nkcsgexi: Probably another option is to keep the version as is if they fall into the invalid gap so that Swift and Clang can later diagnose. hmm

nkcsgexi:
@swift-ci please test

jakepetroules (approved):

nkcsgexi:
@swift-ci please test Windows

nkcsgexi:
@swift-ci please test Windows

nkcsgexi:
@swift-ci please test Windows

PR #2154, opened 2026-06-15
🍒[6.4] Bring OS version mappings up to date

nkcsgexi:
- Explanation: teach swift-driver about two recent OS version mappings: (1) darwin25-26 → macOS 26-27, and (2) version 26 alignment that was introduced last year.
- Scope: these target triple APIs are used by SwiftPM and swift-driver as a standalone executable. 
- Risk: Very Low. We already have OS triple version mappings like these at multiple layers.
- Original PR: https://github.com/swiftlang/swift-driver/pull/2153
- Radar: 177999453
- Reviewer: @jakepetroules

### DIFF
--- a/Sources/SwiftDriver/Utilities/Triple.swift
+++ b/Sources/SwiftDriver/Utilities/Triple.swift
@@ -1573,6 +1573,32 @@ extension Triple {
 // MARK: - Darwin Versions
 
 extension Triple {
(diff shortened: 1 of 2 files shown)


nkcsgexi:
@swift-ci please test

tbkka (approved):

PR #2157, opened 2026-07-09
[SwiftScan] Introduce `SwiftScanCapabilities`

akyrtzi:
This is for querying the capabilities of the libSwiftScan library.

I've only added `supportsCASSandbox` for now since other capabilities are either exposed from other APIs or I'm not sure if they are intended to be exposed or not.

### DIFF
--- a/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
+++ b/Sources/SwiftDriver/SwiftScan/SwiftScan.swift
@@ -120,6 +120,19 @@ private extension String {
   }
 }
 
+/// Used for querying for the capabilities of libSwiftScan.
+public struct SwiftScanCapabilities {
+  let swiftScan: SwiftScan
+
+  public init(dylib path: AbsolutePath? = nil) throws {
+    self.swiftScan = try SwiftScan(dylib: path)
+  }
+
+  public var supportsCASSandbox: Bool {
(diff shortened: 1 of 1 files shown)


akyrtzi:
@swift-ci Please test

cachemeifyoucan (commented):
SwiftScanCAS.supportsCASSandbox is not enough?

akyrtzi:
> SwiftScanCAS.supportsCASSandbox is not enough?

It requires initializing a whole CAS instance, I want something simpler that can be attached as test requirement.

cachemeifyoucan:
> It requires initializing a whole CAS instance, I want something simpler that can be attached as test requirement.

I would either `@testable` import SwiftDriver, or just make the underlying `SwiftScan` API public.

cachemeifyoucan:
Actually, now I think about it, we should just make this a new feature in https://github.com/swiftlang/swift/blob/main/lib/Option/features.json and both driver and build system have existing construct to query that.

akyrtzi:
> Actually, now I think about it, we should just make this a new feature in https://github.com/swiftlang/swift/blob/main/lib/Option/features.json and both driver and build system have existing construct to query that.

SGTM, added "dependency-scan-cas-sandbox" feature in https://github.com/swiftlang/swift/pull/89448

PR #2158, opened 2026-07-14
Darwin needs -isysroot instead of --sysroot to take precedence over SDKROOT

ian-twilightcoder:
Darwin targets prefer -isysroot instead of --sysroot, the latter is only used as a fallback if both `-isysroot` and SDKROOT aren't present. That causes issues where SDKROOT is set in the environment but Run Script build phases run `swiftc -sdk <different SDK>` to try to use a different SDK. swiftc will use the passed SDK, but when it goes to link, clang will use the SDKROOT SDK and the link will fail.

rdar://182106220

### DIFF
--- a/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
+++ b/Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
@@ -222,7 +222,7 @@ extension DarwinToolchain {
 
     // Add the SDK path
     if let sdkPath = targetInfo.sdkPath?.path {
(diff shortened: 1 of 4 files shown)


cachemeifyoucan (commented):
LGTM. There is a test that wasn't updated.

ian-twilightcoder:
> LGTM. There is a test that wasn't updated.

Not sure how I missed that one, fixed now.

ian-twilightcoder:
@swift-ci test

cachemeifyoucan (approved):

PR #2159, opened 2026-07-16
[Driver] Disable scanner prefix mapping when emitting an ObjC header …

cachemeifyoucan:
…with an external bridging header

When compilation caching is enabled with `-scanner-prefix-*` prefix mapping, using an external bridging header (`-import-objc-header` / `-import-bridging-header`) together with `-emit-objc-header` leaks the prefix-mapped bridging header path into the generated Objective-C header. Downstream consumers cannot resolve the mapped path, so the generated header is unusable.

Detect this combination, drop the prefix mapping, and emit a warning. When the frontend supports it, suggest adopting
`-internal-import-bridging-header`, which keeps the bridging header internal and allows prefix mapping to remain enabled.

### DIFF
--- a/Sources/SwiftDriver/Driver/Driver.swift
+++ b/Sources/SwiftDriver/Driver/Driver.swift
@@ -322,9 +322,19 @@ public struct Driver {
                                                        .parentDirectory { // toolchain
       mapping.append((toolchainPath, toolchainMapping))
     }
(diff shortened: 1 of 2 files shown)


cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:
Ping.

benlangmuir (approved):

PR #2162, opened 2026-07-21
IncrementalCompilation: tolerate duplicate input paths in build record

nkcsgexi:
Tentatively fixing: rdar://181716113

### DIFF
--- a/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState+Extensions.swift
+++ b/Sources/SwiftDriver/IncrementalCompilation/IncrementalCompilationState+Extensions.swift
@@ -119,6 +119,14 @@ extension Diagnostic.Message {
     )
   }
 
+  static func warning_incremental_disabled_duplicate_input(
+    _ path: VirtualPath, _ types: [FileType]
+  ) -> Diagnostic.Message {
+    let typeList = types.map { "'\($0)'" }.joined(separator: ", ")
+    return .warning(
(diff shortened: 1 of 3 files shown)


nkcsgexi:
@swift-ci please test

Sources/SwiftDriver/IncrementalCompilation/BuildRecord.swift
```diff
@@ -70,7 +70,9 @@ extension BuildRecord {
       swiftVersion: actualSwiftVersion,
       buildStartTime: timeBeforeFirstJob,
       buildEndTime: timeAfterLastJob,
-      inputInfos: Dictionary(uniqueKeysWithValues: inputInfosArray)
+      // Two inputs may share one path under different FileTypes; both collapse to
```
owenv: I wonder if we should thread through some way to report an error / warning here when this happens, and not try to attempt an incremental compile. Without knowing how we got into this situation it's hard to know if an incremental compile will be sound

nkcsgexi: That's a great idea. Let me try that.

nkcsgexi:
@swift-ci please test

nkcsgexi:
@owenv mind taking another look?

nkcsgexi:
@swift-ci please test Windows

owenv (approved):

PR #2172, opened 2026-08-21
SDK: use DefaultDeploymentTarget in SDKInfo.json for -clang-target triples when the field is present

nkcsgexi:
rdar://184842115


artemcm (approved):
Thank you!

PR #2188, opened 2026-09-11
[IncrementalCompilation] Add a safety margin when checking external dependency currency

artemcm:
`ExternalDependencyCurrencyCache.isCurrentWRTFileSystem` compared an imported swiftmodule's modification time directly against `buildStartTime`. The driver captures `buildStartTime` with sub-second precision from its own clock, but some filesystems only report modification times with whole-second granularity. A dependency rewritten in the same second as `buildStartTime` then reports a truncated mod time that looks older than `buildStartTime`. The driver treats the changed dependency as current and skips recompiling its dependents, which produces a stale module and potentially a link failure.

Require a full second of slack so truncation cannot flip the comparison.

This repository already accounts for coarse timestamp granularity elsewhere; see the `touch` helper in `PathExtensions.swift`.

Resolves rdar://165472777


owenv (approved):

artemcm:

artemcm:
@swift-ci test Windows

artemcm:
@swift-ci test Windows

Sources/SwiftDriver/IncrementalCompilation/ModuleDependencyGraph.swift
```diff
@@ -644,13 +644,23 @@ extension ModuleDependencyGraph {
...
+    private static let filesystemGranularitySafetyMargin = TimePoint.seconds(1)
+
     private func isCurrentWRTFileSystem(_ externalDependency: ExternalDependency) -> Bool {
       guard let depFile = resolvedPath(for: externalDependency),
             let fileModTime = try? self.fileSystem.lastModificationTime(for: depFile)
       else {
         return false
       }
-      return fileModTime < self.buildStartTime
+      return fileModTime + Self.filesystemGranularitySafetyMargin < self.buildStartTime
```
cachemeifyoucan, line 663: Should this just be `<=`? I think we should universally treat the same time as ambiguous, thus conservatively rebuilt. 

In test harness, I already added tooling to make sure file touches are done in a loop till it can be tell apart by the file system.

PR #2189, opened 2026-09-14
[Darwin] Don't pass -add_ast_path under -explicit-module-build

adrian-prantl:
Under an explicit module build the compile job is passed -debug-module-path, which records this module's own binary swiftmodule in its debug info for the debugger to load it from.

The Xcode build system already suppresses -add_ast_path for explicitly built modules; match that in the driver.

rdar://187411108
Assisted-by: Claude

See also https://github.com/swiftlang/swift/pull/92255


Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
```diff
@@ -582,6 +582,46 @@ func getStdlibShimsPaths(_ driver: Driver) throws -> (AbsolutePath, AbsolutePath
     }
   }
 
+  @Test(.requireHostOS(.macosx, comment: "-add_ast_path is only passed on Darwin"))
```
compnerd: I think that this isn't entirely correct - hostOS here is actually build OS, and there is no reason to assume that you aren't cross-compiling (or cross-testing). "`-add_ast_path` requires ld64 which is only available on macOS" is the real reason that this cannot be run.

Sources/SwiftDriver/Jobs/DarwinToolchain+LinkerSupport.swift
```diff
@@ -63,6 +63,10 @@ extension DarwinToolchain {
     sanitizers: Set<Sanitizer>,
     targetInfo: FrontendTargetInfo
   ) throws -> ResolvedTool {
+    // Under an explicit module build the compile job is passed
+    // -debug-module-path, which supersedes -add_ast_path for this module.
+    let shouldAddASTPaths = !parsedOptions.hasArgument(.driverExplicitModuleBuild)
```
cachemeifyoucan: This should also be tied with swift-frontend support `-debug-module-path` since swift-driver can get out of sync with swift-frontend.

adrian-prantl: Done.

Tests/SwiftDriverTests/ExplicitModuleBuildTests.swift
adrian-prantl: If we're being technical, it's the scanning part that requires a macOS SDK, we don't actually invoke the linker here. Either way, I've updated the comment.

artemcm (approved):

artemcm:
> test with [swiftlang/swift#92255](https://github.com/swiftlang/swift/pull/92255) @swift-ci test

I believe that cross-repo testing does not work from this repo, it must be done from the `swift` repo and include this PR.

adrian-prantl:
@swift-ci test

adrian-prantl:
@swift-ci test windows

PR #2191, opened 2026-09-15
DriverAsExecutable: add GNU make jobserver support. 

nkcsgexi:
When swift-driver runs as an executable under `make -jN`, each process independently spawns up to `-j` frontend jobs, so N sibling drivers oversubscribe the CPU by a factor of N. GNU make solves this with a jobserver: a shared token pool advertised in MAKEFLAGS that all participants draw from, bounding total concurrency across the whole build rather than per process.

Participate in that pool when opted in via the new `-experimental-swift-driver-use-gnu-jobserver` flag. Authentication is read from MAKEFLAGS (`--jobserver-auth=` / `--jobserver-fds=`), covering both the inherited pipe descriptors of older make and the named FIFO of make 4.4+. A driver still runs one job on its own implicit token when the pool is empty, so it never deadlocks.

This only affects the executable path (SwiftDriverExecutor/MultiJobExecutor driving `Driver.run`); library clients that plan jobs and run them in their own build engine are unchanged.

Resolves: rdar://187486093


nkcsgexi:
@swift-ci please smoke test

Sources/SwiftDriverExecution/JobServer.swift
```diff
@@ -0,0 +1,500 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2014 - 2025 Apple Inc. and the Swift project authors
```
artemcm:
```suggestion
// Copyright (c) 2014 - 2026 Apple Inc. and the Swift project authors
```

Sources/SwiftOptions/Options.swift
```diff
@@ -587,6 +587,7 @@ extension Option {
   public static let experimentalSkipNonInlinableFunctionBodies: Option = Option("-experimental-skip-non-inlinable-function-bodies", .flag, attributes: [.helpHidden, .frontend], helpText: "Skip type-checking and SIL generation for non-inlinable function bodies")
   public static let experimentalSpiImports: Option = Option("-experimental-spi-imports", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Enable experimental support for SPI imports")
   public static let experimentalSpiOnlyImports: Option = Option("-experimental-spi-only-imports", .flag, attributes: [.helpHidden, .frontend, .noDriver], helpText: "Enable use of @_spiOnly imports")
+  public static let experimentalSwiftDriverUseGnuJobserver: Option = Option("-experimental-swift-driver-use-gnu-jobserver", .flag, attributes: [.helpHidden, .doesNotAffectIncrementalBuild], helpText: "Participate in the GNU make jobserver advertised in MAKEFLAGS, letting its token pool (not -j) bound concurrency across the build")
```
artemcm: See my comment on the corresponding `swift` PR on this flag.

nkcsgexi:
@swift-ci please test

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -339,6 +370,9 @@ public final class MultiJobExecutor {
       workload: workload,
       executorDelegate: executorDelegate,
       jobQueue: jobQueue,
+      jobServerDispatcher: jobServer.map {
```
artemcm: We already check `if jobServer != nil {` above, perhaps we can configure the dispatcher in that conditional block as well.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
```diff
@@ -328,9 +349,19 @@ public final class MultiJobExecutor {
...
+
     let jobQueue = OperationQueue()
     jobQueue.name = "org.swift.driver.job-execution"
-    jobQueue.maxConcurrentOperationCount = numParallelJobs
+    if jobServer != nil {
+      // A jobserver is a build-wide concurrency limit, so when there is one the
+      // tokens we hold -- not `-j` -- bound how many jobs run. Capping the queue
+      // at `numParallelJobs` would cap us below the pool, and since `-j`
+      // defaults to 1 that would serialize the very builds a jobserver widens.
+      jobQueue.maxConcurrentOperationCount = Int.max
```
artemcm, line 366: `dispatchRemainingWithoutTokens` adds *all* jobs, however many there are, to the queue simultaneously. Why not keep/respect `numParallelJobs` here, just in case? I don't think we want to exceed it even with a build-wide concurrency limit imposed by a job server.

Sources/SwiftDriverExecution/JobServer.swift
```diff
@@ -0,0 +1,500 @@
...
+      // token, which adds nothing to the pool for us to read.
+      if hasShutDown { return nil }
+      if let token = takeImplicitToken() { return token }
+
+      let ready = waitForToken()
+      if ready.wasWokenUp { drainWakeup() }
+      guard ready.hasToken else { continue }
+
+      var byte: UInt8 = 0
+      let count = retryOnInterrupt { read(readFD, &byte, 1) }
```
artemcm, line 351: On the inherited-pipe transport this `read` can block: poll in `waitForToken` only says a byte was there, and a concurrent subling driver sharing the read end can take it first. 
Since the descriptor has to stay blocking, we halt here and can't reach the `takeImplicitToken()` check at the top of the loop, so we stay idle even after our own job returns the implicit token.

Sources/SwiftDriverExecution/JobServer.swift
```diff
@@ -0,0 +1,500 @@
...
+
+      if shouldStop {
+        return dispatchRemainingWithoutTokens()
+      }
+
+      // Take the token before claiming the work, so that we never hold a token
+      // with nothing to spend it on. Only this thread removes from `pending`,
+      // so it cannot have been emptied while we waited.
+      guard let token = jobServer.acquire() else {
+        return dispatchRemainingWithoutTokens()
```
artemcm: `brokerLoop` is the entire execution thread workload, so returning here shuts it down. 
But I believe there are cases where `acquire` returns `nil` without indicating shutting down the build. 

Calls to `euqueue` after that (and after this call to `dispatchRemainingWithoutTokens`) will add jobs but no thread is there to wait on it so those jobs won't run at all and never call `taskIsComplete` so I suspect the driver would hang.

Sources/SwiftDriverExecution/JobServer.swift
```diff
@@ -0,0 +1,500 @@
...
+  /// Waits until the pool has a token or the waiter is woken.
+  private func waitForToken() -> (hasToken: Bool, wasWokenUp: Bool) {
+    var descriptors = [
+      pollfd(fd: readFD, events: Int16(POLLIN), revents: 0),
+      pollfd(fd: wakeupReadFD, events: Int16(POLLIN), revents: 0),
+    ]
+    while true {
+      let ready = poll(&descriptors, nfds_t(descriptors.count),
+                       Self.pollIntervalMilliseconds)
+      if ready < 0 && errno == EINTR { continue }
```
artemcm: If this is a `poll` error other than `EINTR`, we will return `(false, false)` in the next line, same as in a timeout, loop back to the `acquireFromPool` loop which will treat it as "no token yet" and end up back here, I believe, where `poll` will likely fail again and we will spin this thread for the rest of the build.

Sources/SwiftDriverExecution/JobServer.swift
```diff
@@ -0,0 +1,500 @@
...
+      guard descriptors.count == 2,
+            let readFD = CInt(descriptors[0]),
+            let writeFD = CInt(descriptors[1]),
+            fcntl(readFD, F_GETFD) != -1,
+            fcntl(writeFD, F_GETFD) != -1 else {
+        close(wakeup[0])
+        close(wakeup[1])
+        return nil
+      }
+      // Our children are compiler frontends rather than jobserver clients, so
```
artemcm, line 223: I think we similarly need to censor `MAKEFLAGS` from the environment we pass to all of the compilation tasks.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
nkcsgexi: Makes sense. Addressed!

Sources/SwiftDriverExecution/JobServer.swift
nkcsgexi: Done!

Sources/SwiftDriverExecution/JobServer.swift
nkcsgexi: Great catch! I think this is a real issue that swift-driver may hang after the brokenLoop's exit. I've changed it so that `dispatchRemainingWithoutTokens()` is the sole exit point of `brokerLoop() ` and all of the remaining jobs are scheduled without waiting on tokens.

Sources/SwiftDriverExecution/JobServer.swift
nkcsgexi: Great catch! I've updated it so we only retry polling when the errno is `EINTR`.

Sources/SwiftDriverExecution/JobServer.swift
nkcsgexi: Looking into this more closely, it seems to me that the blocking `read` here might not be a significant issue. When a sibling steals the byte between our poll and read, the broker parks in read — but the stolen byte isn't gone: the sibling that took it is now running a job on that token and will write it back when the job finishes. I incline to not adding more complexity in addressing this, WDYT?

Sources/SwiftDriverExecution/MultiJobExecutor.swift
nkcsgexi: Please note that here we have to use `Int.max` to make the token count the sole limiter. However, it's a great point that we should adopt `numParallelJobs` in `dispatchRemainingWithoutTokens`, and I've added that in the latest commit.

Sources/SwiftDriverExecution/JobServer.swift
```diff
@@ -0,0 +1,556 @@
...
+
+  /// Guards `pending`, `isShutDown` and `hasBrokerFinished`, and wakes the
+  /// broker thread.
+  private let condition = NSCondition()
+  private var pending: [() -> Void] = []
+  private var isShutDown = false
+  private var hasBrokerFinished = false
+
+  /// How long `shutDown` waits for the broker before giving up. The broker can
+  /// be parked in an uninterruptible read, and hanging teardown is worse than
```
artemcm, line 419:
```suggestion
  /// be parked in an uninterruptible read, and a teardown that can block indefinitely is worse than
```

Sources/SwiftDriverExecution/JobServer.swift
artemcm: This may not deadlock most of the time, but it's still unnecessary to wait for another client's token write back. Also, I think this still could deadlock if that sibling client which took the byte somehow exists without writing it back.

From what I understand, with Make 3.81, the reads from this pipe are always blocking, which means there are going to be multiple clients blocking/waiting to read a byte from this pipe and therefore often competing with each other as a general case.

Sources/SwiftDriverExecution/MultiJobExecutor.swift
artemcm: Thanks for the `dispatchRemainingWithoutTokens` fix. 

On `Int.max`, when the user specifies `-j`, why shouldn't we respect both: have the token count be the actual limit and still treat `-j` as upper bound? 

Also, with `SWIFTC_MAXIMUM_DETERMINISM`, we force set `-j 1`, which this change currently would break and using `numParallelJobs` would still end up preserving automatically.

Sources/SwiftDriverExecution/JobServer.swift
```diff
@@ -0,0 +1,556 @@
...
+  }
+
+  /// Extracts the jobserver authentication string from the value of
+  /// `MAKEFLAGS`, or returns `nil` if it does not advertise a pool.
+  ///
+  /// GNU make 4.2 renamed `--jobserver-fds` to `--jobserver-auth`; the old
+  /// spelling is still what make 3.81 (macOS's `/usr/bin/make`) emits, so both
+  /// are accepted. The last occurrence wins.
+  static func parseAuthentication(from makeFlags: String) -> String? {
+    let prefixes = ["--jobserver-auth=", "--jobserver-fds="]
```
artemcm, line 132: This code is duplicated below in `censoringAuthentication`, we should unify these.

Sources/SwiftDriverExecution/JobServer.swift
```diff
@@ -0,0 +1,556 @@
...
+    var authentication: String? = nil
+    for flag in makeFlags.split(separator: " ") {
```
artemcm, line 134: I think this would not be able to handle strings backslash-escaped whitespace. 
e.g. `--jobserver-auth=fifo:/tmp/js\ audit/GMfifo7872`.

artemcm (commented):
One other thing that just occurred to me is how we handle batch mode. 
In batch mode, we compute batch sizes before the executor and are then committed to specific batches, which is likely to be far from optimal when using the job server. 

When there is an explicit `-j` specified by the user it probably still makes sense to respect it, but when there isn't one, and the driver knows it is joining a jobserver pool, maybe we disable batching (and warn the user, perhaps)?

PR #2192, opened 2026-09-19
Adopt Swift 6 language mode

owenv:
This is easiest to review commit by commit, many of them are very mechanical changes. I can also break it up more if that's easier.

Notable changes:
- The package tools version increases from 5.10 to 6.0, AFAIK this is safe now
- There are ~half a dozen instances of @unchecked Sendable which could probably be eliminated with some more invasive changes
- TSCBasic is imported @preconcurrency for now. With some cleanup of stderr/stdout handling we could drop this
- I switched from main.swift to @main in the driver executable because otherwise @MainActor inference caused a bunch of issues


owenv:
@swift-ci test

owenv:
hmm, differences between how compiler versions interpret Swift 6 mode might make this tricky to adopt. I might need to drop CheckImplementationOnly, will take a closer look in a bit

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test windows

Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/InterModuleDependencyOracle.swift
```diff
@@ -32,7 +32,7 @@ import Dispatch
 // underlying implementation of the interface with the libSwiftScan shared
 // library, allowing us to replace the underlying implementation in the future.
 //
-public class InterModuleDependencyOracle {
+public final class InterModuleDependencyOracle: @unchecked Sendable {
```
DougGregor, line 35: Why is it okay for this to be `@unchecked Sendable`? Is the `SwiftScan` instance here doing its own synchronization?

Sources/SwiftDriver/SwiftScan/SwiftScanCAS.swift
```diff
@@ -22,7 +22,7 @@ import struct Foundation.Data
 import _Concurrency
 #endif
 
-public final class CachedCompilation {
+public final class CachedCompilation: @unchecked Sendable {
```
DougGregor, line 25: Same question about `@unchecked Sendable` here.

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -349,7 +349,7 @@ extension VirtualPath {
   }
 
   /// An implementation of a concurrent path cache.
-  private final class PathCache {
+  private final class PathCache: @unchecked Sendable {
     private var uniquer: [String: VirtualPath.Handle]
     private var table: [VirtualPath]
     private let queue: DispatchQueue
```
DougGregor, line 355: Okay, the queue is what makes sense `@unchecked Sendable` okay. AFAICT we're only doing `sync` access to the queue, so it looks like a case for using `Mutex` instead.

Sources/SwiftDriver/Utilities/VirtualPath.swift
```diff
@@ -479,7 +479,7 @@ extension VirtualPath {
   }
 
   /// A cache of created temporary files
-  private final class TemporaryFileStore {
+  private final class TemporaryFileStore: @unchecked Sendable {
     private var uniqueFileCountDict: [String: Int]
     private var queue: DispatchQueue
```
DougGregor, line 484: Same as before: it's okay with the queue, but this could probably be improved by switching to `Mutex` at some point.

DougGregor (approved):
A couple of questions/comments over `@unchecked Sendable`, but this is looking good to me. Thank you!

Sources/SwiftDriver/ExplicitModuleBuilds/InterModuleDependencies/InterModuleDependencyOracle.swift
owenv: the scanner itself is designed to be thread safe, and we only reassign it on `queue`

Sources/SwiftDriver/SwiftScan/SwiftScanCAS.swift
owenv: This should be immutable and never accessed simultaneously. It should really be a noncopyable struct but that doesn't work well with the older compilers we support

Sources/SwiftDriver/Utilities/VirtualPath.swift
owenv: Can't use Mutex here because the macOS deployment target is still 12.0

DougGregor: Ahhh, okay.

owenv:
@swift-ci test

owenv:
@swift-ci test windows

artemcm (approved):

artemcm:
@swift-ci test windows

owenv:
@swift-ci test

owenv:
@swift-ci test windows

owenv:
@swift-ci test

owenv:
@swift-ci test

owenv:
@swift-ci test macOS

owenv:
this is proving to be unwieldy, I'm going to start peeling off patches to land separately

cachemeifyoucan:
The macOS test is broken by @artemcm https://github.com/swiftlang/swift-driver/pull/2188

PR #2195, opened 2026-09-21
[Caching][PrefixMapping] Partial revert PR #2159

cachemeifyoucan:
In PR 2159, a workaround was added for using three build configurations together: bridging header, emit objc header, and caching prefix mapping, which can cause prefix mapped path leaked inside the objc header.

However, that workaround can cause build time regression due to more module variants now needs to be built, even when the prefix mapping will not caused leaked path. Rephase the warning to point out the possible issue and disable the behavior for automatically turning off the prefix mapping.

rdar://188001267


cachemeifyoucan:
@swift-ci please test

owenv (approved):

cachemeifyoucan:
@swift-ci please test

cachemeifyoucan:

cachemeifyoucan:

PR #2196, opened 2026-09-23
Parse -frontend from response files

keith:
This matches the C++ driver which also expands response files before
parsing any arguments:

https://github.com/swiftlang/swift/blob/df39904186029b97ac50eb845bf98bf1a2f3b1ef/lib/DriverTool/driver.cpp#L476

Previously if you had an invocation like:

```
swiftc @foo.params
```

With `foo.params`:

```
-frontend
-some-frontend-arg
```

This would fail because the mode was not parsed from this file first.


keith:
@swift-ci please test

keith:
should i be checking in these package.swift changes? i didn't make them ofc, it seems like it was out of date

artemcm (approved):
Thanks!

